PR #43, opened 2022-04-07
fix: update azure SWA preset and update node versions detection

danielroe:
### 📚 Description

This is now how we take advantage of the new Node runtimes with Azure SWA: https://docs.microsoft.com/en-us/azure/static-web-apps/configuration#platform

### DIFF
--- a/src/presets/azure.ts
+++ b/src/presets/azure.ts
@@ -26,7 +26,18 @@ async function writeRoutes (nitro) {
     version: '2.0'
   }
 
+  let nodeVersion = '16'
+  try {
+    const currentNodeVersion = fse.readJSONSync(join(nitro.options.rootDir, 'package.json')).engines.node
+    if (['16', '14'].includes(currentNodeVersion)) {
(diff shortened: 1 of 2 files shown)


src/presets/azure.ts
```diff
@@ -26,7 +26,18 @@ async function writeRoutes (nitro) {
     version: '2.0'
   }
 
+  let nodeVersion = '16'
+  try {
+    const currentNodeVersion = fse.readJSONSync(join(nitro.options.rootDir, 'package.json')).engines.node
```
pi0, line 31: Isn't engines.node optional? Are we doing `try/catch` to ignore this?

danielroe: Implementation from https://github.com/unjs/nitropack/blob/305c49853d514ceb00342107a7affdbd9500f5f2/src/presets/firebase.ts#L57-L65.

But maybe might be better to just grab from `process`?

pi0: I like the idea of falling back to process.versions :)

PR #47, opened 2022-04-07
chore: fix term cases

antfu:
The case police is here! 👮‍♂️😜


pi0:
We shall add CI (also maybe nitro to term policy :D)

PR #58, opened 2022-04-08
fix: allow users to override nitro error

danielroe:
### ❓ Type of change

### 📚 Description

Previously it was impossible to override sub-paths (for example, `#nitro/error`) as `#nitro` was added first to the alias object.

~Do not yet merge as there is an error that this reveals in Nuxt that needs to be resolved first:~ resolved in https://github.com/nuxt/framework/pull/4215

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -68,7 +68,6 @@ const NitroDefaults: NitroConfig = {
   commands: {}
 }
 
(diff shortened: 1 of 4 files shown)


pi0:
We might use `#nitro-*` alias for a temporary error alias workaround. I don't think reusing `#nitro` namespace is best idea in terms of stability.

pi0 (approved):

PR #70, opened 2022-04-11
fix: normalize `#build` windows path for dev and prerender presets

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4232

### ❓ Type of change

### 📚 Description

The fix applied to dev preset also needs to be applied to prerender preset. It seemed easiest way was to extract from options and add to preset itself. There may be a better way to do this.

### 📝 Checklist

### DIFF
--- a/src/presets/nitro-dev.ts
+++ b/src/presets/nitro-dev.ts
@@ -1,3 +1,5 @@
+import { pathToFileURL } from 'url'
+import { isWindows } from 'std-env'
 import { defineNitroPreset } from '../preset'
 
 export const nitroDev = defineNitroPreset({
@@ -8,5 +10,13 @@ export const nitroDev = defineNitroPreset({
   },
   externals: { trace: false },
   inlineDynamicImports: true, // externals plugin limitation
-  sourceMap: true
+  sourceMap: true,
+  hooks: {
+    'nitro:rollup:before' (nitro) {
+      if (isWindows) {
+        // Windows dynamic imports should be file:// url
+        nitro.options.alias['#build'] = pathToFileURL(nitro.options.buildDir).href
+      }
+    }
+  }
 })
--- a/src/presets/nitro-prerender.ts
+++ b/src/presets/nitro-prerender.ts
@@ -12,7 +12,7 @@ export const nitroPrerender = defineNitroPreset({
   hooks: {
     'nitro:rollup:before' (nitro) {
       if (isWindows) {
-      // Windows dynamic imports should be file:// url
+       // Windows dynamic imports should be file:// url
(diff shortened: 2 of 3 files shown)


src/presets/nitro-dev.ts
```diff
@@ -8,5 +10,13 @@ export const nitroDev = defineNitroPreset({
   },
   externals: { trace: false },
   inlineDynamicImports: true, // externals plugin limitation
-  sourceMap: true
+  sourceMap: true,
+  hooks: {
+    'nitro:rollup:before' (nitro) {
+      if (isWindows) {
+        // Windows dynamic imports should be file:// url
+        nitro.options.alias['#build'] = pathToFileURL(nitro.options.buildDir).href
```
pi0, line 18: Shared situation with both is `externals: { trace: false }`

src/presets/nitro-prerender.ts
```diff
@@ -6,5 +8,13 @@ export const nitroPrerender = defineNitroPreset({
   output: {
     serverDir: '{{ buildDir }}/prerender'
   },
-  externals: { trace: false }
+  externals: { trace: false },
+  hooks: {
+    'nitro:rollup:before' (nitro) {
+      if (isWindows) {
+      // Windows dynamic imports should be file:// url
```
pi0, line 15:
```suggestion
       // Windows dynamic imports should be file:// url
```

pi0 (approved):
LGTM. Seems a more clear workaround at least.

src/presets/nitro-dev.ts
danielroe: Hm. Perhaps we could just test for that instead in rollup config, or are you happy with the preset option?

src/presets/nitro-prerender.ts
```diff
@@ -6,5 +8,13 @@ export const nitroPrerender = defineNitroPreset({
   output: {
     serverDir: '{{ buildDir }}/prerender'
   },
-  externals: { trace: false }
+  externals: { trace: false },
+  hooks: {
+    'nitro:rollup:before' (nitro) {
+      if (isWindows) {
+       // Windows dynamic imports should be file:// url
```
pi0, line 15:
```suggestion
        // Windows dynamic imports should be file:// url
```

src/presets/nitro-dev.ts
pi0: Improved in https://github.com/unjs/nitro/commit/1ec2ee4cf3dfd355a2b7d8d6d7aa5de6dcb51b7c. Extending alias was too late here.

PR #112, opened 2022-04-14
fix(prerender): only extract relative links

Atinux:
### 🔗 Linked issue

Related to https://github.com/nuxt/nuxt3-docs/issues/21

### ❓ Type of change

### 📚 Description

The crawler should only extract the relative link. I don't know where to add a test case, I did test in the nuxt 3 upcoming docs and fixes the isssue.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -1,5 +1,5 @@
 import { resolve, relative, join } from 'pathe'
-import { parseURL } from 'ufo'
+import { hasProtocol } from 'ufo'
 import { createNitro } from './nitro'
 import { build } from './build'
 import type { Nitro } from './types'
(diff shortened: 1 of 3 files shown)


src/prerender.ts
```diff
@@ -80,7 +80,9 @@ const LINK_REGEX = /href=['"]?([^'" >]+)/g
 function extractLinks (html: string, _url: string) {
   const links: string[] = []
   for (const match of html.matchAll(LINK_REGEX)) {
-    links.push(parseURL(match[1]).pathname)
+    const url = parseURL(match[1])
+    // Extract only relative links
```
pi0, line 84: Currently we don't support relative links either (see `_url` above is unused)

Atinux: What I meant is excluding the url that includes a `host` (external urls).

Atinux: Updated the comment

pi0:
Added tests and an improvement to pass relative links as well.

PR #116, opened 2022-04-15
feat(cache): add `staleMaxAge` option for caching header

Intevel:
### ❓ Type of change

### 📚 Description
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
`stale-while-revalidate` gives you the option to revalidate the response in the background after a certain time.

Example:
`{ swr: true, name: "mySWRData", magAge: 1, staleMaxAge: 30 }`

Will create this `cache-control` headers:
`cache-control: s-maxage=1, stale-while-revalidate=30`

### 📝 Checklist


src/runtime/cache.ts
```diff
@@ -135,7 +136,11 @@ export function defineCachedEventHandler (handler: CompatibilityEventHandler, op
       if (opts.magAge) {
         cacheControl.push(`s-maxage=${opts.magAge}`)
       }
-      cacheControl.push('stale-while-revalidate')
+	  if (opts.staleMaxAge) {
+		  cacheControl.push(`stale-while-revalidate=${opts.staleMaxAge}`)
```
pi0, line 140: I was thinking since we are keeping stale value up to `maxAge` anyway does it make sense to default it to `opts.maxAge`?

Intevel: Mostly you don't use the same values. I think leaving multiple options open to a developer is better.

pi0: I see, however, this is the actual behavior of cachify function right now. We only revalidate until maxAge reaches not before.

Intevel: Yes but that is not the point behind `swr`

pi0: `stale-while-revalidate` header alone tells browser always revalidate no?

Intevel: Yes, but with the options you can define when it should revalidate and not only when the response expired.

pi0: I know but we _already know_ that data is not going to be invalidated until maxAge. But really we can improve default generated value later ;)

Intevel: You have convinced me.

pi0:
Let's iterate with next PRs :)

PR #136, opened 2022-04-21
fix: add parent node_modules directories

antfu:
### 🔗 Linked issue

Partially fix https://github.com/nuxt/framework/issues/4217

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -1,4 +1,5 @@
-import { resolve } from 'pathe'
+import { existsSync } from 'fs'
+import { dirname, join, resolve } from 'pathe'
 import { loadConfig } from 'c12'
 import { klona } from 'klona/full'
 import { camelCase } from 'scule'
@@ -122,7 +123,15 @@ export async function loadOptions (userConfig: NitroConfig = {}): Promise<NitroO
   options.output.publicDir = resolvePath(options.output.publicDir, options)
   options.output.serverDir = resolvePath(options.output.serverDir, options)
 
-  options.nodeModulesDirs.push(resolve(options.rootDir, 'node_modules'))
+  // add parent node_modules directory
+  let currentDir = resolve(options.rootDir)
+  while (currentDir && currentDir !== dirname(currentDir)) {
+    const nodeModule = join(currentDir, 'node_modules')
+    if (existsSync(nodeModule)) {
+      options.nodeModulesDirs.push(nodeModule)
+    }
+    currentDir = dirname(currentDir)
+  }
   options.nodeModulesDirs.push(resolve(pkgDir, 'node_modules'))
   options.nodeModulesDirs = Array.from(new Set(options.nodeModulesDirs))
 


src/options.ts
```diff
@@ -122,7 +123,15 @@ export async function loadOptions (userConfig: NitroConfig = {}): Promise<NitroO
   options.output.publicDir = resolvePath(options.output.publicDir, options)
   options.output.serverDir = resolvePath(options.output.serverDir, options)
 
-  options.nodeModulesDirs.push(resolve(options.rootDir, 'node_modules'))
+  // Scan and add parent node_modules directory
+  let currentDir = resolve(options.rootDir)
+  while (currentDir && currentDir !== dirname(currentDir)) {
```
pi0, line 128: According to the Node.js search algorithm, this should normally happen when all parent `node_modules` are scanned. (https://nodejs.org/api/modules.html#modules_all_together see `NODE_MODULES_PATHS` section). Might this be an issue with rollup plugin?

antfu: Yes, this PR is trying to mimic the behavior of Node. The bug of https://github.com/nuxt/framework/issues/4217 is that the dependencies are hoisted to the upper level so the rollup failed to find it. To me, it sounds reasonable for rollup to only scan the node_modules when you provide them explicitly. I think this is quite an important bug, maybe we could have it for now and raise the discussion to rollup later?

pi0: I'm not confident to add this and masking issue without a minimal reproduction. We use plugin node-resolve that internally uses resolve package. They are supposed to do exactly this! 

https://github.com/browserify/resolve/blob/dc8ad1945c663ba47821a269aefb5549f6177cf7/lib/node-modules-paths.js#L40

pi0: Trying to reproduce. Can you please check this if can find a way to reproduce? https://github.com/pi0/nitro-reproduction

antfu: https://github.com/antfu/nuxt-monorepo-repro

`cd nuxt-app` and `npm run build`

---

We had the same issue in [the modules.nuxtjs.org migration](https://github.com/nuxt/modules/pull/357)  here 

https://vercel.com/nuxt-js/modules/5TEN9v28ppMEzdtFz5whmsnRKKcc

and workaround as
[nuxt/modules@`e2ac251` (#357)](https://github.com/nuxt/modules/pull/357/commits/e2ac25148d7f4868653ceca2014143ac73df4377)

With the suggestions https://github.com/nuxt/framework/issues/4217#issuecomment-1094852657 by @danielroe

pi0: Thanks for repro but my a minimal reproduction for nitro, I meant a _nitro project_. I've updated mine with yarn 3 monorepo however this issue doesn't seems to exist. It is something with nuxt integration.

antfu: Updated your repro. By `cd app && yarn build`

pi0: Thanks! Now i can reproduce it. (More properly with `externals: { inline: ['lodash-es'] }`)

pi0:
Investigating more, currently, resolve package (used by rollup plugin) processes a strange search path:

image

Adding current workaround on top of it, fixes the issue, but also produces even more duplicates:

image

pi0:
I came up with a simpler solution by adding `['node_modules']` to the rollup options, we can create same effect for same path. But definitely need to improve resolver and also introduce [workspaceDir](https://github.com/nuxt/framework/issues/158)

PR #141, opened 2022-04-21
fix: respect `sourceMap` option

danielroe:
### ❓ Type of change

### 📚 Description

We should likely respect `nitro.options.sourceMap` as default value for `esbuild` as well.

Related: https://github.com/nuxt/framework/pull/4509

### 📝 Checklist

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -151,7 +151,7 @@ export const getRollupConfig = (nitro: Nitro) => {
   // esbuild
   rollupConfig.plugins.push(esbuild({
     target: 'es2019',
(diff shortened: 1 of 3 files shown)


src/rollup/config.ts
```diff
@@ -151,7 +151,7 @@ export const getRollupConfig = (nitro: Nitro) => {
   // esbuild
   rollupConfig.plugins.push(esbuild({
     target: 'es2019',
-    sourceMap: true,
+    sourceMap: nitro.options.sourceMap !== false,
```
pi0, line 154: Can we please set the default in `options.ts` to be same for every usage?

PR #179, opened 2022-05-02
fix: improve bundled storage

Atinux:
### ❓ Type of change

### 📚 Description

- Improve `normalizeKey` to remove duplicate `:`
- Use `assets:nitro:bundled:` for the `bundledStorage` to avoid having to prefix with `/` (ex: `cache:content` instead of `/cache/content`)

### DIFF
--- a/src/rollup/plugins/server-assets.ts
+++ b/src/rollup/plugins/server-assets.ts
@@ -3,6 +3,7 @@ import type { Plugin } from 'rollup'
 import createEtag from 'etag'
 import mime from 'mime'
 import { resolve } from 'pathe'
+import { normalizeKey } from 'unstorage'
 import { globby } from 'globby'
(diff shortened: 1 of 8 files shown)


src/rollup/plugins/server-assets.ts
```diff
@@ -71,7 +71,7 @@ for (const asset of serverAssets) {
 }
 
 function normalizeKey (key) {
-  return key.replace(/[/\\]/g, ':').replace(/^:|:$/g, '')
+  return key.replace(/[/\\]/g, ':').replace(/:+/g, ':').replace(/^:|:$/g, '')
```
Atinux, line 74: I don't know if this would be faster:

```suggestion
  return key.replace(/[/\\]/g, ':').replace(/:{2,}/g, ':').replace(/^:|:$/g, '')
```

PR #180, opened 2022-05-03
refactor: improve `esbuild.target` option handling

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/4771

### ❓ Type of change

### 📚 Description

Context: https://github.com/nuxt/framework/issues/507

This sets esbuild target to currently minimum node version required for project, and removes the additional default set within the plugin.

https://esbuild.github.io/api/#target

### 📝 Checklist

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -152,7 +152,6 @@ export const getRollupConfig = (nitro: Nitro) => {
 
   // esbuild
   rollupConfig.plugins.push(esbuild({
-    target: 'es2019',
     sourceMap: nitro.options.sourceMap,
     ...nitro.options.esbuild?.options
   }))
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -152,6 +152,7 @@ export const getRollupConfig = (nitro: Nitro) => {
 
   // esbuild
   rollupConfig.plugins.push(esbuild({
+    target: 'node14.16.0',
     sourceMap: nitro.options.sourceMap,
     ...nitro.options.esbuild?.options
   }))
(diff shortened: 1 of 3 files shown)


pi0:
We might set the default to a fixed value to avoid unpredictable behavior.

It is also not _solving_ linked issue. `import.meta.url` should be avoided regardless in runtime!

danielroe:
Yes, agreed that's a separate issue. But I don't see a downside to defaulting to minimum supported node version. (We could then override in specific presets, e.g. for Cloudflare.) What do you think of the change?

pi0:
No. Node.js should not be esbuild target! It will make strange issues when bundled. Let's track universal `import.meta` support via #181

danielroe:
It is currently set to node12/esbuild2019. I think this should be changed.

pi0:
It is always set to `es2019`. Migration to `es2020` is a good idea but I need to investigate more to ensure there would be no breaking changes. 

What is goal of this PR?

danielroe:
If you want to stay on `es2019`, then I think the changes this PR makes are more in the nature of a tweak than a change.

However, `es2020` should be safe in node 14+. FWIW, I don't believe `node14.16.0` uses _node-specific_ APIs. It's just shorthand for the JS features supported in Node 14.16.0. (See [here](https://esbuild.github.io/content-types/#javascript) for details of the transforms.) What makes you think it would cause issues once bundled?

pi0:
Migration from es2019 to es2020:

- Removes optional chaining / nullish  coalescing transform. It can break some targets
- Adds `import.meta`. It is not safe to use even within context of Node for purpose of `import.meta.url`. (therefore https://github.com/unjs/nitro/issues/181)

What else this change would solve?

image

PR #186, opened 2022-05-03
docs: update cloudflare deploy instructions

manigandham:
The example TOML config is incorrect, updated based on the latest docs: https://developers.cloudflare.com/workers/cli-wrangler/configuration/#build

### DIFF
--- a/docs/deploy/providers/cloudflare.md
+++ b/docs/deploy/providers/cloudflare.md
@@ -19,8 +19,10 @@ bucket = ".output/public"
 entry-point = ".output"
 
 [build]
-command = "true"
-upload.format = "service-worker"
+command = "yarn build"
+
+[build.upload]
+format = "service-worker"
 ```
 
 ## Testing locally


docs/deploy/providers/cloudflare.md
```diff
@@ -19,8 +19,10 @@ bucket = ".output/public"
 entry-point = ".output"
 
 [build]
-command = "true"
-upload.format = "service-worker"
+command = "yarn build"
```
pi0, line 22: We assume `.output` is already built in github actions (this is important because `NITRO_PRESET="cloudflare"` should be set for the correct format.

docs/deploy/providers/cloudflare.md
```diff
@@ -19,8 +19,10 @@ bucket = ".output/public"
 entry-point = ".output"
 
 [build]
-command = "true"
-upload.format = "service-worker"
```
pi0, line 23: I assume this is a shortcut and should work. Doesn't it?

docs/deploy/providers/cloudflare.md
manigandham: I'm not sure about this - depends on where `wrangler publish` is run so it can point to the correct entry.

docs/deploy/providers/cloudflare.md
manigandham: No it doesn't work, the format might have changed over time but I'm using the what's in the CF docs now (and that's what works for my project).

pi0: I'm afraid not. The format is standard toml! We have a working deployment (https://nitro-deployment.pi0.workers.dev/) currently with same config (tried few minutes ago).

docs/deploy/providers/cloudflare.md
pi0: Are you exactly following guide in the [docs](https://nitro.unjs.io/deploy/providers/cloudflare.html#deploy-from-your-local-machine-using-wrangler) or using your own action / CI-CD workflow?

pi0:
Thanks for the PR @manigandham but I'm afraid these changes are not necessary unless you have your own custom deployment flow which might differ in requirements.

PixelBoii:
@pi0 I think the issue here is that setting the `build.command` to `True` works fine for Linux systems, but throws an error on Windows. Using the updated config that this PR suggested works for me on Windows and seems like a good solution imo (altough I am facing some other issues with github actions specificially, but those seem unrelated).

Apologies for the lack of logs here, I just stumbled upon this while on my phone. Hopefully it makes sense at least 👌

manigandham:
```
PS C:\dev\ntest> wrangler publish
 Running true
'true' is not recognized as an internal or external command,
operable program or batch file.
Error: Build failed! Status Code: 1

PS C:\dev\ntest> wrangler publish
Error:  Workers Sites requires using a bundler, and your configuration indicates that you aren't using one. You can fix this by:
* setting your project type to "webpack" to use our automatically configured webpack bundler.
* setting your project type to "javascript", and configuring a build command in the `[build]` section if you wish to use your choice of bundler.
```

docs/deploy/providers/cloudflare.md
manigandham: Yes following the guide does not work. Logs posted in the comment below.

pi0: (the guide is not meant for windows PC but github action with linux)

manigandham:
@pi0 

> (the guide is not meant for windows PC but github action with linux)

The guide has a "Deploy from your local machine using wrangler" section:

image

The guide uses the same config for both local and github actions. I don't understand the issue here. Why deviate from the official format in the Cloudflare docs only to be *less* compatible for people?

pi0:
You can `build` command to `""` for windows compatibility issue (docs updated in https://github.com/unjs/nitro/commit/91f53398c77613206367fae501e648fc40a41e04 thanks @PixelBoii for the hint!)

> Why deviate from the official format in the Cloudflare docs only to be less compatible for people?

Dear @manigandham  There isn't really an issue with config! With nitro, unlike platform-specific CLIs (vercel, netlify, wrangler), nitro has a platform-agnostic dev/build stage + platform-specific deploy. We use wrangler only for deployment which is why not using `build` option in `wrangler.toml`. Nitro's usecase simply doesn't falls under basic usage guide of wrangler.

PR #205, opened 2022-05-06
fix: add missing include path `../../**/*` in generate tsconfig.json

### LINKED ISSUE #204: type hints not working as designed, missing tsconfig include path
### Environment

Nitro: Latest
Node: v16.13.0

### Reproduction

clone and run pnpm dev: https://github.com/didavid61202/nitro-app

go to `routes/index.ts`, and typescript will throw an error:  `Cannot find name 'eventHandler'.`  and type hints are not working as designed.

### Describe the bug

the generated `tsconfig.json` file in `.nitro/types` is missing "include path" of relative path of the nitro.config.buildDir to the nitro.config.rootDir/srcDir (`"../../**/*"` when using basic config setup), causing typescript unable to grab all global types defined in `.nitro/types/nitro.d.ts`.

adding relative path of the nitro.config.buildDir to the nitro.config.rootDir/srcDir to `tsConfig.include` array in `nitro\src\build.ts` should fix this issue.

can be fix by #205

### Additional context

_No response_

### Logs

_No response_

didavid61202:
### 🔗 Linked issue

fix #204

### ❓ Type of change

### 📚 Description
the generated tsconfig.json file in .nitro/types is missing "include path" of "../../**/*", causing typescript unable to grab all global types defined in .nitro/types/nitro.d.ts.  adding "../../**/*" to tsConfig.include array in nitro\src\build.ts should fix this issue.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -101,7 +101,8 @@ export async function writeTypes (nitro: Nitro) {
       },
       include: [
         './nitro.d.ts',
-        '../../**/*'
+        join(relative(join(nitro.options.buildDir, 'types'), nitro.options.rootDir), '**/*'),
+        ...nitro.options.srcDir !== nitro.options.rootDir ? [join(relative(join(nitro.options.buildDir, 'types'), nitro.options.srcDir), '**/*')] : []
       ]
     }
     await writeFile(join(nitro.options.buildDir, 'types/tsconfig.json'), JSON.stringify(tsConfig, null, 2))
--- a/src/build.ts
+++ b/src/build.ts
@@ -100,7 +100,8 @@ export async function writeTypes (nitro: Nitro) {
           : {}
       },
       include: [
-        './nitro.d.ts'
+        './nitro.d.ts',
+        '../../**/*'
       ]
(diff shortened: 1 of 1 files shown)


src/build.ts
```diff
@@ -100,7 +100,8 @@ export async function writeTypes (nitro: Nitro) {
           : {}
       },
       include: [
-        './nitro.d.ts'
+        './nitro.d.ts',
+        '../../**/*'
```
danielroe, line 104: This is right, but it should be based on the relative path of the buildDir to the root directory rather than hard-coded. See https://github.com/nuxt/framework/blob/2d202b59e33e77174102cc0f6114abd700874f37/packages/nuxi/src/utils/prepare.ts#L27-L31 for example.

didavid61202: ah yes, thanks for pointing this out! 👍

didavid61202: Resolved.

pi0 (approved):

pi0:
Thanks!

Issue #196, opened 2022-05-05
cloudflare-pages-static with Functions


pi0:
Reference: nuxt/framework#110

> @[warflash](https://github.com/warflash)
> Cloudflare pages for static assets + [functions](https://blog.cloudflare.com/cloudflare-pages-goes-full-stack/) for ssr and dynamic content would be awesome!

warflash:
Nice! Really looking forward to that.

Related discussions for easy future reference:
https://github.com/nuxt/framework/discussions/2471
https://github.com/nuxt/framework/discussions/1528
https://github.com/unjs/nitro/discussions/110

@DaniFoldi 🎉

DaniFoldi:
I'll look into implementing this sometime this week (I have a test repo that _works_), probably create a draft PR, would be very helpful if someone with more nitro knowledge could take a look

DaniFoldi:
@warflash I've got #210 drafted, looking for feedback now

warflash:
Awesome, great work!
I can't give much feedback on the preset itself but what I did notice is the use of `_workers.js`.
That reminded me of a conversation on the [cloudflare discord](https://discord.com/channels/595317990191398933/967071916160610344/970000768667090954) I read:

> Functions in particular has been interesting (to me, at least). SvelteKit opted for using _worker.js, whereas Remix and Qwik have opted for functions/[[all]].ts.
Plugins should technically work with the later two, but won’t (currently) with SvelteKit, because of that decision. Maybe we encourage SvelteKit to move to the functions folder, or maybe we offer some other way into the Plugins ecosystem. Haven’t quite worked that bit out yet.

That makes it sounds like `functions/[[all]].ts` is the recommended way forward. I haven't used plugins yet nor am I experienced with nitro presets so I'm not sure how big of a change that is, just figured I'd drop it here.

Again, nicely done 🚀

AlexDicy:
> Awesome, great work!
> I can't give much feedback on the preset itself but what I did notice is the use of `_workers.js`.
> That reminded me of a conversation on the [cloudflare discord](https://discord.com/channels/595317990191398933/967071916160610344/970000768667090954) I read:
> 
> > Functions in particular has been interesting (to me, at least). SvelteKit opted for using _worker.js, whereas Remix and Qwik have opted for functions/[[all]].ts.
[...]

PR #210, opened 2022-05-07
feat: add cloudflare-pages preset

DaniFoldi:
### 🔗 Linked issue

#196 

### ❓ Type of change

### 📚 Description

Related to discussions #110 #196, a few in nuxt/framework

This PR intends to add a preset for Cloudflare Pages Functions - which is currently in Open Beta. Docs can be found at https://developers.cloudflare.com/pages/platform/functions/ .

Pages Functions are based on the `modules` worker type, so a different deployment preset will be needed.

_update since first drafted, removed no longer relevant beta warnings_

Functions are still in beta, however Direct Upload is now available for Pages, and I fixed all the issues I could find.

_update again_

Docs have now been updated (feedback welcome), the newest version of std-env should allow for automatic detection (only via git integration).

Have a good day.

### 📝 Checklist


tuarrep:
Hey @DaniFoldi 👋 

Do you need help to test / finish this one?

I'm currently investigating to migrate to Cloudflare pages and I require this preset.

DaniFoldi:
Hi!

I started writing the docs for the preset earlier today, so it should be ready today or tomorrow - the one remaining thing after this PR will be figuring out how to give api routes the request context.

DaniFoldi:
@tuarrep I believe it's ready for testing, and feel free to leave any feedback on the docs as well.
(vercel had a problem deploying, but doesn't tell what the problem was, so if it's something that I messed up (probably) could someone with access let me know?

src/runtime/entries/cloudflare-pages.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const url = new URL(request.url)
+  let body
+  if (requestHasBody(request)) {
+    body = await useRequestBody(request)
+  }
+
+  const r = await nitroApp.localCall({
+    env,
+    context,
```
pi0, line 34: FYI this currently has no effect ([src](https://github.com/unjs/unenv/blob/e389fb5d73af908c6e15a4c1a9dddd6770326e31/src/runtime/fetch/call.ts#L18)). unenv only picks known keys from input. But we should definitely add cf context in later changes.

pi0 (approved):
Thanks for this nice work 💯  I guess we might have problems with public directory handling (`.output/public`) should be picked. Nuxt used top level `dist/` which is probably if you tested and working. We could do the same btw with preset. Will have to try on nitro-edge.

ricky11:
does this mean by adding ./server/api/function.js and with git integration the preset will be auto set as cloudflare_pages, and the /function directory will be created with the function.js api?   I have tested it here, and while the function folder is created, no function is actually deployed on cloudflare network . Tried running CF_PAGES=1 npm run generate in this stackblitz terminal https://stackblitz.com/edit/github-dxg5qg?file=server%2Fapi%2Fcustomer.get.js

Plinpod:
> 

Seeing same issue with non-get requests returning 405

PR #212, opened 2022-05-08
fix: improve error handler to include stacktrace

tobiasdiez:
### ❓ Type of change

### 📚 Description

If an unhandled error occurs somewhere, nito only prints the error message. This makes it hard to locate where the error occurs and how to fix it. With this PR, the full stacktrace is printed for an error.

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/nitro-dev.ts
+++ b/src/runtime/entries/nitro-dev.ts
@@ -34,5 +34,5 @@ server.listen(listenAddress, () => {
   })
 })
 
-process.on('unhandledRejection', err => console.error('[nitro] [dev] [unhandledRejection] ' + err))
-process.on('uncaughtException', err => console.error('[nitro] [dev] [uncaughtException] ' + err))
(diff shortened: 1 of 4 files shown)


pi0:
Hi @tobiasdiez, and thanks for this PR. But normally, all Promise errors should be handled. These handles are just a fallback to ensure Node.js doesn't exit the process. 

We might support showing full stacks on fallback behind an environment variable flag like `proces.env.DEBUG`. 

But I would also like to know your usage that ends with unhandled flags.

tobiasdiez:
Very early in the migration to nuxt3, I've encountered an error that was only printed through these "unhandled" error handlers. I sadly don't remember the exact context, but I added the the detailed error reporting in https://github.com/JabRef/JabRefOnline/commit/20fac0a6f7fb88acd4941d28727ad60066fdc40f, so maybe something related to the runtime config.

If I encounter this again, I'll let you know where/why it occurred.

Issue #88, opened 2022-04-12
`prerender:` hooks


pi0:
https://github.com/nuxt/framework/issues/523

Most importantly `prerender:route` hook which can be useful to modify and inspect pretender results. (eg: for service-worker to inject it's script)

farnabaz:
We have a `prerender:route` hoo, but it is not useful.
https://github.com/unjs/nitro/blob/cbc7ba3835df539116ca6560a85365615af55fbb/src/prerender.ts#L99

But this should move into `generateRoute` function, before persisting content into file

https://github.com/unjs/nitro/blob/cbc7ba3835df539116ca6560a85365615af55fbb/src/prerender.ts#L66-L68

WDYT @pi0

pi0:
Nice find @farnabaz. `prerender:route` is called outside as for generic error handling and stats purposes. We might introduce another predenrer hook and use before `writeFileCall` to allow extending to skipping.

PR #213, opened 2022-05-09
feat: add prerender:route hook

TotomInc:
### ❓ Type of change

### 📚 Description

Add a new hook `prerender:route` that is called when a single page has been successfully prerendered.

This hook can be used externally to create specific logic tied to route generation, such as editing output of a route.

`generateRoute` now returns `{ contents, route, filePath }` instead of nothing, in order to:

- Access the saved file directly by its `filePath`.
- Read the `contents` that have been generated.

I can see more potential for SSG (fully static).

A quick example usage could be to dynamically download and replace static assets, stored on an API server, that you don't want to use in production (parse HTML content, download image, replace image path with the one downloaded locally).

Please, let me know if this change is not intended for nitro.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -67,6 +67,8 @@ export async function prerender (nitro: Nitro) {
         }
       }
     }
+
+    return { route, contents, filePath }
   }
 
   nitro.logger.info(nitro.options.prerender.crawlLinks
@@ -76,9 +78,16 @@ export async function prerender (nitro: Nitro) {
   for (let i = 0; i < 100 && routes.size; i++) {
     for (const route of Array.from(routes)) {
       const start = Date.now()
-      const error = await generateRoute(route).catch(err => err)
-      const end = Date.now()
-      nitro.logger.log(chalk.gray(`  ├─ ${route} (${end - start}ms) ${error ? `(${error})` : ''}`))
+      let end: number | null = null
+
+      try {
+        const { contents, filePath } = await generateRoute(route)
+        end = Date.now()
+        nitro.logger.log(chalk.gray(`  ├─ ${route} (${end - start}ms)`))
+        nitro.hooks.callHook('prerender:route', { route, contents, filePath })
+      } catch (error) {
+        nitro.logger.log(chalk.gray(`  ├─ ${route} (${end - start}ms) ${error ? `(${error})` : ''}`))
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/prerender.ts
```diff
@@ -76,9 +78,16 @@ export async function prerender (nitro: Nitro) {
...
-      const error = await generateRoute(route).catch(err => err)
-      const end = Date.now()
-      nitro.logger.log(chalk.gray(`  ├─ ${route} (${end - start}ms) ${error ? `(${error})` : ''}`))
+      let end: number | null = null
+
+      try {
+        const { contents, filePath } = await generateRoute(route)
+        end = Date.now()
+        nitro.logger.log(chalk.gray(`  ├─ ${route} (${end - start}ms)`))
+        nitro.hooks.callHook('prerender:route', { route, contents, filePath })
```
pi0, line 87: We need to await here
```suggestion
        await nitro.hooks.callHook('prerender:route', { route, contents, filePath })
```

src/prerender.ts
```diff
@@ -76,9 +78,16 @@ export async function prerender (nitro: Nitro) {
...
+      } catch (error) {
+        nitro.logger.log(chalk.gray(`  ├─ ${route} (${end - start}ms) ${error ? `(${error})` : ''}`))
```
pi0, line 89: I would also allow hook access to the errors rendered with error.

pi0:
Thanks! This also resolves https://github.com/unjs/nitro/issues/88

src/prerender.ts
TotomInc: Thanks, added!

src/prerender.ts
TotomInc: Thanks for the feedback.

I've made the return response of the `prerender:routes` hook to be an object or an error. I also thought of introducing a hook to handle errors such as `prerender:route:error`. If you think it's cleaner to do this, I can mek the change.

pi0: I was thinking, that by using a single hook, we can allow:

- Custom error (hooking to validate and modify context to extract an error being thrown)
- Clear error (hooking to allow still prerendering some pages with errors)

TotomInc: Alright, I'm not sure I have understood correctly, but please let me know if my last changes seems good for you.

pi0:
❤️

PR #247, opened 2022-05-17
feat(dev): add `x-forwarded` headers

abdullrahman02000:
This is very important since every reverse proxy should provide a way sending client ip, port, ip family.
It caused issue in nuxt3 which uses nitro because it is not possible to get client info in development mode

I am sorry I did not add more details, It is the first time I contribute, I have already modified it on my PC and tested it.
Look at this link:
https://github.com/http-party/node-http-proxy#listening-for-proxy-events


pi0 (approved):

pi0:
Thanks!

PR #252, opened 2022-05-19
feat(handlers): allow method suffix in handler file path

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/framework/issues/5020

### ❓ Type of change

### 📚 Description

Enhance the server handlers.
Allow (if the method is not specified) to use the method suffix as the handler's route method

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -45,7 +45,15 @@ export async function writeTypes (nitro: Nitro) {
 
   const middleware = [
     ...nitro.scannedHandlers,
-    ...nitro.options.handlers
+    ...nitro.options.handlers.map((handler) => {
+      if (handler.method) { return handler }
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for this pull request @huang-julien (and sorry to check late!). I think we could directly add this to nuxt kit [addServerHandler](https://github.com/nuxt/framework/blob/59da4c6dd4759881377de8f0f83f21ef86fdd66e/packages/kit/src/server.ts#L26). Nitro automatically uses this suffix but only for discovered files. Seems too much magic to me for options. We can first try to do this on nuxt kit and and useful enough backport as nitro behavior. (and if we do, we can do the normalization in single place of `src/config.ts`).

PR #253, opened 2022-05-21
feat: add option to define additional presets

jarvainen:
### ❓ Type of change

### 📚 Description

Added new property to NitroConfig, which allows user to define own presets. Changed `#internal/nitro` alias to mandatory, so when user defines own alias to import his own preset from projects root directory, `#internal/nitro` does not get overwritten.

I felt the need for this kind of option, when I was working on a project with nuxt3, and had to start websocket server on the same server instance that was running nitro, and there was no easy way to do this without writing a new preset.

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -57,9 +57,7 @@ const NitroDefaults: NitroConfig = {
   },
 
   // Rollup
-  alias: {
-    '#internal/nitro': runtimeDir
-  },
+  alias: {},
   unenv: {},
   analyze: false,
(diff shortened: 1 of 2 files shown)


pi0:
Hi @jarvainen, and thanks for the pull request. I understand that providing a local preset might be useful but I'm not sure about adding a new `presets` option. Can you please elaborate more about your usecase in a issue or discussion?

Also why one would override `#internal/nitro` alias? If they do, I guess there was clearly a reason but in normal conditions, this should something obvious to avoid unless knowing what we are doing. same as rest of defaults...

PR #265, opened 2022-06-01
fix(cloudflare): use `@cloudflare/wrangler` until we support wrangler 2

Atinux:
### ❓ Type of change

### 📚 Description

We do not support wrangler 2 so far:

Screenshot 2022-06-01 at 16 55 47@2x

This PR fixes the deploy command displayed in order to work properly (which is currently displayed when using Nuxt 3 as well).

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -8,7 +8,7 @@ export const cloudflare = defineNitroPreset({
   entry: '#internal/nitro/entries/cloudflare',
   commands: {
     preview: 'npx miniflare ./server/index.mjs --site ./public',
-    deploy: 'cd ./server && npx wrangler publish'
+    deploy: 'npx @cloudflare/wrangler publish'
   },
   hooks: {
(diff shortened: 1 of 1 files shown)


src/presets/cloudflare.ts
```diff
@@ -8,7 +8,7 @@ export const cloudflare = defineNitroPreset({
   entry: '#internal/nitro/entries/cloudflare',
   commands: {
     preview: 'npx miniflare ./server/index.mjs --site ./public',
-    deploy: 'cd ./server && npx wrangler publish'
+    deploy: 'npx @cloudflare/wrangler publish'
```
danielroe, line 11: do we not need to cd with `@cloudflare/wrangler`?

Atinux: Actually not it was not needed since the `wrangler.toml` actually gives output to read from

danielroe (approved):

pi0 (approved):

PR #271, opened 2022-06-08
fix(cache): invalidate entry before calling resolver

clarkdo:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/5328

### ❓ Type of change

### 📚 Description

Resolves https://github.com/nuxt/framework/issues/5328

When cache is expired and multiple request come in at the same time, 1st request is pending at the handler, then 2nd request will reuse invalid response from previous expired value and 404 response may be returned.

This pr is fixing the invalid cached response by removing previous response value before handler is called.

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -52,8 +52,11 @@ export function defineCachedFunction <T=any> (fn: ((...args) => T | Promise<T>),
 
     const _resolve = async () => {
       if (!pending[key]) {
-        // Remove cached value for preventing using expired cache on concurrent requests
+        // Remove cached entry to prevent using expired cache on concurrent requests
         entry.value = undefined
+        entry.integrity = undefined
+        entry.mtime = undefined
+        entry.expires = undefined
         pending[key] = Promise.resolve(resolver())
       }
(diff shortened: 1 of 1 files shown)


src/runtime/cache.ts
```diff
@@ -52,6 +52,8 @@ export function defineCachedFunction <T=any> (fn: ((...args) => T | Promise<T>),
 
     const _resolve = async () => {
       if (!pending[key]) {
+        // Remove cached value for preventing using expired cache on concurrent requests
+        delete entry.value
```
pi0, line 56:
```suggestion
        entry.value = undefined
```
```suggestion
        delete entry.value
```

pi0, line 56:
```suggestion
        entry.value = undefined
```
```suggestion
        delete entry.value
```

clarkdo: Thanks, is this OK to be merged ?

pi0 (changes requested):
Thanks <3

pi0 (approved):

PR #289, opened 2022-06-14
fix(prerender): avoid adding duplicate baseURL for local fetch

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/content/issues/1252

### ❓ Type of change

### 📚 Description

When following links with a baseURL set, we were following them directly, resulting in double-prefix of baseURL.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -1,6 +1,6 @@
 import { pathToFileURL } from 'url'
 import { resolve, join } from 'pathe'
-import { joinURL, parseURL } from 'ufo'
+import { joinURL, parseURL, withoutBase } from 'ufo'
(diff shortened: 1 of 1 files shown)


pi0:
Do you mind creating an issue with minimum (nitro) reproduction? I believe base URL has to be fetched as-is when configured but might be wrong.

danielroe:
We are adding the baseURL here:

https://github.com/unjs/nitropack/blob/main/src/prerender.ts#L55

So if (in the Nitro response) there is a URL with that baseURL already present (as would be required for routing to work) then it will add the baseURL again at the point of generating the route.

PR #292, opened 2022-06-15
chore: disable `import/no-named-as-default` eslint rule

antfu:
For `defu`


pi0:
Going to use named export soon. Let's keep the rule :)

Issue #294, opened 2022-06-15
Enable auto import for scanDirs


pi0:
History: https://github.com/nuxt/framework/pull/5042#issuecomment-1152357425, https://github.com/unjs/nitro/pull/277

We need to allow auto importing from user code in all layers (when extending from `node_modules`). But it made some regressions with auto imports (https://github.com/nuxt/nuxt.js/issues/14137) feature to be enabled back once solved.

pi0x:
Auto-imports were removed entirely in Nitro v3 (#4573), so there is no auto-import layer left to extend to scanDirs. See the migration guide: https://nitro.build/docs/migration#auto-imports-are-removed

_Triaged with AI assistance._

PR #295, opened 2022-06-15
fix(options):  correct `include` empty value

cinob:
### ❓ Type of change

### 📚 Description

If `include` is omitted or has zero length, filter will return `true` by default.
Resulted in scanning the entire directory.

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -136,7 +136,7 @@ export async function loadOptions (userConfig: NitroConfig = {}): Promise<NitroO
     options.scanDirs = [options.srcDir]
   }
 
-  options.autoImport.include = [
+  const include = [
     ...Array.isArray(options.autoImport.include)
       ? options.autoImport.include
       : [options.autoImport.include].filter(Boolean),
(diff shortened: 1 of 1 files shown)


pi0:
Hi dear @cinob and thanks for PR. We currently reverted this bits of code as hotfix (see #294).

I think you discovered a potential bug in underlying [unjs/unimport](https://github.com/unjs/unimport) library that with an empty array implicitly is changing the behavior. Do you think we can move it there? /cc @antfu

antfu:
I wouldn't actually consider it a bug, as `undefined` should be treated differently from `[]`. To me `undefined` means to use the default value, while `[]` indicates that users want to override the default as empty. Otherwise, there would be no way to override the default.

pi0:
But `include: []` shouldn't mean including everything either! What do you mean by overriding default? It can be something like `include: true` if you are thinking of such behavior.

pi0:
Thanks for the reference @antfu. Shall we make it more semantic for unimport? I can understand that it is the behavior of used rollup filter util but their API is not semantic that caused a wrong use by us and also needs a patch similar to this PR to just overcome. (regardless we could improve _exclude_ pattern to improve).

pi0:
Makes sense @antfu. Did you had a chance to work on it?

antfu:
@pi0 included in `unimport@0.4.4`

pi0:
Thanks. I think we then don't need this fix here since unimport should already use defaults with length:0

PR #301, opened 2022-06-20
fix: export runtime types directly from `nitropack`

danielroe:
### ❓ Type of change

### 📚 Description

Otherwise users wanting to use a type like `NitroApp` or `NitroAppPlugin` would have to do something like:

```js
import { NitroApp } from 'nitropack/dist/runtime/app'
```

...or use an internal nitro alias.

### 📝 Checklist

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1,6 +1,4 @@
 export * from './fetch'
 export * from './nitro'
 export * from './handler'
-export type { NitroApp } from '../runtime/app'
-export type { CacheEntry, CacheOptions, ResponseCacheEntry } from '../runtime/cache'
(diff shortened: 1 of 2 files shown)


src/types/index.ts
```diff
@@ -1,3 +1,6 @@
 export * from './fetch'
 export * from './nitro'
 export * from './handler'
+export type { NitroApp } from '../runtime/app'
+export type { CacheEntry, CacheOptions, ResponseCacheEntry } from '../runtime/cache'
+export type { NitroAppPlugin } from '../runtime/plugin'
```
pi0, line 6: Refactor: Can we (re)expose these from `runtime/types.ts`?

PR #328, opened 2022-07-06
fix: allow overriding specific build output files with aliases

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/5013

### ❓ Type of change

- x ] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

In some cases, such as the linked issue (which we could resolve by adding an alias to a stub file in the case of `ssr: false`), it is helpful to be able to override individual `#build/some-file.js` aliases, and to do this we need `#build` to be defined last in the object.

### 📝 Checklist

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -219,13 +219,13 @@ export const plugins = [
   }
   rollupConfig.plugins.push(alias({
     entries: resolveAliases({
-      '#build': buildDir,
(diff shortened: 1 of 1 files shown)


pi0:
We have dynamic virtual imports (https://nitro.unjs.io/config/#virtual) that can be used in order to override any import. (I made this feature initially for linked issue in fact)

`env.alias` can be used to override other aliases for any similar situation.

PR #329, opened 2022-07-06
fix(prerender): remove `baseURL` from generated file paths

farnabaz:
### ❓ Type of change

### 📚 Description
Generating a project using custom baseURL results unwanted directory inside `.output/public`.
For example, if you set `baseURL = romi-project`, you will see a `romi-project` directory in the generated website. https://github.com/hacknug/romi-project/tree/gh-pages

### 📝 Checklist


farnabaz:
Maybe I'm wrong about the behavior of Nuxt pretender. Should it generate a subdirectory inside `.output/public`?

What I think is that the generated website should not have a subdirectory inside it. When I set `baseURL` in nuxt config, I want to deploy it under a directory in a host. Generating this subdirectory causes duplicate `baseURL`

Feel free to close this PR if it does not look good :)

farnabaz:
Yes removing baseURL from `filePath` will do the trick too.

But I think the best way is to remove baseURL from routes and add it in the fetch call. This way we can support URLs like this

```
/baseUrl/baseUrl/something
```

Imagine user have a page with same name is baseURL

danielroe:
I think that's a good point. But we will still need to update the filenames

danielroe (approved):

pi0 (approved):
LGTM. We might add a symbolic link to avoid breaking change and allow cases when public structure is expected to have base url

PR #335, opened 2022-07-12
fix: call `rollup:before` before generating rollup config

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/content/pull/1346#issuecomment-1180431720

### ❓ Type of change

### 📚 Description

Likely in `rollup:before` users might want to modify nitro options with effect of changing rollup config. For example:

https://github.com/nuxt/framework/blob/f321a56e63be508e5f478b1de5227a5add2b977f/packages/nuxt/src/core/nitro.ts#L124-L134

This makes that possible.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -35,8 +35,8 @@ export async function copyPublicAssets (nitro: Nitro) {
 }
 
 export async function build (nitro: Nitro) {
-  const rollupConfig = getRollupConfig(nitro)
   await nitro.hooks.callHook('rollup:before', nitro)
+  const rollupConfig = getRollupConfig(nitro)
   return nitro.options.dev ? _watch(nitro, rollupConfig) : _build(nitro, rollupConfig)
 }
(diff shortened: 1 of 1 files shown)


src/build.ts
```diff
@@ -35,8 +35,8 @@ export async function copyPublicAssets (nitro: Nitro) {
 }
 
 export async function build (nitro: Nitro) {
-  const rollupConfig = getRollupConfig(nitro)
   await nitro.hooks.callHook('rollup:before', nitro)
+  const rollupConfig = getRollupConfig(nitro)
```
pi0, line 39: I think put hook before to allow extending resolved config but not possible. We can use new hook for that :)

pi0 (approved):

src/build.ts
danielroe: Makes sense 👍

PR #347, opened 2022-07-14
fix(config): enable dotenv in development

### LINKED ISSUE #246: .env file don't load in config
### Environment

Node v16.15.0
npm 8.5.5
nitropack ^0.4.4

### Reproduction

Reproducing this is simple, FYI, I'm using Nitro without Nuxt.
If you follow the documentation you would end up with something like this in `nitro.config.ts`:
```
import { defineNitroConfig } from 'nitropack'

  runtimeConfig: {
    supabase: {
      api_url: '',
      api_anon_key: ''
    }
  }
  // extends: [
  //   'config.dev.ts'
  // ]
})
```

And `.env`file:
```
NITRO_SUPABASE_API_URL='xx'
NITRO_SUPABASE_API_ANON_KEY='yy'
```

Trying to read the values in a route with:
```
export default () => {
  const config = useRuntimeConfig()
  return config
}
```

you get empty values (the defaults).
```
{
  "app": {
    "baseURL": "/"
  },
  "nitro": {
    "routes": {}
  },
  "supabase": {
    "api_url": "",
    "api_anon_key": ""
  }
}
```

While the expected behavior is to find the values from the `.env` file:
```
{
  "app": {
    "baseURL": "/"
  },
  "nitro": {
    "routes": {}
  },
  "supabase": {
    "api_url": "xx",
    "api_anon_key": "yy"
  }
}
```

### Describe the bug

I figured from reading the code, that Nitro suppose to get values from `.env` using [unjs/c12](https://github.com/unjs/c12) & [dotenv](https://github.com/motdotla/dotenv) while looping for values in the file that start with `NITRO_` prefix, but it doesn't work (at least on my local)

I found a solution that works for me, which is importing dotenv at the start of `nitro.config.ts`, which solves this issue.

If this is the standard way to do so, then I assume an example of it in the documentation would be helpful (it took me a long time to figure it out myself)

If this is not the standard way, which is what I believe to be so, then importing `dotenv` in the beginning of the `[src/options.ts](https://github.com/unjs/nitro/blob/b153cea1f2e0805088bdc512f490539cccec69c4/src/options.ts)` could solve the issue.

### Additional context

What solved it for me is adding `import 'dotenv/config'` at the top:
```
import 'dotenv/config'
import { defineNitroConfig } from 'nitropack'

  runtimeConfig: {
    supabase: {
      api_url: '',
      api_anon_key: ''
    }
  }
...
```

### Logs

_No response_

kevinmarrec:
### 🔗 Linked issue

Fixes #246

### 🔗 Linked discussion

Closes #238 

### ❓ Type of change

### 📚 Description

This enables `dotenv` setup feature from `c12` in development (`nitro dev` only), making `.env` files working out of the box.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -91,6 +91,7 @@ export async function loadOptions (userConfig: NitroConfig = {}): Promise<NitroO
     name: 'nitro',
     defaults: NitroDefaults,
     cwd: userConfig.rootDir,
+    dotenv: userConfig.dev,
     resolve (id: string) {
       type PT = Map<String, NitroConfig>
       let matchedPreset = (PRESETS as any as PT)[id] || (PRESETS as any as PT)[camelCase(id)]


pi0 (approved):
Nice find. Thanks!

PR #355, opened 2022-07-19
fix(options): exclude `buildDir` for auto-imports

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/pull/5981, https://github.com/nuxt/bridge/issues/421

### ❓ Type of change

### 📚 Description

This disables processing the build output of the previous step from auto-imports. Credit to @cinob for spotting the issue; this just moves the fix up to Nitro if we think that makes sense.

cc: @antfu 

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -148,9 +148,9 @@ export async function loadOptions (userConfig: NitroConfig = {}): Promise<NitroO
   // ]
 
   if (options.autoImport) {
-    options.autoImport.exclude = options.autoImport.exclude || [/[\\/]node_modules[\\/]/, /[\\/]\.git[\\/]/]
+    options.autoImport.exclude = options.autoImport.exclude || []
     if (Array.isArray(options.autoImport.exclude)) {
-      options.autoImport.exclude.push(options.buildDir)
+      options.autoImport.exclude.push(options.buildDir, /[\\/]node_modules[\\/]/, /[\\/]\.git[\\/]/)
     }
   }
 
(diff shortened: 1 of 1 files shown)


antfu (approved):

src/options.ts
```diff
@@ -147,6 +147,13 @@ export async function loadOptions (userConfig: NitroConfig = {}): Promise<NitroO
   //     .map(i => new RegExp(`(^|\\/)${escapeRE(i.split('node_modules/').pop())}(\\/|$)(?!node_modules\\/)`))
   // ]
 
+  if (options.autoImport) {
+    options.autoImport.exclude = options.autoImport.exclude || [/[\\/]node_modules[\\/]/, /[\\/]\.git[\\/]/]
```
pi0, line 151: Default config merging strategy is joining (not override) also we are already pushing buildDir. I think we can add node_modules and .git to the defaults as well.

danielroe: Actually, are we sure it doesn't override?

https://github.com/unjs/unimport/blob/78e8d7416b2ea723167777cc3a1175109e850358/src/unplugin.ts#L26-L29

pi0: Yes, I'm aware unimport does override. Nitro/Nuxt options extend in the other hands. Also good idea for duplicating defaults for predictable behavior 👍🏼

danielroe: Oh, so you were suggesting we always push `.git` and `node_modules` as well?

danielroe: Just checking 😉

pi0 (approved):

PR #357, opened 2022-07-20
fix(aws-lambda): return outgoing cookies on response objects

### LINKED ISSUE #245: aws-lambda preset should use cookies vs set-cookie header in response
### Environment

aws-lambda deployment setting

nitro via nuxt@3.0.0-rc3

### Reproduction

See below

### Describe the bug

It's not possible to pass multiple cookies with the `set-cookie` header via AWS Lambda, so the `normalizeOutgoingHeaders` does not help with cookies in responses.

```
const handler = async function handler2(event, context) {
  const url = withQuery(event.path || event.rawPath, event.queryStringParameters || {});
  const method = event.httpMethod || event.requestContext?.http?.method || "get";
  if ("cookies" in event && event.cookies) {
    event.headers.cookie = event.cookies.join(",");
  }
  const r = await nitroApp.localCall({
    event,
    url,
    context,
    headers: normalizeIncomingHeaders(event.headers),
    method,
    query: event.queryStringParameters,
    body: event.body
  });
  return {
    statusCode: r.status,
    headers: normalizeOutgoingHeaders(r.headers),
    body: r.body.toString()
  };
};
function normalizeIncomingHeaders(headers) {
  return Object.fromEntries(Object.entries(headers).map(([key, value]) => [key.toLowerCase(), value]));
}
function normalizeOutgoingHeaders(headers) {
  return Object.fromEntries(Object.entries(headers).map(([k, v]) => [k, Array.isArray(v) ? v.join(",") : v]));
}
```

This results in issues like the following for some requests (e.g. adapted @auth0/nextjs code for SSR-compatible serverless auth), as only the first cookie is actually set:
```
{"url":"/api/auth/callback?code=xxx&state=yyy","statusCode":400,"statusMessage":"checks.state argument is missing","message":"checks.state argument is missing","description":""}
```

### Additional context

Documentation here: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.v2

I was able to fix the issue with this hack to the handler, which is very inelegant:
```
const handler = async function handler2(event, context) {
  const url = withQuery(event.path || event.rawPath, event.queryStringParameters || {});
  const method = event.httpMethod || event.requestContext?.http?.method || "get";
  if ("cookies" in event && event.cookies) {
    event.headers.cookie = event.cookies.join("; ");
  }
  const r = await nitroApp.localCall({
    event,
    url,
    context,
    headers: normalizeIncomingHeaders(event.headers),
    method,
    query: event.queryStringParameters,
    body: event.body
  });
  const headers = normalizeOutgoingHeaders(r.headers);
  return Object.assign({
    statusCode: r.status,
    body: r.body.toString()
  }, headers);
};
function normalizeIncomingHeaders(headers) {
  return Object.fromEntries(Object.entries(headers).map(([key, value]) => [key.toLowerCase(), value]));
}
function normalizeOutgoingHeaders(headers) {
  const obj = {
    headers: {},
    cookies: []
  };
  for (const k in headers) {
    const v = headers[k];
    if (k.toLowerCase() === 'set-cookie') {
      if (Array.isArray(v)) {
        obj.cookies = obj.cookies.concat(v);
      } else {
        obj.cookies = obj.cookies.concat(v.split(','));
      }
    } else {
      obj.headers[k] = v;
    }
  }
  return obj;
}
```

Note that I also changed the `.join()` to use `; ` because the `cookie` packages expects that format.

Apologies if this should be on `nuxt/framework` instead. This took me many hours to find out, and I just wanted to make sure it was documented somewhere to eventually get fixed, so that my hack is no longer required as a post-build step.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #245

### ❓ Type of change

### 📚 Description

This PR returns an array of cookies to set from aws-lambda preset rather than keeping them on the headers. More info in [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.v2).

@brtinney - would be grateful if you could confirm this resolves the issue for you.

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/aws-lambda.ts
+++ b/src/runtime/entries/aws-lambda.ts
@@ -25,7 +25,11 @@ export const handler = async function handler (event: Event, context: Context): 
     body: event.body // TODO: handle event.isBase64Encoded
   })
 
+  const outgoingCookies = r.headers['set-cookie']
+  const cookies = Array.isArray(outgoingCookies) ? outgoingCookies : outgoingCookies?.split(',') || []
+
   return {
+    cookies,
     statusCode: r.status,
     headers: normalizeOutgoingHeaders(r.headers),
     body: r.body.toString()
@@ -37,5 +41,7 @@ function normalizeIncomingHeaders (headers: APIGatewayProxyEventHeaders) {
 }
 
 function normalizeOutgoingHeaders (headers: Record<string, string | string[] | undefined>) {
-  return Object.fromEntries(Object.entries(headers).map(([k, v]) => [k, Array.isArray(v) ? v.join(',') : v!]))
+  return Object.fromEntries(Object.entries(headers)
+    .filter(([key]) => !['set-cookie'].includes(key))
+    .map(([k, v]) => [k, Array.isArray(v) ? v.join(',') : v!]))
 }


brtinney:
@danielroe: It appears to properly set the cookies now, although I still have to use: `event.headers.cookie = event.cookies.join("; ");` over `event.headers.cookie = event.cookies.join(",");` in order to use the `cookie` library to parse them from the event, as the (more common?) convention is to use `;` to separate them.

That said, I'm not sure if I can necessarily insist that `nitro` should be using `;` to separate the cookies. I can at least get away with doing `parse((req.headers.cookie || '').replace(/,/g, ';'))` in my actual code instead of having to modify the built `mjs` for that particular issue. (Which should be safe since neither `,` or `;` are supposed to actually appear in cookies....)

Thanks!

danielroe:
@brtinney I think the other issue should be resolved with https://github.com/unjs/nitro/pull/356.

pi0 (approved):
Finally hadn't a chance to test but hope change is working

datake914:
@pi0 
I have tried this and it doesn't work.
I configured Amazon API Gateway proxy integration to work with an AWS Lambda function. When I call my endpoint, I receive a configuration error and an HTTP 502 status code.

```
Execution failed due to configuration error: Malformed Lambda proxy response
```

This is due to the fact that cookies are returned in the response object.
For API Gateway to handle a Lambda function's response, the function must return output according to the following JSON format. Additional fields are not allowed.

```
{
    "isBase64Encoded": true|false,
    "statusCode": httpStatusCode,
    "headers": { "headerName": "headerValue", ... },
    "body": "..."
}
```

PR #387, opened 2022-08-05
fix(externals): resolve relative imports when checking for externals

### LINKED ISSUE #386: npm run generate breaks when import contains re-export `export { default as xxx } from 'ooo'`
<table><td>

&nbsp;&nbsp;&nbsp;Forwarded from downstream issue:
- https://github.com/nuxt/nuxt.js/issues/14125 by @mmis1000

</td></table><br>

### Environment

Nuxt project info: 

- Operating System: `Darwin`
- Node Version:     `v14.18.2`
- Nuxt Version:     `3.0.0-rc.3`
- Package Manager:  `npm@6.14.17`
- Builder:          `vite`
- User Config:      `typescript`, `publicRuntimeConfig`, `buildModules`, `alias`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `@pinia/nuxt@0.1.9`

### Reproduction

https://codesandbox.io/s/angry-johnson-3h59db?file=/app.vue

Run `npm run generate`  from the terminal.

```html
<template>
  <div>
    <NuxtWelcome />
  </div>
</template>
<script>
// the swiper js package has re-export in entry point
import { Pagination } from "swiper";
console.log(Pagination);

export default {};
</script>
```

### Describe the bug

If package you used contains re-export syntex, then `npm run generate` didn't work.

### Additional context

The import was incorrectly resolved to `import Pagination from './modules/pagination/pagination.js';` in the `/sandbox/.nuxt/prerender/chunks/app/server.mjs` . and thus break the build

### Logs

```shell
ERROR  [nuxt] [request error] Cannot find module '/sandbox/.nuxt/prerender/chunks/app/modules/pagination/pagination.js' imported from /sandbox/.nuxt/prerender/chunks/app/server.mjs
  at new NodeError (internal/errors.js:322:7)
  at finalizeResolution (internal/modules/esm/resolve.js:308:11)
  at moduleResolve (internal/modules/esm/resolve.js:731:10)
  at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:842:11)
  at Loader.resolve (internal/modules/esm/loader.js:89:40)
  at Loader.getModuleJob (internal/modules/esm/loader.js:242:28)
  at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:76:40)
  at link (internal/modules/esm/module_job.js:75:36)
```
```

### PR DESCRIPTION
**Pull request by danielroe**
### 🔗 Linked issue

resolves nuxt/framework#386, https://github.com/nuxt/nuxt.js/issues/14125

### ❓ Type of change

### 📚 Description

Previously relative imports from modules were skipped/inlined, resulting in this behaviour:

```ts
// app.vue
import { Pagination } from "swiper";
// swiper is an external, but this becomes, in `server.mjs`:
import Pagination from './modules/pagination/pagination.js';
// with this PR:
import Pagination from 'file:///my/project/node_modules/.pnpm/swiper@8.3.2/node_modules/swiper/modules/pagination/pagination.js';
```

### 📝 Checklist


pi0:
@danielroe Do you have any update on this? 🙏🏼 Need to finalize state of this PR :)

pi0:
Checking back on this issue, interestingly it is only reproducible on external repo + using custom `srcDir` + `externals.trace=false` (enabled for prerender). 

Changes by your PR fixes the issue and generate import with right cde but I'm afraid we are missing some other root causes. One of them is that swiper is detected by mlly as invalid esm import (hense auto inlined) while it is valid.

danielroe:
Nice finds. Would be very happy to look into those issues as well 👍 Do you think that means we shouldn't merge this PR, or that we should both merge it and resolve them?

pi0:
I would suggest first fixing the root causes and see if we can make a fix limited to `trace: false` (which is actually causing problem) instead of doing modification to originalId.

pi0:
Fix for swiper wrong detection: https://github.com/unjs/mlly/commit/8e866c3459c04163bb59dd56469f8468add899b3

pi0:
Thanks for confirming! I am always up to revising this when needed (which i hopefully by following up the inline-first approach will be even less likely!

Issue #378, opened 2022-07-29
Error building with `cloudflare`, `vercel-edge` or `netlify-edge` presets


maoosi:
### Environment

- Node.js v16.14.2
- Nuxt 3.0.0-rc.6
- Nitropack 0.4.12

### Reproduction

[Click here to access reproduction on Stackblitz](https://stackblitz.com/edit/nitro-preset-cloudflare-issue?file=src/app.vue), then run the below command:

```shell
# working as expected
npm run build

# build errors
NITRO_PRESET=cloudflare npm run build
NITRO_PRESET=vercel-edge npm run build
NITRO_PRESET=netlify-edge npm run build
```

### Describe the bug

Nuxt3 build is failing when running with edge-related presets such as:

- `NITRO_PRESET=cloudflare`
- `NITRO_PRESET=vercel-edge`
- `NITRO_PRESET=netlify-edge`

### Additional context

Probably related to:

- https://github.com/unjs/nitro/issues/280.
- https://github.com/unjs/nitro/issues/293

### Logs

Example logs when using `NITRO_PRESET=cloudflare`:

```sh
ERROR  Rollup error: Cannot resolve "/home/projects/nitro-preset-cloudflare-issue/node_modules/unenv/runtime/node/stream/index/web.cjs" from "/home/projects/nitro-preset-cloudflare-issue/node_modules/cross-undici-fetch/dist/create-node-ponyfill.js" and externals are not allowed!

ERROR  Cannot resolve "/home/projects/nitro-preset-cloudflare-issue/node_modules/unenv/runtime/node/stream/index/web.cjs" from "/home/projects/nitro-preset-cloudflare-issue/node_modules/cross-undici-fetch/dist/create-node-ponyfill.js" and externals are not allowed!
```

warflash:
#293 is probably also be related. Seems like unfortunately the cloudflare preset is pretty dead at the moment

maoosi:
I've edited my initial post to include #293. I also noticed the same issue with `NITRO_PRESET=vercel-edge`, so this issue is probably affecting all edge-related presets.

warflash:
@maoosi Did you get a chance to verify the fix yet? I just updated my nuxt brige app to use nitro 0.4.18 and am still running into the same rollup `externals not allowed` error on edge presets. Wondering if I'm experiencing something unrelated or if the error still persists

danielroe:
[...]

PR #388, opened 2022-08-05
fix: sort aliases to ensure priority is given to more specific aliases

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/378, see also https://github.com/unjs/unenv/pull/45

### ❓ Type of change

### 📚 Description

A couple of issues have come up recently about alias ordering. This PR applies a fix to order aliases by specificity (e.g. number of slashes) meaning we should always have more specific alias in front of general (`fs/promises` in front of `fs`).

The upstream PR to unenv resolves it more specifically within the unenv presets but this is still worth merging here as there are other potential aliases involved.

### 📝 Checklist

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -135,7 +135,7 @@ export function readPackageJson (
 export function resolveAliases (_aliases: Record<string, string>) {
   // Sort aliases from specific to general (ie. fs/promises before fs)
   const aliases = Object.fromEntries(Object.entries(_aliases).sort(([a], [b]) =>
-    scoreAlias(b) - scoreAlias(a)
+    (b.split('/').length - a.split('/').length) || (b.length - a.length)
   ))
   // Resolve alias values in relation to each other
   for (const key in aliases) {
@@ -149,9 +149,4 @@ export function resolveAliases (_aliases: Record<string, string>) {
     }
   }
   return aliases
-}
-
-const scoreAlias = (alias: string) => {
-  const segments = alias.split('/')
-  return segments.length + (segments[segments.length - 1].length / 1000)
 }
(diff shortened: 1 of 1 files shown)


src/utils/index.ts
```diff
@@ -132,7 +132,12 @@ export function readPackageJson (
   }
 }
 
-export function resolveAliases (aliases: Record<string, string>) {
+export function resolveAliases (_aliases: Record<string, string>) {
+  // Sort aliases from specific to general (ie. fs/promises before fs)
+  const aliases = Object.fromEntries(Object.entries(_aliases).sort(([a], [b]) => {
+    return b.split('/').length - a.split('/').length
```
pi0, line 138: Maybe fallback to last subpath length if both have same segments?

danielroe: Why?

pi0: I mean if there is `foo/bar` `foo/bar-123`, they have same length of 2. Then we can compare them by their length. (In case alias is not a file but prefix)

src/utils/index.ts
```diff
@@ -134,9 +134,9 @@ export function readPackageJson (
 
 export function resolveAliases (_aliases: Record<string, string>) {
   // Sort aliases from specific to general (ie. fs/promises before fs)
-  const aliases = Object.fromEntries(Object.entries(_aliases).sort(([a], [b]) => {
-    return b.split('/').length - a.split('/').length
```
pi0: `(b.split('/').length - a.split('/').length) || (b.length - a.length)`

This should have same effect :D

pi0: `['a/foo', 'a/foo-bar'].sort((a, b) => b.split('/').length - a.split('/').length || b.length - a.length)`

pi0 (approved):

PR #431, opened 2022-08-22
fix(externals)!: fall back to `mlly` resolver in more cases

danielroe:
### 🔗 Linked issue

context: https://github.com/nuxt/framework/issues/6733

### ❓ Type of change

### 📚 Description

In the linked issue above, the presence of a _directory_ `express` was skipping the mlly resolution process as the path existed. I've added a couple of additional conditions here:

1. resolve with `mlly` if id is not absolute (`express.js`) won't map to `~/express.js` in node terms
2. resolve with `mlly` if id is a _directory_ - as directory imports won't work in esm and we need to resolve `/index.js`
3. only call `normalizeid` on absolute paths as otherwise (if `mlly` absolutely can't resolve it) we can end up with something like `file://express` 

### 📝 Checklist


pi0:
Would you please provide a (nitro) reproduction? 🙏🏼

danielroe:
I believe the fix makes sense on its own for the reasons mentioned in the description.

But here you go: https://stackblitz.com/edit/nuxt-starter-qlsvtz.

src/rollup/plugins/externals.ts
```diff
@@ -64,7 +65,7 @@ export function externals (opts: NodeExternalsOptions): Plugin {
       const resolved = await this.resolve(originalId, importer, { ...options, skipSelf: true }) || { id }
 
       // Try resolving with mlly as fallback
-      if (!existsSync(resolved.id)) {
+      if (!isAbsolute(resolved.id) || !existsSync(resolved.id) || await isDirectory(resolved.id)) {
         resolved.id = await _resolve(resolved.id).catch(() => resolved.id)
```
pi0, line 69: Calling resolver on non-absolute paths is probably also wrong (URL from _resolve is node_modules). Can we instead simply test `id` against npm package name regex and try resolving?

src/rollup/plugins/externals.ts
```diff
@@ -77,7 +78,7 @@ export function externals (opts: NodeExternalsOptions): Plugin {
       if (opts.trace === false) {
         return {
           ...resolved,
-          id: normalizeid(resolved.id),
+          id: isAbsolute(resolved.id) ? normalizeid(resolved.id) : resolved.id,
```
pi0, line 81: Here we could also have explicit check for possible npm import name before normalizing (or it could be also handled from mlly.normalizeid to not wrongly add file protocol for them)

danielroe: I checked normalizeid but its description says that it is meant to always add a protocol.

src/rollup/plugins/externals.ts
danielroe: What kind of regexp did you have in mind?

danielroe: Ah. The edge case is where the input does _not_ look like an absolute path but does exist in the FS, e.g. `express` when there is a folder called `express`. I think it is sufficient for the linked bug to check whether it is a directory or not.

But I think we shouldn't just check for directory. After all, we _require_ an absolute path and we really should re-resolve it if it isn't one at this point. If we get a result like `express/index.js` this may be a file on the disk but we can't import the file that way.

src/rollup/plugins/externals.ts
danielroe: This really should never ever be hit, but if it _is_ hit it is guaranteed to be a bug. A non-absolute path at this point will become `file://someid` which is an invalid URL and cannot be imported. So I think the check is right.

pi0: It still doesn't makes sense to me why we add a condition for something is not supposed to happen and hiding the issue.. Anyway not worth discussing returning as is for non absolute is okay.

pi0 (approved):
LGTM to include in next semver-major since this change is risky somehow.

For next PRs, please use a minimal **nitro** reproduction to help me review faster 🙏🏼

PR #443, opened 2022-08-26
feat: use CachedEventHandler options in nitro routes config

LouisHaftmann:
### 🔗 Linked issue

#442

### ❓ Type of change

### 📚 Description

This applies all nitro route options to the CachedEventHandler instead of just checking and enabling `swr`

Resolves #442

### 📝 Checklist


src/types/nitro.ts
```diff
@@ -61,10 +63,7 @@ export interface NitroConfig extends DeepPartial<NitroOptions> {
   extends?: string | string[] | NitroPreset
 }
 
-export interface NitroRouteOption {
-  swr?: boolean | number
-  redirect?: string
-}
+export type NitroRouteOption = Omit<CacheOptions<ResponseCacheEntry<unknown>>, 'getKey'>
```
pi0, line 66: I think we can support more options as `{ cache }`. Route Options is not only for caching finally.

LouisHaftmann: Of course there will be more options in the future but swr is currently the only used option. Maybe instead of all the cache options we could just use `maxAge` and `swr` or make the `swr` option a string/number (eg. `'10 min'`/`600`, see this [Nuxt discussion](https://github.com/nuxt/framework/discussions/560))

pi0: It was kinda experimental and indeed we can keep limited shortcuts top level but for the purpose of this PR to allow all cache related options, namespace seems more safe against naming conflict and semantical way of writing rules.

danielroe (approved):

pi0:
Thanks for the nice PR @LouisHaftmann. Both docs and shortcuts supported are updated in the main supporting `cache` options as well as `swr: number` shortcut. Because of merge conflict, i will add docs from main.

pi0:
https://github.com/unjs/nitro/commit/221c7478f90a1e3d69be03f3df8bc9df5785e574

PR #450, opened 2022-08-31
feat: allow specifying wasm plugin options

sisou:
I'm using nitro in Nuxt 3 and I found that the default options only work when the WASM file is smaller than the default inlining-threshold. Otherwise the Nuxt3 NodeJS server environment (and other server targets I'd imagine) chokes on the wasm plugin using `require()`calls (how dare them use `require` in NodeJS! :D). And even if that would be fixed upstream in the `@rollup/plugin-wasm` package, it's still nice to be able to change the default options here.

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -3,6 +3,7 @@ import { dirname, join, normalize, relative, resolve } from 'pathe'
 import type { InputOptions, OutputOptions } from 'rollup'
 import { defu } from 'defu'
 import { terser } from 'rollup-plugin-terser'
+import type { RollupWasmOptions } from '@rollup/plugin-wasm'
 import commonjs from '@rollup/plugin-commonjs'
 import { nodeResolve } from '@rollup/plugin-node-resolve'
 import alias from '@rollup/plugin-alias'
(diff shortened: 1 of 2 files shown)


src/rollup/config.ts
```diff
@@ -130,7 +130,8 @@ export const getRollupConfig = (nitro: Nitro) => {
 
   // WASM import support
   if (nitro.options.experimental.wasm) {
-    rollupConfig.plugins.push(wasmPlugin())
+    const options = typeof nitro.options.experimental.wasm === 'object' ? nitro.options.experimental.wasm : {}
```
pi0, line 133:
```suggestion
    const options = { ...nitro.options.experimental.wasm }
```

sisou:
Jaha, I thought so, too. But spreading a boolean is not allowed, sadly.

pi0:
Try this `{ ...true }`. JS is land of wonders :D

sisou:
I tried that too before and saw that it works just fine. Only TS is not allowing it. I don't know if you accept `// @ts-expect-error` in your codebase?

src/rollup/config.ts
```diff
@@ -130,7 +130,8 @@ export const getRollupConfig = (nitro: Nitro) => {
 
   // WASM import support
   if (nitro.options.experimental.wasm) {
-    rollupConfig.plugins.push(wasmPlugin())
+    const options = { ...nitro.options.experimental.wasm }
```
pi0, line 133:
```suggestion
    const options = { ...nitro.options.experimental.wasm as any }
```

pi0:
We can also use `as any|[Type]`. The main reason is that later it makes it easier to assign some default without `?:` branches.

PR #454, opened 2022-08-31
feat: expanded devServer's versatility

richardeschloss:
This change suggests moving the `createDevServer`'s `app` initialization to a function parameter which defaults to `createApp`. The reason for this is there may be large projects that are structured like:

"Massive express-based app --> build/start Nuxt"

For those massive apps to migrate to Nuxt3, where they may already have their own `app` and `app.use` stuff, they'll probably want to provide Nitro their app instance so that they can app.use the Nitro things, like [publicAssets](https://github.com/unjs/nitro/blob/6bfbb22d569bd6ab6bbb844805d6d8e93b343460/src/dev/server.ts#L115). 

Then, massive projects can simply do: `createDevServer(nitro, app)`, and most Nuxt3 things should work assuming there are no naming conflicts. 
 
### 🔗 Linked issue

#461 

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -67,7 +67,7 @@ async function killWorker (worker?: NitroWorker) {
   }
 }
 
-export function createDevServer (nitro: Nitro): NitroDevServer {
+export function createDevServer (nitro: Nitro, app: App = createApp()): NitroDevServer {
   // Worker
   const workerEntry = resolve(nitro.options.output.dir, nitro.options.output.serverDir, 'index.mjs')
 
@@ -98,9 +98,6 @@ export function createDevServer (nitro: Nitro): NitroDevServer {
     return reloadPromise
   })
   nitro.hooks.hook('dev:reload', reload)
-
-  // App
-  const app = createApp()
 
   // Dev-only handlers
   for (const handler of nitro.options.devHandlers) {


pi0:
Thanks for PR @richardeschloss but h3 is the building block for nitro output in both dev and production and it has support for express middleware. Adding the ability to replace h3 app with another interface makes inconsistent behavior when we internally use it.

Here is an example sandbox to register Express as a Nitro middleware: https://stackblitz.com/edit/github-cuverk?file=middleware%2Fexpress.ts


```ts
// middleware/express.ts
import express from 'express';
import { toEventHandler } from 'h3';

const app = express();

app.use('/test.json', (req, res) => res.json({ express: 'works' }));

export default toEventHandler(app.handle.bind(app));
```

richardeschloss:
> Thanks for PR @richardeschloss but h3 is the building block for nitro output in both dev and production and it has support for express middleware. Adding the ability to replace h3 app with another interface makes inconsistent behavior when we internally use it.
> 
> Here is an example sandbox to register Express as a Nitro middleware: https://stackblitz.com/edit/github-cuverk?file=middleware%2Fexpress.ts
> 
> ```ts
> // middleware/express.ts
> import express from 'express';
> import { toEventHandler } from 'h3';
> 
> const app = express();
> 
> app.use('/test.json', (req, res) => res.json({ express: 'works' }));
> 
> export default toEventHandler(app.handle.bind(app));
> ```

Ok, I appreciate your response!

PR #482, opened 2022-09-08
fix(node-server): default production host to `localhost`

KubesDavid:
### ❓ Type of change

### 📚 Description

Since the host address is hard coded to `0.0.0.0` (IPv4), it is not possible to connect using `[::1]` (IPv6). Setting `localhost` will allow connection via both IPv4 and IPv6.

### 📝 Checklist


pi0:
Thanks for the PR @KubesDavid. It makes sense to listen to ipv6 as well but using `localhost`, will makes issues when we want to expose the port externally like inside containers.

`localhost` resolves to `127.0.0.1  for IPV4 and `::1` for IPv4 and they listen loopback only. `0.0.0.0` and `::0` have special meaning to listen to **all interfaces** and it is for default host.

PR #494, opened 2022-09-13
fix: properly serve `/index.html` static asset

danielroe:
### 🔗 Linked issue

#7467

### ❓ Type of change

### 📚 Description

`/` can't be served via assets because it becomes `//index.html`, meaning we were totally skipping static `/` routes. In fact, we probably want to use `joinURL` so that the trailing slash status is not taken into account when joining with `index.html`.

### 📝 Checklist


michealroberts:
Thanks @danielroe Legend! 🙌

PR #495, opened 2022-09-13
fix: use the prerendering preset even if NITRO_PRESET is present

tobiasdiez:
### ❓ Type of change

### 📚 Description

If one explicitly changes the preset using the `NITRO_PRESET` variable, then prerendering does no longer work since the nitro prerender server then uses the preset specified in that variable instead of the correct `nitro-prerender` preset.
This is a result of the changes in https://github.com/unjs/nitro/commit/92d711fe936fda0ff877c23d8a0d73ed4ea4adc4

### 📝 Checklist


pi0:
Thanks for noticing issue @tobiasdiez and PR. I've fixed it differently to respect config overrides over environment variable (https://github.com/unjs/nitro/commit/82a2b1e9139084dd19f74bd4789850addc3fb66e#diff-2123f97b7bf9480a61717c843425e483e8d97819592e8c04e4e149eb65b451e8R91)

PR #500, opened 2022-09-14
fix(vercel, azure): serve prerendered routes statically

danielroe:
### ❓ Type of change

### 📚 Description

Current hybrid vercel deployments will always server-render prerendered HTML (though you can get the static html by adding `/index.html` to the URL). Vercel provides the `overrides` property to allow us to map clean URLs (without `.html` or `index.html` suffix) to static files: https://vercel.com/docs/build-output-api/v3#build-output-configuration/supported-properties/overrides. This PR implements that.

Might also be worth a consistent implementation of accessing which HTML files have been prerendered. (See [this line](https://github.com/unjs/nitropack/blob/ba39ffe9d5fe9516825ccbafd55c18211c78a424/src/presets/azure.ts#L66-L69) for how we were doing it in Azure - I could easily create a shared context that both presets could use - happy to pick this up.)

### 📝 Checklist

### DIFF
--- a/src/presets/vercel.ts
+++ b/src/presets/vercel.ts
@@ -11,7 +11,7 @@ export const vercel = defineNitroPreset({
   entry: '#internal/nitro/entries/vercel',
   output: {
     dir: '{{ rootDir }}/.vercel/output',
-    serverDir: '{{ output.dir }}/functions/__nuxt.func',
+    serverDir: '{{ output.dir }}/functions/__nitro.func',
     publicDir: '{{ output.dir }}/static'
   },
   commands: {
@@ -47,7 +47,7 @@ export const vercelEdge = defineNitroPreset({
   entry: '#internal/nitro/entries/vercel-edge',
   output: {
     dir: '{{ rootDir }}/.vercel/output',
-    serverDir: '{{ output.dir }}/functions/__nuxt.func',
+    serverDir: '{{ output.dir }}/functions/__nitro.func',
     publicDir: '{{ output.dir }}/static'
   },
   commands: {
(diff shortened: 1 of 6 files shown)


src/presets/vercel.ts
```diff
@@ -11,38 +11,23 @@ export const vercel = defineNitroPreset({
   entry: '#internal/nitro/entries/vercel',
   output: {
     dir: '{{ rootDir }}/.vercel/output',
-    serverDir: '{{ output.dir }}/functions/index.func',
+    serverDir: '{{ output.dir }}/functions/__nuxt.func',
```
pi0, line 14: Would using index as function name make conflicts?

pi0: (it could be __nitro or __server in this case)

src/presets/vercel.ts
```diff
@@ -11,38 +11,23 @@ export const vercel = defineNitroPreset({
...
     publicDir: '{{ output.dir }}/static'
   },
   commands: {
     deploy: '',
     preview: ''
   },
   hooks: {
+    'prerender:generate' (route, nitro: any) {
+      nitro._routes = nitro._routes || new Set<string>()
```
pi0, line 23: We could move it to prerender.ts (but call it `_routeFiles` or something like this)

src/presets/vercel.ts
danielroe: Yes, it does; that's why I changed it. 👍

danielroe: oh of course :facepalm:

pi0 (approved):

PR #529, opened 2022-09-23
fix(externals): fallback traced files for minor/patch versions

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/526, https://github.com/nuxt/framework/issues/7729

### ❓ Type of change

### 📚 Description

Currently, if we trace file `a` from v 7.1 and file `b` from v 7.0 of a putative package, we discard `b` entirely. This PR changes the behaviour to copy across `b` from v 7.1. This isn't a perfect fix as we stil may have issues with major version differences, but it can resolve small differences, for example in https://github.com/nuxt/framework/issues/7729.

### 📝 Checklist


Atinux (approved):

pi0:
I'm reverting this PR. Mixing two files of two different package versions is not a nice idea!

At least for mentioned node-fetch native, we might find better solution and for rest of packages, think of supporting hosting for trace extraction.

danielroe:
Very happy for this to be reverted. It was only meant to be a hotfix. The current externals tracing implementation is fairly fragile; even differing patch versions of a package may cause 500 errors on deploy.

There are also other issues regarding differing _major_ versions.

Probably all of these can be solved by reworking the implementation.

pi0:
Indeed it is fraglie... Moved discussion to https://github.com/unjs/nitro/issues/572 would be nice if you can share your input on plan and also if any other open issues are relevant to link.

danielroe:
Also, just in case you missed it, this _didn't_ copy files across from two different versions.

PR #537, opened 2022-09-30
fix: retry writing traced files if there are conflicts

### LINKED ISSUE #536: resource busy or locked when copying traced files on windows
### Environment

windows GitHub Action CI, latest nitropack (https://github.com/nuxt/framework)

### Reproduction

https://github.com/nuxt/framework/actions/runs/3127025329/jobs/5073193013

### Describe the bug

nuxt build outputs:

```
Error: EBUSY: resource busy or locked, copyfile 'D:\a\framework\framework\node_modules\@vue\compiler-ssr\package.json' -> 'D:\a\framework\framework\test\fixtures\basic\.nuxt\ud0cz0\output\server\node_modules\@vue\compiler-ssr\package.json'
```

### Additional context

I've struggled to reproduce this locally (even running same test suite as in CI) so at least partially due to test environment. However, I think we can possibly fix root issue in nitropack based on the error.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #536

### ❓ Type of change

### 📚 Description

It seems that when multiple traced files in the same directory are copied on windows, in some circumstances, we have a race condition which causes a 'resource busy or locked' error. It's possible this is very much an edge case, such as when testing. But this change seems safe; it adds a small 2ms timeout and will retry a file up to 3 times, which will hopefully reduce any flakiness. I've [confirmed](https://github.com/nuxt/framework/pull/7927) (with `yarn patch`) that it does resolve the issue in framework repo.

(Another option would be synchronous file copying.)

### 📝 Checklist

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -146,3 +146,15 @@ export function resolveAliases (_aliases: Record<string, string>) {
   }
   return aliases
 }
+
+export async function retry (fn: () => Promise<void>, retries: number) {
+  let retry = 0
+  let error: any
+  while (retry++ < retries) {
+    try { return await fn() } catch (err) {
+      error = err
+      await new Promise(resolve => setTimeout(resolve, 2))
+    }
+  }
+  throw error
+}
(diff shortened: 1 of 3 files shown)


src/utils/index.ts
```diff
@@ -146,3 +146,15 @@ export function resolveAliases (_aliases: Record<string, string>) {
   }
   return aliases
 }
+
+export async function retry (fn: () => Promise<void>, retries: number) {
+  let retry = 0
+  let error: any
+  while (retry++ < retries) {
+    try { return await fn() } catch (err) {
+      error = err
```
pi0, line 155: We might always show error as warning. If it is happening too often not hiding the issue.

Issue #93, opened 2022-04-12
netlify-on demand preset


pi0:
https://github.com/nuxt/framework/issues/1539

Alternatively we can integrate with route caching rules for single preset

Rigo-m:
Also, `netlify_builder` preset lacks a [TTL Parameter](https://docs.netlify.com/configure-builds/on-demand-builders/#time-to-live-ttl)

PR #538, opened 2022-09-30
feat: support `redirect`, `headers`, and `cors` route rules

danielroe:
### 🔗 Linked issue

resolves #535, resolves #539, resolves #95

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This implements redirect route rules (including platform-specific support for vercel + netlify cdns).

It also implements `headers` and `cors` route rules. Apologies @pi0, I spotted too late you had self-assigned the `cors` route rule. 😬 Not my intention; happy to revert it - or feel free to amend PR.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #540, opened 2022-10-02
feat(netlify): incremental static regeneration + swr

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/93

### ❓ Type of change

### 📚 Description

This PR adapts the `netlify` preset to support the `swr`/`static` route rule for dynamic on-demand builder configuration.

For example, the following config would use builder functions only for the blog & articles routes. The blog routes would be generated only once and then persisted; the articles routes would be regenerated every 360 seconds.

```js
  routes: {
    '/blog/**': { static: true },
    '/articles/**': { swr: 360 }
  }
})
```

Here's a deployment using this PR: https://main--melodious-klepon-3a2ccb.netlify.app

~~**Note**: we pass the value of `swr` directly through to `ttl` but it is expressed in the Netlify docs in _seconds_ and we should likely specify what the number means - e.g. seconds or milliseconds, etc.~~

### 📝 Checklist


Atinux:
I believe seconds makes more sense personally :)

Atinux (approved):

src/runtime/entries/netlify.ts
```diff
@@ -0,0 +1,59 @@
...
+import { nitroApp } from '../app'
+import { useRuntimeConfig } from '../config'
+
+export const handler: Handler = async function handler (event, context) {
+  const config = useRuntimeConfig()
+  const routerOptions = createMatcher({ routes: config.nitro.routes })
+
+  const query = { ...event.queryStringParameters, ...event.multiValueQueryStringParameters }
+  const url = withQuery(event.path, query)
+  const routeOptions = routerOptions.lookup(url) || {}
```
pi0, line 15: Note: Same issue mentioned in https://github.com/unjs/nitro/pull/538#discussion_r993783351 applies here.

Issue #383, opened 2022-08-03
Expose `event.context.azure` for logging


tobiasdiez:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.16.0`
- Nuxt Version:     `3.0.0-rc.5`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `alias`, `nitro`, `ssr`, `meta`, `css`, `tailwindcss`, `modules`, `watch`, `runtimeConfig`, `serverHandlers`, `storybook`, `hooks`, `vite`
- Runtime Modules:  `@nuxtjs/tailwindcss@5.1.3`, `@pinia/nuxt@0.3.1`, `./modules/storybook`, `-runtime-compiler-module`
- Build Modules:    `-`

### Reproduction

Just have a function with `console.log("something")`.

### Describe the bug

Azure requires that logs are attached to a function run via `context.log` (instead of say `console.log`), see https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2-v3-v4-export%2Cv2-v3-v4-done%2Cv2%2Cv2-log-custom-telemetry%2Cv2-accessing-request-and-response%2Cwindows-setting-the-node-version#write-trace-output-to-logs.
But as far as I can see there is no wrapper to provide framework independent logging nor a way to get the azure context from within nitro/nuxt.

### Additional context

Maybe one could use the async hooks to introduce a `useLogger` method (similar to `useEvent` of https://github.com/unjs/nitro/pull/1546). This would allow to bound the logging output to the correct function invocation (see https://github.com/BrianRosamilia/azure-function-log-intercept/issues/4 for background).

PR #542, opened 2022-10-04
fix(azure-functions): redirect console to azure context for logging

mcremer-able:
### 🔗 Linked issue

#383 

### ❓ Type of change

### 📚 Description

At the moment there is no easy way to run nitro in an azure-funtions context and retain the ability to log out.
This pr redirects the built in `console.log(),console.warn(),console.info() and console.error()` to the azure provided logging context.
This adresses #383 ,but should be refined in the future to allow a more flexible log target

### 📝 Checklist


src/runtime/entries/azure-functions.ts
```diff
@@ -1,9 +1,28 @@
...
+/**
+ * intercepts all calls to console.* and redirects them to the azure context logger
+ */
+const interceptLogging = context =>
+  levels.forEach(m => higherOrderLog(m, context))
+
 export async function handle (context, req) {
   const url = '/' + (req.params.url || '')
-
+  interceptLogging(context)
```
danielroe, line 25: Will this have issues with logs being associated to the wrong request if there are multiple parallel requests?

If not (that is, if there is only one aggregated logging bucket), then do we need to run this on every incoming request or could we early return if the interceptor has already been run?

mcremer-able: As I understand it, it is a single logging bucket for the whole function app.

We could optimize and dont intercept ,if we are already intercepting.
The gains do not outweight the potentiell bugs, in my opinion, with the added complexity we get.

I rather intercept twice than miss a critical log.

danielroe:
Regarding my comment:

> Will this have issues with logs being associated to the wrong request if there are multiple parallel requests?

I'm checking with the Azure team to ensure that there are no downsides to this approach.

pi0:
Any updates on this?

mcremer-able:
ah F thats not what I wanted
Rebase was the wrong command to fetch upstream changes
Any git wizards that can help?

About the concerns that we associate logs of different requests with each other.
The Microsoft docs say that context.log does disriminate based on request, but console.log does **not**
see [docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2-v3-v4-export%2Cv2-v3-v4-done%2Cv2%2Cv2-log-custom-telemetry%2Cv2-accessing-request-and-response%2Cwindows-setting-the-node-version#write-trace-output-to-logs)

danielroe:
In that case I think we need a different approach, almost certainly not overwriting `console.log` but instead exposing azure utilities on `event.context.azure`, much like we are discussing doing with cloudflare.

mcremer-able:
I will close this as the long-term option will be an `event.context.azure` object with the necesarry properties.

For the short term I will continue to patch nitro in my CI with the `console.log` interceptor

tobiasdiez:
I agree that `event.context.azure` (or a more general purpose `event.context.log`) is a good idea and solves the issue for manual log statements. However, often a library that one uses logs output via `console.log` and this output is simply lost when deployed to azure, making it really hard to debug/investigate issues. Thus, there should also be a way to redirect output to the correct logger.

PR #583, opened 2022-10-15
feat: normalize route options and shortcuts

pi0:
Context: #538

By normalizing route options, we can support more shortcuts and also avoid duplicate logic per vendor for native support.

This PR normalizes 3 shortcuts:
- cors => converted to headers
- swr / static => converted to (new) cache options (rework of #443)
- [redirect is normalized]

Also fixes some issues with netlify and vercel generated rules

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -183,6 +183,19 @@ export async function loadOptions (configOverrides: NitroConfig = {}): Promise<N
     })
   }
 
+  // Normalize route rules
+  for (const rule of Object.values(options.routes)) {
+    if (rule.cors) {
+      rule.headers = {
+        ...rule.headers,
+        'access-control-allow-origin': '*',
+        'access-control-allowed-methods': '*',
+        'access-control-allow-headers': '*',
+        'access-control-max-age': '0'
+      }
+    }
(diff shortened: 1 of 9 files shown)


src/options.ts
```diff
@@ -183,6 +183,19 @@ export async function loadOptions (configOverrides: NitroConfig = {}): Promise<N
...
+  // Normalize route rules
+  for (const rule of Object.values(options.routes)) {
+    if (rule.cors) {
+      rule.headers = {
+        ...rule.headers,
+        'access-control-allow-origin': '*',
+        'access-control-allowed-methods': '*',
+        'access-control-allow-headers': '*',
+        'access-control-max-age': '0'
+      }
```
danielroe, line 195:
```suggestion
      rule.headers = {
        'access-control-allow-origin': '*',
        'access-control-allowed-methods': '*',
        'access-control-allow-headers': '*',
        'access-control-max-age': '0',
        ...rule.headers
      }
```

danielroe: I think allowing headers to override cors defaults is likely better, so users can edit the defaults

PR #589, opened 2022-10-16
fix(azure): raw body no longer returns parsed json

tobiasdiez:
### 🔗 Linked issue

None

### ❓ Type of change

### 📚 Description

Azure functions with `httpTrigger` automatically JSON-parse the body. This is not explicitly mentioned in the docs (at least I couldn't find it), but in one of [the examples](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=in-process%2Cfunctionsv2&pivots=programming-language-javascript#example) they use `req.body.name` showing that `req.body` is indeed an object. This then leads to problems when used with h3's `useRawBody` method, which actually returns the object instead of the stringy body when deployed to azure.

This is fixed by using `req.rawBody` in the azure handler. It's hard to find documentation on this property, but `rawBody` is used in an official test file:
https://github.com/Azure/azure-functions-host/blob/71ecbb2c303214f96d7e17310681fd717180bdbb/test/WebJobs.Script.Tests.Integration/TestScripts/Node/HttpTrigger/index.js#L13

Example `console.log(req)` in the handler:
```
....
[api] [2022-10-16T13:33:53.998Z]   body: {
[api] [2022-10-16T13:33:53.998Z]     query: 'query MeE2ENotLoggedIn {\n  me {\n    id\n  }\n}',
[api] [2022-10-16T13:33:53.998Z]     operationName: 'MeE2ENotLoggedIn'
[api] [2022-10-16T13:33:53.998Z]   },
[api] [2022-10-16T13:33:53.998Z]   rawBody: '{"query":"query MeE2ENotLoggedIn {\\n  me {\\n    id\\n  }\\n}","operationName":"MeE2ENotLoggedIn"}'
....
```

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/azure-functions.ts
+++ b/src/runtime/entries/azure-functions.ts
@@ -8,6 +8,7 @@ export async function handle (context, req) {
     url,
     headers: req.headers,
     method: req.method,
+    // https://github.com/Azure/azure-functions-host/issues/293
     body: req.rawBody
   })
 
--- a/src/runtime/entries/azure-functions.ts
+++ b/src/runtime/entries/azure-functions.ts
@@ -8,7 +8,7 @@ export async function handle (context, req) {
     url,
     headers: req.headers,
     method: req.method,
-    body: req.body
+    body: req.rawBody
   })
 
   context.res = {
--- a/src/runtime/entries/azure.ts
+++ b/src/runtime/entries/azure.ts
@@ -18,6 +18,7 @@ export async function handle (context, req) {
     url,
     headers: req.headers,
     method: req.method,
+    // https://github.com/Azure/azure-functions-host/issues/293
     body: req.rawBody
   })
 
--- a/src/runtime/entries/azure.ts
+++ b/src/runtime/entries/azure.ts
@@ -18,7 +18,7 @@ export async function handle (context, req) {
     url,
     headers: req.headers,
     method: req.method,
-    body: req.body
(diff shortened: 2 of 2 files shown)


src/runtime/entries/azure-functions.ts
```diff
@@ -8,7 +8,7 @@ export async function handle (context, req) {
     url,
     headers: req.headers,
     method: req.method,
-    body: req.body
+    body: req.rawBody
```
danielroe, line 11:
```suggestion
    // https://github.com/Azure/azure-functions-host/issues/293
    body: req.rawBody
```

src/runtime/entries/azure.ts
```diff
@@ -18,7 +18,7 @@ export async function handle (context, req) {
     url,
     headers: req.headers,
     method: req.method,
-    body: req.body
+    body: req.rawBody
```
danielroe, line 21:
```suggestion
    // https://github.com/Azure/azure-functions-host/issues/293
    body: req.rawBody
```

danielroe (approved):

danielroe (approved):
Thanks for finding! ❤️

brendonmatos:
@danielroe Hi!
Should useRawBody always return the body in string?

if yes, the same is happening in cloudflare deploys

Issue #422, opened 2022-08-15
Deno preset


pi0:
(not same as `deno-deploy`)

We can have a new preset for deno to make output working with `deno` CLI. 

The main catch, is that we need to transform imports to nitro. We could either continue working on [rollup-plugin-node-deno](https://github.com/unjs/rollup-plugin-node-deno) or inline transform into preset.

bartlomieju:
Hi, Bartek from the Deno team here.

What is the problem with imports you are facing? We have been working on improving our Node compatibility significantly in the recent weeks and it feels like there might not be a need to transform imports anymore.

Let me know if there's something the Deno team could help you with.

so1ve:
It seems that the readAsset function is not implemented?

oceaann:
> > Hi, Bartek from the Deno team here.
> > What is the problem with imports you are facing? We have been working on improving our Node compatibility significantly in the recent weeks and it feels like there might not be a need to transform imports anymore.
> > Let me know if there's something the Deno team could help you with.
> 
> I tried running `deno task dev`
> 
> ```js
> error: Dir F:\geohints\geohints\node_modules\.deno\unimport@3.0.6\node_modules\unimport\dist to F:\geohints\geohints\node_modules\.deno\unimport@3.0.6_1\node_modules\unimport\dist
> 
> Caused by:
>     0: Hard linking F:\geohints\geohints\node_modules\.deno\unimport@3.0.6\node_modules\unimport\dist\addons.cjs to F:\geohints\geohints\node_modules\.deno\unimport@3.0.6_1\node_modules\unimport\dist\addons.cjs
>     1: Incorrect function. (os error 1)
> ```


For some reason there is `\unimport@3.0.6\` trying to import things from `\unimport@3.0.6_1\` -  @bartlomieju I think its the reason why it doesn't work

bartlomieju:
@oceaann this issue should be solved by https://github.com/denoland/deno/pull/18824

oceaann:
> @oceaann this issue should be solved by [denoland/deno#18824](https://github.com/denoland/deno/pull/18824)

I used `deno upgrade --canary` and exact same error (after running `deno task dev` on nuxt project
```js
error: Dir F:\geohints\geohints\node_modules\.deno\unimport@3.0.6\node_modules\unimport\dist to F:\geohints\geohints\node_modules\.deno\unimport@3.0.6_1\node_modules\unimport\dist

Caused by:
[...]

PR #592, opened 2022-10-16
feat: add deno presets and entries

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/422

### ❓ Type of change

### 📚 Description

This PR adds a deno preset + deno server entry. You can see it in assembled action in a Nuxt project in https://github.com/danielroe/nuxt-deno - just run `pnpm i && pnpm build && pnpm preview`.

### 📝 Checklist


src/runtime/entries/deno-server.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { handler } from './deno'
+import { useRuntimeConfig } from '#internal/nitro'
+
+const cert = process.env.NITRO_SSL_CERT
+const key = process.env.NITRO_SSL_KEY
+const port = destr(process.env.NITRO_PORT || process.env.PORT) || 3e3
+const hostname = process.env.NITRO_HOST || process.env.HOST
+
+if (cert && key) {
```
bartlomieju, line 16: Is the idea is to get this file being written in "idomatic Deno"? If so I can surely make some suggestions.

Is there a reason to import `process` polyfill other than registering `unhandledRejection` and `uncaughtException` events? Does any of the imported code depend on it?

danielroe: Many of the Vue dependencies rely on `process.env.NODE_ENV` to determine whether to return prod/dev code. This of course needs to be updated there to use production/development import conditions but for now it's what we're working with.

danielroe: We also provide `import.meta.env` via our own polyfill from `process.env`

src/presets/deno.ts
```diff
@@ -0,0 +1,35 @@
+import deno from 'rollup-plugin-node-deno'
+import MagicString from 'magic-string'
+import { defineNitroPreset } from '../preset'
+
+export const denoServer = defineNitroPreset({
+  extends: 'node-server',
+  entry: '#internal/nitro/entries/deno-server',
+  commands: {
+    preview: 'deno run --unstable --allow-net --allow-read --allow-env ./server/index.mjs'
```
bartlomieju, line 9: Where can I find `./server/index.mjs` file?

danielroe: When compiled, you can find it in `.output/server/index.mjs`

src/presets/deno.ts
```diff
@@ -1,7 +1,13 @@
+import { isAbsolute } from "node:path";
```
pi0, line 1: use pathe?

src/presets/deno.ts
```diff
@@ -1,7 +1,13 @@
...
+import { builtinModules } from "node:module";
+
+import MagicString from "magic-string";
+import { findStaticImports } from "mlly";
+import inject from "@rollup/plugin-inject";
 import { defineNitroPreset } from "../preset";
 
 export const deno = defineNitroPreset({
```
pi0, line 9: Can you please rename it to denoDeploy with an alias to deno?

danielroe:
Okay, we're ready to go on deno. Sorry for the delay.

If you want to test with a Nuxt project, make sure to transpile `vue` or deno will pick the client version when running the project.

```
// https://nuxt.com/docs/api/configuration/nuxt-config
  build: {
    transpile: [/@vue/, /\bvue\b/]
  }
})
```

**Note**: deno doesn't use the `node_modules` folder we create, but turning off `trace` means we include full path in the output. Neither is great. There may be an improvement we can do here to turn tracing off but leave ids unchanged, or alternatively we can leave `node_modules` in place - it may be possible for deno to respect what we create via https://deno.land/manual@v1.30.2/node/npm_specifiers#--node-modules-dir-flag or similar.

src/presets/deno.ts
```diff
@@ -12,11 +18,119 @@ export const deno = defineNitroPreset({
...
+        modules: {
+          process: "process",
+          global: "global",
+          Buffer: ["buffer", "Buffer"],
+          setTimeout: ["timers", "setTimeout"],
+          clearTimeout: ["timers", "clearTimeout"],
+          setInterval: ["timers", "setInterval"],
+          clearInterval: ["timers", "clearInterval"],
+          setImmediate: ["timers", "setImmediate"],
+          clearImmediate: ["timers", "clearImmediate"],
```
pi0, line 52: It will use "native" process, timers, buffer correct? (not browserify)

bartlomieju:
@danielroe in the next minor release of Deno, coming out in two weeks, Deno should be able to pick up `package.json` file and use local `node_modules/` directory automatically in such case. If you are not in a hurry, I suggest to wait a bit and we can provide some more help to make the integration nicer.

src/runtime/entries/deno-server.ts
```diff
@@ -0,0 +1,41 @@
...
+// @ts-expect-error unknown global Deno
+const cert = Deno.env.get("NITRO_SSL_CERT");
+// @ts-expect-error unknown global Deno
+const key = Deno.env.get("NITRO_SSL_KEY");
+// @ts-expect-error unknown global Deno
+const port = destr(Deno.env.get("NITRO_PORT") || Deno.env.get("PORT")) || 3e3;
+// @ts-expect-error unknown global Deno
+const hostname = Deno.env.get("NITRO_HOST") || Deno.env.get("HOST");
+
+function onListen({ port, hostname }) {
```
pi0, line 15: Can you please avoid destructure to avoid confusing with top level vars?

src/presets/deno.ts
```diff
@@ -12,11 +18,119 @@ export const deno = defineNitroPreset({
...
+          if (s.hasChanged()) {
+            return {
+              code: s.toString(),
+              map: s.generateMap({ includeContent: true }),
+            };
+          }
+        },
+      },
+      {
+        name: "inject-process",
```
pi0, line 106: I'm wondering what is the cases that provide plugin fails to inject?

pi0:
> Note: deno doesn't use the node_modules folder we create, but turning off trace means we include full path in the output. Neither is great. 

While deno supporting `node_modules` to enable, can we make it opt-in and use inline strategy? (with chunks)

src/presets/deno.ts
danielroe: it's not that it fails to inject, it's that the importMeta plugin adds it within `generateBundle` - which is too late for the inject plugin.

src/presets/deno.ts
danielroe: yes, it generates (for example) in final bundle:

```js
import { setTimeout, setInterval } from "node:timers";
```

danielroe:
> While deno supporting node_modules to enable, can we make it opt-in and use inline strategy? (with chunks)

The problem is that then we miss out on the great work with node compat from Deno - so `node:` imports will fail as they can't be inlined - and we end up needing nodeless environment and any corresponding compatibility issues.

pi0:
I think we can try to only externalize `node:` (and http/node) protocol and inline any other dependency.

septatrix:
Is this stale or still under development? Denos compatibility has only gotten better since the last comment here

danielroe:
@bartlomieju `deno` is still not picking up on the traced node_modules we've produced. Any suggestions for the best way to ensure it's picking up the `node_modules` in the `.output/server` folder? Does it depend on the working directory for the `deno` command - or do we need to ensure we list dependencies explicitly?

pi0 (approved):

bartlomieju:
@danielroe I need to debug it on my side, could you give me some instructions how you were running this PR?

danielroe:
@bartlomieju You can run:

```sh
git clone git@github.com:unjs/nitro.git
git checkout feat/deno
pnpm i
NITRO_PRESET=deno-server pnpm nitro build test/fixture
# you should then be able to run
deno run --unstable --allow-net --allow-read --allow-env test/fixture/.output/server/index.mjs
```

Note that the `package.json` is in `test/fixture/.output/server/package.json` and uses the `bundledDependencies` key rather than `dependencies`.

bartlomieju:
Thanks @danielroe, I'm pretty sure we don't have support for `bundledDependencies` at the moment. We'll prioritize it and try to release it next week.

danielroe:
I guess this isn't necessarily a blocker to merge as things still work, but it would be a nice perf improvement when it lands. If we do go ahead and merge this, would you let me know when you add support, and I can confirm/adapt if required.

@pi0 Happy either to keep this pending or merge, as you like.

bartlomieju:
Hey, sorry for slow turnaround @danielroe. I just tested the reproduction you provided and the server starts up properly but shows 404 error. I assume this is the problem at hand that Deno doesn't discover modules in `.output/server/node_modules`?

danielroe:
@bartlomieju No, it's not a bug - there is no index route in the test fixture. Try `/api/hello` or one of the other routes in `test/fixture/api` or `test/fixture/routes`.

bartlomieju:
Thanks. I just tried and they all seem to work. Sorry but I'm a bit confused now about the problem here. Is the problem that Deno doesn't install `bundleDependencies` that are present in `package.json`?

danielroe:
The problem is that Deno _does_ install them. Nitro has produced a `node_modules` folder which is ultra-minimal. It's traced just the files that are used, which results in a hugely decreased server `node_modules` folder. It also in some cases correctly resolves different versions of packages in a way that works with tree-based node_modules resolution algorithm. Ideally we would like to use this pre-selected node_modules folder. Otherwise it may be (1) larger than needed, which is suboptimal but not a disaster, (2) it may resolve to a single version of a package, when multiple versions may be required by the build - this could be breaking for users.

danielroe:
@pi0 I know you've approved this previously - is there anything I can do at this point to help get this PR over the line?

danielroe:
@pi0 Let me know when you are ready to review and I will happily resolve merge conflicts 🙏

pi0:
I have made few refactors to keep `deno-server` implementation fully separated making it easier to move forward also added `deno.json` with `start` task and tests.

I think we can move forward since it basically works but there are two important things:

- Versions are not picked from `package.json`. We might need to add `importMaps` to the `deno.json` OR at least add the version to `npm:` specifier. (local hoist test is skipped since failing)
- There are some rollup issues with inject plugin. We likely can fix it by reordering again since basic build works seems good to me to move forward.

bartlomieju:
@pi0 sorry for a slow response, I missed notitication in my inbox. Is there something we can do on Deno side to make it easier for you folks to handle the problem? Could we somehow limit what is discovered from `package.json` by Deno or provide a JS library that would allow to leverage `node_modules/` structure generated by Deno?

PR #593, opened 2022-10-17
refactor!: rename routeOptions to routeRules

danielroe:
### ❓ Type of change

### 📚 Description

This renames a number of route rule related utilities + config to use `route rules` rather than `route options`. This is a breaking change only for users tracking `nitropack-edge`.

### 📝 Checklist


src/runtime/route-options.ts
```diff
@@ -2,35 +2,35 @@ import { eventHandler, H3Event, sendRedirect, setHeaders } from 'h3'
 import defu from 'defu'
```
pi0, line 2: (we might rename file as well)

src/types/nitro.ts
```diff
@@ -165,7 +165,7 @@ export interface NitroOptions extends PresetOptions {
   // Routing
   baseURL: string,
   handlers: NitroEventHandler[]
-  routes: { [path: string]: NitroRouteOptions }
+  routes: { [path: string]: NitroRouteRules }
```
pi0, line 168: Do you mind also renaming `routes` (we can have backward compatibility for a while)

danielroe: shall we also have config.nitro.routeRules or keep as config.nitro.routes?

danielroe: I mean runtime config.

pi0: Sure lets rename it!

PR #605, opened 2022-10-20
fix: add production export condition to commands

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/8362

### ❓ Type of change

### 📚 Description

As we trace dependencies via the `production` export condition, if users do not run their node output with that condition, it may cause issues for them. We may add the condition to our documentation and preview commands to prevent these issues (and possibly overall increase performance).

### 📝 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ npx nitropack build
 4️⃣ Output is in the `.output` directory and ready to be deployed on almost any VPS with no dependencies. You can locally try it too:
 
 ```bash
-node .output/server/index.mjs
+node -C production .output/server/index.mjs
 ```
 
 That's it you got it! Read the [documentation](https://nitro.unjs.io) to learn more.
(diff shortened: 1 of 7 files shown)


danielroe:
Production deployments should be launched with the condition as well.

pi0:
Is there an argument for example for vercel to do it?

danielroe:
I'm not sure. I'll follow up with them directly. But I would _guess_ that setting a `NODE_OPTIONS` environment variable would allow it to be done.

danielroe:
... and I believe we can do that from the FS API. Will check. Shall I add to this PR or make a separate one?

pi0:
Indeed we may find solutions per (some) providers too. My point is, probably right solution instead of requiring the condition is to rewrite written `package.json` exports to make `production` default. We are already doing copy it is not that hard and is a universal fix.

danielroe:
Sounds like a good idea 👍

pi0:
I think we can rework this with new externals plugin and fine-gained access to package.json writing. https://github.com/unjs/nitro/issues/844 (feel free to pick it)

PR #606, opened 2022-10-21
feat: autocompletion for the `preset` config

DamianGlowala:
### ❓ Type of change

### 📚 Description

PR provides autocompletion for the available preset names, and yet allowing to enter arbitrary values for the custom ones.

### 📝 Checklist


src/types/nitro.ts
```diff
@@ -113,6 +113,20 @@ export interface NitroOptions extends PresetOptions {
...
+    | 'cloudflare'
+    | 'digital-ocean'
+    | 'firebase'
+    | 'heroku'
+    | 'layer0'
+    | 'netlify'
+    | 'render-com'
+    | 'stormkit'
+    | 'vercel'
+    | (string & {})
```
danielroe, line 129: can we infer from the actual list of presets - e.g. `keyof typeof presets`?

DamianGlowala: Sure! How would you approach the difference between camel case and dash case though?

image

danielroe: How about excluding presets with capitals like this?

```ts
type DeCamel<T extends string> = T extends Lowercase<T> ? T : never
```

We could also write a slightly more complex type to _convert_ camel -> kebab, but I think that's probably overkill.

DamianGlowala: That works great, but does it mean we would only autocomplete preset names without capitals?

danielroe: Threw this together for fun. And definitely not to use here, but ...

```ts
type UppercaseLetter = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K' | 'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'W' | 'X' | 'Y' | 'Z'
type KebabCase<T extends string> = T extends Lowercase<T>
  ? T
  : T extends `${infer Start}${UppercaseLetter}${string}`
    ? Start extends Lowercase<Start>
      ? T extends `${Start}${infer U}${infer Rest}`
        ? Start extends ''
          ? `${Lowercase<U>}${KebabCase<Rest>}`
          : `${Start}-${Lowercase<U>}${KebabCase<Rest>}`
        : never
      : never
    : never
```

danielroe: Oh, good point... might be worth just hard-coding 😆

DamianGlowala: That was quick! :)

Found a shorter one which also seems to work fine, any thoughts?
https://stackoverflow.com/questions/66140451/typescript-add-kebab-case-types-form-actual-camel-case-keys
```
type Kebab<T extends string, A extends string = ""> =
    T extends `${infer F}${infer R}` ?
    Kebab<R, `${A}${F extends Lowercase<F> ? "" : "-"}${Lowercase<F>}`> :
    A
```
image

danielroe: Oh, nice! Seems a much more reasonable level of complexity.

pi0 (approved):
LOVE IT!

PR #627, opened 2022-10-30
fix(firebase): could not find `functions.yaml`

DevJoghurt:
### 🔗 Linked issue

nuxt/framework#236

### ❓ Type of change

- [x ] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Problem was described in this issue nuxt/framework#236

For internal reasons Firebase cli calls a node script 'spawn ./node_modules/.bin/firebase-functions'.
Nitro cannot know about this and does not copy the necessary files via '@vercel/nft'. 
I have written a small function that copies all the missing necessary files so that the deployment can be performed.

### 📝 Checklist


DevJoghurt:
Related discussions and issues:
https://github.com/nuxt/framework/issues/8301
https://github.com/nuxt/framework/issues/6163
https://github.com/nuxt/framework/discussions/1873
https://github.com/nuxt/framework/discussions/6608

aaronjsutton:
@DevJoghurt Awesome! I also just fixed this issue and wrote the exact same helper function. Was about to submit a PR, guess I don't have to!

danielroe (commented):
This is great - thank you!

Would you try instead whether it's possible to include these files by adding them to the `externals.traceInclude` array rather than running a manual copy?

DevJoghurt:
@danielroe 
Thanks for your review.
I tried to include 'firebase-functions' by something like this:
```javascript
export const firebase = defineNitroPreset({
  entry: '#internal/nitro/entries/firebase',
  externals: {
    traceInclude: ['firebase-functions/lib/bin/firebase-functions.js']
  },
  commands: {
    deploy: 'npx firebase deploy'
  },
  hooks: {
    async 'compiled' (ctx) {
      //resolveRequiredNodeModules(ctx)
      await writeRoutes(ctx)
    }
  }
})
```
It is failing with the following error:
Could not resolve import "firebase-functions/lib/bin/firebase-functions.js" in undefined using exports defined in /Users/XXX/Documents/Code/nitro/node_modules/firebase-functions/package.json.

I think the problem is that the firebase cli needs the following file to work correctly: node_modules/.bin/firebase-functions.
This file links to 'firebase-functions/lib/bin/firebase-functions.js'. This is defined as a binary and not as an export in package.json.
I'm not sure if this can be done by tracing. What do you think?

pi0:
Kind reminder for review @danielroe 

If it is not relevant anymore or you think an alternative method could be better, would be happy to move it to an issue to track.

danielroe:
Apologies - I missed this.

@DevJoghurt I think you can pass paths to binaries as well - they just need to be full paths (that is, including `node_modules`). Would you try?

DevJoghurt:
> Apologies - I missed this.
> 
> @DevJoghurt I think you can pass paths to binaries as well - they just need to be full paths (that is, including `node_modules`). Would you try?

Sure, I'll give it a try and then give you feedback.

DevJoghurt:
@danielroe
Ok, this seems to work in general. Unfortunately, in the firebase-functions.js file the first line is like this: '#!/usr/bin/env node'. This seems to cause a problem with '@rollup/plugin-inject'. Tried to exclude the file via exclude, but this was not possible with the current config options in nitro.
Do you have any idea what else I can try?

Hebilicious:
@DevJoghurt , could you please provide a reproduction ? from my latest tests with nitro and nuxt, the firebase preset works for me without additional modifications.

DevJoghurt:
Yes, that is correct. It seems to work with the latest firebase cli version.
So I will close the pull request now.

PR #629, opened 2022-11-01
fix(cloudflare): pass raw body instead of parsing it

brendonmatos:
### ❓ Type of change

### 📚 Description

In cases where it is necessary to read the body of the raw request, the cloudflare deploy was always parsing the body even using `useRawBody`.
Implementations like the stripe's webhook that needs to read the raw body would break.

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/cloudflare-pages.ts
+++ b/src/runtime/entries/cloudflare-pages.ts
@@ -1,5 +1,5 @@
 import '#internal/nitro/virtual/polyfill'
-import { requestHasBody, useRequestBody } from '../utils'
+import { requestHasBody } from '../utils'
 import { nitroApp } from '../app'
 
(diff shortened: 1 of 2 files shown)


pi0:
`Buffer` for workers is supported via lots of unenv polyfills. If possible, it would be better to use arrayBuffer directly and handle it from h3 utility.

brendonmatos:
thanks for the path you gave me!
the behaviour was implemented in both entries

barisbora:
Can we pass CF records too?

https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties

pi0 (approved):

pi0:
Thanks!

pi0:
> Can we pass CF records too?

Another enhancement, but sure we shall pass them to `event.context.cf`

PR #642, opened 2022-11-07
refactor: nitro/src

Simon-He95:
I have removed wrapper functions, but in fact in runtime this code `function isObject (input: unknown) {
  return typeof input === 'object' && !Array.isArray(input)
}` as wrapper functions is already the case in your repo :)

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -20,10 +20,9 @@ import { compressPublicAssets } from './compress'
 
 export async function prepare (nitro: Nitro) {
   await prepareDir(nitro.options.output.dir)
-  if (!nitro.options.noPublicDir) {
-    await prepareDir(nitro.options.output.publicDir)
-  }
-  await prepareDir(nitro.options.output.serverDir)
+  await prepareDir(!nitro.options.noPublicDir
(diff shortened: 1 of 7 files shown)


src/build.ts
```diff
@@ -20,10 +20,9 @@ import { compressPublicAssets } from './compress'
 
 export async function prepare (nitro: Nitro) {
   await prepareDir(nitro.options.output.dir)
-  if (!nitro.options.noPublicDir) {
-    await prepareDir(nitro.options.output.publicDir)
-  }
-  await prepareDir(nitro.options.output.serverDir)
+  await prepareDir(!nitro.options.noPublicDir
+    ? nitro.options.output.publicDir
+    : nitro.options.output.serverDir)
```
pi0, line 25: Both dirs need to be created not either of them

pi0:
Thanks for the PR. Since we have new [eslint-config](https://github.com/nuxt/eslint-config) and prettier defaults, i think would keep using it. Also in some conditions intentionally not merging if blocks to improve readability (for myself mostly...)

Issue #641, opened 2022-11-07
Dev Proxy does not work for some pages


manniL:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nitro Version:    `0.6.1`
- Package Manager:  `npm@7.17.0`
------------------------------

### Reproduction

1. Open [the StackBlitz](https://stackblitz.com/edit/unjsnitroissues641)
2. Go to /proxy
3. See that proxy to `http://example.com` does not work
4. Go to `/proxy-2`
5. See that the proxy to `https://api.albionstatus.com` works

Config:

```
import { defineNitroConfig } from 'nitropack';

export default defineNitroConfig({
  devProxy: {
    '/proxy': { target: 'http://example.com', changeOrigin: true },
    '/proxy-2': { target: 'https://api.albionstatus.com', changeOrigin: true },
  },
});
```

### Describe the bug

The `devProxy option` does not work for some pages 🤔

### Additional context

Also: Would be nice to add some tests for the proxy routes. Fixture is there but no test yet.

Related: https://github.com/nuxt/framework/issues/8763#issuecomment-1305351343

manniL:
Tested locally. Seems that some proxy-settings don't work in StackBlitz due to CORS. If you rebuild the linked StackBlitz locally, it will work fine ☺️

PR #644, opened 2022-11-08
test: add tests for dev server and proxy

manniL:
### 🔗 Linked issue

#641 - not resolving though

### ❓ Type of change

### 📚 Description

Initial plan was to check why the proxy is failing. In the local setup, the proxy did not fail.
I've added some test for the proxy and for the dev preset in general to ensure

### 📝 Checklist


test/tests.ts
```diff
@@ -63,11 +71,21 @@ export async function setupTest (preset) {
...
   console.log('>', ctx.server!.url)
 }
 
+export async function startDevServer (ctx: Context) {
+  const devServer = createDevServer(ctx.nitro)
+  ctx.server = await devServer.listen({})
+  await prepare(ctx.nitro)
+  await build(ctx.nitro)
+  // TODO: How to await that dev server is ready?!
+  await new Promise(resolve => setTimeout(resolve, 1000))
```
manniL, line 83: Did not find a better way 🙈

test/tests.ts
```diff
@@ -63,11 +69,21 @@ export async function setupTest (preset) {
...
```
danielroe, line 83: wdyt @pi0?

```suggestion
  const ready = new Promise<void>((resolve) => { ctx.nitro.hooks.hook('dev:reload', () => resolve()) })
  await build(ctx.nitro)
  await ready
```

pi0: `dev:reload` hook seems a nice idea! Addtionally we need to move this test to dev preset.

danielroe: @manniL would you move this to the dev preset? I think @pi0 just means extracting this function and inlining it in the dev preset test file.

manniL: Ohh, yes sure 👍🏻

danielroe (approved):

pi0 (approved):
Thanks!

PR #650, opened 2022-11-13
perf(aws-lambda): inline dynamic chunks to reduce first invokation time

WinterYukky:
### 🔗 Linked issue

None

### ❓ Type of change

### 📚 Description

This PR will reduce AWS Lambda runtime and improve cost savings and performance for customers.

By default, inlineDynamicImports is false in nitro. Therefore, if we define an API that uses any module under the `server/api` directory and run it, the module will be loaded at runtime and the first time execution will take a long time. In fact, I created and run an API that just scans DynamoDB using aws-sdk v3, it will take more than 3 seconds the first time. This exceeds the default timeout time of AWS Lambda, which causes the API Gateway response to be an Internal Server Error. To avoid this, set inlineDynamicImports to true in the aws-lambda preset. This change speeds up the API from taking 3 seconds to 300 milliseconds.

### 📝 Checklist


danielroe:
Would you provide an example of some code that takes that additional time? Inlining dynamic imports will make the entrypoint larger, so this is paradoxical behaviour.

pi0:
Thanks for the PR. Dynamic chunks loading in general (both browser and Node) has a natural overhead. But it is essential to make timings stable when bundle size does grow while if we inline dependencies, it increases size even when not used. So whenever possible with Nitro presets, we tend to use chunks.

As @danielroe mentioned, would be nice if you can provide a public reproduction of your project to test timing differences.

pi0:
Thanks for sharing links and stats @WinterYukky ❤️  Also linking [this](https://aws.amazon.com/blogs/compute/operating-lambda-performance-optimization-part-1/). 

When assuming most of the chunks of the server bundle will be used during every invocation, inlining them makes perfect sense and reduces latency. However, when the number of server handlers (and individual libraries and other assets they use) grows, I believe this adds to the cold-start time.

I'm positive this would be a nice idea to enable inlining by default but please let me run some more tests and check alternatives if we can keep using chunks but warmup opon code-initialization for entry critical chunks.

Hebilicious:
@pi0 We could add this as`nitro.options.aws` so user can decide for themselves.

pi0:
I have added docs for aws deployment for now to note about `inlineDynamicImports` option linking to this PR (https://github.com/unjs/nitro/commit/ff38dc02fcd0ea1cfdd2c693edfbe2e6fc78d0a4).

Issue #391, opened 2022-08-08
Cloudflare `request.cf`


jjjrmy:
I'd like to be able to access the [Incoming Request CF Properties](https://developers.cloudflare.com/workers//runtime-apis/request#incomingrequestcfproperties)

pi0:
Latest unenv contains a fix that allows accessing raw event object. As a temporary workaround, you can access it using `event.node.req.__unenv__.event.request.cf` (requires fresh lock)

PR #677, opened 2022-11-20
feat: extended cloudflare context

barisbora:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/391

### ❓ Type of change

### 📚 Description

Extending context object with cloudflare CF Properties and headers

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/cloudflare.ts
+++ b/src/runtime/entries/cloudflare.ts
@@ -1,6 +1,9 @@
 import '#internal/nitro/virtual/polyfill'
 import { getAssetFromKV, mapRequestToAsset } from '@cloudflare/kv-asset-handler'
 import { withoutBase } from 'ufo'
+import { RequestInit } from '@cloudflare/workers-types'
+import { createCall } from 'unenv/runtime/fetch/index'
+import { NodeListener, App, createEvent, createError, isError, sendError, H3EventContext } from 'h3'
 import { requestHasBody } from '../utils'
 import { nitroApp } from '../app'
 import { useRuntimeConfig } from '#internal/nitro'
@@ -22,7 +25,14 @@ async function handleEvent (event: FetchEvent) {
     body = Buffer.from(await event.request.arrayBuffer())
   }
 
-  const r = await nitroApp.localCall({
+  const headers = Object.fromEntries(event.request.headers.entries())
+
+  const localCall = createCall(toNodeListener(nitroApp.h3App, {
+    headers,
+    cf: (event.request as unknown as RequestInit)?.cf
+  }) as any)
+
+  const r = await localCall({
     event,
     url: url.pathname + url.search,
     host: url.hostname,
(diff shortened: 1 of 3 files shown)


src/runtime/entries/cloudflare.ts
```diff
@@ -22,7 +25,11 @@ async function handleEvent (event: FetchEvent) {
     body = Buffer.from(await event.request.arrayBuffer())
   }
 
-  const r = await nitroApp.localCall({
```
pi0, line 25: We could simply extend `event.context.cf = event.cf` in line before i guess.

src/runtime/entries/cloudflare.ts
```diff
@@ -60,3 +67,28 @@ const baseURLModifier = (request: Request) => {
 function normalizeOutgoingHeaders (headers: Record<string, string | string[] | undefined>) {
   return Object.entries(headers).map(([k, v]) => [k, Array.isArray(v) ? v.join(',') : v])
 }
+
+function toNodeListener (app: App, context: H3EventContext): NodeListener {
+  const toNodeHandle: NodeListener = async function (req, res) {
+    const event = createEvent(req, res)
+    event.context = { ...event.context, ...context }
```
danielroe, line 74: I think we might want to add CF context as `event.context.cloudflare` for better type support.

timhanlon:
```suggestion
    event.context = { ...event.context, cloudflare: context }
```

danielroe: Oops - didn't mean to edit. What about?

```suggestion
    event.context.cloudflare = context
```

barisbora: Thanks for feedback, implemented 🙌

src/runtime/entries/cloudflare.ts
```diff
@@ -60,3 +67,28 @@ const baseURLModifier = (request: Request) => {
 function normalizeOutgoingHeaders (headers: Record<string, string | string[] | undefined>) {
   return Object.entries(headers).map(([k, v]) => [k, Array.isArray(v) ? v.join(',') : v])
 }
+
+function toNodeListener (app: App, context: H3EventContext): NodeListener {
+  const toNodeHandle: NodeListener = async function (req, res) {
+    const event = createEvent(req, res)
+    event.context.cloudflare = context
```
danielroe, line 74: Should this be:

```suggestion
    event.context.cloudflare = context.cf
```

timhanlon: I think the answer to this depends on where you'll want the [fetch event context](https://developers.cloudflare.com/workers/runtime-apis/fetch-event/#parameters) to live when module syntax support lands in #681.

pi0:
Thanks for working on this @barisbora ❤️ Adding support with a slightly more generic approach in #927

Issue #685, opened 2022-11-22
Type routes correctly based on their method


dsluijk:
### Describe the feature

Currently if you have multiple routes with the same name but a different method, the return type is set to a combination of the two. This requires you to change the types after a request has been send, which is not a great developer experience. As the method is given when sending the request, it is possible to return the correct types based on the method.

This is also required for https://github.com/nuxt/framework/issues/8538.

### Additional information

- [X] Would you be willing to help implement this feature?

PR #686, opened 2022-11-22
feat(types): correctly type $fetch based on the method

dsluijk:
### 🔗 Linked issue

#685

### ❓ Type of change

### 📚 Description

This feature fixed the types of the routes when used with $fetch, when there are multiple handlers with different methods for the same route. It detects the method given with the options, and returns the correct set of types for that call. This also works if you don't specify a method or if the handler does not specify a method.

This might be a breaking change, as it changes the format of `InternalApi`.

### 📝 Checklist

I don't think a documentation update is needed here.


dsluijk:
I've fixed the failing test and added a few new ones to assert the new types. The `method` option in a request is now also typed correctly with the methods available.

src/build.ts
```diff
@@ -101,8 +113,14 @@ export async function writeTypes(nitro: Nitro) {
     "declare module 'nitropack' {",
     "  type Awaited<T> = T extends PromiseLike<infer U> ? Awaited<U> : T",
     "  interface InternalApi {",
-    ...Object.entries(routeTypes).map(
-      ([path, types]) => `    '${path}': ${types.join(" | ")}`
+    ...Object.entries(routeTypes).map(([path, methods]) =>
+      [
+        `    '${path}': {`,
+        ...Object.entries(methods).map(
+          ([method, imports]) => `      '${method}': ${imports.join(" | ")}`
```
danielroe, line 120:
```suggestion
          ([method, types]) => `      '${method}': ${types.join(" | ")}`
```

danielroe (approved):
This looks very nice - thanks! 😊

PR #687, opened 2022-11-23
fix(plugin): Fixes imports of plugins under windows platform

grbd:
### 🔗 Linked issue

  * https://github.com/nuxt-themes/docus/issues/684

### ❓ Type of change

- [X ] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Docus has recently been made public, but this has highlighted some issues when building under windows.
One of them is that nitro I think has trouble parsing plugin paths under windows due to a lack of the "file:///" prefix when passed to the import

Currently it's being called by
https://github.com/Tahul/pinceau/blob/e99c19bbca4ef4626c59d35bbd5bf652df6d2753/src/nuxt.ts#L44

### 📝 Checklist

- [ X] I have linked an issue or discussion.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -264,7 +264,7 @@ export async function loadOptions (configOverrides: NitroConfig = {}): Promise<N
   options.plugins = options.plugins.map(p => resolvePath(p, options))
 
   if (isWindows) {
-    options.plugins = options.plugins.map(p => pathToFileURL(p).toString())
+    options.plugins = options.plugins.map(p => pathToFileURL(p).href)
   }
 
   return options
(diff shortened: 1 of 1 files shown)


src/options.ts
```diff
@@ -261,6 +263,10 @@ export async function loadOptions (configOverrides: NitroConfig = {}): Promise<N
   // Resolve plugin paths
   options.plugins = options.plugins.map(p => resolvePath(p, options))
 
+  if (isWindows) {
```
danielroe, line 266: It might be worth confirming whether we need to do this _when building_ as well as in development.

danielroe (commented):
Thank you. It feels like it might be useful to enable CI testing on windows to spot some of these issues earlier - wdyt @pi0?

pi0:
Hi @grbd and thanks for the PR. It should be solved by #732 in next release.

PR #694, opened 2022-11-25
fix(deno): implement readAsset

so1ve:
### ❓ Type of change

### 📝 Checklist


so1ve:
I'm not sure if path.join should be used instead of hardcoded id.

pi0 (approved):
Thanks!

pi0:
> I'm not sure if path.join should be used instead of hardcoded id.

id should be already normalized when passed to bundle. would be nice to double check it for windows however

> Also, it seems that process.env is not replaced by Deno.env.toObject() in the deno press

Yeah we still need unenv/nitro preset for polyfills @danielroe is working on it. Feel free to open an issue for it.

so1ve:
Hi @pi0,
actually deno provides official polyfills here: https://deno.land/std/node

pi0:
yes, std is what we need to port as unenv preset for mapping.

PR #708, opened 2022-11-29
refactor: tidy up

Simon-He95:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -64,12 +64,10 @@ export async function prerender (nitro: Nitro) {
   // Start prerendering
   const generatedRoutes = new Set()
   const canPrerender = (route: string = '/') => {
-    if (generatedRoutes.has(route)) { return false }
-    if (route.length > 250) { return false }
-    for (const ignore of nitro.options.prerender.ignore) {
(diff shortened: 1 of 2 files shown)


manniL:
Thanks for your effort but I feel like the proposed improvements make the code less readable than before. Similar to previously proposed PRs. 🙁

Simon-He95:
> Thanks for your effort but I feel like the proposed improvements make the code less readable than before. Similar to previously proposed PRs. 🙁

Why is it more difficult to understand, I just combined the conditions of the same return result, and then replaced the for that satisfies the conditions with some

manniL:
Because chained logical statements can make code less readable. I'd rather have longer code that shows that the statements are separate concerns / cases than figuring out these cases in a longer chained or-statement.

If necessary, some conditions would even have an own variable that describes it. Longer code, but easier to understand.

Simon-He95:
I think finding a unique condition for returns, which just satisfies some function, can express this meaning. In addition, since they all return false, it will be more concise to use the if condition to merge and express in a new line. Returning in this case, I don't think he loses readability, it's ok, everyone's understanding is different.

PR #732, opened 2022-12-06
fix: normalize nitro plugin paths to url in development

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/8461
https://github.com/Baroshem/nuxt-security/issues/42

### ❓ Type of change

### 📚 Description

We need to normalize plugins that are absolute paths (`C:\test\something`) to file URLs on windows before we generate the imports from them in `dev/index.mjs`. 

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -7,7 +7,7 @@ import { defu } from 'defu'
 import { resolveModuleExportNames, resolvePath as resolveModule } from 'mlly'
 // import escapeRE from 'escape-string-regexp'
 import { withLeadingSlash, withoutTrailingSlash, withTrailingSlash } from 'ufo'
-import { isTest, isDebug, isWindows } from 'std-env'
+import { isTest, isDebug } from 'std-env'
 import { findWorkspaceDir } from 'pkg-types'
 import { resolvePath, detectTarget } from './utils'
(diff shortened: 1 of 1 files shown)


src/options.ts
```diff
@@ -259,7 +260,13 @@ export async function loadOptions (configOverrides: NitroConfig = {}): Promise<N
   }
 
   // Resolve plugin paths
-  options.plugins = options.plugins.map(p => resolvePath(p, options))
+  options.plugins = options.plugins.map((p) => {
+    const path = resolvePath(p, options)
+    if (isWindows && options.dev && isAbsolute(path)) {
+      return pathToFileURL(path).href
```
pi0, line 266: Would it broken on non windows using file urls?

danielroe: seems to work fine in the repro in one of the linked issues., but I seem to recall previously there were issues with file urls outside of windows environments (but that might have been in a nuxt/vite environment 🤔). if you think it better I am happy to remove the condition.

pi0: I think it was vite since we already use file URLs for dev normally. Let's try with consistent behavior and limit to windows if was causing trouble

PR #742, opened 2022-12-09
fix: use auto-imports to enable global `$fetch`

danielroe:
### ❓ Type of change

### 📚 Description

There are currently ssues with using `vercel-edge` and other presets where we are adding globals into `globalThis` - e.g. https://github.com/pi0/nuxt-on-the-edge/blob/main/app.vue#L11. This PR is a starting point for discussion.

I propose that if we go down this route, we add a warning if user explicitly disables nitro auto-imports, stating that they will need to import `$fetch` explicitly in that case.

### 📝 Checklist


pi0:
Can we please keep global injection behavior _in addition_ to auto imports?

pi0:
globalThis issue with vercel's edge is so strange indeed. Can you please point out "and other presets" currently suffering this issue? 

Regardless even if going down this path, i think we can keep hybrid approach to both inject and auto import.

danielroe:
Personally I only experienced the issue with vercel-edge but I imagined it might be replicable on netlify-edge or cloudflare.

src/types/fetch.ts
```diff
@@ -93,15 +93,4 @@ export interface $Fetch<
   ): $Fetch<T, R>;
 }
 
-declare global {
```
pi0, line 96: Can we please still keep this? I like to make types working as much as possible before prepare command.

danielroe: I am worried that the types may conflict but I'm happy to restore it and see.

harlan-zw:
This would be nice, vercel-edge does throw the `undefined $fetch`. If you import directly from `ofetch` it breaks the internal route handling.

Workaround was adding `globalThis.` to any requests

```ts
await globalThis.$fetch(path)
```

pi0:
I finally got a chance to make a minimal reproduction. Seems `globalThis` support issues are really limited to `vercel-edge` preset and other workers are ok. Wile using auto imports, might handle edge case of `$fetch` on vercel-deploy, there are many more cases we are already using globalThis to polyfill and such workaround cannot work for them.

Let's track issue in https://github.com/unjs/nitro/issues/921. 

(as a better workaround maybe we can only for `vercel-edge` transform `$fetch` to `globalThis.` using rollup provide?)

PR #744, opened 2022-12-10
feat(cache): allow setting custom `getKey` for `defineCachedEventHandler`

### LINKED ISSUE #691: Allow passing getKey method to route rules
### Describe the feature

I want to use a custom cache key for my route cache (coming from nuxt but open to implementing a nitro plugin).

Naive me thought I could just pass it like this in my nitro (well, nuxt) config:
```js
  routeRules: {
    '/**': {
      cache: {
        swr: true,
        maxAge: 600,
        name: 'pagecache',
        getKey: (e: any) => {
          console.log('hello', e);
          return 'dum';
        },
      },
    },
  },
});
```
that doesn't work because nitro can't serialize my method. fair enough.

my next try is a lot more ugly, but I basically did this as a nitro plugin:
```js
  let h = handlers.find(h => h.route === "/**");
  let handler = h.lazy ? lazyEventHandler(h.handler) : h.handler;
  console.log("installing handler", h.route);

  nitroApp.router.use(h.route, cachedEventHandler(handler, {
        group: "nitro/routes",
        swr: true,
        maxAge: 600,
        name: "pagecache",
        getKey: (e: any) => {
          console.log('hello', e);
          return 'dum';
        },
      }
  ), h.method);
});
```
it doesn't use any api and depends on implementation detail (handlers and cachedEventHandler aren't in scope in my plugin but are in scope in the compiled node-server.mjs), but it would be acceptable to me.

sadly not even that works, because the getKey method in my options object is overriden here: https://github.com/unjs/nitro/blob/main/src/runtime/cache.ts#L120

I am close to commenting that out in the compiled code via sed, but I figured I would open a feature request here first :D

I can't think of a nice api to make my plugin code nicer right now, since I don't have the handler cause it gets created by nuxt, but I would already be happy if cache.ts could be adjusted to not override getKey methods from the outside.

open to other ideas tho, I am willing to help brainstorming (I am MiniDigger#3086 on discord if that helps) and even implement this (haven't touched nitro code before but seems straight forward), my company allows me to work/contribute on this problem on work time.

### Additional information

yassilah:
### 🔗 Linked issue
resolves #691 

### ❓ Type of change

### 📝 Checklist


src/runtime/cache.ts
```diff
@@ -142,14 +139,16 @@ export function defineCachedEventHandler<T = any>(
...
-      const friendlyName = decodeURI(parseURL(url).pathname)
-        .replace(/[^\dA-Za-z]/g, "")
-        .slice(0, 16);
-      const urlHash = hash(url);
-      return `${friendlyName}.${urlHash}`;
-    },
+    getKey:
+      opts.getKey ||
+      ((event) => {
+        const url = event.req.originalUrl || event.req.url;
```
pi0, line 145: Maybe we can call `getKey` inside this function? this way if it returns null, we can fallback to the default key behavior.

We also need to escape invalid chars also in all conditions (including `..`) as they can lead to security issues.

yassilah: Yes, I think it makes sense to call it inside and fallback to the default; as for the escaping of invalid chars, do you mean escaping them from the returned key?

pi0: Yes, to only allow ascii charachters like later regex we have.

src/runtime/cache.ts
```diff
@@ -142,11 +146,9 @@ export function defineCachedEventHandler<T = any>(
     ...opts,
     getKey: (event) => {
       const key = opts.getKey?.(event)
-      if (key) { return key }
+      if (key) { return escapeKey(key) }
```
yassilah: like this? @pi0

pi0 (approved):

pi0:
Thanks <3

MiniDigger:
thank you, now I can't wait to get back to work next year and remove all my disgusting hacks, lmao

Issue #745, opened 2022-12-10
Add an option to force a cache refresh when defining a cached event handler


yassilah:
### Describe the feature

I have tried to use the existing options to allow for refreshing a cached endpoint output but haven't really succeeded in doing so. I would suggest adding a simple `refresh` option to the existing ones to allow for forcing a given endpoint to refresh its original state and store it for future requests. Here's an example usage:

```ts
export default defineCachedEventHandler(event => {
  return Math.floor(Math.random() * 5)
}, {
  refresh: event => getQuery(event).refresh === 'true'
})
```

`/api/foo => cache initial value`
```json
3
```

`/api/foo?refresh=true => renew cache`
```json
4 
```

`/api/foo => use latest cache`
```json
4
```

### Additional information

- [X] Would you be willing to help implement this feature?

PR #746, opened 2022-12-10
feat: add `shouldInvalidateCache ` option to cache utils

yassilah:
### 🔗 Linked issue
#745 

### ❓ Type of change

### 📝 Checklist


src/runtime/cache.ts
```diff
@@ -69,6 +71,7 @@ export function defineCachedFunction<T = any>(
     }
 
     const expired =
+      refresh ||
```
pi0, line 74: I think it should be called? Also i'm thinking if we can give it a more generic name such as `skipCache(..)` as it is a passive checker (refresh seems like an active function that does refresh when called)

yassilah: It is being called but in the returned async function below following the same pattern as `getKey` :) As for the name, I may be wrong but I feel like `skipCache` implies it is just a one-time thing rather than a permanent cache override; what do you think?

pi0: Right. I think it could be prefixed with `do` or `force` to clarify it is not same as option but a flag.

Agreed about skipCache naming. What about `shouldSkipCache` or `shouldInvalidateCache` or something similar?

yassilah: I like `shouldInvalidateCache`, I think it's clear enough!

pi0 (approved):

PR #761, opened 2022-12-17
feat: include only compressible mime types

nathanchase:
### 🔗 Linked issue
https://github.com/nuxt/framework/discussions/9626

### ❓ Type of change

### 📚 Description

The code for `compressPublicAssets` was double-compressing all image types (including .jpg, .png, .gif, .webp, etc.) when it really should only be compressible static file types that are text-based. I updated the code to use the following list of MIME types, and also added it to the docs for `compressPublicAssets`:

text/html
text/css
text/plain
text/xml
text/x-component
text/javascript
application/x-javascript
application/javascript
application/json
application/manifest+json
application/vnd.api+json
application/xml
application/xhtml+xml
application/rss+xml
application/atom+xml
application/vnd.ms-fontobject
application/x-font-ttf
application/x-font-opentype
application/x-font-truetype
image/svg+xml
image/x-icon
image/vnd.microsoft.icon
font/ttf
font/eot
font/otf
font/opentype

### 📝 Checklist

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -72,5 +72,24 @@ function isTextMime(mimeType: string) {
 }
 
 function isCompressableMime(mimeType: string) {
-  return /image|text|font|json|xml|javascript/.test(mimeType);
+  const compressibleMimeTypes = [
+    'atom',
+    'css',
+    'eot',
+    'htc',
+    'html',
+    'ico',
+    'js',
+    'json',
+    'mjs',
+    'otf',
+    'rss',
+    'svg',
+    'text',
+    'ttf',
+    'webmanifest',
+    'xml',
+  ];
+
+  return compressibleMimeTypes.includes(mimeType);
 }
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -71,7 +71,8 @@ function isTextMime(mimeType: string) {
   return /text|javascript|json|xml/.test(mimeType);
 }
 
-const COMPRESSIBLE_MIMES_RE = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/;
+const COMPRESSIBLE_MIMES_RE =
+  /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/;
 
 function isCompressableMime(mimeType: string) {
   return COMPRESSIBLE_MIMES_RE.test(mimeType);
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -71,7 +71,7 @@ function isTextMime(mimeType: string) {
   return /text|javascript|json|xml/.test(mimeType);
 }
 
-const COMPRESSIBLE_MIMES_RE = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/
+const COMPRESSIBLE_MIMES_RE = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/;
 
 function isCompressableMime(mimeType: string) {
   return COMPRESSIBLE_MIMES_RE.test(mimeType);
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -71,9 +71,57 @@ function isTextMime(mimeType: string) {
   return /text|javascript|json|xml/.test(mimeType);
 }
 
-const COMPRESSIBLE_MIMES_RE =
-  /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/;
+// Reference list of compressible MIME types from AWS
+// https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-cloudfront-file-types
+const COMPRESSIBLE_MIMES_RE = new Set([
+  "application/dash+xml",
+  "application/eot",
+  "application/font",
+  "application/font-sfnt",
+  "application/javascript",
+  "application/json",
+  "application/opentype",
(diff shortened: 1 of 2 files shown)


src/compress.ts
```diff
@@ -72,5 +72,24 @@ function isTextMime(mimeType: string) {
...
+    'mjs',
+    'otf',
+    'rss',
+    'svg',
+    'text',
+    'ttf',
+    'webmanifest',
+    'xml',
+  ];
+
```
danielroe, line 93: I think we can extract this from the function so we don't redeclare the array each time:

```suggestion
```

src/compress.ts
```diff
@@ -72,5 +72,24 @@ function isTextMime(mimeType: string) {
 }
```
danielroe, line 73:
```suggestion

  const compressibleMimeTypes = [
    'atom',
    'css',
    'eot',
    'htc',
    'html',
    'ico',
    'js',
    'json',
    'mjs',
    'otf',
    'rss',
    'svg',
    'text',
    'ttf',
    'webmanifest',
    'xml',
  ];

```

src/compress.ts
```diff
@@ -71,6 +71,25 @@ function isTextMime(mimeType: string) {
...
+    'svg',
+    'text',
+    'ttf',
+    'webmanifest',
+    'xml',
+  ];
+
 function isCompressableMime(mimeType: string) {
-  return /image|text|font|json|xml|javascript/.test(mimeType);
+  return compressibleMimeTypes.includes(mimeType);
```
danielroe, line 94: I'm not sure this will work as written, because aren't we testing a longer string like `text/xml`? Maybe something like:

```suggestion
  return compressibleMimeTypes.some(type => mimeType.includes(type));
```

nathanchase: We could keep with the regex approach that was previously used and do something like:
```
const compressibleMimeTypesRegex = compressibleMimeTypes.join('|');
  return new RegExp(compressibleMimeTypesRegex).test(mimeType);
```

danielroe: Good idea. But let's use a regexp where we define it directly then to safe the overhead of converting a string => regexp.

nathanchase: So... something like this should work:

```
const compressibleMimeTypesRegex = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/

function isCompressible(mimeType) {
  return compressibleMimeTypesRegex.test(mimeType);
}
```

danielroe: Perfect.

src/compress.ts
```diff
@@ -71,6 +71,8 @@ function isTextMime(mimeType: string) {
   return /text|javascript|json|xml/.test(mimeType);
 }
 
+const compressibleMimeTypesRegex = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/
```
danielroe, line 74:
```suggestion
const COMPRESSIBLE_MIMES_RE = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/
```

src/compress.ts
```diff
@@ -71,6 +71,8 @@ function isTextMime(mimeType: string) {
...
+
 function isCompressableMime(mimeType: string) {
-  return /image|text|font|json|xml|javascript/.test(mimeType);
+  return compressibleMimeTypesRegex.test(mimeType);
```
danielroe, line 77:
```suggestion
  return COMPRESSIBLE_MIMES_RE.test(mimeType);
```

src/compress.ts
```diff
@@ -71,6 +71,8 @@ function isTextMime(mimeType: string) {
   return /text|javascript|json|xml/.test(mimeType);
 }
 
+const COMPRESSIBLE_MIMES_RE = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/
```
danielroe, line 74:
```suggestion
const COMPRESSIBLE_MIMES_RE = /atom|css|eot|htc|html|ico|js|json|mjs|otf|rss|svg|text|ttf|webmanifest|xml/;
```

danielroe (approved):

manniL (approved):

pi0:
Thanks for the PR @nathanchase  and review @manniL @danielroe.

I think currently, docs is not consistent with implementation. We shall maintain an explicit list (array) of supported mimes. This also allows making it easier to customize by users whereas regex is harder to extend. Also we might include webp to the list as is text based.

nathanchase:
@pi0 We could use https://www.npmjs.com/package/compressible ?

nathanchase:
Here's a list that AWS uses: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#compressed-content-cloudfront-file-types

nathanchase:
> Also we might include webp to the list as is text based.

I was under the understanding that using Brotli/Gzip on .webp files would just increase the file size. I'm unable to find a clear answer to that question in my research, other than `image/webp` not appearing on any compressible type lists.

pi0:
[compressible](https://unpkg.com/browse/compressible@2.0.18/index.js) uses mime-db, which nitro already stubs into [this](https://github.com/unjs/unenv/blob/main/src/runtime/npm/mime-db.ts). 

AWS list, seems a great start 👍🏼 Let's not over complicate this.

> I was under the understanding that using Brotli/Gzip on .webp files would just increase the file size. I'm unable to find a clear answer to that question in my research, other than image/webp not appearing on any compressible type lists.

You are right. Totally my bad. I meant SVG 🤦🏼‍♂️

nathanchase:
Take a look at the latest commits and see what you think. Should be good to go.

pi0 (approved):
💯

Issue #762, opened 2022-12-17
Set a default Cache-Control for public static assets


nathanchase:
### Describe the feature

For all public static assets, we can define a best practice value for the Cache-Control header, to fall in line with the Google Lighthouse scoring of [serving assets with an efficient cache policy](https://developer.chrome.com/en/docs/lighthouse/performance/uses-long-cache-ttl/).

`Cache-Control: max-age=31536000, immutable`

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
Hi. This is possible since Nitro 2.1 to specify `maxAge` for `publicAssets`. Since we are not sure if public assets are properly hashed or not, nitro has no default value but Nuxt sets it for `_nuxt`/buildDir  https://github.com/nuxt/nuxt/blob/f225ebd6435d1d9305d1d143f835ea921798a2ba/packages/nuxt/src/core/nitro.ts#L89

PR #763, opened 2022-12-17
feat: set default Cache-Control header with max-age of 1 year

nathanchase:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/762, https://github.com/nuxt/nuxt.js/issues/15465

### ❓ Type of change

### 📚 Description

For all public static assets, we can define a best practice value for the Cache-Control header, to fall in line with the Google Lighthouse scoring of [serving assets with an efficient cache policy](https://developer.chrome.com/en/docs/lighthouse/performance/uses-long-cache-ttl/).

`Cache-Control: max-age=31536000, immutable`

This checks if the asset is public, and not already defined, and sets the header accordingly.

### 📝 Checklist

### DIFF
--- a/src/runtime/static.ts
+++ b/src/runtime/static.ts
@@ -92,11 +92,10 @@ export default eventHandler((event) => {
     event.res.setHeader("Content-Length", asset.size);
   }
 
-  // TODO: Asset dir cache control
-  // if (isBuildAsset) {
-  // const TWO_DAYS = 2 * 60 * 60 * 24
-  // event.res.setHeader('Cache-Control', `max-age=${TWO_DAYS}, immutable`)
-  // }
+  if (isPublicAssetURL(id) && !event.res.getHeader("Cache-Control")) {
+    const ONE_YEAR = 365 * 24 * 60 * 60;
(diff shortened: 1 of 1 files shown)


src/runtime/static.ts
```diff
@@ -92,11 +92,10 @@ export default eventHandler((event) => {
     event.res.setHeader("Content-Length", asset.size);
   }
 
-  // TODO: Asset dir cache control
-  // if (isBuildAsset) {
-  // const TWO_DAYS = 2 * 60 * 60 * 24
-  // event.res.setHeader('Cache-Control', `max-age=${TWO_DAYS}, immutable`)
-  // }
+  if (isPublicAssetURL(id) && !event.res.getHeader("Cache-Control")) {
```
danielroe, line 95: for a public asset base of `/` (which would be the case with Nuxt's `/public` directory) this will cache all html/assets, including things like `robots.txt`, `favico.ico` or even `/about/index.html`. I think that is probably too broad a net to cast.

danielroe: I think @pi0 has some thoughts about this, but maybe allow cache headers to be set per asset folder? So Nuxt could push in `/_nuxt`, for example, with the appropriate long cache headers.

PR #767, opened 2022-12-19
fix: remove base url before calculating route rules

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/9762

### ❓ Type of change

### 📚 Description

This PR removes baseURL before calculating route rules.

### 📝 Checklist

### DIFF
--- a/src/runtime/route-rules.ts
+++ b/src/runtime/route-rules.ts
@@ -1,6 +1,7 @@
 import { eventHandler, H3Event, sendRedirect, setHeaders } from "h3";
 import defu from "defu";
 import { createRouter as createRadixRouter, toRouteMatcher } from "radix3";
+import { withoutBase } from "ufo";
 import { useRuntimeConfig } from "./config";
 import type { NitroRouteRules } from "nitropack";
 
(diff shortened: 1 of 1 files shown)


src/runtime/route-rules.ts
```diff
@@ -32,7 +33,7 @@ export function getRouteRules(event: H3Event): NitroRouteRules {
   event.context._nitro = event.context._nitro || {};
   if (!event.context._nitro.routeRules) {
     const path = new URL(event.req.url, "http://localhost").pathname;
-    event.context._nitro.routeRules = getRouteRulesForPath(path);
+    event.context._nitro.routeRules = getRouteRulesForPath(withoutBase(path, useRuntimeConfig().app.baseURL));
```
pi0, line 36: I'm wondering if we should do this directly inside `getRouteRulesForPath`?

danielroe: I felt it was more natural here as we have access to `event`; `getRouteRulesForPath` feels like more of a 'pure' function. Also, looking at usage, it is used with an already normalised url, meaning if we move this there we are likely to double-remove a baseURL:

https://github.com/unjs/nitropack/blob/6541373545ee2716566bce46c7568ad2c3164cdf/src/runtime/app.ts#L58-L60

pi0 (approved):

PR #768, opened 2022-12-19
fix(netlify, vercel): order route rules from most specific + avoid double-rendering root

danielroe:
### ❓ Type of change

### 📚 Description

This orders route rules within vercel + netlify presets so that (particularly glob) rules do not shadow rules below them. In addition, if the default vercel route rule is `/**: { static: true }` we don't also need to output a second, SSR rule.

cc: @antfu - if you clone this and checkout commit c7f8f39972b43a8ddb0b06d558b3b9f3b19f0a45, then run `vitest run -u` the test passes and the inline snapshots are not updated. I've also noticed some other weird issues like a different snapshot within a file being updated than the one it corresponds to (e.g. the next snapshot below).

### 📝 Checklist

### DIFF
--- a/src/presets/netlify.ts
+++ b/src/presets/netlify.ts
@@ -85,7 +85,7 @@ async function writeRedirects(nitro: Nitro) {
   let contents = "/* /.netlify/functions/server 200";
 
   const rules = Object.entries(nitro.options.routeRules)
-    .sort((a, b) => b[0].split("/").length - a[0].split("/").length)
+    .sort((a, b) => a[0].split("/").length - b[0].split("/").length)
 
   // Rewrite static cached paths to builder functions
   for (const [key] of rules.filter(
--- a/src/presets/netlify.ts
+++ b/src/presets/netlify.ts
@@ -85,7 +85,7 @@ async function writeRedirects(nitro: Nitro) {
   let contents = "/* /.netlify/functions/server 200";
 
   const rules = Object.entries(nitro.options.routeRules)
-    .sort((a, b) => a[0].split("/").length - b[0].split("/").length)
+    .sort((a, b) => a[0].split(/\/(?!\*)/).length - b[0].split(/\/(?!\*)/).length)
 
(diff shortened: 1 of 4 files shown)


src/presets/netlify.ts
```diff
@@ -125,9 +126,10 @@ async function writeHeaders(nitro: Nitro) {
   const headersPath = join(nitro.options.output.publicDir, "_headers");
   let contents = "";
 
-  for (const [path, routeRules] of Object.entries(
-    nitro.options.routeRules
-  ).filter(([_, routeRules]) => routeRules.headers)) {
+  const rules = Object.entries(nitro.options.routeRules)
+    .sort((a, b) => b[0].split("/").length - a[0].split("/").length)
```
pi0, line 130: Small issue with sorting `/test/a` is more specific than `/test/*` but if * is before, it remains before and shadows.

src/presets/vercel.ts
```diff
@@ -173,10 +173,12 @@ function generateBuildConfig(nitro: Nitro) {
           src: key.replace(/^(.*)\/\*\*/, "(?<url>$1/.*)"),
           dest: generateEndpoint(key) + "?url=$url",
         })),
-      {
-        src: "/(.*)",
-        dest: "/__nitro",
-      },
+      ...(!nitro.options.routeRules['/**']?.cache || !(nitro.options.routeRules['/**'].cache.swr || nitro.options.routeRules['/**']?.cache.static)
```
pi0, line 176: Can you please explain it better in line above? Do we need to also handle other route rules for catch-all such as headers?

src/presets/netlify.ts
danielroe: nice catch - thank you ❤️

pi0 (approved):

pi0:
Thanks for the updates. About sorting, it should be same as radix3 structure while hoisting placeholders in regex, might be less accurate i think we can merge it to unblock downstream issue for nuxt.com :)

PR #758, opened 2022-12-15
fix(externals): keep both version of conflicting dependencies

mahdiboomeri:
### 🔗 Linked issue
resolves #722 
potentially resolves #612 

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
This PR keeps both versions of conflicting dependencies by nesting them inside their parent.
It also introduces a new nitro option called `optimizeExternals` which allows you to manually include or exclude specific packages to the optimization process. 

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #782, opened 2022-12-23
feat!: rewrite external copy with multi version hoisting support

### LINKED ISSUE #572: Issues with flatten extrernals tracing and multiple versions
On production builds, we trace used package files from `node_modules` and extract them to `.output/server/node_modules` but at the same time, flatten the structure by choosing the latest requested version of each package.

This strategy has two major issues:
 1. In the case of two major versions (sub-dependencies)
 2. In the case of two minor versions (eg #259) since we already traced, if the version is not compatible one will fail as we already traced paths back to their actual package path

For solving 1, only possible solution is hoisting sub-dependencies into `node_modules/parent/node_modules/child`. It can be also done by **respecting package manager structure without flattening it** or less fragile by doing what package managers do for hosting with more stable output with different pms!

For solving 2, we need to **re-trace internal package paths** in case of a conflict is detected as **external exports and API** of a package should be compatible in semver-minor/patch but since we already traced them to internal structure, it will be broken.

pi0:
Resolves #572

Based on amazing work from @mahdiboomeri (#758) this PR introduces a full rewrite of the external copy mechanism.

Previously if multiple versions of an external library were being used by traced externals, we used this strategy:
- In case of minor version conflicts, pick the latest to use for all libs
- In case of major version conflicts, warn and pick the latest semver-major one

This causes unresolvable issues especially when sub-dependencies are conflicting.

With this refactor, we respect hoisted structure of package manager and map it into `.output/server/node_modules`:
- Create a normalized list of all traced files and their parents
- Create a normalized list of all relevant package/versions from traced files
- For each package
  - If there is only one traced version detected, copy it as as
  - If more than one traced version is detected, copy to `node_modules/{name}@{version}` and symlink to `node_modules/{parent}/node_modules/{name}` for each parent. If no parent modules are detected, hoist it to top level as default (user's version)

In case of having issues with new implementation in nitropack v2, you can opt to use legacy externals:

```js
experimental: {
  // Use nitropack v1 behavior
  legacyExternals: true
}
```


pi0:
@mahdiboomeri If you check latest refactors, do you think we would still need an option for optimizeExternals?

mahdiboomeri:
@pi0 I added the options as a workaround if they were some edge cases that weren't covered (such as a package that doesn't follow semver), but IMO the new approach seems reliable enough to skip that.

src/rollup/plugins/externals.ts
```diff
@@ -197,102 +188,187 @@ export function externals(opts: NodeExternalsOptions): Plugin {
...
+                await writePackage(tracedPackage.name, version);
+              } else {
+                // Write alternative version into node_modules/{name}@{version}
+                await writePackage(
+                  tracedPackage.name,
+                  version,
+                  `${tracedPackage.name}@${version}`
+                );
+                // For each parent, link into node_modules/{parent}/node_modules/{name}
+                for (const parentPath of parentPkgs) {
```
mahdiboomeri, line 345: Linking breaks if the nested packages also have conflicting children.
Consider the following structure:

```
└── node_modules
    ├── nitro-dep-a
    │   └── nitro-lib#v1.0.0
    │       └── nested-lib#v1.0.0
    ├── nitro-dep-b
    │   └── nitro-lib#v2.0.0
    │       └── nested-lib#v2.0.0
    └── nitro-lib#v2.9.0 
```

itpropro:
Any updates for this @pi0 ?

src/rollup/plugins/externals.ts
pi0: Should be fixed in latest version using parent-level hoisting

pi0:
Updated implementation with parent-level hosting for nested support, windows support using junctions (not portable), and alias support.

Although new implementation should be much better for handling hoisting situations and allow better enhancements using new clearer data-structures, it can be buggy without testing on real-world installs. So let's get it a spin. 

Also there is new `experimental.legacyExternals` flag to allows an escape hatch to use v1 behavior.

Issue #787, opened 2022-12-28
Cloudflare page will always get 405 error for non-GET method


DanSnow:
### Environment

Nitro: v1.0.0
Node: v16.18.1

### Reproduction

repo: https://github.com/DanSnow/nitro-bug
Cloudflare Page: https://nitro-bug.pages.dev
1. create an API event handler
2. use POST request to the event handler
3. will get `405 method not allow` error

### Describe the bug

I think it's because here only test 404 error, but Cloudflare will respond 405 if the method is not GET here
https://github.com/unjs/nitro/blob/main/src/runtime/entries/cloudflare-pages.ts#L25

PR #796, opened 2023-01-03
fix: unable to handle non-GET request in cloudflare page

### LINKED ISSUE #497: Nuxt 3.0.0-rc.11 with Nitro 0.5.12> /Server/api functions fail on POST requests, GET methods are OK when publishing to Cloudflare pages functions)
### Environment

Nuxt 3.0.0-rc.10-27718444.40d0907 with Nitro 0.5.2-27718085.c24dbcf  
Cloudflare pages with functions

### Reproduction

https://stackblitz.com/edit/github-dxg5qg?file=server/api/customer.get.js

### Describe the bug

Deployment of my Nuxt 3 app that contains following folder server folder structure

/server/api/customer.post.js <- POST fail with 405 method not allowed on cloudflare pages
/server/api/customer.get.js <-- GET works.
/server/api/test.js  <-- should response to all routes

Nitro preset :
nitro: {
    preset: 'cloudflare_pages',
  },

Upon pushing to github, cloudflare initiates a page build with the following successful log entries on cf pages:

```

Nuxt 3.0.0-rc.11-27719556.94377b1 with Nitro 0.5.3-27719444.ff99330
--
18:54:51.303 | ℹ Client built in 2421ms
18:54:51.304 | ℹ Building server...
18:54:52.513 | ✔ Server built in 1209ms
18:54:52.572 | ✔ Generated public .output/public
18:54:52.588 | ℹ Building Nitro Server (preset: cloudflare_pages)
18:54:59.396 | ✔ Nitro server built
18:54:59.419 | └─ functions/path.js (414 kB) (132 kB gzip)
...
18:55:16.586 | Success: Assets published!
18:55:18.083 | Success: Your site was deployed!
```
When visiting the domain all looks good and page is published!
Now time to test the /functions folder

GET  https://www.domain.com/api/customer HTTP/1.1
 
RESULT : 200 OK

POST https://www.domain.com/api/customer HTTP/1.1

RESULT : 405 NOT ALLOWED

POST https://www.domain/api/incorrectpath HTTP/1.1

RESULT : 405 NOT ALLOWED (should return 404 NOT FOUND)

Talking to Cloudflare discord support they say <quote>
----
Let them know that if they want to discuss it then feel free to pop on here.
Cloudflare Pages never returns 405s, so this is Nitro
- No POST handlers work, either in the .post.js file or the .js file
- It works in dev (so in Vite), not prod
- The command (make sure they know the command is build not generate) 
---

Strangely when I POST to a route that does not exist like POST https://www.domain.com/api/blah, 
the result is also 405 NOT ALLOWED, hence it seems all POST requests are being block before the function is actually triggered, but for some reason all GET routes are OK.

**to recreate the issue**

- use the stackblitz template above : https://stackblitz.com/edit/github-dxg5qg?file=server/api/customer.get.js
- run `npm run build` , not generate, as generate will not create /functions folders
- `npx wrangler pages dev .output/public` OR `npx wrangler pages publish .output/public`
- visit xxx.com/api/customer  -- this will be a GET and it will work
- make a postman POST request to xx.com/api/customer - it will fail with 405 (if not pls let me know)

According to the nitro docs : [](https://nitro.unjs.io/deploy/providers/cloudflare) it should just work.

### Additional context

Cf discord channel has been informed of this issue and says they would be happy to connect with nuxt core team, please contact the discord community member @isaac-mcfadyen#4321  on cloudflare-pages discord channel.

DanSnow:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=kE2uAWh)

### 🔗 Linked issue

#497
#787

### ❓ Type of change

### 📚 Description

Fix an issue that nitro won't be able to handle POST, PUT, etc requests on the Cloudflare page

Resolves #497 #787

### 📝 Checklist


Plinpod:
When making these changes, I was able to get 200 responses for the API call however, the server API code never actually get executed. Have you run into that?

DaniFoldi:
👋 Hi,

Thanks for creating the PR to fix this - however I'm not sure if the approach you have taken is the one that should be used going forward. Pages will still invoke a Function for static assets, which defeats the purpose of using Pages, so instead, we should take advantage of their internal static asset handling.

For more context please see my comment [here](https://github.com/unjs/nitro/issues/497#issuecomment-1370686485)

src/runtime/entries/cloudflare-pages.ts
```diff
@@ -22,7 +22,7 @@ export async function onRequest(ctx: CFRequestContext) {
   try {
     // const asset = await env.ASSETS.fetch(request, { cacheControl: assetsCacheControl })
     const asset = await ctx.next();
```
danielroe, line 24:
```suggestion
    if (ctx.request.method === "GET") {
      const asset = await ctx.next();
```

src/runtime/entries/cloudflare-pages.ts
```diff
@@ -22,7 +22,7 @@ export async function onRequest(ctx: CFRequestContext) {
...
-    if (asset.status !== 404) {
+    if (asset.status < 400) {
       return asset;
     }
```
danielroe, line 27:
```suggestion
      if (asset.status !== 404) {
        return asset;
      }
    }
```

danielroe (changes requested):
I agree with @DaniFoldi that we need to do more than this, but regardless will still need to merge this to handle assets that are not prefixed or known in advance. The bug linked is caused by the fact that we shouldn't be getting assets by anything other than a GET request anyway so it may be better to add a condition.

Linking related PR to avoid hitting function for prefixed assets: https://github.com/unjs/nitro/pull/965

PR #799, opened 2023-01-03
fix(prerender): decode urls to allow comma in the `x-nitro-prerender` header

posixpascal:
Hey, first timer here 🎉.

### 🔗 Linked issue

https://github.com/nuxt/image/pull/701

### ❓ Type of change

### 📚 Description

When using `@nuxt/image-edge` with Nuxt 3 the default image provider is [IPX](https://github.com/unjs/ipx) which allows arbitrary image manipulation through URLs.

When using multiple transformations on an image the different transformations are seperated by a comma like this in the final image URL:
```
/_ipx/q_50,s_10x10/images/nitropack.png
```

`@nuxt/image-edge` appends these URLs to the `X-Nitro-Prerender` header without encoding the URL.
The header will then be split by "," when prerendering the site which leads to the following URLs being prerendered:
- _ipx/q_50
- s_10x10/images/nitropack.png

This results in some images not being properly prerendered and/or missing in the final output build when running `yarn run generate`

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -257,7 +257,10 @@ function extractLinks(
 
   // Extract from x-nitro-prerender headers
   const header = res.headers.get("x-nitro-prerender") || "";
-  _links.push(...header.split(",").map((i) => i.trim()));
+  _links.push(...header.split(",")
+      .map((i) => i.trim())
+      .map((i) => decodeURIComponent(i));
+
 
   for (const link of _links.filter(Boolean)) {
     const parsed = parseURL(link);


src/prerender.ts
```diff
@@ -257,7 +257,9 @@ function extractLinks(
 
   // Extract from x-nitro-prerender headers
   const header = res.headers.get("x-nitro-prerender") || "";
-  _links.push(...header.split(",").map((i) => i.trim()));
+  _links.push(...header.split(",")
+      .map((i) => decodeURIComponent(i))
+      .map((i) => i.trim()));
```
pi0, line 262: I think we should move trim before decode

posixpascal: Yes, good catch. I've updated the PR.

pi0 (approved):
THanks!

src/prerender.ts
```diff
@@ -257,7 +257,10 @@ function extractLinks(
 
   // Extract from x-nitro-prerender headers
   const header = res.headers.get("x-nitro-prerender") || "";
-  _links.push(...header.split(",").map((i) => i.trim()));
+  _links.push(...header.split(",")
+      .map((i) => i.trim())
+      .map((i) => decodeURIComponent(i));
```
MiniDigger, line 262: you are missing a ) here:
```suggestion
      .map((i) => decodeURIComponent(i)));
```

posixpascal: Oh my bad. Fixed it in #800. 😔

PR #807, opened 2023-01-06
fix: filter unique scanned handlers

### LINKED ISSUE #806: Only keep a single route/method handler when scanning directories
### Describe the feature

Right now, when using Nuxt layers to extend the server API endpoints, the "top-level" endpoints do not override "sub-level" endpoints; instead they're just being appended to the list of handlers. They're then being added to the server output file even though they will never be called as Nitro will first match the top-level handlers, which increases the file size and the transpiling process duration. A very simple fix would be to filter unique route/method pairs here to make sure only a single handler per route is being added: 
https://github.com/unjs/nitro/blob/1c4d6e1fba7be6923e80cefbde541f47fc9582f2/src/scan.ts#L20
 

### Additional information

yassilah:
### 🔗 Linked issue
Resolves #806 

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/scan.ts
+++ b/src/scan.ts
@@ -17,9 +17,14 @@ export async function scanHandlers(nitro: Nitro) {
     scanRoutes(nitro, "routes", "/"),
   ]).then((r) => r.flat());
 
-  nitro.scannedHandlers = handlers.flatMap((h) => h.handlers).filter(({ route, method }, index, array) => {
-    return array.findIndex((h) => h.route === route && h.method === method) === index;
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks 💯

yassilah:
Hi! I think this PR unintentionally caused that important issue: https://github.com/nuxt/nuxt/discussions/18515. I believe the filtering function should only be applied to api routes and not middlewares

PR #812, opened 2023-01-07
fix: resolve and include scanDirs within `node_modules` for auto import

yassilah:
### 🔗 Linked issue

Fixes a Nuxt related issue: https://github.com/nuxt/framework/issues/9961
I'm not sure this is the best way to fix it but this should at least allow for fully customizing that exclude array.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -5,7 +5,7 @@ import { klona } from "klona/full";
 import { camelCase } from "scule";
 import { defu } from "defu";
 import { resolveModuleExportNames, resolvePath as resolveModule } from "mlly";
-// import escapeRE from 'escape-string-regexp'
+import escapeRE from "escape-string-regexp";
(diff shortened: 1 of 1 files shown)


pi0:
Enabling auto imports for all node_modules can both impact performance and increase chances of unwanted auto import behavior.

Similar to Nuxt, we shall explicitly enable auto imports for `options.dirs` (~layers in nuxt 3)

yassilah:
Are folders explicitly set in `options.dirs` included despite `options.exclude`? e.g. would this still apply the plugin here: 

```ts
imports: {
    dirs: ['node_modules/foo/composables']
    exclude: [/node_modules/]
}
```

yassilah:
Thanks so much for taking the time 💛

PR #819, opened 2023-01-09
fix(build): publicAssets priority on build

antfu:
### ❓ Type of change

### 📚 Description

#817 was not a complete fix. As Vite will make an initial copy for the public file, ignoring `overrides` will instead keep the version from Vite's initial copy.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -36,12 +36,11 @@ export async function copyPublicAssets(nitro: Nitro) {
   if (nitro.options.noPublicDir) {
     return;
   }
-  for (const asset of nitro.options.publicAssets) {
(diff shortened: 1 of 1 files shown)


antfu:
I think we could disable the public copy for Vite, but I am afraid it will affect how asset paths are resolved by Vite. Removing the public before copy might not work as Vite will merge the bundle and assets, which would be hard to identify.

I will do a bit more research to see if it's possible on Vite side.

antfu:
It seems there is a `copyPublicDir` options introduced for SvelteKit, close in favor of https://github.com/nuxt/framework/pull/10013

PR #823, opened 2023-01-10
fix(rumtime): disable server-timing header via options.timing

clarkdo:
### 🔗 Linked issue

Fix https://github.com/nuxt/framework/issues/9824

### ❓ Type of change

- [x]   📖 Documentation (updates to the documentation or readme)
- [x]  🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

`timing` config is for enabling bundles loading time instead of timing header, this pr is improving the accuracy of description.

### 📝 Checklist

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -40,7 +40,7 @@ function createNitroApp(): NitroApp {
     onError: errorHandler,
   });
 
-  if(config.timing) {
+  if (config.timing) {
     h3App.use(config.app.baseURL, timingMiddleware);
   }
 
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -14,7 +14,6 @@ import {
 } from "unenv/runtime/fetch/index";
 import { createHooks, Hookable } from "hookable";
 import { useRuntimeConfig } from "./config";
-import { timingMiddleware } from "./timing";
 import { cachedEventHandler } from "./cache";
 import { createRouteRulesHandler, getRouteRulesForPath } from "./route-rules";
 import { plugins } from "#internal/nitro/virtual/plugins";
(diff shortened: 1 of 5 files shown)


src/runtime/app.ts
```diff
@@ -40,7 +40,9 @@ function createNitroApp(): NitroApp {
     onError: errorHandler,
   });
 
-  h3App.use(config.app.baseURL, timingMiddleware);
+  if (config.timing) {
```
danielroe, line 43: config in this context refers to `runtimeConfig` and not the nitro config.

I agree it would be good to remove the header, but ideally as part of the build and not at runtime (wdyt @pi0?). At the moment the only if/then that happens is this:

https://github.com/unjs/nitro/blob/5831707ca653a1383fe52a0212ec9247424e8601/src/rollup/config.ts#L137-L139

danielroe: it feels like it would be best to implement timing as a nitro plugin so it doesn't need to be present in the `app` file at all.

clarkdo: @danielroe Thanks for the hint, I've moved it to plugin, is this what you suggested ?

src/runtime/timing.ts
```diff
@@ -1,32 +1,37 @@
...
   start: () => 0,
   end: () => 0,
   metrics: [],
 };
 
-// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing
-export const timingMiddleware = eventHandler((event) => {
-  const start = globalTiming.start();
+export const timingPlugin = defineNitroPlugin((nitro) => {
+  nitro.h3App.use(
```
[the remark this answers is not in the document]
clarkdo: Good point, I've made change, does it make sense ?

It is still in the new start phase and needs more familiarity with the new architecture.

src/runtime/app.ts
danielroe: exactly! ❤️

src/runtime/timing.ts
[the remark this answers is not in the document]
danielroe: this looks really nice to me.

pi0 (approved):
Thanks 💯 Nice idea using plugins to opt-in

PR #828, opened 2023-01-11
refactor: update `event.` to `event.node.`

cpreston321:
### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

This will resolve deprecated use of `event.req|res` to the new way which is `event.node`. To make sure it stays up to date.

### 📝 Checklist

### DIFF
--- a/src/dev/error.ts
+++ b/src/dev/error.ts
@@ -1,14 +1,14 @@
 import { NitroErrorHandler } from "../types";
 
 function errorHandler(error, event) {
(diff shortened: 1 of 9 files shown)


pi0 (approved):
Thanks. (on a related note, the main goal of deprecating was to encourage using utilities instead of interacting with low-level `req/res`. we can progressively do this)

PR #829, opened 2023-01-11
fix(netlify, vercel): explicit server rendering with disabled cache/swr

danielroe:
### ❓ Type of change

### 📚 Description

At the moment if you have a default cache rule, you can't override it negatively at a more specific level.

This PR should resolve that.

### 📝 Checklist

### DIFF
--- a/src/presets/netlify.ts
+++ b/src/presets/netlify.ts
@@ -89,12 +89,18 @@ async function writeRedirects(nitro: Nitro) {
   );
 
   // Rewrite static cached paths to builder functions
-  for (const [key] of rules.filter(
-    ([_, routeRules]) =>
-      routeRules.cache && (routeRules.cache?.static || routeRules.cache?.swr)
+  for (const [key, value] of rules.filter(
+    ([_, value]) =>
+      value.cache === false ||
+      (value.cache && value.cache.swr === false) ||
+      (value.cache && (value.cache?.static || value.cache?.swr))
   )) {
     contents =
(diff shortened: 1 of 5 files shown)


Atinux:
Related to https://github.com/nuxtlabs/nuxt.com/blob/1b11f30460c93c39e8edd35155a68fe269c5d3d3/nuxt.config.ts#L164

src/presets/netlify.ts
```diff
@@ -89,12 +89,18 @@ async function writeRedirects(nitro: Nitro) {
...
 
   // Rewrite static cached paths to builder functions
-  for (const [key] of rules.filter(
-    ([_, routeRules]) =>
-      routeRules.cache && (routeRules.cache?.static || routeRules.cache?.swr)
+  for (const [key, value] of rules.filter(
+    ([_, value]) =>
+      value.cache === false ||
+      (value.cache && value.cache.swr === false) ||
+      (value.cache && (value.cache?.static || value.cache?.swr))
```
pi0, line 96: I think this can be simplified for swr check only.

---

Asking ChatGPT: 🤓 

image

image

pi0 (approved):
LGTM for vercel hotfix. But we really need to improve cache<>builder mapping system and make it generic and better testable in the future...

src/presets/netlify.ts
pi0: Yeah. The main problem is that we are not deeply normalizing the route rules at the moment and with _undifined_ conditions, such simplification is wrong. Cache should be either `false` or an object. And I'm thinking the static flag could be moved outside. Lets improve later.

PR #837, opened 2023-01-13
ci: run tests against windows as well

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

There are some significant differences with windows, particularly around file URLs and absolute paths. This is a starting point to test nitro on Windows to catch these issues early.

Here are two key differences noticed so far:

- hash function (for etag) gives different results - this may be okay as I'm not sure we need consistency across OSes
- dependency hoisting appears to be broken - may be resolved with https://github.com/unjs/nitro/pull/782

cc: @pi0 

### 📝 Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.txt	text eol=lf
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,10 @@ on:
 
 jobs:
   ci:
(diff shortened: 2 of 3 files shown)


pi0:
Nice. With #782 I can confirm at leat the tests pass however the way we link modules using junction links is not portable. So it is false report almost. But I think we can accept it as a known limitation at least.

Re etags, my guess is CRLF issue. Which could be fixed.

PR #857, opened 2023-01-20
feat: support `staleMaxAge: -1` to always respond stale value

### LINKED ISSUE #856: [feature]: add a no-stale option to allow sending stale value
### Describe the feature

Currently, SWR option works great but it would be great to allow sending a stale value, while updating the cache in the background to avoid any downtime or timeout requests when querying large amounts of data. It also allows deploying to Vercel on the Hobby plan, which has a 10s task execution limit, while still querying updated data in the background 😊

### Additional information

yassilah:
### 🔗 Linked issue

Resolves #856 

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -78,7 +78,7 @@ export function defineCachedFunction<T = any>(
 
     const _resolve = async () => {
       if (!pending[key]) {
-        if (entry.value && (opts.staleMaxAge ?? 0) > 0) {
+        if (entry.value && (opts.staleMaxAge ?? 0) >= 0) {
           // Remove cached entry to prevent using expired cache on concurrent requests
           entry.value = undefined;
           entry.integrity = undefined;
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -29,6 +29,7 @@ export interface CacheOptions<T = any> {
   swr?: boolean;
   staleMaxAge?: number;
   base?: string;
+  allowStale?: boolean;
 }
 
 const defaultCacheOptions = {
@@ -36,6 +37,7 @@ const defaultCacheOptions = {
   base: "/cache",
   swr: true,
   maxAge: 1,
+  allowStale: false
 };
 
 export function defineCachedFunction<T = any>(
@@ -78,11 +80,13 @@ export function defineCachedFunction<T = any>(
 
     const _resolve = async () => {
       if (!pending[key]) {
-        // Remove cached entry to prevent using expired cache on concurrent requests
-        entry.value = undefined;
-        entry.integrity = undefined;
-        entry.mtime = undefined;
-        entry.expires = undefined;
+        if (entry.value && !opts.allowStale) {
+          // Remove cached entry to prevent using expired cache on concurrent requests
+          entry.value = undefined;
+          entry.integrity = undefined;
+          entry.mtime = undefined;
+          entry.expires = undefined;
+        }
         pending[key] = Promise.resolve(resolver());
       }
       entry.value = await pending[key];
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -37,7 +37,7 @@ const defaultCacheOptions = {
   base: "/cache",
   swr: true,
   maxAge: 1,
-  allowStale: false
+  allowStale: false,
 };
 
 export function defineCachedFunction<T = any>(
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -78,7 +78,7 @@ export function defineCachedFunction<T = any>(
 
     const _resolve = async () => {
       if (!pending[key]) {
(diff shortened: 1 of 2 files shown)


src/runtime/cache.ts
```diff
@@ -78,11 +80,13 @@ export function defineCachedFunction<T = any>(
...
+        if (entry.value && !opts.allowStale) {
+          // Remove cached entry to prevent using expired cache on concurrent requests
+          entry.value = undefined;
+          entry.integrity = undefined;
+          entry.mtime = undefined;
+          entry.expires = undefined;
+        }
         pending[key] = Promise.resolve(resolver());
       }
       entry.value = await pending[key];
```
yassilah, line 92: Actually, this is not relevant for this PR but I just noticed that everything after the pending request is done should probably be done only once after the first time is it requested, no? Because right now it seems that if I request twice the same data, it is only resolving the promise once, but it is then setting `mtime`, `integrity`, validating and setting the result in the storage for each call; is it intended?

src/runtime/cache.ts
```diff
@@ -78,11 +80,13 @@ export function defineCachedFunction<T = any>(
 
     const _resolve = async () => {
       if (!pending[key]) {
-        // Remove cached entry to prevent using expired cache on concurrent requests
-        entry.value = undefined;
-        entry.integrity = undefined;
-        entry.mtime = undefined;
-        entry.expires = undefined;
+        if (entry.value && !opts.allowStale) {
```
pi0, line 83: Does it makes sense that we directly use `!opts.swr` here?

src/runtime/cache.ts
pi0: Nice spot. However both value and time would be the same when promise resolves, we might move to to `Promise.resolve(resolver()).then(() => {...})` to do once

yassilah: Yes I was more thinking in terms of validating + setting the value in storage, which could be expensive especially if the storage driver is not local

src/runtime/cache.ts
yassilah: You mean instead of `entry.value && !opts.allowStale`?

src/runtime/cache.ts
pi0: Indeed putting lock on remote storage to sync state would be expensive. We can however support later a flag that keeps initial time in promise and before writing to cache check the meta first and avoid write if there is a newer entity. considering cost at write.

yassilah: In the meantime couldn't we simply do something like:

```ts
const isPending = pending[key] // check whether there was a key (false the first time, true every other time)

if (!isPending) {
    ...
    pending[key] = Promise.resolve(resolver())
}

entity.value = await pending[key]

if (!isPending) {
    // do everything but only if there wasn't a key to begin with
}
```

src/runtime/cache.ts
pi0: Yes. I suppose it should be expected behavior of SWR to be nonblocking after `maxAge`. (To be precise, also before `staleMaxAge` but it is not implemented yet in handler)

pi0: You are definitely not wrong. We use cache options for both Browser<>Server And Server<>Handler caching strategy. And these utils handle Server<>Handler SWR caching strategy _similar_ to browsers (regardless of HTTP client which could be another API or browser)

----

SWR Cache Timeline: `0....maxAge` (A) .... `maxAge...staleMaxAge` (B) ... after maxAge/staleMaxAge (C)

Stage A: Fresh Cache. Use value without revalidate
Stage B: Stale Cache. Respond and revalidate in background
Stage C: Expired Cache. Block and revalidate

Since we didn't implement `staleMaxAge` yet, it is either B or C. New `allowStale` flag is identical to `staleMaxAge: Infinity`.

How about supporting `staleMaxAge: -1` as a way to support without new flag?

yassilah: Alright, yes I do see what you mean and it would make sense to use the existing `staleMaxAge` flag to avoid multiplying the options; i'll update the PR

src/runtime/cache.ts
```diff
@@ -78,11 +78,13 @@ export function defineCachedFunction<T = any>(
 
     const _resolve = async () => {
       if (!pending[key]) {
-        // Remove cached entry to prevent using expired cache on concurrent requests
-        entry.value = undefined;
-        entry.integrity = undefined;
-        entry.mtime = undefined;
-        entry.expires = undefined;
+        if (entry.value && (opts.staleMaxAge || 0) >= 0) {
```
pi0, line 81: (falsy values such as `0`, `false` and `null` can also be a value for result)

```suggestion
        if (entry.value !== undefined && (opts.staleMaxAge || 0) >= 0) {
```

pi0 (approved):

pi0:
Thanks!

PR #860, opened 2023-01-20
feat: support `maxAge` for public assets

pi0:
Allow using `maxAge` for `publicAssets: []` option to provide caching headers:

```js
  publicAssets: [
    {
      baseURL: "dist",
      dir: "public/dist",
      maxAge: 3600,
    },
  ],
```

Remarks:
- Using route rules in production we can also map caching to native providers
- Default (and unused) 60sec is removed for public dirs allowing opt-in control of this feature
- This feature is auto-disabled for `/` prefix and `fallthrough` paths since another handler might respond
- This feature is only effective on production to reduce chance of misconfiguration and development mode issues
- `immutable` and `public` conditions enabled
- One can directly use routeRules to apply for development or any other custom beahior 

Another fix is also included that allows to modify `nitro.options.routeRules` after init syncing with runtimeConfig object.

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -134,7 +134,15 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
   // Serve asset dirs
   for (const asset of nitro.options.publicAssets) {
     const url = joinURL(nitro.options.runtimeConfig.app.baseURL, asset.baseURL);
-    app.use(url, fromNodeMiddleware(serveStatic(asset.dir)));
+    app.use(
+      url,
+      fromNodeMiddleware(
+        serveStatic(asset.dir, {
+          maxAge: asset.fallthrough ? undefined : asset.maxAge * 1000,
+          immutable: asset.fallthrough ? undefined : true,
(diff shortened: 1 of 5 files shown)


gabrielcipriano:
what is fallthrough in this context?

pi0:
When a request comes to a public asset dirs prefix with fallthrough enabled, and not matching, router continues to make a SSR request (useful for cases when a base URL is shared for public assets and API routes)

gabrielcipriano:
Cool, greate stuff!!

kkukelka:
@pi0 is this supposed to work for SSG as well? if yes, is there anything regarding the `dir` and `baseUrl` paths I have to be aware of?

pi0:
in SSG, there is no Nitro server running. Alotugh since we integrate with route rules, for the supported platforms (vercel, netlify, cf pages) we generated static rules for maxAge to be handled by their edge CDN.

Default TTL is not set for base asset dirs with `/` only to prevent unwanted caching issues.

PR #861, opened 2023-01-20
fix: only update cache in storage once per pending request

### LINKED ISSUE #859: bug: cache is updated multiple times per pending request
Actually, this is not relevant for this PR but I just noticed that everything after the pending request is done should probably be done only once after the first time is it requested, no? Because right now it seems that if I request twice the same data, it is only resolving the promise once, but it is then setting `mtime`, `integrity`, validating and setting the result in the storage for each call; is it intended?

_Originally posted by @yassilah in https://github.com/unjs/nitro/pull/857#discussion_r1082442157_

yassilah:
### 🔗 Linked issue

Resolves #859

### ❓ Type of change

### 📝 Checklist


pi0 (approved):
Thanks!

PR #865, opened 2023-01-23
feat(externals): support tracing orphan files in `node_modules`

pi0:
### ❓ Type of change

### 📚 Description

There are some files in `node_modules` without a parent `package.json` (perisma for example generates `node_modules/.prisma` without `.prisma/package.json`.

This PR supports tracing and copying them using a virtual package

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/externals.ts
+++ b/src/rollup/plugins/externals.ts
@@ -239,34 +239,45 @@ export function externals(opts: NodeExternalsOptions): Plugin {
         versions: Record<
           string,
           {
+            pkgJSON: Record<string, any>;
             path: string;
             files: string[];
           }
         >;
       };
       const tracedPackages: Record<string, TracedPackage> = {};
-      await Promise.all(
-        Object.values(tracedFiles).map(async (tracedFile) => {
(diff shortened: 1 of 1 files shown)


src/rollup/plugins/externals.ts
```diff
@@ -239,34 +239,45 @@ export function externals(opts: NodeExternalsOptions): Plugin {
...
-          if (!tracedPackageVersion) {
-            tracedPackageVersion = { path: tracedFile.pkgPath, files: [] };
-            tracedPackage.versions[pkgJSON.version] = tracedPackageVersion;
-          }
-          tracedPackageVersion.files.push(tracedFile.path);
-          tracedFile.pkgName = pkgName;
-          tracedFile.pkgVersion = pkgJSON.version;
-        })
-      );
+      for (const tracedFile of Object.values(tracedFiles)) {
```
manniL, line 249: is it intentional that this is not running in parallel anymore?

pi0: Yes we need to sync virtual packages and include all traced files.

pi0: Copy is still parallel

PR #867, opened 2023-01-23
feat(externals): apply `production` condition to package.exports

### LINKED ISSUE #844: apply `production` contraint to traced `package.json` files
We trace and extract package files with `production` condition however node, does not apply it by default without passing `-C production` flag and we cannot always pass a node.js flag.

When extracting packages, we shall apply it to `exports` that production is picked by default.

More context: https://github.com/unjs/nitro/pull/605#issuecomment-1286660793

pi0:
### 🔗 Linked issue

resolves #844

### ❓ Type of change

### 📚 Description

Rewrite `package.json` with production condition in exports field before writing to be consistent with the condition of resoluton before. Tested against two linked packages.

### 📝 Checklist

### DIFF
--- a/test/unit/externals.test.ts
+++ b/test/unit/externals.test.ts
@@ -1,0 +1,85 @@
+import { expect, describe, it } from "vitest";
+import { applyProductionCondition } from "../../src/rollup/plugins/externals";
+
+describe("externals:applyProductionCondition", () => {
+  const applyProductionConditionCases = [
+    {
+      name: "vue-router@4.1.6",
+      in: {
+        ".": {
+          types: "./dist/vue-router.d.ts",
+          node: {
+            import: {
+              production: "./dist/vue-router.node.mjs",
+              development: "./dist/vue-router.node.mjs",
+              default: "./dist/vue-router.node.mjs",
+            },
+            require: {
+              production: "./dist/vue-router.prod.cjs",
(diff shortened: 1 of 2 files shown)


test/unit/externals.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    {
+      name: "vue-router@4.1.6",
+      in: {
+        ".": {
+          types: "./dist/vue-router.d.ts",
+          node: {
+            import: {
+              production: "./dist/vue-router.node.mjs",
+              development: "./dist/vue-router.node.mjs",
+              default: "./dist/vue-router.node.mjs",
```
danielroe, line 15:
```suggestion
              production: "./dist/vue-router.node.mjs",
              development: "./dist/vue-router.dev.node.mjs",
              default: "./dist/vue-router.dev.node.mjs",
```

danielroe (approved):
Nice. 👌

test/unit/externals.test.ts
pi0: I took it from here https://unpkg.com/browse/vue-router@4.1.6/package.json it doesn't have dev however would be nice if it had. Are you suggesting this for test clarity?

danielroe: Exactly. Otherwise it doesn't show a difference between in and out.

PR #870, opened 2023-01-24
docs(netlify): clarify placement of `_redirects` file

OlaAlsaker:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.deploy/providers/netlify.md
+++ b/docs/content/2.deploy/providers/netlify.md
@@ -18,7 +18,7 @@ For new sites, Netlify will detect that you are using Nitro and set the publish 
 
 If you are upgrading an existing site you should check these and update them if needed.
 
-If you want to add custom redirects, you can do so with [`routeRules`](/config#routerules) or by adding a [`_redirects`](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) file.
+If you want to add custom redirects, you can do so with [`routeRules`](/config#routerules) or by adding a [`_redirects`](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) file to your `public` directory.
 
 For deployment, just push to your git repository [as you would normally do for Netlify](https://docs.netlify.com/configure-builds/get-started/).
 


docs/content/2.deploy/providers/netlify.md
```diff
@@ -18,7 +18,7 @@ For new sites, Netlify will detect that you are using Nitro and set the publish
 
 If you are upgrading an existing site you should check these and update them if needed.
 
-If you want to add custom redirects, you can do so with [`routeRules`](/config#routerules) or by adding a [`_redirects`](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) file.
+If you want to add custom redirects, you can do so with [`routeRules`](/config#routerules) or by adding a [`_redirects`](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) file to your `public` directory.
```
manniL, line 21: Actually, I'd suggest to *only* mention the `routeRules` here as it is a vendor-independent way for designing redirects which will work even when changing the deploy preset.

Any thoughts @pi0 @danielroe?

OlaAlsaker: In that case, `routeRules` will need a more descriptive documentation. I tried `routeRules`, but it did not solve the issue I had. I had to use `_redirects`. Just bringing my thoughts 😄

OlaAlsaker: To be more precise: in my project, I have a completely separated SPA that lives in the `public/admin` directory. As with other SPAs, I have to make a redirect rule that work like this:
```
/admin/*   ->   /admin/index.html
```

I tried using `routeRules` for this, but did not succeed.

pi0: I'm pretty sure this can be achieved using route rules too. We can keep this as last resort (howver consider the file might be overriden)

Do you mind to make a PR @manniL to add it in addition?

pi0 (approved):

docs/content/2.deploy/providers/netlify.md
manniL: Will do ☺️

PR #879, opened 2023-01-25
feat(vercel): auto-detect runtime version

smaeda-ks:
Currently, it's setting `nodejs16.x` statically, so any serverless functions are deployed with Node.js v16.x regardless of the project's Node version setting:
https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version

This PR changes the behavior and starts producing the `.vc-config.json` file (for Vercel Build Output API) with the running OS's Node runtime version. This prevents Node version mismatch between build time and runtime.

### ❓ Type of change

### 📝 Checklist


src/types/presets.ts
```diff
@@ -45,5 +45,8 @@ export interface VercelBuildConfigV3 {
 export interface PresetOptions {
   vercel: {
     config: VercelBuildConfigV3;
+    function: {
+      runtime: string;
```
pi0, line 49: Since we are only dealing with node, maybe simplify option to `nodeVersion`?

src/presets/vercel.ts
```diff
@@ -27,12 +27,16 @@ export const vercel = defineNitroPreset({
       const buildConfig = generateBuildConfig(nitro);
       await writeFile(buildConfigPath, JSON.stringify(buildConfig, null, 2));
 
+      const systemNodeVersion = process.versions.node.split('.')[0];
```
pi0, line 30: This is smart! However, I think the Node.js version during the build (both locally and the one Vercel/CI) does use, can differ. I'm wondering if they provide an environment variable helper.

src/types/presets.ts
smaeda-ks: Good point! My intention was because this setting essentially allows users to set different Node versions between runtime version at build time and serverless environment (although they should be the same), I wanted to make it clear that it is intended for serverless functions, if that makes sense...!

src/presets/vercel.ts
smaeda-ks: Hmm, I don't think there's such an environment variable support as of today 🤔 AFAIK, when building a Nuxt 3 project on Vercel (Zero-Config), their virtual environment would pick up the desired Node version either from the project settings or `package.json` file at its init stage: https://vercel.com/docs/concepts/functions/serverless-functions/runtimes/node-js#node.js-version
so reading `process.versions` at build time should tell the right Node version, and the above documentation also mentions the use of `process.versions`.

I'm not too concerned about the local build, as they should be using the same major Node version. For unique environments where it's the concern, they can set the new `vercel.function.runtime` option, as it takes precedence over the auto-detected Node version.

styfle: What about using `import { getNodeVersion } from '@vercel/build-utils'` helper function?

Right now that is expected to be used from a Builder so it accepts `config` and `meta`, but we can probably make this work without those parameters for use with Build Output API like this use case.

Would that solve the problem?

smaeda-ks: @styfle Thank you! Sounds like a better plan. Are you going to make some changes there?

src/types/presets.ts
pi0: `runtimeNodeVersion` should also be a good option 👍🏼 (i mainly like to avoid nesting config when not necessary)

src/presets/vercel.ts
pi0: It seems landed on `@vercel/build-utils@6.0.0`. Checking implementation, the utility is using `engines` field from the nearest `package.json` + checking against vercel supported ranges 👌🏼 (if v6 is not being used, you can temporary use `getPackageJSON` from `pkg-types` which is already installed as Nitro dep with same behavior)

styfle: Since there is no `config` object, we can set the default here: https://github.com/vercel/vercel/pull/9332

bastian-friedrich:
@smaeda-ks any news on this PR? We really need this in a prod env.

pi0:
@smaeda-ks Can we split auto detection part to merge new configuration sooner? 🙏🏼

src/types/presets.ts
styfle: @smaeda-ks @pi0 I think this is a bad idea. The build time an runtime must match. So you shouldn't let the user configure it unless you can also swap out the build version of Node.js too.

bastian-friedrich: @styfle with this approach, you theoretically could:
- set the node version in vercel / engines: node in package.json (build)
- set vercel node version with new config param (runtime)
That would lead to different node versions by configuration. Not sure if anybody needs that but it would probably ship a faster "fix" to our problem right now, like @pi0 mentioned in his comment above.

pi0: Is there any special reason for this matching @styfle other than natives matching? Nitro uses rollup/esbuild which can generates backward/forward compatible syntax based on expected runtime.

styfle: @pi0 The reason this PR exists at all is because there was a mismatch between build and runtime. The reason mismatches cause problems native `dependencies` will break (basically any dependency that isn't pure JS).

pi0: Then maybe we should rename the option to `vercel.nodeVersion` and make a runtime warning if build-time node is not matching?

styfle: That sounds painful for the user. I suggest using `@vercel/build-utils` to set the correct PATH env before executing any `npm install` or `npm run build`. 

Here's an example:

https://github.com/vercel/vercel/blob/e8385566fad2ff56efec64fd7250909789662790/packages/next/src/index.ts#L258-L259

smaeda-ks: Thanks! Going to remove this additon and use the `getNodeVersion` helper function instead👍

src/presets/vercel.ts
smaeda-ks: Thanks @styfle @pi0!
Updated this PR to use `@vercel/build-utils`.

smaeda-ks:
@pi0 @styfle Could you please review this updated version? Thanks!

src/presets/vercel.ts
```diff
@@ -2,6 +2,7 @@ import fsp from "node:fs/promises";
 import { dirname, relative, resolve } from "pathe";
 import { defu } from "defu";
 import { withoutLeadingSlash } from "ufo";
+import vercelBuildUtils from '@vercel/build-utils';
```
styfle, line 5:
```suggestion
import { getNodeVersion } from '@vercel/build-utils';
```

src/presets/vercel.ts
```diff
@@ -27,12 +28,15 @@ export const vercel = defineNitroPreset({
       const buildConfig = generateBuildConfig(nitro);
       await writeFile(buildConfigPath, JSON.stringify(buildConfig, null, 2));
 
+      const { getNodeVersion } = vercelBuildUtils;
```
styfle, line 31:
```suggestion
```

src/presets/vercel.ts
```diff
@@ -27,12 +28,15 @@ export const vercel = defineNitroPreset({
       const buildConfig = generateBuildConfig(nitro);
```
styfle, line 28: Is there some build command that gets executed like `nuxt build`? 

We need to make sure that its spawned with the correct PATH so it uses the correct version of Node.js

src/presets/vercel.ts
smaeda-ks: Thanks, @styfle 
I get the following with your suggestion
```
 ERROR  Named export 'getNodeVersion' not found. The requested module '@vercel/build-utils' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@vercel/build-utils';
const { getNodeVersion } = pkg;
```

src/presets/vercel.ts
smaeda-ks: I'm not sure if that's something nitro should be thinking of. This PR is about to generate a `.vc-config.json` file with the right Node version in it, and I don't think Nuxt would use it for their own build.

styfle: In that case, you should change this to use `process.node.version` like you had before, but you could compare against some allow list `[18,16,14]`. Otherwise you could end up with a mismatch between build time and runtime.

smaeda-ks: @styfle Thank you!
Just to be clear, are you suggesting not to use `getNodeVersion` (was suggested by you, though)? I'm a bit confused about the next action here, so your advice would be appreciated! :bow: 

> Otherwise you could end up with a mismatch between build time and runtime.

This is the case for local builds, correct? So `getNodeVersion` doesn't take into account the running OS's Node version but pulls it either from the `package.json` file or Vercel project settings, so if the local Node version differs from them that's the problem.

> you could compare against some allow list [18,16,14]

Can this be optional? I believe Vercel validates that anyway upon deployments and would throw an error. So I'm not sure if nitro has to double-check that.

Once it's confirmed, I'm fine with bringing back my initial change (without additional config options).

pi0:
Hi! Is there any updates on this? Today we are going to release Nitropack 2.1 / Nuxt 3.2

styfle:
Do we have example code that reproduces this issue? How can we confirm this PR fixes the issue?

src/presets/vercel.ts
smaeda-ks: ok, I'm going to revert the changes now cc: @pi0

pi0:
A local build with node v16.19.0 results into this output:

```
NITRO_PRESET=vercel p nitro build playground

> nitropack@2.1.1 nitro /Users/pooya/Code/nitro
> jiti ./src/cli.ts "build" "playground"

✔ Generated public playground/.vercel/output/static                                                                                        nitro 15:31:15
ℹ Building Nitro Server (preset: vercel)                                                                                                   nitro 15:31:15
✔ Nitro server built                                                                                                                       nitro 15:31:15
  ├─ playground/.vercel/output/functions/__nitro.func/package.json (496 B) (253 B gzip)
  ├─ playground/.vercel/output/functions/__nitro.func/index.mjs (16.6 kB) (5.03 kB gzip)
  ├─ playground/.vercel/output/functions/__nitro.func/chunks/index.mjs.map (186 B) (164 B gzip)
  └─ playground/.vercel/output/functions/__nitro.func/chunks/index.mjs (160 B) (150 B gzip)
Σ Total size: 579 kB (140 kB gzip)
Warning: Detected "engines": { "node": "^14.16.0 || ^16.11.0 || ^17.0.0 || ^18.0.0 || ^19.0.0" } in your `package.json` that will automatically upgrade when a new major Node.js Version is released. Learn More: http://vercel.link/node-version
```

Internal detected:

```
{ nodeVersion: { major: 18, range: '18.x', runtime: 'nodejs18.x' } }
```

(warning is from [monorepo engined](https://github.com/unjs/nitro/blob/860aa76f2786c6f53a6b875bdf83ef2e121c1aaa/package.json#L132))

@styfle Can you please confirm if this seems an expected behavior? Would be happy to move and test with edge release channel.

src/presets/vercel.ts
smaeda-ks: done https://github.com/unjs/nitro/pull/879/commits/8e90828656d1cb6ce238a98594cba6271161f21c

pi0 (approved):
I trust `process.versions.node.split('.')[0]` is expected behavior from vercel..

styfle (approved):
Looks good, thanks! 

Lets try it out with a real deployment and try setting `engines` to see if that is respected.

smaeda-ks:
Tested that the PR works as expected: https://github.com/smaeda-ks/nuxt3-vercel-nitropack-pr-879/pull/1

preview: https://nuxt3-vercel-nitropack-pr-879-git-test-smaeda-ks.vercel.app/api/hello

bastian-friedrich:
Awesome! It works! :) Thank you guys so much!

PR #892, opened 2023-01-28
feat: add generation of route input types

### LINKED ISSUE #841: Types for route inputs
### Describe the feature

Since Nitro generates some types for the response of routes, it would be great if it could also generate types for the input.
As an example:
```typescript
// routes/index.ts
// This is just a normal eventhandler
    const { greet } = (await readBody(event)) satisfies Input;
    return {
        hello: greet,
    };
});
// here comes the fun part:
export type Input = {
    greet: string;
};
```

Nitro already generates the types for the response from the event handler:
```typescript
// .nitro/types/nitro.d.ts
interface InternalApi {
    '/': Awaited<ReturnType<typeof import('../../routes/index').default>>
}
```

With the types for the input this could look something like this:
```typescript
interface InternalApi {
    '/': Awaited<ReturnType<typeof import('../../routes/index').default>>
}
interface InternalApiInputs {
    '/': import("../../routes/index").Input
}
```

### Additional information

Blechlawine:
### 🔗 Linked issue

#841 

### ❓ Type of change

### 📚 Description
Add generated types for inputs to routes. A more detailed explanation is in the issue #841.

Resolves #841 

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -67,6 +67,10 @@ export async function writeTypes(nitro: Nitro) {
     string,
     Partial<Record<RouterMethod | "default", string[]>>
   > = {};
+  const routeInputTypes: Record<
+    string,
+    Partial<Record<RouterMethod | "default", string[]>>
+  > = {};
 
   const middleware = [...nitro.scannedHandlers, ...nitro.options.handlers];
 
@@ -82,13 +86,22 @@ export async function writeTypes(nitro: Nitro) {
     if (!routeTypes[mw.route]) {
       routeTypes[mw.route] = {};
     }
+    if (!routeInputTypes[mw.route]) {
+      routeInputTypes[mw.route] = {};
+    }
 
     const method = mw.method || "default";
     if (!routeTypes[mw.route][method]) {
       routeTypes[mw.route][method] = [];
     }
+    if (!routeInputTypes[mw.route][method]) {
+      routeInputTypes[mw.route][method] = [];
+    }
     routeTypes[mw.route][method].push(
       `Awaited<ReturnType<typeof import('${relativePath}').default>>`
+    );
+    routeInputTypes[mw.route][method].push(
+      `import('${relativePath}').Input`
     );
   }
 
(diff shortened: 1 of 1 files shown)


danielroe (commented):
What about using a generic for `defineEventHandler`? We could have:

```js
export default defineEventHandler<{ id: string }>(event => {
  const data = await readBody(event)
  // data is typed
})
```

This would need to be implemented in https://github.com/unjs/h3 (with backwards-compatible generic implementation), and then this PR could be (very slightly) updated to use the generic instead of exported `Input` type.

Blechlawine:
> What about using a generic for `defineEventHandler`? We could have:
> 
> ```js
> export default defineEventHandler<{ id: string }>(event => {
>   const data = await readBody(event)
>   // data is typed
> })
> ```

This would be ideal, but how would we access the generic type through the type of the exported eventHandler?

danielroe:
We can use a query along the lines of `T extends EventHandler<infer Input>` (pseudo-code, I'd have to check the name of the type).

Blechlawine:
> Hello, I think, for type-safety we cannot trust that the data sent by the user is the same type we are inferring.
> 
> My thought is that it would be nice if we would expose a util eg-`parseBody / validateBody` which takes in a function that returns the parsed body. The function's return type will be inferred as the type of body
> 
> From this there will be one more benefit: we can assign type to the `useFetch` composable in nuxt to automatically infer type of input data as we do with response.

This was exactly my thinking, but I wanted to leave that up to the developer to decide if they want to validate the request body. A helper could easily be built around my original idea:
```typescript
// defineZodRoute.ts (a custom file which should probably not be included in nitro)
import { z } from "zod";

function defineZodRoute<TInput, TOutput, TResponse>(
    input: z.ZodType<TOutput, z.ZodTypeDef, TInput>,
    resolver: (input: TOutput) => TResponse
) {
    return defineEventHandler(async (event) => {
        const body = await readBody(event);
        const parsedInput = input.safeParse(body);
        if (parsedInput.success) {
            return resolver(parsedInput.data);
        } else {
            return parsedInput.error;
        }
    });
}

export default defineZodRoute;
```

anuragkumar19:
@Blechlawine 
Yeah. Right. Sorry, I missed some details in last comment. We cannot do that way. I was just trying to avoid explicit mention of types and any make it e2e type-safe with zod so developer will not shoot himself in foot. I wanted nuxt to automatically infer types as it infer response types.

pi0:
Hi, dear @Blechlawine. Thanks for the nice write-up in #841 and this PR.

New typed events handlers in https://github.com/unjs/h3/pull/417 will help to implement auto-type generation. First step is https://github.com/unjs/nitro/pull/1532

I am closing this one for maintenance but thanks again ❤️

PR #897, opened 2023-02-01
fix(cache): validate using entry `expires`

harlan-zw:
### ❓ Type of change

### 📚 Description

This PR more accurately invalidates files based on their set expiry time, rather than using the maxAge + modified time.

The previous logic works correctly when using the function as intended.

However, I need to implement dynamic maxAge times. It's a bit of a hack but I'm using `validate(entry)` to modify the expires time that is stored. The issue is that the previous logic looks like it won't honor my custom `expires` time and will always reset it.

Having a purposeful exposed function to dynamically set the entry options before they are stored would be nice, but out of the scope of this PR.

Here is the code sample:

```ts
 const generateSvg = defineCachedFunction<{ options: OgImageOptions }>(
    async () => {
      const options = await fetchOptions(path)
      setHeader(e, 'Content-Type', 'image/svg+xml')
      const provider = await useProvider(options.provider!)
      return { options, svg: provider.createSvg(withBase(path, useHostname(e)), options) }
    },
    {
      validate(entry) {
        const options = entry.value?.options || {}
        if (options.maxAge) {
          // modify the expires to user provided maxAge
          entry.expires = Date.now() + options.maxAge * 1000
        }
        else if (options?.static) {
          // modify the mtime to 1 hour
          entry.expires = Date.now() + 60 * 60 * 1000
        }
        return true
      },
      getKey: () => `nuxt-og-image:svg:${path}`,
      maxAge: 0,
      swr: true,
    },
  )
```

It's possible I haven't considered something, but I believe that this logic makes more sense generally. There doesn't seem to be any tests around the caching, so please sanity-check my logic.

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -67,14 +67,14 @@ export function defineCachedFunction<T = any>(
       ((await useStorage().getItem(cacheKey)) as any) || {};
 
     const ttl = (opts.maxAge ?? opts.maxAge ?? 0) * 1000;
-    if (ttl) {
+    if (ttl && !entry.expires) {
       entry.expires = Date.now() + ttl;
     }
 
     const expired =
       shouldInvalidateCache ||
       entry.integrity !== integrity ||
-      (ttl && Date.now() - (entry.mtime || 0) > ttl) ||
+      (entry.expires && Date.now() > entry.expires) ||
       !validate(entry);
 
     const _resolve = async () => {


src/runtime/cache.ts
```diff
@@ -67,14 +67,14 @@ export function defineCachedFunction<T = any>(
       ((await useStorage().getItem(cacheKey)) as any) || {};
 
     const ttl = (opts.maxAge ?? opts.maxAge ?? 0) * 1000;
-    if (ttl) {
+    if (ttl && !entry.expires) {
       entry.expires = Date.now() + ttl;
```
pi0, line 71: We might need to create a local `expires` variable that takes `ttl` into account as well

harlan-zw: Would you mind sharing some more context? Still getting across the complexity of this function

pi0: Hmm thinking more, in your example, you are generating an `expires` based on `Date.now()` + a certain ttl (from options or 60*60). Would it makes sense to allow directly modify ttl?

BTW to be honest it is a hacky workaround as you mentioned. validate is passing opts without expecting it to be modified.

harlan-zw: The feature is allowing you to dynamically set how long something can exist in the cache based on the entry value. Don't really mind how it's accomplished. 

Basically, the end user can modify the cache time of the svg, but we can only find out the cache time once it's been fetched initially.

Happy to rework this PR to introduce a feature to do it. Not sure the best name for it though, `onCacheSubmit(entry)` ?

pi0:
Hi, dear @harlan-zw. I just noticed this PR is blocked for a while. While I really agree supporting custom expiry would be a really useful addition, I am still not confident with the current solution.

I have moded it to https://github.com/unjs/nitro/issues/1544 to keep track but be sure will get into it ASAP specially that plan to extract caching functionality of nitro into a standalone and well testable project it might be much easier to move forard.

harlan-zw:
Thanks! Sounds good. I ended up just using a custom cache function as a workaround so not too urgent.

Any improvements around this or just composing your own cache functions are a plus for DX.

Issue #905, opened 2023-02-03
SWR does not work in per route setup. Node preset


oneart-dev:
### Environment

Tested on:
 - Node.js version: 16.14.2 && v18.13.0
 - Nuxt ^3.1.2

### Reproduction

Minimal Reproduction:
https://stackblitz.com/edit/github-fuonyg?file=pages/index.vue

```bash
npm run build && npm run preview
```

### Describe the bug

SWR setting of 1800 seconds do not work as every page refresh we still getting freeze while promise resolved.
**Not working**
```js
routeRules: {
   '/': { swr: 1800 },
},
```

**Working**
Instant refresh as expected
```js
routeRules: {
   '/**': { swr: 1800 },
},
```

manniL:
This does work as expected with a basic nitro example

https://stackblitz.com/edit/unjsnitroissues905?file=nitro.config.ts,routes%2Findex.ts,package.json

oneart-dev:
@manniL ok, it does but it doesn't work with nuxt and very similar setup. What should I do? Create an issue in nuxt repository?

manniL:
> @manniL ok, it does but it doesn't work with nuxt and very similar setup. What should I do? Create an issue in nuxt repository?

Actually yes, that'd be the best as it seems to be an integration issue on the "nuxt side" 😋

oneart-dev:
Ok, issue created. I will leave this issue open for a bit in case ping pong between repositories send me back to nitro =)

manniL:
I'd prefer closing it as the example above shows that it must be a nuxt-related topic 😄 
Worst case, we will reopen 😋

PR #906, opened 2023-02-03
fix: apply cached rules to overlaping wildcard patterns

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/905
https://github.com/nuxt/nuxt/issues/18745

### ❓ Type of change

### 📚 Description

If a cached route rule is created for a path that would match but is not equal to a glob pattern, we do not currently cache it (because the entire handler can't be cached).

We can work around this by adding a duplicate handler that we _will_ cache, that matches the more specific variant.

An example is where we have a glob `/**` handler but only wish to cache the path `/foo` or `/foo/**`. In this case we need to add new handlers to cover these more specific prefixes.

**Note**: I've added this in the runtime code because that's where the existing processing of route rules happens. It would of course be possible to move all of this code straightforwardly to https://github.com/unjs/nitro/blob/6541373545ee2716566bce46c7568ad2c3164cdf/src/rollup/plugins/handlers.ts#L20-L26, but that is also true of [these lines of code](https://github.com/unjs/nitro/blob/a8b6ec54dd011eaedd3054457367fee69dd62964/src/runtime/app.ts#L86-L88) and I think we must have a good reason to keep them there (such as potentially setting route rules at runtime).

### 📝 Checklist


pi0:
This is a smart solution!  I agree. I think would be better that we move this to handler generation in build time and reduce runtime overhead. Route rules should be pre-determined in build time if we expect them universally so should be fine...

danielroe:
Updated. ✅

One comment - it's occurred to me that we _also_ need to track nested conditions, for example:

```bash
/posts/test: { cache: false }
/posts/**: { cache: true }
/**: { cache: false }
```

My last commit should hopefully handle this as well.

pi0 (approved):

PR #916, opened 2023-02-06
feat: support async plugins

### LINKED ISSUE #915: Make plugins async
For the benefit of bootstrapping plugins it is usefull to make plugins load async.

https://github.com/unjs/nitro/blob/ed3a48026e8a27ecfb86e406ba39bdeb418f8573/src/runtime/app.ts#L89-L91

Can someone explain to me if there is a good reason to load plugins synchronously?

MarijnMensinga:
### 🔗 Linked issue

#915

### ❓ Type of change

### 📚 Description

Plugins should be loaded async to make the plugins bootstrap propperly. 

This fixes potential race-conditions when creating plugins that are called before they are bootstrapped correctly.

Resolves #915

### 📝 Checklist

- [x] I have updated the documentation accordingly. (not needed)

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -29,7 +29,7 @@ export interface NitroApp {
   localFetch: ReturnType<typeof createLocalFetch>;
 }
 
-function createNitroApp(): NitroApp {
+async function createNitroApp(): NitroApp {
   const config = useRuntimeConfig();
 
   const hooks = createHooks();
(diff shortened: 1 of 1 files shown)


pi0:
This is a nice feature to have! The main issue is that support for top-level await is limited and many providers are currently depending on nonasync API. 

Putting a pending label for now. It is likely we could only support async plugins in Nitropack v3 and some breaking changes.

Hebilicious:
@pi0 Can't we work around the issue with a closure ? 

```ts
const createNitroApp = () => {

const installPlugins = async () => {
  for (const plugin of plugins) {
    await plugin(app);
  }
}

installPlugins()
}
```

pi0:
@Hebilicious no we lose order this way.

pi0:
Thanks for PR again dear @MarijnMensinga. It is not mergeable without breaking changes. Let's keep track via https://github.com/unjs/nitro/issues/915 (i might have some ideas to introduce as non breaking)

Issue #113, opened 2022-04-14
Built-in proxy support


pi0:
Adding as a feature to route rules.

DamianGlowala:
@pi0, would you be able to give any ETA on this one? 🙏 Would be super useful to have it working in Nuxt 3.

ironytr:
can't use nuxt3@latest because proxy issues, non module working rn. is nitro proxy ready yet?

makkarpov:
devProxy doesn't seem to be supporting relative URL in fetch, is this by design?

I.e.

```javascript
// config:
export default defineNuxtConfig({
    nitro: {
        devProxy: {
            '/api/': {
                target: process.env.API_TARGET,
                changeOrigin: true
            }
        }
    }
})

// ======================
// later, on server side:
// ======================

const r = await (await fetch('/api/test')).json();
// => 500 Invalid URL
```

frasza:
Any indication when it will be fully implemented?

makkarpov:
@pi0

> Yes. You need to pass full URLs because dev proxy is outside of worker and needs to make an actual request.

But this breaks code unification between client and server. In browser, when I make a `/api/xxx` request, browser will resolve it to my domain automatically. So it's natural to assume same behavior in SSR context - relative requests should finally end up being proxied. Or resolved to absolute URLs via proxy config and executed normally.

makkarpov:
It's also desirable (but far less annoying) to proxy WebSocket request for same reason. This time things are even more complicated: you have to invent ad-hoc logic _on client_. This means, you need to put all your backend paths into a `runtimeConfig`, check if it's dev or prod, use relative paths on prod (nginx is completely fine with websockets), match correct backend on dev, build an URL.

pi0:
Web socket layer and runtime proxy support (with direct fetch ability) are on the way :)

PavlosMac:
> Web socket layer and runtime proxy support (with direct fetch ability) are on the way :)

Available yet?

Arturexe:
Will this be available soon? I can't send auth requests to my backend because of it. Before this is integrated, I can't use Nuxt 3.

bogumilkorek:
Some ETA would me much appreciated. Our team is also waiting for this.

paulvonber:
Any news on it?

MiniDigger:
bumping this every day certainly will not make it faster, plus it annoys everyone who subscribed to this issue for actual updates, so please, just don't.

michauds:
[...]

PR #926, opened 2023-02-08
feat: support runtime proxy using route rules

pi0:
### 🔗 Linked issue

#113

### ❓ Type of change

### 📚 Description

Support runtime proxy using route rules and `h3.proxyRequest` (universally works with fetch)

```js
{
  routeRules: {
    '/proxy/example': { proxy: 'https://example.com' },
    "/proxy/api/**": { proxy: '/api/proxied/**' },
  }
}
```

Local support is pending https://github.com/unjs/h3/pull/321

### 📝 Checklist

### DIFF
--- a/docs/content/3.config/index.md
+++ b/docs/content/3.config/index.md
@@ -1,17 +1,16 @@
 ---
 title: Configuration
 aside: false
-description: 'Customize your Nitro app with a configuration file!'
+description: "Customize your Nitro app with a configuration file!"
 ---
 
 In order to customize nitro's behavior, we create a file named `nitro.config.ts`.
 
 ```js
 // nitro.config.ts
-import { defineNitroConfig } from 'nitropack'
+import { defineNitroConfig } from "nitropack";
 
-export default defineNitroConfig({
-})
+export default defineNitroConfig({});
 ```
 
 ## Config Reference
@@ -219,11 +218,11 @@ Path to a custom runtime error handler. Replacing nitro's built-in error page.
 **Example:**
 
 ```js [nitro.config]
-import { defineNitroConfig } from 'nitropack'
+import { defineNitroConfig } from "nitropack";
 
 export default defineNitroConfig({
-  errorHandler: '~/error'
-})
+  errorHandler: "~/error",
+});
 ```
 
 ```js [error.ts]
@@ -253,7 +252,8 @@ When `cache` option is set, handlers matching pattern will be automatically wrap
     '/blog/**': { cache: { /* cache options*/ } },
     '/assets/**': { headers: { 'cache-control': 's-maxage=0' } },
     '/api/v1/**': { cors: true, headers: { 'access-control-allowed-methods': 'GET' } },
-    '/old-page': { redirect: '/new-page' }
+    '/old-page': { redirect: '/new-page' },
+    '/proxy/**': { proxy: 'https://example.com' }
   }
 }
 ```
--- a/docs/content/3.config/index.md
+++ b/docs/content/3.config/index.md
@@ -253,7 +253,8 @@ When `cache` option is set, handlers matching pattern will be automatically wrap
     '/assets/**': { headers: { 'cache-control': 's-maxage=0' } },
     '/api/v1/**': { cors: true, headers: { 'access-control-allowed-methods': 'GET' } },
     '/old-page': { redirect: '/new-page' },
-    '/proxy/**': { proxy: 'https://example.com' }
+    '/proxy/example': { proxy: 'https://example.com' },
+    "/proxy/api/**": { proxy: '/api/proxied/**' },
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


yancyhuang:
Hi @pi0 

Very wonderful feature!
BTW, does it support proxy web socket now? In ticket #113 you mentioned web socket proxy feature is on the way, not sure if it's released or not.

Thank you,
Yancy

pi0:
@yancyhuang It is partially with `devProxy` options. We still need a unified layer for WS support in servers to make it happen.

antlionguard:
Will path rewrite support come? Like devProxy. 

```
rewrite: (path) => path.replace(/^\/api/, '')
```

or i'm missing something about this topic?

pi0:
@antlionguard Something like this has same effect in route rules: `"/proxied/api/**": "https://backend/**`.

jonkri:
> We still need a unified layer for WS support in servers to make it happen.

Hi! 👋 Is there an issue that tracks this?

vushe:
> Something like this has same effect in route rules: `"/proxied/api/**": "https://backend/**`.

Is there a way set header on proxy request to an external resource?

With [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) we could do something like this
```js
onProxyReq: (proxyReq, req, res) => {
    proxyReq.setHeader('x-data', req.cookies.data);
}
```

blowsie:
@pi0 , Thanks for this unless I missed something, whilst his feat is called "support **runtime proxy** using route rules", it only actually supports rules that are defined at **build time**.

Would it be possible to acheive something like this using variables?
```ts
  routeRules: {
    '/api': { proxy: NUXT_API_PROXY },
  }
})
```

pi0:
@blowsie For dynamic target, you should make a custom route handler and use `sendProxy(event)` utility.

imzedi:
@pi0, any way I can use the same for static build? (by `ssr: false` and `"nuxt build --prerender`)

manniL:
@imzedi With a static build, you platform or reverse proxy has to handle that as no "nuxt/nitro" server is running ☺️

manniL:
@pi0 e.g. for netlify you mean?

pi0:
Yes like [this](https://docs.netlify.com/routing/redirects/rewrites-proxies/#proxy-to-another-service) we can use for static presets.

imzedi:
Thanks @pi0 @manniL .
Got this whole point.
Anyhow we will need a server  whether its CDN  or serverless function provided by deployment platform or we manage our own.

JoniLieh:
Hello, thanks for maintaining such an awesome library!
Is there support to proxy websockets like socket.io?
I'm using nuxt and this works, but not for websockets (even with `ws: true`)
```
"/socket.io/**": {
        proxy: {
          to: "http://backend:3001/**",
          headers: {
            "X-API-KEY": NUXT_API_SECRET,
          },
}
```

Thanks!

keryigit:
This doesn't seem to work for websockets. The HTTP upgrade message is directly responded instead of forwarded to the proxy address.

pi0:
@keryigit Can you please open new feature request so i don’t forget this? 🙏

pi0:
Made trackers for WS Proxy support:
- https://github.com/nitrojs/nitro/issues/2886
- https://github.com/unjs/crossws/issues/93

pi0:
@JoniLieh thanks for the response, but i don't recommend this solution, it will likely break with future versions of Nitro nor works with other runtimes..

JoniLieh:
@pi0  Thanks for the info, its for nuxt as well.
- Which solution do you recommend to proxy websockets? :)

dschreij:
@pi0 I am curious to the question of @JoniLieh as well, as his solution was the only viable one I found at this time, for proxying websocket requests through nitro. I kept on looking for other ways to do this after you mentioned it wasn't the recommended way, but I couldn't find any.

I tried the `proxyRequest` function from h3 / nitro itself, but it doesn't appear to handle websockets at all, even if you pass it the `duplex: full` or `sendStream: true` option parameters.

I also tried setting the `{ nitro: { experimental: { websockets: true } } }` flag in nuxt.config.ts, and this causes the websocket connection from the client to be successfully established, but with nitro, not with the server it should proxy to. I couldn't find any documentation on how to proxy the websocket connection from nitro onwards to our other service.

I ended up with creating a `./server/api/ws.ts` file that accepts the websocket request and passes it on to the server using http-proxy-middleware. I use sidebase-auth for authentication, which stores the jwt token only in nitro, which I still need to add to the header of the websocket request (which is quite neat to able to I must say). If you have any small pointers on how to do this differently or where to look for a more future-proof solution, please let me know :)

```ts
import { getToken } from '#auth'
import { createProxyMiddleware } from 'http-proxy-middleware'

const config = useRuntimeConfig()

  const apiUrl = config.apiUrl

  const token = await getToken({ event })
  if (!token) {
    return createError({
      statusCode: 401,
      statusMessage: 'Unauthorized',
    })
  }
  const accessToken = token.accessToken
  if (!accessToken) {
    return createError({
      statusCode: 401,
      statusMessage: 'Unauthorized',
    })
  }

  const proxy = createProxyMiddleware({
    target: apiUrl,
    headers: {
      authorization: `Bearer ${accessToken}`,
    },
    changeOrigin: true,
    ws: true,
    pathRewrite: {
      '^/api': '',
    },
    logger: console,
    on: {
      error: (err) => {
        console.error('WebSocket proxy error:', err)
      },
    },
  })
  const { req, res } = event.node
  return proxy(req, res, () => {
    console.error('WebSocket proxy error')
  })
})
```

PR #927, opened 2023-02-08
feat(cloudflare): expose `event.context.cf`

### LINKED ISSUE #391: Cloudflare `request.cf`

pi0:
### 🔗 Linked issue

resolves #391

closes #677

### ❓ Type of change

### 📚 Description

This PR exposes cloudflare [request context](https://developers.cloudflare.com/workers//runtime-apis/request#incomingrequestcfproperties) as `event.context.cf`. 

Implementation is based on new unenv support. Any other platform based on localFetch, can expose the same context by passing `{ context: {} }` to the request init.

A new generic handler is added to nitro. It will be used later for other purposes as well.

Note: For nested `$fetch`, we should explicitly pass the `cf` to be proxied:

```js
  await $fetch("/test", { context: { cf: event.context.cf } });
```

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "source-map-support": "^0.5.21",
     "std-env": "^3.3.2",
     "ufo": "^1.0.1",
-    "unenv": "^1.1.0",
+    "unenv": "^1.1.1",
     "unimport": "^2.2.4",
     "unstorage": "^1.1.3"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -81,7 +81,7 @@ importers:
       typescript: ^4.9.5
       ufo: ^1.0.1
       unbuild: ^1.1.1
-      unenv: ^1.1.0
+      unenv: ^1.1.1
       unimport: ^2.2.4
       unstorage: ^1.1.3
       vitest: ^0.28.4
@@ -141,7 +141,7 @@ importers:
       source-map-support: 0.5.21
       std-env: 3.3.2
       ufo: 1.0.1
-      unenv: 1.1.0
+      unenv: 1.1.1
       unimport: 2.2.4_rollup@3.14.0
       unstorage: 1.1.3
     devDependencies:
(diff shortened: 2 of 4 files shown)


altryne:
Is there a way to use this to get bindings from .env? 

I saw a screenshot from Atinux about storing state in cloudflare KV workers, so it seems that conceptually that's possible? 

But the above exposed request.cf does not include the .env (nor the secrets stored in cloudflare) nor the service, KV, queue bindings. 

Is there a way that I'm just missing?

pi0:
You can use storage layer to store data in KV, Filesystem, memory and anywhere else.

- https://github.com/Atinux/atinotes/blob/ee88ffa8f9dc1d5254ccef10411b815343fdce6a/nuxt.config.ts#L10
- https://github.com/Atinux/atinotes/blob/ee88ffa8f9dc1d5254ccef10411b815343fdce6a/server/api/pages/%5Bslug%5D.get.ts#L5

dario-piotrowicz:
@altryne you're right you cannot access KVs or other Cloudflare services from what it's exposed, we need to pass the env and ctx to the user as I am doing here for the cloudflare-pages adaptor: https://github.com/unjs/nitro/pull/997

After I (hopefully) get the cloudflare-pages adaptor PR merged I can look into adding access to the env and ctx to the cloudflare worker adaptor as well, although in order to do so we'd fist need to convert the adaptor to the ESM syntax as now it uses the old service worker one in which the various variables are just put in the global scope (and some like DO aren't even available).

altryne:
I actually found a way! it's possible via self.

https://discord.com/channels/473401852243869706/1002541842409193563/1077016381662433372

dario-piotrowicz:
ah wow, that's interesting! :open_mouth: 

are you deploying your Nuxt app via the worker adaptor? (not the pages one)
because if you are yeah I do imagine that the bindings are just added to the global scope by Cloudflare so they end up being accessible by `self` :thinking:

altryne:
Yeah not via the pages one.

dario-piotrowicz:
ah ok I see, that makes sense, thanks :slightly_smiling_face::+1:

altryne:
> You can use storage layer to store data in KV, Filesystem, memory and anywhere else.

@pi0 I'm so sorry I missed your comment on this, however, this isn't answering the question of accessing cloudflare bindings on .env. 

The storage layer adds storage, however, on the .env, cloudflare is storing the following things that are not storage: 
- secrets https://developers.cloudflare.com/workers/platform/environment-variables/#environment-variables-via-wrangler
- Service bindings https://developers.cloudflare.com/workers/runtime-apis/service-bindings/
- Queues https://developers.cloudflare.com/queues/examples/send-errors-to-r2/

All those things are needed when developing a server side application on cloudflare, all their examples aways require to take this information from the request context and I think it would be great to have them, and not have to do hacks like this one: 
https://discord.com/channels/473401852243869706/1002541842409193563/1077016381662433372


I think for many worker devs, this would be more important than the context of the IP and country (which is also important)

PR #951, opened 2023-02-16
fix: don't render json response if url contains `/api/`

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/19093

### ❓ Type of change

### 📚 Description

Although a helpful idea, I think this is too broad a net for rendering an API response. Even routes starting with `/api` might still benefit from an HTML response if requested from a browser.

### 📝 Checklist

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -41,8 +41,7 @@ export function isJsonRequest(event: H3Event) {
     hasReqHeader(event, "accept", "application/json") ||
     hasReqHeader(event, "user-agent", "curl/") ||
     hasReqHeader(event, "user-agent", "httpie/") ||
-    event.node.req.url?.endsWith(".json") ||
-    event.node.req.url?.includes("/api/")
+    event.node.req.url?.endsWith(".json")
   );
 }
 


pi0 (approved):

dargmuesli:
Isn't that a breaking change? With Nuxt requests to `server/api` endpoint now return html instead of the expected json.

danielroe:
~> see https://github.com/unjs/nitro/pull/969

dargmuesli:
Thank you for the background, I see :+1:  Intuitively I'd expect "root" `/api` routes to keep the previous json behavior, not matter how they're accessed.

I have an endpoint that either returns a specific content type (`text/calendar`) or throws a `createError`. I'd expect json for errors, even when calling that endpoint by manual input into the browser's address bar.
I don't want to rename the endpoint to `*.json` as it can return a different mime type. I also don't want to set the `Content-Type` header manually before errors. I'm not sure if the cors-related fix will solve that question, but I'll try later :raised_hands: 
Is there something I should do instead? Maybe I don't see the obvious way to do it now.

danielroe:
I understand what you mean, though equally I would not expect users to access an `/api/` endpoint directly in the browser and receive JSON (but only in the event of an error).

Would you open an issue to discuss adding a `.startsWith()` test? We might also consider allowing handlers to _indicate_ their preference somehow, or something else.

cc: @pi0 who may have a better idea.

pi0:
Reverting in https://github.com/unjs/nitro/pull/971 for now.

dargmuesli:
I've created https://github.com/unjs/nitro/discussions/978 as a place where a discussion can continue.

PR #965, opened 2023-02-20
fix(cloudflare-pages): exclude assets from function call

danielroe:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/497

### ❓ Type of change

### 📚 Description

We can generate a `_routes.json` file to prevent hits on assets directories from hitting the function at all. We might also be able to add individual assets that are outside of these directories to the json but this runs the risk of hitting the 100 rule limit. (Maybe we can allow extending the default config, suggested by @colonelbundy.)

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -1,6 +1,7 @@
-import { resolve } from "pathe";
+import { join, resolve } from "pathe";
 import fse from "fs-extra";
-import { joinURL } from "ufo";
+import { joinURL, withLeadingSlash } from "ufo";
+import { globby } from "globby";
 import { writeFile } from "../utils";
 import { defineNitroPreset } from "../preset";
 import type { Nitro } from "../types";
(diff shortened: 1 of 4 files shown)


src/presets/cloudflare.ts
```diff
@@ -56,6 +67,46 @@ export const cloudflarePages = defineNitroPreset({
...
+
+      const routes: CloudflarePagesRoutes = {
+        version: 1,
+        include: ["/*"],
+        exclude: [],
+      };
+
+      // Exclude public assets from hitting the worker
+      const explicitPublicAssets = nitro.options.publicAssets.filter(
+        (i) => !i.fallthrough
```
pi0, line 79: User options as well as root path / check are applied in nitro.ts. `fallthrough` flag specified final behavior

pi0 (approved):
Thanks.

PR #967, opened 2023-02-20
refactor(externals): sort `bundledDependencies` keys in output `package.json`

danielroe:
### ❓ Type of change

### 📚 Description

This reduces change between two builds of the same package. It's quite a minor change but can help ensure reproducible nitro builds.

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/externals-legacy.ts
+++ b/src/rollup/plugins/externals-legacy.ts
@@ -296,7 +296,7 @@ export function externals(opts: NodeExternalsOptions): Plugin {
             name: "nitro-output",
             version: "0.0.0",
             private: true,
-            bundledDependencies: [...tracedPackages.keys()],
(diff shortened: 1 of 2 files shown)


pi0:
Keys order are expected to remain as is from original package. Is there a more specific reason doing this?

danielroe:
When I run the same build more than once, these keys are reordered, so this is not currently deterministic.

pi0:
LGTM. Only note that we might probably drop `bundledDependencies` property it was mainly for debugging purposes but also to introduce new manifest file in top level `.output` containing timestamp of build time. 

I wish knew more about initiatives behind this PR...

danielroe:
> I wish knew more about initiatives behind this PR...

I was debugging a bundle size diff for a renovate update PR and spotted the discrepancy.

PR #968, opened 2023-02-21
fix(cloudflare-pages): handle assets only for get requests

danielroe:
### 🔗 Linked issue

closes https://github.com/unjs/nitro/pull/796 - this is a remake as I had no permission to push changes to the branch (make sure to give credit when merging)
resolves https://github.com/unjs/nitro/issues/497
resolves https://github.com/unjs/nitro/issues/787

### ❓ Type of change

### 📚 Description

CF will throw a 405 error when fetching assets with a non-GET request. I think we can guard them here safely, with the note that we should also merge https://github.com/unjs/nitro/pull/965 alongside this to avoid hitting endpoint for known assets/asset prefixes.

### 📝 Checklist


src/runtime/entries/cloudflare-pages.ts
```diff
@@ -21,9 +21,11 @@ interface CFRequestContext {
 export async function onRequest(ctx: CFRequestContext) {
   try {
     // const asset = await env.ASSETS.fetch(request, { cacheControl: assetsCacheControl })
-    const asset = await ctx.next();
-    if (asset.status !== 404) {
-      return asset;
+    if (ctx.request.method === "GET") {
```
pi0, line 24: I think we might also support for HEAD 
```suggestion
    if (ctx.request.method === "GET" || ctx.request.method === "HEAD") {
```

danielroe: No, that is not allowed by Cloudflare either.

pi0: Hmm thanks for confirming.

This is strange because normal worker has fine gained API to control caching headers via head. (and i suppose would be useful for cors and options)

danielroe: It would indeed be useful. Maybe they can add more options down the line to configuration via `_routes.json`?

pi0: Okay, let's revise it later GET is good enough.

PR #973, opened 2023-02-22
fix(cloudflare-pages): only allow 100 rules

danielroe:
### ❓ Type of change

### 📚 Description

CF only supports 100 rules, which includes the 'include' as well as the 'exclude' rule. Seems like this might have been lost in the refactor.

### 📝 Checklist


pi0:
Maybe we can minus the include length? (100-include)

PR #976, opened 2023-02-22
feat(vercel): add functions config

smaeda-ks:
Currently, some serverless function options are not customizable, such as `memory` and `maxDuration`:
https://vercel.com/docs/build-output-api/v3#vercel-primitives/serverless-functions/configuration

This PR adds support so users can optionally set those options in their `nitro.config.ts` file:

```ts
// nitro.config.ts
import { defineNitroConfig } from 'nitropack'
  "vercel": {
        "functions": {
            "memory": 1024,
            "maxDuration": 10,
        }
    }
})
```

### ❓ Type of change

### 📝 Checklist


pi0:
This is nice addition thanks! Any reason we don' allow specify all function configurations?

src/presets/vercel.ts
```diff
@@ -38,6 +40,8 @@ export const vercel = defineNitroPreset({
         handler: "index.mjs",
         launcherType: "Nodejs",
         shouldAddHelpers: false,
+        ...(customMemory && { memory: customMemory }),
+        ...(customMaxDuration && {maxDuration: customMaxDuration}),
```
pi0: TIP: `JSON.stringify({ test: undefined })` is `{}`. We can simply allow undefined to be set

smaeda-ks: Ah, thanks for the tip! Will change them to write that way.

smaeda-ks:
> This is nice addition thanks! Any reason we don' allow specify all function configurations?

@pi0 Thanks for asking! Other options such as `environment` and `regions` should be managed in the project settings:
https://vercel.com/docs/concepts/projects/project-configuration#env

> We recommend against using this property. To add custom Environment Variables to your Project [define them in the Project Settings](https://vercel.com/docs/concepts/projects/environment-variables).

So adding `memory` and `maxDuration` would be enough, I believe.

pi0 (approved):

PR #989, opened 2023-02-23
fix(cloudflare): split `Set-Cookie` header into multiple headers

### LINKED ISSUE #988: Set Cookie merged into one header (Cloudflare Workers)
### Environment

Cloudflare Workers entry

### Reproduction

Make proxy which sets more than 2 cookies 

### Describe the bug

When you set cookies all you cookies merged into one header, while it’s not allowed by RFC. You need to pass one header per cookie 

### Additional context

I gonna open PR and fix it, I will link it 

### Logs

_No response_

oleghalin:
### 🔗 Linked issue

#988

### ❓ Type of change

### 📚 Description

Splits Set-Cookie for each upstream header provided. 

Resolves #988 

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/cloudflare.ts
+++ b/src/runtime/entries/cloudflare.ts
@@ -4,7 +4,7 @@ import {
   mapRequestToAsset,
 } from "@cloudflare/kv-asset-handler";
 import { withoutBase } from "ufo";
-import { splitCookiesString } from "set-cookie-parser"
+import { splitCookiesString } from "set-cookie-parser";
 import { requestHasBody } from "../utils";
 import { nitroApp } from "#internal/nitro/app";
 import { useRuntimeConfig } from "#internal/nitro";
@@ -77,15 +77,15 @@ function normalizeOutgoingHeaders(
   const result = new Headers();
 
   for (const [k, v] of Object.entries(headers)) {
-    if(k === 'set-cookie') {
+    if (k === "set-cookie") {
       for (const cookie of splitCookiesString(v)) {
-        result.append('set-cookie', cookie)
+        result.append("set-cookie", cookie);
       }
 
       continue;
     }
 
-    result.append(k, Array.isArray(v) ? v.join(',') : v)
+    result.append(k, Array.isArray(v) ? v.join(",") : v);
   }
 
   return result;
--- a/src/runtime/entries/cloudflare.ts
+++ b/src/runtime/entries/cloudflare.ts
@@ -3,12 +3,13 @@ import {
   getAssetFromKV,
   mapRequestToAsset,
 } from "@cloudflare/kv-asset-handler";
-import { withoutBase } from "ufo";
+import { parseURL, withLeadingSlash, withoutBase, withoutTrailingSlash } from "ufo";
 import { splitCookiesString } from "set-cookie-parser";
+import { createError } from "h3";
 import { requestHasBody } from "../utils";
 import { nitroApp } from "#internal/nitro/app";
 import { useRuntimeConfig } from "#internal/nitro";
-import { getPublicAssetMeta } from "#internal/nitro/virtual/public-assets";
+import { getPublicAssetMeta, isPublicAssetURL } from "#internal/nitro/virtual/public-assets";
 
 addEventListener("fetch", (event: any) => {
   event.respondWith(handleEvent(event));
@@ -23,6 +24,7 @@ async function handleEvent(event: FetchEvent) {
   } catch {
     // Ignore
   }
+
 
   const url = new URL(event.request.url);
   let body;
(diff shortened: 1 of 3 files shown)


oleghalin:
I suppose that these changes should be applied also to other presets where `.join(',')` used.

oleghalin:
I checked codebase a little bit deeper and i think that it should be changed on unenv level and response.headers (which is HeadersObject) should implement Headers instead of `{ [key: string]: string | string[] | undefined };` and with this we will need to make such transformations on preset level.

src/runtime/entries/cloudflare.ts
```diff
@@ -71,9 +73,20 @@ const baseURLModifier = (request: Request) => {
...
-  return Object.entries(headers).map(([k, v]) => [
-    k,
-    Array.isArray(v) ? v.join(",") : v,
-  ]);
+): Headers {
+  const result = new Headers();
+
+  for (const [k, v] of Object.entries(headers)) {
+    if (k === "set-cookie") {
+      for (const cookie of splitCookiesString(v)) {
```
pi0, line 81: We have recently introduced [same util](https://github.com/unjs/h3/blob/80f2c611918a17d1c303636924ce39ea7f246489/src/utils/internal/cookies.ts#L10) in h3. It might be one less dependency and bundle size if we expose it from h3 package.

oleghalin: Thanks, what is proper way to sync versions of h3 and current nitro? Should we wait for h3 version tag or?

pi0: Sure, releasing next version until this weekend. (at the same time thinking about your other comments, we need to really fix header normalization across entries)

oleghalin: As i see normalize function looks same for all other entries, but i think something is different on provider level, i suppose that some of same do cookie splitting themselves. We can drop normalize function to utils and use it every entry until make changes with type change for `HadersObject`.

pi0: Indeed there are slight behavior diferences. Also we never properly tested header behavior on [nitro-deploys](https://github.com/unjs/nitro-deploys) 😓

oleghalin: Hello, @pi0. Do you have any plans for releasing new versions soon? Gonna continue work on module and project, but this issue blocks me testing any env except local one :)

oleghalin:
Hello, @pi0. I also pushed to this PR change for not caching 404 responses if its asset directory, can you check it too please

oleghalin:
Hello, @pi0 any updates on it? Issue with Set-Cookie header and 404 caching its making Nuxt absolutely unusuable with Cloudlare Workers

Stupremee:
Hey @oleghalin,
I just saw this PR because I have the same problems with Nuxt on CF Workers, which is soo annoying.

I noticed that the new `h3` version used in nitro now exports the `splitCookeString` function, so maybe you can remove the `set-cookie-parser` dependency and this PR can move forward.

oleghalin:
Hello, @pi0. I just just updated my PR and dropped third-party dep

oleghalin:
Hey, @pi0. I suppose that it makes many problems for all users who use CF workers to deploy, can you take a look on it? It solves 2 big problems with caching and cookies

jhd4tmr:
+1 I had the same bug that bothered me for days，it's not just about ```nitro.routeRules```, ```proxyRequest``` has the same bug

Hebilicious:
@oleghalin This is on my radar, apologies for the delay. Can you provide me with a reproduction ? I would like to see if this behaviour affects the other cloudflare presets too.

Hebilicious (changes requested):
This doesn't include any tests or reproduction, so this is hard to test.

oleghalin:
Not sure why you need reproduction, Set-Cookie header should be merged into one because browsers read each Set-Cookie header and set 1 cookie per each Set-Cookie header, as soon as there is only 1 header delimited by coma it will write only first cookie. As I know Vercel and node server parse it on their end and thats why it works, but it doesn't work for CF.

Hebilicious:
Thanks for the additional information and the PR.
A reproduction is useful to test the proposed changes. Without one, it takes more maintainer time to review and get things merged.

src/runtime/entries/cloudflare.ts
```diff
@@ -3,27 +3,38 @@ import {
...
-    });
-  } catch {
-    // Ignore
+  const url = new URL(event.request.url);
+  const id = decodeURIComponent(
+    withLeadingSlash(withoutTrailingSlash(url.pathname))
+  );
+
+  // Fetch public assets from KV only
+  if (isPublicAssetURL(id)) {
```
Hebilicious, line 27: I agree this needs to happen, but it should be handled in a different PR (https://github.com/unjs/nitro/pull/1236/files)

Hebilicious (commented):
This is absolutely a necessary change! 

- The check for publicAsset needs to happen in another PR https://github.com/unjs/nitro/pull/1236 (feedback welcome)

- This affects multiple preset and needs to be applied everywhere

PR #995, opened 2023-02-24
fix: remove rejected cache promise from pending list. 

tdekoning:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/18926

### ❓ Type of change

### 📚 Description

Resolved https://github.com/nuxt/nuxt/issues/18926.

Cached events that have rejected were kept inside the `pending` list, causing later requests to the same event to reject as well. This led to events on the same path to never recover, because the code of the event was never called again after an error.

### 📝 Checklist


danielroe (approved):

playground/routes/error-route.ts
```diff
@@ -0,0 +1,9 @@
+export default defineCachedEventHandler(
```
manniL, line 1: We sould probably remove the test route from the playground file 😋

tdekoning: Done! Thanks for the suggestion

manniL (approved):

tdekoning:
@pi0 do you have time to review this change? It seems to me as quite a small fix for a very impactful problem.

src/runtime/cache.ts
```diff
@@ -90,18 +90,26 @@ export function defineCachedFunction<T = any>(
...
+          delete pending[key];
+          if (validate(entry)) {
+            useStorage()
+              .setItem(cacheKey, entry)
+              .catch((error) => console.error("[nitro] [cache]", error));
+          }
         }
+      } catch (error) {
+        // Make sure entries that reject get removed.
+        delete pending[key];
```
pi0, line 109: I think we could limit scope of try catch around L94 only.

pi0:
Thanks for the PR dear @tdekoning and sorry took long to review.

Issue #773, opened 2022-12-20
Pass ctx in preset cloudflare-pages


domlen2003:
### Describe the feature

In the preset fore cloudflare-pages there already seems to be a TODO for adding the cf context in the [File](https://github.com/unjs/nitro/blob/main/src/runtime/entries/cloudflare-pages.ts) was originally implemented by @DaniFoldi 

This would really help me with my current nuxt projekt where i really need a KV binding. 
In the normal cloudflare-functions adapter this is already possible tho our whole project is based on cloudflare-pages. 

### Additional information

- [x] Would you be willing to help implement this feature?

domlen2003:
I just saw this is probably not the ctx but the env. Anyway the cf datastore options should be available

domlen2003:
Since there is no easy way to pass context to an event handler (every additional info given in the adapter gets dropped in unenv/runtime/fetch/call.ts callHandle, when creating req) 

I would suggest adding:
`Object.entries(ctx.env).forEach(([k, v]) => process.env[k] = v)`
just before invoking the call in the adapter.

This would make it somewath similar to the context.env version in pages functions.

dario-piotrowicz:
PS: I think this issue is a duplicate of https://github.com/unjs/nitro/issues/391

manniL:
Let's track in https://github.com/unjs/nitro/issues/773 then

dario-piotrowicz:
@manniL sorry about the confusion but this issue is different from #391 and it hasn't been resolved by #677

I've got confused by the `context` terminology, but anyways in the other issue/PR they were referring to the `cf` object present in requests' headers but the context mentioned in this issue is different (is the one needed to access Cloudflare bindings)

could you please reopen this issue? :pray:

Hebilicious:
Fixed by https://github.com/unjs/nitro/pull/1004

PR #997, opened 2023-02-25
fix(cloudflare-pages): pass CfRequestContext to nitro localCall

dario-piotrowicz:
### 🔗 Linked issue

#773

### ❓ Type of change

### 📚 Description

in the cloudflare-pages adapter I've added the CfRequestContext that the onRequest function receives from Cloudflare to the nitroApp localCall so that users can access such context in their nitro application

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/cloudflare-pages.ts
+++ b/src/runtime/entries/cloudflare-pages.ts
@@ -44,8 +44,9 @@ export async function onRequest(ctx: CFRequestContext) {
     host: url.hostname,
     protocol: url.protocol,
     body,
-    // TODO: Allow passing custom context
-    // cf: ctx,
+    context: {
+      cf: ctx,
+    },
     // TODO: Handle redirects?
     // redirect: ctx.request.redirect
   });


src/runtime/entries/cloudflare-pages.ts
```diff
@@ -44,8 +44,7 @@ export async function onRequest(ctx: CFRequestContext) {
     host: url.hostname,
     protocol: url.protocol,
     body,
-    // TODO: Allow passing custom context
-    // cf: ctx,
+    context: ctx,
```
dario-piotrowicz, line 47: **Note:** this needs to be called `context` and not `cf` (as suggested in the comment) because unenv takes this specific context field and puts it in the `req.__unenv__`: https://github.com/unjs/unenv/blob/4ed234cf8fd88264852739d14b40bebf7ce31afe/src/runtime/fetch/call.ts#L51
(which is where then users can have access to it)

pi0: Sure but it can be nested like here: https://github.com/unjs/nitro/blob/7bcfc362a77764aed566b9f266ae702ccd0dceed/src/runtime/entries/cloudflare.ts#L36

dario-piotrowicz:
@pi0 sorry to bother you, could you help me understand why the deployment's failing?

pi0:
Failed docs deployment is okay :)

src/runtime/entries/cloudflare-pages.ts
pi0: It is intentionally verbose because we are accessing a platform-api. `cf` is already added for normal worker target. we need to add it it for pages too (i suppose interfaces are same)

dario-piotrowicz: ah by the way, regarding the `cf` we I don't think we need to manually pass it since it will accessible from the `ctx.request` anyways

eg.
Screenshot at 2023-02-27 15-15-19

so it looks like `context: ctx,`  should actually be enough, what do you think?

or alternatively we can and choose what fields from the CfRequestContext pass along, in that case we could construct the `context` ourselves, but I am not sure if that's needed/beneficial

pi0: `event.context.cf` can have all the keys and context passed by cloudflare. that simple :)

dario-piotrowicz: mh.... ok, but if we do that then if someone uses the worker adaptor then the `event.context.cf` will contain only the cf object whilst with the pages adaptor `event.context.cf` will contain the `CfRequestContext`, so this might lead to confusion?

if you are ok with this misalignment of the two I am as well, I'd also want to update the worker adaptor anyways in a followup PR to make it use the ESM syntax, so I could re-align them there (but I wanted to get this pages change out of the way asap so that people can start using Nuxt on Cloudflare pages, the worker adaptor refactoring is not that pressing)

dario-piotrowicz:
No longer needed after https://github.com/unjs/nitro/pull/1004

Issue #1000, opened 2023-02-27
Auto infer types from `.toJSON` if property exisits


pi0:
When returning a value in API handlers, we use `JSON.stringify` for it finally. If `toJSON` exists, it can have more precise type.

PR #1002, opened 2023-02-27
feat(types): type $fetch to match json serialization output

### LINKED ISSUE #804: Improve type-safety of $Fetch
### Describe the feature

Currently, the API routes type is returned as is, even though it's being JSON serialized when being passed over the wire. This can cause some problems in terms of DX, It can break out of the type-safe mental model and introduces some overhead in the case of being careful about managing un-serializable types (such as Date, Map, Set, etc.).

Example:
```ts
// api
export default defineEventHandler(() => {
  return {
    createdAt: new Date()
  }
})

// client
const data = $fetch('/api/posts') // `data.createdAt` is returned as `Date` even though the real type is `string`
```

My proposed solution is one of the following:

1. Allow passing a custom serializer to h3.
2. Following similar API solutions in the space (such as remix and trpc), we can convert the return type to match JSON serialization output. ([I did a similar thing for Nuxt `useFetch`](https://gist.github.com/mahdiboomeri/36bbd43a1c4fab8e01910ed475bff6ea))

Related Issues: [nuxt/nuxt.js#15590](https://github.com/nuxt/nuxt.js/issues/15590), https://github.com/nuxt/nuxt/issues/12831

### Additional information

mahdiboomeri:
### 🔗 Linked issue

resolves #804, resolves https://github.com/unjs/nitro/issues/1000

### ❓ Type of change

### 📚 Description

This PR transforms the output type of $Fetch to better match what JSON.stringify returns (using trpc and remix implementation).

This is potentially a breaking change.

### 📝 Checklist

### DIFF
--- a/src/types/serialize.ts
+++ b/src/types/serialize.ts
@@ -1,0 +1,56 @@
+/**
+ * @link https://github.com/remix-run/remix/blob/2248669ed59fd716e267ea41df5d665d4781f4a9/packages/remix-server-runtime/serialize.ts
+ */
+type JsonPrimitive =
+  | string
+  | number
+  | boolean
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | String
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | Number
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | Boolean
+  | null;
+// eslint-disable-next-line @typescript-eslint/ban-types
+type NonJsonPrimitive = undefined | Function | symbol;
+
+/*
+ * `any` is the only type that can let you equate `0` with `1`
+ * See https://stackoverflow.com/a/49928360/1490091
+ */
+type IsAny<T> = 0 extends 1 & T ? true : false;
+
+type FilterKeys<TObj extends object, TFilter> = {
+  [TKey in keyof TObj]: TObj[TKey] extends TFilter ? TKey : never;
+}[keyof TObj];
+
+// prettier-ignore
+export type Serialize<T> =
+ IsAny<T> extends true ? any :
+ T extends JsonPrimitive ? T :
+ T extends Map<any,any> | Set<any> ? object :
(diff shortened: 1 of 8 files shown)


manniL (approved):

src/types/serialize.ts
```diff
@@ -0,0 +1,56 @@
...
+
+type FilterKeys<TObj extends object, TFilter> = {
+  [TKey in keyof TObj]: TObj[TKey] extends TFilter ? TKey : never;
+}[keyof TObj];
+
+// prettier-ignore
+export type Serialize<T> =
+ IsAny<T> extends true ? any :
+ T extends JsonPrimitive ? T :
+ T extends Map<any,any> | Set<any> ? object :
```
danielroe, line 32: I think this will always serialize to an empty object:

```suggestion
 T extends Map<any,any> | Set<any> ? Record<string, never> :
```

danielroe (approved):
This is really great.

I only have one suggestion, so we can avoid the super-permissive `object` type. Otherwise 👌

mahdiboomeri:
@danielroe Thanks 💚, I've also updated the tests. It should be ready to go.

danielroe:
@mahdiboomeri Just to let you know we had to revert part of this as a hotfix owing to increased type complexity: see https://github.com/unjs/nitro/pull/1050. If you have any ideas, a PR would be most welcome 🙏

mahdiboomeri:
@danielroe I think moving `Serialize` (and `Simplify`) to `nitro-routes.d.ts` can help. I did a quick experiment and it seems like it improves type checking duration.

Before: 
Screenshot 2023-03-15 at 20-00-53 trace json (1 MB) - Perfetto UI

After:
Screenshot 2023-03-15 at 20-00-33 trace json (1 MB) - Perfetto UI

Is there any reproduction on this?

danielroe:
Previously, it reproduced in the Nuxt `test:types` check.

PR #1004, opened 2023-03-01
feat(cloudflare-pages): rewrite with module syntax

pi0:
### ❓ Type of change

### 📚 Description

This PR introduces several improvements to the (currently experimental) cloudflare pages preset using module syntax and [advanced mode](https://developers.cloudflare.com/pages/platform/functions/advanced-mode/) `_worker.js` output.

While testing, I noticed zero-configs is broken with `dist` link (it was only supported for Nuxt also since it was doing `dist` symlinking and has it as default in CF pages). New output used top-level `dist` instead of `.output/public` also making it more consistent with their [other FW presets](https://developers.cloudflare.com/pages/platform/build-configuration/#framework-presets). This is a slightly breaking change for the experimental preset if `.output/public` was hardcoded should be changed to `dist`

Next steps: Supporting [`_redirects`](https://developers.cloudflare.com/pages/platform/redirects) and [`_headers`](https://developers.cloudflare.com/pages/platform/headers)

Also new in this preset: We use local fetch to directly generate response. it should make normalization more consistent in single place ([in unenv](https://github.com/unjs/unenv/blob/e3999ebaf1016b1bf9d716b0098be770abf5b8bf/src/runtime/fetch/index.ts#L17)

### 📝 Checklist


pi0:
Waiting for CI

image

dario-piotrowicz:
@altryne yes if you use the pages adaptor, I don't think the improvement has been applied on the worker one (as you can see from https://github.com/unjs/nitro/issues/1054)

PR #1021, opened 2023-03-06
feat: allow customising generated tsconfig path

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/13920

### ❓ Type of change

### 📚 Description

This allows customising the `types/tsconfig.json` path in case we want it to be in a different location or file name.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -175,12 +175,7 @@ export async function writeTypes(nitro: Nitro) {
       ],
     };
     await writeFile(
-      join(
-        nitro.options.buildDir,
-        typeof nitro.options.typescript.generateTsConfig === "string"
-          ? nitro.options.typescript.generateTsConfig
-          : "types/tsconfig.json"
-      ),
+      join(nitro.options.buildDir, nitro.options.typescript.tsconfigPath),
       JSON.stringify(tsConfig, null, 2)
     );
   }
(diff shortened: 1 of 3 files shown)


src/types/nitro.ts
```diff
@@ -228,7 +228,7 @@ export interface NitroOptions extends PresetOptions {
   // Advanced
   typescript: {
     internalPaths?: boolean;
-    generateTsConfig?: boolean;
+    generateTsConfig?: boolean | string;
```
pi0, line 231: Maybe we can use an explicit option like `tsconfigPath` ?

src/build.ts
```diff
@@ -175,7 +175,7 @@ export async function writeTypes(nitro: Nitro) {
       ],
     };
     await writeFile(
-      join(nitro.options.buildDir, "types/tsconfig.json"),
+      join(nitro.options.buildDir, nitro.options.typescript.tsconfigPath),
```
pi0, line 178: I think we should switch to `resolve` now to allow absolute paths wdyt?

```suggestion
      resolve(nitro.options.buildDir, nitro.options.typescript.tsconfigPath),
```

pi0 (approved):

PR #1027, opened 2023-03-06
refactor: split out api types

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/13920

### ❓ Type of change

### 📚 Description

This is a follow-on from https://github.com/unjs/nitro/pull/1021, adding support for splitting explicit `nitropack` augmentations from auto-import types (which affect the global scope). This should mean that a consuming framework, like Nuxt, can pull in `types/nitro-augmentations.d.ts` _safely_ for typed `$fetch` without also adding the auto-import types to the global namespace.

I've implemented this (with `/// <reference />`) to keep the change backward-compatible in case users are explicitly referencing `nitro.d.ts`, but with this change `nitro-augmentations.d.ts` also becomes part of the public API (so better naming ideas are definitely welcome)...

### 📝 Checklist


pi0:
Nice change. Can we with the same PR, split to explicit `nitro-imports.d.ts` too? (then nitro can reference both)

(further, i think we might split `nitro-routes` and `nitro-config` too. logic in same file is getting longer over the time)

Issue #1029, opened 2023-03-07
Prerender crawler breaks for links with spaces


juliesaia:
### Environment

Nitro: 2.2.3

### Reproduction

https://stackblitz.com/edit/nuxt-starter-1qwjem?file=pages/index.vue

### Describe the bug

Links with spaces in the `href` get cut off by the regex in `prerender.ts`. In the min repro, it should prerender `/link with space` but it prerenders `/link` instead.

### Logs

```
ℹ Initializing prerenderer                                                                                                       nitro 05:48:07
ℹ Prerendering 1 initial routes with crawler                                                                                     nitro 05:48:09
  ├─ / (103ms)                                                                                                                   nitro 05:48:09
  ├─ /link (8ms)
```

pi0:
Thanks for issue and reproduction.

For reference, space is not allowed in HTML spec for values: 

<img width="1109" alt="image" src="https://user-images.githubusercontent.com/5158436/223406636-982aed2b-c61d-4a70-9609-52d25e2a13b7.png">

PR #1030, opened 2023-03-07
fix(prerender): allow spaces in `href` value regex

juliesaia:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1029

### ❓ Type of change

### 📚 Description

Removes the space character from the end group character set in `LINK_REGEX`. Only single/double quotes can end the match now, allowing for `href` paths with spaces in them.

Resolves https://github.com/unjs/nitro/issues/1029

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -239,7 +239,7 @@ export async function prerender(nitro: Nitro) {
   }
 }
 
-const LINK_REGEX = /href=["']?([^ "'>]+)/g;
+const LINK_REGEX = /href=["']?([^"'>]+)/g;
 
 function extractLinks(
   html: string,


pi0 (approved):
LGTM. BTW space is not normally an allowed char (https://github.com/unjs/nitro/issues/1029#issuecomment-1457987803)

manniL:
Should we emit a warning in this case (and allow it)?

pi0:
> Should we emit a warning in this case (and allow it)?

We might do it in higher order (prerenderer) with an opt-in validate feature. Seems browsers can handle it at least apart of the spec.

PR #1032, opened 2023-03-07
docs: add `prerender.ignore` description

hanneskuettner:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Add short description for the `prerender.ignore` option implemented in https://github.com/unjs/nitro/commit/a6cbbbe482243a3d8dc4c3917ec406b74fdb2a8f. 

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #1033, opened 2023-03-07
feat: support regex for `prerender.ignore`

### LINKED ISSUE #308: allow excluding routes from prerenderer crawling
Would be nice if something is intended to be dynamic.

Perhaps we could support `prerender.exclude` with RegExp, function or string signature? Or a render hook with option to skip rendering (https://github.com/unjs/nitro/issues/88)?

thezzisu:
### 🔗 Linked issue

Resolves #308

### ❓ Type of change

### 📚 Description

Enhance prerender.ignore with the ability to use RegExp.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -112,7 +112,11 @@ export async function prerender(nitro: Nitro) {
 
     // Check for explicitly ignored routes
     for (const ignore of nitro.options.prerender.ignore) {
-      if (route.startsWith(ignore)) {
+      if (typeof ignore === "string") {
+        if (route.startsWith(ignore)) {
+          return false;
+        }
+      } else if (ignore.test(route)) {
         return false;
       }
     }
--- a/docs/content/3.config/index.md
+++ b/docs/content/3.config/index.md
@@ -261,9 +261,11 @@ When `cache` option is set, handlers matching pattern will be automatically wrap
 
 ### `prerender`
 
-Default: `{ crawlLinks: false, routes: [] }`
+Default: `{ crawlLinks: false, ignore: [], routes: [] }`
 
 Prerendered options. Any route specified will be fetched during the build and copied to the `.output/public` directory as a static asset.
+
+The `ignore` option accepts an array of string or regex to specify routes to ignore. Routes that starts with the string or matches the regex will be ignored.
 
 If `crawlLinks` option is set to `true`, nitro starts with `/` by default (or all routes in `routes` array) and for HTML pages extracts `<a href="">` tags and prerender them as well.
 
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -208,7 +208,7 @@ export interface NitroOptions extends PresetOptions {
   devErrorHandler: NitroErrorHandler;
   prerender: {
     crawlLinks: boolean;
-    ignore: string[];
+    ignore: (string | RegExp)[];
     routes: string[];
   };
 


thezzisu:
Notice: maybe conflict with #1032

Atinux:
Will first merge #1032 to rebase your branch

Atinux:
Fixed the conflicts and improve the docs at the same time.

src/prerender.ts
```diff
@@ -112,7 +112,11 @@ export async function prerender(nitro: Nitro) {
 
     // Check for explicitly ignored routes
     for (const ignore of nitro.options.prerender.ignore) {
-      if (route.startsWith(ignore)) {
+      if (typeof ignore === "string") {
+        if (route.startsWith(ignore)) {
+          return false;
+        }
+      } else if (ignore.test(route)) {
```
Atinux, line 119: Shall we test if `ignore.test` is a function to make sure it does not break anything?

Atinux:
Wondering if we could add a test case for it as well.

pi0:
I'm thinking we might be better first allow `**` and `*` in ignore patterns to keep it consistent with route rules. We can also introduce a hook / function filter to allow ignoring routes on demand for more advanced cases.

Atinux:
Allowing the pattern `**` and `*` will solve your usecase @thezzisu ?

Atinux:
So what are the next steps?

src/prerender.ts
thezzisu: For now, ignore would be neither string or Regex if the provided configuration is checked by TypeScript. So there's no need to test wether `ignore.test` exists.

thezzisu:
As @pi0 said, custom hooks/callbacks would be the most powerful solution; but, for most use cases, RegEx should be enough.

> Allowing the pattern `**` and `*` will solve your usecase @thezzisu ?

In fact, my use cases needs the power of "excluded ignore", which is not easily solvable with route rules matching. The underlying library, whose radix-tree-matching do not produce result with the same order as in the input.

thezzisu:
> Wondering if we could add a test case for it as well.

Should I implement more strict input checking? If the user uses TypeScript, the current solution will work fine.

Atinux:
I think it's fine for types using TS.

If I understand correctly, you don't want to support regex but the ability to give a function instead @pi0 ?

thezzisu:
> A hook would be nice for advanced usecases.
> 
> I prefer to keep `ignore` strict string pattern format as much as possible. This (same as route rules) is an advantage to be able static rules to provider rules which by introducing regexes to route defs we lose.
> 
> What do you think about introducing `prerender.include` as well:
> 
> ```js
> prerender: {
>   include: ['/static/**'],
>   exclude: ['/**']
> }
> ```

But how about the priority? Is the 'include' a 'negative exclude' like `!blahblah` syntax in `.gitignore`?

Since this option won't affect how route options work (in fact, it's a option for the crawler, not the server), I think RegExs should be fine.

thezzisu:
> > But how about the priority?
> 
> The most precise rule takes into priority
> 
> > Since this option won't affect how route options work
> 
> The point is route rule definition consistency and having most simple way to do things when possible.

OK. How about add a `prerenderRules` like `routeRules` that using the similar syntax `Record<string, SomeOptions>`?

thezzisu:
> > > But how about the priority?
> > 
> > 
> > The most precise rule takes into priority
> > > Since this option won't affect how route options work
> > 
> > 
> > The point is route rule definition consistency and having most simple way to do things when possible.
> 
> OK. How about add a `prerenderRules` like `routeRules` that using the similar syntax `Record<string, SomeOptions>`?

The reason is that, routeRules is **for server**, but the prerender is a standalone thing though.

pi0:
We do actually support `prerender: true` for `routeRules` (route rules are unified routing rules not specific to server).

We can indeed support `prerender: false` to allow overiridng for specific patterns 👍🏼

thezzisu:
> We do actually support `prerender: true` for `routeRules` (route rules are unified routing rules not limited to server).
> 
> We can indeed support `prerender: false` to allow overiridng for specific patterns 👍🏼

So the problem is I want to make `/` not prerendered (SSR) when using the server built by nuxt build, but to be prerendered when using nuxt generate.

pi0:
Yes that would be possible! (SSR renderer won't be affected by `prerender` rule only `nitro prerender`/`nuxt generate` uses this specific rule).

thezzisu:
Sorry for the delay. Now the problems becomes that the https://github.com/unjs/radix3#route-matcher is not matching the most specific rule! Let's see a example below:

image
image

Simple swapping the rules will affect the output.

Atinux:
Nice catch @thezzisu 

It seems we need a way to prioritize the routes for best matching strategy, similar to what vue-router 4 is doing 🤔

thezzisu:
I think this would be a better solution than supporting regex. I'll close this pull request, and open a issue on [unjs/radix3](https://github.com/unjs/radix3#route-matcher) instead.

pi0:
Radix 3 basically does that :) but we don't use it internally in nitro prerender inference...

Atinux:
So open back an issue in Nitro pre-render to support this I guess 😅

pi0:
~> https://github.com/unjs/nitro/issues/1078

PR #1038, opened 2023-03-08
fix(cache): use stale value by default when `swr` is enabled

Atinux:
### ❓ Type of change

### 📚 Description

**To be properly tested**.

So far, when `swr: true`, we should return the stale data for all incoming request until the cache is updated in background.

Also, I am not sure to understand the `(opts.staleMaxAge || 0) >= 0` condition here.

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -80,7 +80,11 @@ export function defineCachedFunction<T = any>(
     const _resolve = async () => {
       const isPending = pending[key];
       if (!isPending) {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the PR.

`staleMaxAge` is the time when we should stop responding to stale responses.

https://github.com/unjs/nitro/pull/857 supported `-1` value as a quick solution to allow setting stale time to **inifnity**. 

I think we can set it as default.

pi0 (approved):
I think this is a good hotfix to be back to previous behavior until handling staleMaxAge properly

PR #1062, opened 2023-03-16
fix(azure): support custom baseURL

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/1061
https://github.com/nuxt/nuxt/issues/19385

### ❓ Type of change

### 📚 Description

This adds support for custom baseURLs by nesting the public directory with the correct prefix. (We still have to write our empty `index.html` in the root, though.)

### 📝 Checklist

### DIFF
--- a/src/presets/azure.ts
+++ b/src/presets/azure.ts
@@ -8,6 +8,7 @@ export const azure = defineNitroPreset({
   entry: "#internal/nitro/entries/azure",
   output: {
     serverDir: "{{ output.dir }}/server/functions",
+    publicDir: "{{ output.dir }}/public/{{ baseURL }}",
   },
   commands: {
     preview:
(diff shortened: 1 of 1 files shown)


src/presets/azure.ts
```diff
@@ -135,6 +136,14 @@ async function writeRoutes(nitro: Nitro) {
     JSON.stringify(config, null, 2)
   );
   if (!indexFileExists) {
-    await writeFile(resolve(nitro.options.output.publicDir, "index.html"), "");
+    const baseURLSegments = nitro.options.baseURL.split("/").filter(Boolean);
+    const relativePrefix = baseURLSegments.map(() => "..").join("/");
```
pi0, line 140: Could we handle `..` using `new URL()`?

danielroe: you mean, like this?

```js
const baseURLSegments = nitro.options.baseURL.split("/").filter(Boolean);
const relativePrefix = baseURLSegments.map(() => "..").join("/");
const url = new URL(
  joinURL(relativePrefix, "index.html"),
  pathToFileURL(nitro.options.output.publicDir)
);
```

Yes, that works too, but means we convert path -> file URL -> path and I would assume that resolve is just as good at handling the relative prefix.

pi0 (approved):

PR #1070, opened 2023-03-21
docs(routing): update add `$fetch` usage note

Draichi:
After some confusion trying to identify which command to use to fetch an API using only Nitro, I was not sure if the same command on Nuxt would work here.

To avoid this confusion and to make it clear to anybody that `$fetch` is the command to use I believe this should be in the documentation

### ❓ Type of change

### 📝 Checklist


pi0:
Thanks!

PR #1071, opened 2023-03-21
fix(vercel-edge): fix route generation

smaeda-ks:
It seems `routeRules` generates invalid routes when using the `vercel-edge` preset.

This is because nitro only creates pretender functions for the `vercel` preset:
https://github.com/unjs/nitro/blob/4585ca856b0fb4d5601e30fc36f9dbde4aef03de/src/presets/vercel.ts#L51-L85

and the `generateEndpoint` function isn't considering the `vercel-edge` preset:
https://github.com/unjs/nitro/blob/4585ca856b0fb4d5601e30fc36f9dbde4aef03de/src/presets/vercel.ts#L226-L234

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/presets/vercel.ts
+++ b/src/presets/vercel.ts
@@ -193,7 +193,10 @@ function generateBuildConfig(nitro: Nitro) {
           }
           return {
             src,
-            dest: generateEndpoint(key) + "?url=$url",
+            dest:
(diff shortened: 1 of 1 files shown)


danielroe (approved):

pi0 (approved):
Thanks 💯

Issue #818, opened 2023-01-09
vercel-static


pi0:
https://github.com/nuxt/nuxt.js/issues/12353

https://github.com/unjs/nitro/issues/797

This preset should generate proper manifest for vercel static deployment without functions and might be also auto enabled when `static: true` (new) is passed from CLI (nuxt)

PR #1073, opened 2023-03-21
feat: add `vercel-static` and `netlify-static` presets

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/818
also https://github.com/nuxt/image/issues/638, https://github.com/nuxt/image/issues/617, https://github.com/nuxt/image/issues/757

### ❓ Type of change

### 📚 Description

This adds new `vercel-static` and `netlify-static` presets that will use platform features to handle redirects, headers, etc. For Vercel, this will also include Vercel Images config and solve https://github.com/nuxt/image/issues/638.

It is not fully auto-detectable - frameworks that use it will need to pass `build: false` or otherwise configure this (which they can easily do if user is running some kind of `generate` command). But with that, we can then automatically enable provider-specific configuration for zero-config 'static+' deploys.

### 📝 Checklist


yunyuyuan:
Does vercel-static support Vercel serverless functions?From my tests, it seems that currently `vercel` or `vercel-edge` only support `nuxt build` and do not work with `nuxt generate`, when calling the `/server/api/xxx`, it returns a 405 error

danielroe:
No, this PR is purely static and does not render any serverless functions, just static assets, redirects, and headers.

src/utils/index.ts
```diff
@@ -107,10 +107,19 @@ const autodetectableProviders = {
...
+const autodetectableStaticProviders = {
+  vercel: "vercel-static",
+};
+
 export function detectTarget() {
   return autodetectableProviders[provider];
 }
 
+// TODO: update when https://github.com/unjs/nitro/pull/1072 merges
+export function detectStaticTarget() {
```
pi0, line 119: PR mergeed, also refactor, we might support `{ static }?` for `detectTarget` itself to pas conditions like this without introducing new one

pi0:
Thanks for the work on this PR @danielroe ❤️ More I'm thinking it would be nice if we (in an initial PR) introduce `static` preset + `static` flag that basically stops building server when is set. We can then extend `vercel-static` from it. Wdyt?

danielroe:
Seems good to me! Shall I refactor this PR to include that as well, or would you value a separate PR?

pi0:
A seperate PR would be nicer if you don't mind ❤️

src/utils/index.ts
pi0: I mean not all options but detect conditions (practically merging two utils into one)

src/options.ts
```diff
@@ -149,7 +148,7 @@ export async function loadOptions(
   options.preset =
     presetOverride ||
     (layers.find((l) => l.config.preset)?.config.preset as string) ||
-    defaultPreset;
+    (detectTarget({ static: !options.build }) ?? "node-server");
```
pi0, line 151: I am thinking since we didn't release nitropack yet, would it made sense that we introduced `static: boolean` flag instead of top level `build` flag? It is mainly for sake of static support we introduced flag...

src/presets/vercel.ts
```diff
@@ -163,50 +181,60 @@ function generateBuildConfig(nitro: Nitro) {
...
-            return {
-              src,
-              dest: "/__nitro",
-            };
-          }
+    ],
+  });
+
+  // Early return without serverless functions if we are using a static preset
+  if (nitro.options.preset === "vercel-static") {
```
pi0, line 188: (with that change, this could be like `if (nitro.options.static)` in a more generic way

src/options.ts
danielroe: yes, that works 👍

src/presets/vercel.ts
danielroe: ah, yes, this should have been instead `!nitro.options.build` like in netlify preset.

src/options.ts
danielroe: you still prefer top-level static? very happy to change it if, on reflection, you feel that's the best.

pi0 (approved):
LGTM. Let's do static option refactor in another change

PR #1097, opened 2023-03-29
fix: return file URLs for absolute externals

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/19453

### ❓ Type of change

### 📚 Description

We already normalize absolute paths to file URLs lower down; this ensures we do the same if we can early-return by virtue of having an id explicitly in `opts.external`.

### 📝 Checklist


pi0:
Following up via https://github.com/nuxt/nuxt/issues/19453#issuecomment-1745410511, seems reproduction is not happening anymore on Nuxt. 

For sake of maintenance, I'm thinking of closing this PR but definitely up to consider (safer) fix if there is still an unresolvable issue in Nuxt land.

PR #1115, opened 2023-04-05
fix(dev): ensure socket path is accessable

sbittmann:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/15164

### ❓ Type of change

### 📚 Description

On windows we are facing the issue, that the devServer returns status 503. This is caused by the `existsSync()` call, which returns false if the the api is under load. With `fs.promises.access` I didn't face this problem anymore.

### 📝 Checklist


src/dev/server.ts
```diff
@@ -173,7 +173,14 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
     eventHandler(async (event) => {
       await reloadPromise;
       const address = currentWorker && currentWorker.address;
-      if (!address || (address.socketPath && !existsSync(address.socketPath))) {
+      if (
+        !address ||
+        (address.socketPath &&
+          !(await fsp.access(address.socketPath).then(
```
pi0, line 179: Maybe we can try sync version to avoid a stalled status? (access can be not available in next async think we call actual error handler)..

Hebilicious: Do you mean using https://nodejs.org/api/fs.html#fsaccesssyncpath-mode instead of the promise version ?
This change should fix https://github.com/nuxt/nuxt/issues/15164

datismoa: @Hebilicious 

You cannot just use `accessSync` since it may throw an error:

> If any of the accessibility checks fail, an Error will be thrown. Otherwise, the method will return undefined. (https://nodejs.org/api/fs.html#fsaccesssyncpath-mode)

This causes a dev server to be completely down. To avoid this we should wrap `accesSync` in try-catch. 

Even with that done, the problem does not seem to be fixed. The server cannot handle 200 requests in my case:

```javascript
for (let i = 0; i < 200; i++) {
  fetch('http://localhost:3000/api/test')
}
```

```typescript
// server/api/test.get.ts

export default eventHandler(() => {
  return '123098'
})
```

> POST http://localhost:3000/api/test 503 (Server Unavailable)

I've done more stressful tests with [autocannon](https://www.npmjs.com/package/autocannon). The results:
~ 81k requests in 5.05s, 1.95 MB read
~ 64k errors

pi0:
I am aware this is an important DX fix. However needs to be tested more if we can find a solution without per/request overhead of checking sockets and that this works universally. Delaying to next patch/minor release.

csprance:
This issue still seems to be happening even with this fix.

pi0:
Tracking via https://github.com/unjs/nitro/issues/1454

PR #1124, opened 2023-04-11
feat(vercel, netlify): introduce `isr` route rule

### LINKED ISSUE #946: Dedicated cache rules for `vercel` ISR and `netlify` Builders
Mapping built-in `static` and `swr` cache rules to platform special cache strategies currently introduced lots of confusion for users about how and where caching works with nitro. (should be everywhere, and consistent)

The current mapping strategy is not exactly the same behavior. Netlify `swr` and `static` are mapped to `builders` does not specifies ttl and vercel maps to a strategy that supports expiration but we do not provide stale time which can be different). 

Apart from user confusion and not being able to map the strategies precisely, the built-in caching layer can be used universally, _also_ for the vercel and netlify platforms. Allowing to cache in custom storage (could be useful particularly if we want to keep cache across deployments which is not possible currently by map)

I think we need to either introduce platform flags or a new cache flag that can be mapped to ISR/Builder which is not `swr`/`static` and next major version of nitro, avoid this mapping.

pi0:
### 🔗 Linked issue

Resolves #946

### ❓ Type of change

### 📚 Description

We had been experimenting vercel ISR and netlify Builders feature via an implicit cache route rule. 

Introducing a new `isr` flag, unblocks to combine nitro native cache (SWR) with platform native rules and giving the flexibility to use the most effective caching mechanism. (Currently, on these platforms nitro cache/storage cannot be used)

Note: This is a semi-breaking change for this platforms however switching `swr` to built-in behavior still preserves in memory cache so caching keeps working only can be changed to either isr or permanent by mounting `/storage` for production.

This PR also drops (almost unused) `static` cache rule.

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -310,11 +310,6 @@ export async function loadOptions(
         routeRules.cache.maxAge = routeConfig.swr;
       }
     }
-    // Cache: static
-    if (routeConfig.static) {
-      routeRules.cache = routeRules.cache || {};
-      routeRules.cache.static = true;
-    }
     // Cache: false
     if (routeConfig.cache === false) {
       routeRules.cache = false;
--- a/src/presets/netlify.ts
+++ b/src/presets/netlify.ts
@@ -83,19 +83,15 @@ async function writeRedirects(nitro: Nitro) {
     (a, b) => a[0].split(/\/(?!\*)/).length - b[0].split(/\/(?!\*)/).length
   );
 
-  // Rewrite static cached paths to builder functions
+  // Rewrite static ISR paths to builder functions
   for (const [key, value] of rules.filter(
-    ([_, value]) =>
-      value.cache === false ||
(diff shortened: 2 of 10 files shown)


pi0:
> But maybe worth (if not using vercel/netlify) setting native nitro cache options if isr is enabled?

It would be indeed a nice idea if we could _emulate_ isr functionality for other production platforms. Let's iterate over this. I think one main difference that swr lacks is to invalidate/segment cache based on deployment id.

Feel free to make PR for both ideas btw.

Hebilicious:
It should be possible to implement ISR for the cloudflare providers by leveraging the Cloudflare cache API like [here](https://developers.cloudflare.com/workers/examples/cache-api/). The isr value can be forwarded to the cache-control header as well.
This would be very neat 👌

PR #1125, opened 2023-04-11
feat: experimental sqlite support

pi0:
### ❓ Type of change

### 📚 Description

Nitro provides a platform-agnostic abstraction layer for many features including unstorage for powerful and versatile KV storage. However, there are limitations to supporting full relational DBs via KV.

The idea behind this new future is to provide a simple [SQLite](https://sqlite.org/index.html) database out of the box making it available for development time and also for small production deployments. 

Same as other parts of Nitro, DB provider can be abstract and from any service provider. In this PR, i have added a playground for [cloudflare D1](https://developers.cloudflare.com/d1/) integration. (needs refactor!)

Nitro provides no ORM but we can integrate with a driver for 3rd party orms like [drizzle](https://github.com/drizzle-team/drizzle-orm) easily. (We might also build some simple conventions in server folder for migrations, etc in later steps)

Usage:

```ts
  const db = await useSQliteDB();
  await db.exec(
    "CREATE TABLE IF NOT EXISTS users (id TEXT PRIMARY KEY, firstName TEXT, lastName TEXT, email TEXT)"
  );
  const id = String(Math.round(Math.random() * 10000));
  await db.prepare("INSERT INTO users VALUES (?, 'John', 'Doe', '')").run(id);
  const row = await db.prepare("SELECT * FROM users WHERE id = ?").get(id);
  return {
    row,
  };
});

```

Local playing:

- Switch to `feat/exp-db` branch
- Development: `pnpm nitro dev examples/sqlite`
- Production: 
   - `NITRO_PRESET=cloudflare-module pnpm nitro build examples/sqlite`
   - `npx wrangler dev examples/sqlite/.output/server/index.mjs --local`

Credits to @Atinux for pushing the initiatives behind this idea ❤️ 

### 📝 Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ staticwebapp.config.json
 .eslintcache
 
 test/fixture/functions
+
+*.sqlite*
+*.db
--- a/examples/sqlite/nitro.config.ts
+++ b/examples/sqlite/nitro.config.ts
@@ -1,0 +1,5 @@
+import { defineNitroConfig } from "nitropack";
+
+export default defineNitroConfig({
+  minify: false,
+});
--- a/examples/sqlite/package.json
+++ b/examples/sqlite/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "example-sqlite",
+  "private": true,
+  "scripts": {
+    "dev": "nitro dev",
+    "build": "nitro build"
+  },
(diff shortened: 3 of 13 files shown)


pi0:
Moving to https://github.com/unjs/nitro/pull/1351

Issue #798, opened 2023-01-03
GitHub Pages preset


manniL:
### Describe the feature

The GitHub Pages preset should be based on the static preset (#797) and i.e. include a `.nojekyll` file after generation.

Related: https://github.com/nuxt/nuxt.js/issues/12480

### Additional information

- [X] Would you be willing to help implement this feature?

PR #1133, opened 2023-04-12
feat: add `github-pages` preset

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/798

https://github.com/nuxt/nuxt/issues/12480
https://github.com/nuxt/nuxt/issues/12570

### ❓ Type of change

### 📚 Description

This adds a basic GitHub Pages preset. It creates a `.nojekyll` file and by default uses https://github.com/tschaub/gh-pages as deploy command.

### ✅ TODO

- [x] docs (including GH Actions workflow?)
- [ ] rethinking deploy command
- [ ] base URL support - this may need to be documentation

### 📝 Checklist

### DIFF
--- a/src/presets/github-pages.ts
+++ b/src/presets/github-pages.ts
@@ -7,6 +7,12 @@ export const githubPages = defineNitroPreset({
   commands: {
     deploy: "npx gh-pages -d ./public",
   },
+  prerender: {
+    routes: [
+      // https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site
+      "/404.html",
+    ],
+  },
   hooks: {
     async compiled(nitro) {
       await fsp.writeFile(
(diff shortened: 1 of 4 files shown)


src/presets/github-pages.ts
```diff
@@ -0,0 +1,18 @@
+import fsp from "node:fs/promises";
+import { join } from "pathe";
+import { defineNitroPreset } from "../preset";
+
+export const githubPages = defineNitroPreset({
+  extends: "static",
+  commands: {
+    deploy: "npx gh-pages -d ./public",
+  },
+  hooks: {
```
pi0, line 10: Maybe we could add 404.html to prerender routes by default too?

danielroe: I guess we do it by default in Nuxt, but does it make sense for Nitro to do so when there might not be an HTML renderer?

src/presets/static.ts
```diff
@@ -8,6 +8,7 @@ export const _static = defineNitroPreset({
   },
   prerender: {
     crawlLinks: true,
+    routes: ["/404.html"],
```
pi0, line 11: This only makes sense for github i guess for default behavior (not sure what other static providers respect `404.html` as fallback)

PR #1142, opened 2023-04-14
feat: add firebase functions v2 preset

### LINKED ISSUE #90: Support firebase regions
https://github.com/nuxt/nuxt.js/issues/13341

https://firebase.google.com/docs/functions/manage-functions#modify-region

### LINKED ISSUE #666: [preset: firebase] Allow setting of memory allocation for Firebase functions
### Describe the feature

I would like to be able to set the memory allocation and timeout for Firebase functions generated by Nitro. For Nuxt, Nitro generates the `server` function which is used for SSR.

The default memory allocation for Firebase functions is 256MB, however this results in poor performance and slow loading for SSR Nuxt applications. I would like to be able to increase this memory allocation to improve application speed.

As per [Firebase documentation](https://firebase.google.com/docs/functions/manage-functions#set_timeout_and_memory_allocation), this is normally done by setting `Runtime Options` within the `runWith` parameter. However, as far as I can tell, we do not have the ability to modify these currently. I would like these settings exposed on the Nitro API if possible.

Note that it is currently possible to change the memory allocation in the Google Cloud Console, but every time the Nitro Firebase functions are built and deployed, the memory allocation is reset to the default (256MB). This is probably because the `Runtime Options` are not specified in the source code.

### Additional information

### LINKED ISSUE #1141: Support Firebase functions v2 (2nd Gen) as a deploy target.
### Describe the feature

Almost exactly the same as the current `firebase` preset, but make the generated cloud function a 2nd gen function and expose some config to set region, min/max instances etc.

The cloud functions v2 are in beta at the moment and the documentation is here:
https://firebase.google.com/docs/functions/beta

Though Google claim that they are in GA here:
https://cloud.google.com/blog/products/serverless/cloud-functions-2nd-generation-now-generally-available

Proposal is a new preset `firebaseV2` that will extend the current firebase preset, but just use a different import to take advantage of the new features of firebase functions v2.

There has been a few issues #1098 #666 that express a valid requirement for configuring cloud functions, so this feature also includes an easy way to inject options into the cloud function call through the nirto runtime config. This allows the developer to control all options of the firebase function such as region, min/max instances etc.

The options are detailed here:
https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.https.httpsoptions

### Additional information

JamieCurnow:
### 🔗 Linked issue
 
#1141 

### ❓ Type of change

### 📚 Description

This adds a new preset for firebase functions v2 - the second generation of firebase functions that is now in [GA](https://cloud.google.com/blog/products/serverless/cloud-functions-2nd-generation-now-generally-available).

The preset name is `firebaseV2` though I'm not sure if it should be `firebase-v2` or `firebase-next` or `firebase-gen-2` 🤷 .

I've just extended the existing firebase preset as much of the functionality / requirements are the same, just the entry file needs to import functions from a slightly different place and we now have the option of passing in an object to configure the function which is super helpful, rather than chaining.

The new version of functions has many new features and will, no doubt, be the default version soon now that it's in GA. One of the main features is the ability to take concurrent requests which is a *big* win for apps running on cloud functions so that there is less cold start time.

Using v2 could functions should see a speed improvement and gives the developer more flexibility and control over the server.

Resolves #1141 
Resolves #666
Resolves #90

### 📝 Checklist

I'm not 100% on the convention on passing config to a preset. Other presets advocate a config file, that is read during deploy, but with firebase functions it seems that the configuration needs to be in the code. I've used the runtimeConfig object to pass down config. Seems like a sensible place, but I understand that it may also need documentation / changes to the types and I'm not sure that's the road you want to go down...

There were no tests for the firebase preset, so I've not made any tests for this firebaseV2 preset. I have tested that it works in a nuxt3 app and it works great. Lemme know on tests 😉


src/runtime/entries/firebaseV2.ts
```diff
@@ -0,0 +1,11 @@
+import "#internal/nitro/virtual/polyfill";
+// @ts-ignore
+import functions from "firebase-functions/v2";
```
sisou, line 3: Why is a `@ts-ignore` required here?
Also, I believe the best practice nowadays is to use `@ts-expect-error` with a reason.

sisou (commented):
Looks good, I'll test it later!

src/runtime/entries/firebaseV2.ts
JamieCurnow: @sisou I'm not sure actually if/why it's needed.... I used the existing firebase entry as boiler for this and it was in there...:
https://github.com/unjs/nitro/blob/main/src/runtime/entries/firebase.ts#L2

JamieCurnow: It's because the 'firebase-functions' package is not installed in nitro. Rather it should be installed by the parent package that is using nitro. I've updated to ts-expect-error with a comment 👍

JamieCurnow:
@danielroe any chance of looking at this one? Would really appreciate your advice on using runtimeConfig to pass options through - I've not seen this done yet in nitro and wonder if it's the correct approach?

sisou:
`NitroOptions` actually extends `PresetOptions`, which already includes a section for Vercel. I'd say it can go there, next to the `vercel` key. The types would have to be copied in, or linked indeed.

sisou (approved):
My test was successful, btw. Deployed a Firebase V2 function from this branch :+1:

JamieCurnow:
@sisou I've updated this PR to add the firebase functions options to the nitro config. I've added the firebase-functions package to import the types from because it seems a bit mental to be copying them over... Let me know if this is the correct approach.

CC @danielroe

JamieCurnow:
I've just merged main in. I had a merge conflict in the `pnpm-lock.yaml` so accepted all the incoming changes from main in the merge and then did a fresh `pnpm install` before pushing up again.

Is there anything else I need to do before this is ready to go in @danielroe @pi0 @sisou ?

JamieCurnow:
@danielroe @pi0 , anything I need to do to get this moving? Would be great to start using it 😊👌🤘

JamieCurnow:
@pi0 I've merged main back into this branch - is there anything I need to do for this to be merged in? Firebase functions V2 are out of beta now and I'd love to start using them for our nuxt projects 😊

docs/content/2.deploy/providers/firebase.md
```diff
@@ -90,3 +92,36 @@ Deploying to Firebase Hosting is a simple matter of just running the `firebase d
 NITRO_PRESET=firebase yarn build
 firebase deploy
 ```
+
+## Firebase Functions V2
+
+**Preset:** `firebaseV2` ([switch to this preset](/deploy/#changing-the-deployment-preset))
+
+This preset will deploy to [firebase functions V2](https://firebase.google.com/docs/functions/beta/get-started) - the 2nd generation of cloud functions.
```
Hebilicious, line 100: This link doesn't work anymore, maybe we could link to https://firebase.google.com/docs/functions/version-comparison

docs/content/2.deploy/providers/firebase.md
```diff
@@ -90,3 +92,36 @@ Deploying to Firebase Hosting is a simple matter of just running the `firebase d
...
+
+It works the same as the [`firebase`](#firebase) preset above, but uses the V2 functions runtime.
+
+**Note**: Follow the instructions for the `firebase` preset above for setup and deploy.
+
+The `firebaseV2` preset also supports passing options to the firebase function via the nitro config. These options are defined in the nitro config object under `firebaseV2.httpRequestOptions`.
+
+The `httpRequestOptions` object is passed to `functions.https.onRequest()` function as the first argument as per the [firebase docs](https://firebase.google.com/docs/functions/beta/http-events#trigger_a_function_with_an_http_request).
```
Hebilicious, line 108: This link works but is outdated

docs/content/3.config.md
```diff
@@ -469,3 +469,9 @@ Rollup specific option.
 ### `commonJS`
 
 Rollup specific option. Specifies additional configuration for the rollup CommonJS plugin.
+
+## Preset options
+
+### `firebaseV2`
```
Hebilicious, line 475: Maybe this should be nested under `firebase` to be consistent with `vercel` @pi0

src/types/presets.ts
```diff
@@ -1,3 +1,5 @@
+import { HttpsOptions } from "firebase-functions/lib/v2/providers/https";
```
Hebilicious, line 1:
```suggestion
import type { HttpsOptions } from "firebase-functions/lib/v2/providers/https";
```

src/presets/index.ts
```diff
@@ -10,6 +10,7 @@ export * from "./deno-deploy";
 export * from "./deno-server";
 export * from "./digital-ocean";
 export * from "./firebase";
+export * from "./firebaseV2";
```
Hebilicious, line 13: This preset should be renamed `firebase-v2` for consistency.

src/types/presets.ts
```diff
@@ -65,4 +67,11 @@ export interface PresetOptions {
       maxDuration: number;
     };
   };
+  firebaseV2: {
```
Hebilicious, line 70: should be firebase

package.json
```diff
@@ -76,6 +76,7 @@
     "esbuild": "^0.18.10",
     "escape-string-regexp": "^5.0.0",
     "etag": "^1.8.1",
+    "firebase-functions": "^4.3.1",
```
Hebilicious, line 79: move to dev dependency

Hebilicious (changes requested):

Hebilicious:
I was able to succesfully deploy the v2 preset using the https://github.com/unjs/nitro-deploys repo : 

- https://nitro-39729.web.app/

A few things to note : 

- The deploy command suggested by Nitro, `npx firebase deploy` is wrong, as the packages is called [`firebase-tools`](https://www.npmjs.com/package/firebase-tools)
- Deploying from the CLI with `firebase deploy` is broken, the `.output/server/package.json` and `node_modules` directories needs to be altered manually to remove `fsevents`. I assume that this happens on macos only, windows and linux should be fine.
- Deploying from the CLI requires a special command to enable the cloud functions v2 permissions, which requires the `gcloud` cli : `gcloud functions add-invoker-policy-binding server --project="<project-id>" --region="<region>" --member="allUsers"`. This is documented in the gcloud dashboard in the cloud functions section, but not in the firebase console. We should definitely document that and save people an headache. For some reasons the permissions tab in the gcloud UI doesn't reflect what this command and the permissions can't be enabled from the GUI.
- Streaming is not working due to [google cloud limitations](https://stackoverflow.com/questions/48333855/cloud-function-response-stream/48431968#48431968)

package.json
JamieCurnow: @Hebilicious I agree it should be a dev dep, but I am getting a nitro build error when it is a dev dependancy. The build runs fine when it's a normal dependancy. I'm not sure where to start with that one?

The error I get when it's a dev dependancy is:

```
nitro ⚡ pnpm build                                                                                                              17:34:54

> nitropack@2.5.2 build /Users/JCGeek/Documents/Projects/nitro
> unbuild

ℹ Building nitropack                                                                                                        5:35:05 PM
Error building /Users/JCGeek/Documents/Projects/nitro: SyntaxError: Unexpected token (13:7) in /Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/firebase-functions@4.4.1_firebase-admin@11.9.0/node_modules/firebase-functions/lib/v2/providers/https.d.ts
SyntaxError: Unexpected token (13:7) in /Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/firebase-functions@4.4.1_firebase-admin@11.9.0/node_modules/firebase-functions/lib/v2/providers/https.d.ts
    at pp$4.raise (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:21396:13)
    at pp$9.unexpected (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:18607:8)
    at pp$9.expect (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:18601:26)
    at pp$8.parseExportSpecifiers (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:19655:8)
    at Parser.parseExport (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:25858:32)
    at pp$8.parseStatement (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:18783:74)
    at pp$8.parseTopLevel (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:18664:21)
    at Parser.parse (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:18436:15)
    at Function.parse (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:18486:35)
    at Graph.contextParse (file:///Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/rollup@3.25.3/node_modules/rollup/dist/es/shared/node-entry.js:25522:38) {
  pos: 587,
  loc: {
    column: 7,
    file: '/Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/firebase-functions@4.4.1_firebase-admin@11.9.0/node_modules/firebase-functions/lib/v2/providers/https.d.ts',
    line: 13
  },
  raisedAt: 596,
  frame: '11:  * Options that can be set on an onRequest HTTPS function.\n' +
    '12:  */\n' +
    '13: export interface HttpsOptions extends Omit<GlobalOptions, "region"> {\n' +
    '           ^\n' +
    '14:     /**\n' +
    '15:      * If true, do not deploy or emulate this function.',
  id: '/Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/firebase-functions@4.4.1_firebase-admin@11.9.0/node_modules/firebase-functions/lib/v2/providers/https.d.ts',
  hook: 'resolveId',
  code: 'PLUGIN_ERROR',
  plugin: 'commonjs--resolver',
  watchFiles: [
    '/Users/JCGeek/Documents/Projects/nitro/src/index.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/config.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/cli/cli.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/utils/tree.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/utils.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/fetch.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/preset.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/aws-lambda.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/azure-functions.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/azure.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/base-worker.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/bun.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/cloudflare-module.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/cloudflare-pages.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/cloudflare.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/deno-deploy.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/import-meta.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/deno-server.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/digital-ocean.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/firebase.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/firebase-v2.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/heroku.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/edgio.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/netlify.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/nitro-dev.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/nitro-prerender.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/node-cli.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/node-server.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/node.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/render-com.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/service-worker.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/stormkit.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/presets.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/vercel.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/cleavr.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/layer0.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/flightcontrol.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/lagon.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/static.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/github-pages.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/presets/index.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/utils/index.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/externals.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/esbuild.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/app-config.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/config.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/storage.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/storage.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/cache.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/route-rules.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/plugin.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/plugins.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/error-handler.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/server-handlers.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/app.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/renderer.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/types.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/handler.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/nitro.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/global.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/h3.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/serialize.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/types/index.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/dirs.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/replace.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/virtual.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/dynamic-require.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/externals-legacy.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/timing.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/public-assets.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/server-assets.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/handlers.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/raw.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/storage.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/plugins/app-config.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/rollup/config.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/scan.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/storage.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/compress.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/build.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/imports.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/options.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/nitro.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/dev/vfs.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/dev/error.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/dev/server.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/prerender.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/public-assets.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/runtime/virtual/server-assets.d.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/cli/common.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/cli/commands/dev.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/cli/commands/build.ts',
    '/Users/JCGeek/Documents/Projects/nitro/src/cli/commands/prepare.ts',
    '/Users/JCGeek/Documents/Projects/nitro/node_modules/.pnpm/firebase-functions@4.4.1_firebase-admin@11.9.0/node_modules/firebase-functions/lib/v2/providers/https.d.ts'
  ]
}
```

Hebilicious (approved):
LGTM

package.json
Hebilicious: Oh. I think you can leave it as a regular dependency then !

JamieCurnow:
@Hebilicious 

> * Streaming is not working due to [google cloud limitations](https://stackoverflow.com/questions/48333855/cloud-function-response-stream/48431968#48431968)

👍 Shall I note this in the docs?

> * The deploy command suggested by Nitro, `npx firebase deploy` is wrong, as the packages is called [`firebase-tools`](https://www.npmjs.com/package/firebase-tools)

`npx firebase deploy` works for me 🤔 Did you run `npm install -D firebase-admin firebase-functions firebase-functions-test` as per the docs? Maybe it's because I already have firebase-tools installed globally...

> * Deploying from the CLI with `firebase deploy` is broken, the `.output/server/package.json` and `node_modules` directories needs to be altered manually to remove `fsevents`. I assume that this happens on macos only, windows and linux should be fine.

I've seen the issue and the workaround, and can confirm that it's not an issue on linux as it works when the build is in  github actions. I'll make a PR for the linked issue - have tested filtering out `fsevents` in the firebase preset locally on Mac OS and it works well.
Shall I also / instead add the fix to this PR?

> * Deploying from the CLI requires a special command to enable the cloud functions v2 permissions [...]

I did not get this issue... 🤔 I believe it's to do with your local firebase setup. There is this warning on this page in the firebase docs on setting up cloud functions and hosting:
https://firebase.google.com/docs/hosting/functions

---

Caution: New HTTP and HTTP callable functions deployed with any Firebase CLI lower than version 7.7.0 are private by default and throw HTTP 403 errors when invoked. Either explicitly [make these functions public](https://cloud.google.com/functions/docs/securing/managing-access-iam#allowing_unauthenticated_http_function_invocation) or [update your Firebase CLI](https://firebase.google.com/docs/cli#setup_update_cli) before you deploy any new functions.

---

I've added this warning to the nitro preset docs for v2 firebase preset 😊

Hebilicious:
> @Hebilicious
> 
> > * Streaming is not working due to [google cloud limitations](https://stackoverflow.com/questions/48333855/cloud-function-response-stream/48431968#48431968)
> 
> 👍 Shall I note this in the docs?
> 
> > * The deploy command suggested by Nitro, `npx firebase deploy` is wrong, as the packages is called [`firebase-tools`](https://www.npmjs.com/package/firebase-tools)
> 
> `npx firebase deploy` works for me 🤔 Did you run `npm install -D firebase-admin firebase-functions firebase-functions-test` as per the docs? Maybe it's because I already have firebase-tools installed globally...
> 
> > * Deploying from the CLI with `firebase deploy` is broken, the `.output/server/package.json` and `node_modules` directories needs to be altered manually to remove `fsevents`. I assume that this happens on macos only, windows and linux should be fine.
> 
> I've seen the issue and the workaround, and can confirm that it's not an issue on linux as it works when the build is in github actions. I'll make a PR for the linked issue - have tested filtering out `fsevents` in the firebase preset locally on Mac OS and it works well. Shall I also / instead add the fix to this PR?
> 
> > * Deploying from the CLI requires a special command to enable the cloud functions v2 permissions [...]
> 
> I did not get this issue... 🤔 I believe it's to do with your local firebase setup. There is this warning on this page in the firebase docs on setting up cloud functions and hosting: [firebase.google.com/docs/hosting/functions](https://firebase.google.com/docs/hosting/functions)
> 
> Caution: New HTTP and HTTP callable functions deployed with any Firebase CLI lower than version 7.7.0 are private by default and throw HTTP 403 errors when invoked. Either explicitly [make these functions public](https://cloud.google.com/functions/docs/securing/managing-access-iam#allowing_unauthenticated_http_function_invocation) or [update your Firebase CLI](https://firebase.google.com/docs/cli#setup_update_cli) before you deploy any new functions.
> 
> I've added this warning to the nitro preset docs for v2 firebase preset 😊

This is looking good so far !

- Streaming : Let's add it in another PR 👍🏽  I'm thinking of making a preset streaming compatibility issue. So we can track everything in one place.
- npx firebase deploy : It only works with firebase-tools installed indeed, we should mention that.
- `fsevent` bug : No, this should be fixed with another PR as this affects both presets.
- Thanks, but the docs you mentioned don't fix the issue fore me, I did start everything with a fresh setup and the latest version of the CLI. I would add the command as this was the only way for me to make it work, and I'm sure someone running into this will appreciate it, given how non obvious it was.`gcloud functions add-invoker-policy-binding server --project="<project-id>" --region="<region>" --member="allUsers"`

src/runtime/entries/firebase-v2.ts
```diff
@@ -0,0 +1,10 @@
+import "#internal/nitro/virtual/polyfill";
+import functions from "firebase-functions/v2";
+import { toNodeListener } from "h3";
+import { nitroApp } from "../app";
+import { useRuntimeConfig } from "#internal/nitro";
+
+export const server = functions.https.onRequest(
+  useRuntimeConfig()?.firebase?.httpRequestOptions || {},
```
sisou, line 8: I believe this line needs an update, now that the `firebase` config is no longer under the `runtimeConfig` Nitro setting. Because as I am testing it right now, my config of a different-than-default region is not being respected in the emulator.

sisou (changes requested):

src/runtime/entries/firebase-v2.ts
JamieCurnow: @sisou do you know how to access the nitro config from this entry file? I can't work it out... @Hebilicious any advice?

JamieCurnow:
>npx firebase deploy : It only works with firebase-tools installed indeed, we should mention that.

I've added some more info to the docs 👍 

>fsevent bug : No, this should be fixed with another PR as this affects both presets.

Cool, I'll add the fix to another PR linked to the other issue.

>Thanks, but the docs you mentioned don't fix the issue fore me, I did start everything with a fresh setup and the latest version of the CLI. I would add the command as this was the only way for me to make it work, and I'm sure someone running into this will appreciate it, given how non obvious it was.gcloud functions add-invoker-policy-binding server --project="<project-id>" --region="<region>" --member="allUsers". I believe this could be related to passing HTTP options without explicitly setting invoker:public... Could you try that ?

🤔 You're right, I just got this error deploying to a new project too. I wonder what I did differently to the previous new project that made it work.

Observations:
- Deploy without httpRequestOptions - worked with the first new project, not the second new project. I may have gone through the hosting setup in the firebase console in the first new project... Maybe that has something to do with it.

- Deploy with httpRequestOptions object without an `invoker` key - seems to be not publicly accessible.
- Deploy with httpRequestOptions object with `invoker: "public"` - works
- Deploy again with httpRequestOptions object without an `invoker` key - works still 🤔 
- Delete the function and then deploy again with httpRequestOptions object without an `invoker` key - works still 🤔

As a side note @Hebilicious, I was adding the firebase options inside the `runtimeConfig` in the nitro config file. This is how I had it working to start with before @sisou suggested to put the config on the root nitro config object to align with the vercel config. The issue with that is that I'm now not sure how to access the nitro config object from the firebaseV2 entry file as @sisou has pointed out [it needs to change](https://github.com/unjs/nitro/pull/1142/files#r1250620968).

Maybe that is the root of the weirdness here... Once we figure that out I'll do some more testing and see what the behaviour is to document it. But you really shouldn't need to be running that gcloud command to add a policy binding. Maybe we should think about making the `invoker` setting to `"public"` by default?

src/runtime/entries/firebase-v2.ts
JamieCurnow: I guess the nitro config is build time whereas we're trying to target runtime here... So maybe it should be in runtimeConfig?

Hebilicious: You're right I missed that 🤦🏽 I don't think we can access the Nitro options here.

In the preset file, you can do something similar to vercel with the `rollup:before` hook : 
And you should be able to use the rollup replace plugin to inject the configuration...

If this doesn't work, I have a few other suggestions : 

1) Use the rollup config, add these options to the virtual files, and import them in the entry file. 
2) Leave the options in the Nitro config, attach them to the AppConfig somewhere in the build, then import that in your preset
3) Use the runtimeConfig key directly.

But I think @pi0 advice would be very valuable here, as I'm not sure what's the best way. Probably 2).

Hebilicious (changes requested):
- Find a clean way to pass the httpRequestOptions to the presets
- set invoker: public by default
- document the permissions errors

JamieCurnow:
@Hebilicious I've updated the preset and entry to use appConfig. Works really well - thanks for pointing me in the right direction with that one 🙌 

I've also implemented defaulting the invoker to public and tested deploying on 3 more test projects from scratch and they all worked and were served publicly by default which I think solved the issue with having to run the gcloud command.

Here's the commit - would love to hear your feedback 👍 
https://github.com/unjs/nitro/pull/1142/commits/bd29d152d055318a2d755fe3e9b6904713da9afd

docs/content/3.config.md
```diff
@@ -469,3 +469,9 @@ Rollup specific option.
 ### `commonJS`
 
 Rollup specific option. Specifies additional configuration for the rollup CommonJS plugin.
+
+## Preset options
+
+### `firebase`
+
+The options for the firebase functions v2 preset. See [ Preset](/deploy/providers/firebase#firebase-functions-v2)
```
Hebilicious, line 477: Extra space in Preset

docs/content/2.deploy/providers/firebase.md
```diff
@@ -82,11 +110,81 @@ NITRO_PRESET=firebase yarn build
...
+
+This preset will deploy to [firebase functions V2](https://firebase.google.com/docs/functions/version-comparison) - the 2nd generation of cloud functions.
+
+It works the same as the [`firebase`](#firebase) preset above, but uses the V2 functions runtime.
+
+### Setup
+
+Follow the instructions for the `firebase` preset above for setup and deploy.
+
+**Caution**: New HTTP and HTTP callable functions deployed with any Firebase CLI lower than version 7.7.0 are private by default and throw HTTP 403 errors when invoked. Either explicitly [make these functions public](https://cloud.google.com/functions/docs/securing/managing-access-iam#allowing_unauthenticated_http_function_invocation) or [update your Firebase CLI](https://firebase.google.com/docs/cli#setup_update_cli) before you deploy.
```
Hebilicious, line 134: nit: could add a line about consulting the Cloud Functions console just in case

Hebilicious (commented):
Amazing, this is looking great !

JamieCurnow:
@Hebilicious - I've added a line in the docs about the functions permissions pointing the user to the google cloud console 👍

Hebilicious (approved):
Great work, everything lgtm! @pi0 will take it from there.

sisou (approved):
Works :+1:

src/types/presets.ts
```diff
@@ -65,4 +67,11 @@ export interface PresetOptions {
       maxDuration: number;
     };
   };
+  firebase: {
+    /**
+     * Firebase functions v2 (2nd gen) Configuration
+     * @see https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.https.httpsoptions
+     */
+    httpRequestOptions: HttpsOptions;
```
pi0, line 75: We need to prefix this with version (`v2`)

docs/content/2.deploy/providers/firebase.md
```diff
@@ -20,19 +26,41 @@ This file should then be committed to version control. You can also create a `.f
...
+yarn:
+
+```bash
+yarn global add firebase-tools
+```
+
+npm:
+
+```bash
+npm install -g firebase-tools
```
pi0, line 58: We might instead use `npx firebase-tools@[version]`

src/types/presets.ts
JamieCurnow: @pi0 do you want to go with:

```ts
firebase: {
  v2: {
    httpRequestOptions: HttpsOptions
  }
}
```

OR

```ts
firebase: {
  v2HttpRequestOptions: HttpsOptions
}
```

docs/content/2.deploy/providers/firebase.md
JamieCurnow: Have updated the docs in [Change docs to recommend using npx firebase-tools](https://github.com/unjs/nitro/pull/1142/commits/424fe2ddecc0d37a2485a6b1ea9e2759419c3300)

src/types/presets.ts
JamieCurnow: OR

```
firebaseV2: {
  httpRequestOptions: HttpsOptions
}
```

JamieCurnow: There is a PR for adding another option that will affect both firebase and firebase-v2 presets which has been proposed as:

```
firebase: {
  serverFunctionName: string
}
```

Incase that sways the decision 😉

pi0: Talking with @Hebilicious, i think `v2: {}` as namespace would be safest for longer run 👍🏼  (we might later introduce additional top level shortcuts for common options for better DX)

JamieCurnow: Awesome 👍 I've updated this in [nest httpRequestOptions under the v2 key in the firebase preset options](https://github.com/unjs/nitro/pull/1142/commits/efc5bd0c0107c020388da5598a7b2112f8229cfe)

JamieCurnow:
@pi0 Thanks for the review 🙌 Should all be amended now 👍

docs/content/2.deploy/providers/firebase.md
```diff
@@ -82,11 +104,92 @@ NITRO_PRESET=firebase yarn build
...
+
+This preset will deploy to [firebase functions V2](https://firebase.google.com/docs/functions/version-comparison) - the 2nd generation of cloud functions.
+
+It works the same as the [`firebase`](#firebase) preset above, but uses the V2 functions runtime.
+
+### Setup
+
+Follow the instructions for the `firebase` preset above for setup and deploy.
+
+**Caution**: New HTTP and HTTP callable functions deployed with any Firebase CLI lower than version 7.7.0 are private by default and throw HTTP 403 errors when invoked. Either explicitly [make these functions public](https://cloud.google.com/functions/docs/securing/managing-access-iam#allowing_unauthenticated_http_function_invocation) or [update your Firebase CLI](https://firebase.google.com/docs/cli#setup_update_cli) before you deploy. If you deploy the function and it is not publicly accessible, then you will need to visit the [Functions Console](https://console.cloud.google.com/functions/list) on Google cloud platform, click your function name, select the 'Permissions' tab and follow the instructions on how to make the function public.
```
luc122c, line 137: It might be worth making this warning a bit more noticeable. Example [here](https://nitro.unjs.io/deploy/providers/bun). 

```suggestion
::alert{type="warning"}
New HTTP and HTTP callable functions deployed with any Firebase CLI lower than version 7.7.0 are private by default and throw HTTP 403 errors when invoked. Either explicitly [make these functions public](https://cloud.google.com/functions/docs/securing/managing-access-iam#allowing_unauthenticated_http_function_invocation) or [update your Firebase CLI](https://firebase.google.com/docs/cli#setup_update_cli) before you deploy. If you deploy the function and it is not publicly accessible, then you will need to visit the [Functions Console](https://console.cloud.google.com/functions/list) on Google cloud platform, click your function name, select the 'Permissions' tab and follow the instructions on how to make the function public.
::
```

src/types/presets.ts
```diff
@@ -65,4 +67,13 @@ export interface PresetOptions {
       maxDuration: number;
     };
   };
+  firebase: {
```
posva, line 70: What would be the difference with runtimeConfig? I see other presets using `useRuntimeConfig()` but none using `useAppConfig()`

posva (commented):
Hello,

Disclaimer: I'm the maintainer of VueFire and occasionally work with Firebase.

One of the things we were planning to release with Firebase is a preset for nitro that handles the different generations. I think this is a great addition as it will unblock anybody trying to use the faster version of Cloud functions 😄

However, I don't think it's a good idea to name this `firebase-v2` because it sounds as the successor of the `firebase` preset but if there is a 3rd generation of cloud functions, there will be a `firebase-v3` preset and so on. In practice, **1st and 2nd generation cloud functions can be used on the same project** so I think it's better to name it something else like `firebase-2nd-gen` and release it as a temporary and experimental preset. Even better, release it as a separate package so you have full control over it and unblock yourself and other users.

Because of that, I also think it's better not to document it just like another preset as the goal should be to make the firebase one work in all scenarios.

In the future, I think having the official nitro firebase preset as a separate package will be better as it will allow the preset to have breaking changes such as defaulting to the 2nd generation of cloud functions.

docs/content/2.deploy/providers/firebase.md
```diff
@@ -82,11 +104,94 @@ NITRO_PRESET=firebase yarn build
...
+The `httpRequestOptions` object is passed to `functions.https.onRequest()` function as the first argument as per the [firebase docs](https://firebase.google.com/docs/functions/http-events?gen=2nd#trigger_a_function_with_an_http_request).
+
+For example, to set the region and max instances of the function, you can add the following to your `nitro.config.ts`:
+
+```ts
+import { defineNitroConfig } from 'nitropack/config'
+
+export default defineNitroConfig({
+  // Nitro options...
+  firebase: {
```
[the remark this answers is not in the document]
JamieCurnow: I'll let @pi0 weigh in on this, as we've already been through a review on it. See the resolved issue here @posva :
https://github.com/unjs/nitro/pull/1142#pullrequestreview-1512458039

src/runtime/entries/firebase-v2.ts
```diff
@@ -0,0 +1,10 @@
+import "#internal/nitro/virtual/polyfill";
+import functions from "firebase-functions/v2";
```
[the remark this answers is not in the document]
JamieCurnow: Nice! I didn't realise you could do this 👍 Going to do a quick test locally to make sure it all works/builds and then will commit

src/types/presets.ts
JamieCurnow: @posva I guess you're referencing this line where we add the config to the appConfig. The only reason we do this is to be able to reference it in the entry file here:
https://github.com/unjs/nitro/pull/1142/files#diff-89af85ba779addbc7c61eaecf298b36da9e6d1d5753f25ff55416ef9f50d833eR8

It's all put under a key that isn't intended to be used by the end user. This was implemented on @Hebilicious suggestion here:
https://github.com/unjs/nitro/pull/1142#discussion_r1250797577

We took the options out of runtimeConfig to put them along side other preset config like vercel:
https://github.com/unjs/nitro/pull/1142#issuecomment-1512736670

Nitro options seem like the best place to put these as you are configuring the nitro build at this point rather than runtime options. Then I guess putting them either in app config or runtime config doesn't really matter as it's just a way to get those options into the entry file?

Happy to change it though if it's better practice @Hebilicious @pi0 ?

posva: Thanks! This really helps to follow the flow of the PR!

docs/content/2.deploy/providers/firebase.md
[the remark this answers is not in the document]
AlexLavoie42: Just my 2 cents, but for #1377 my idea was to have the config option shared between v1 and v2 functions, so you could seamlessly switch between them without worrying about the config. As someone using this in production, it would be nice to switch between v1 and v2 without worrying about accidentally overwriting function deployments.

src/runtime/entries/firebase-v2.ts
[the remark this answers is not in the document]
JamieCurnow: That works well 🙌

docs/content/2.deploy/providers/firebase.md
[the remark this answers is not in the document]
posva: Could you detail those circumstances? I don't think I see any reason not to upgrade the currently named `server` function that renders the Nuxt app to gen 2 and then rollback to gen 1 🤔 

Publishing a gen 2 function with the same name will error, you have to name it differently

> The main purpose behind sharing the config would be to simply upgrade to v2 without changing config.

Totally 🙌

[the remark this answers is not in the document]
AlexLavoie42: > Publishing a gen 2 function with the same name will error, you have to name it differently

This is good to know, so in that case you would not be able upgrade to v2 with the same function name unless you delete the original first? Perhaps in that case it should actually be a different config value to avoid an error. Maybe even the default name should be different as well then?

As for downgrading to v1, I am imagining a case where we downgrade temporarily either to debug or rollback in case of any bugs that arise after upgrading. Granted, I don't think this scenario is very likely.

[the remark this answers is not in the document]
posva: IMO the name should be different yes.

> so in that case you would not be able upgrade to v2 with the same function name unless you delete the original first? 

Not exactly: upgrading is a bit different as you want to go through a one-time step where you have both gen functions co-existing. In more complex scenarios: https://firebase.google.com/docs/functions/2nd-gen-upgrade.
You could also just redeploy the second gen with a different name and remove the 1st one in one deployment. I'm not completely sure but I imagine you risk losing a fraction of traffic during that deployment, which in some projects is fine, in others isn't.

src/types/presets.ts
JamieCurnow: @Hebilicious any thoughts on this?

Hebilicious: > @Hebilicious any thoughts on this?

Everything looked good to me the last time I reviewed it, I believe @posva is adding automated tests in nitro deploys and @pi will be in charge of merging.

pi0:
Thanks all for contribution to firebase v2 preset <3 Let's track from one PR #1500 and hopefully soon main branch.

PR #1144, opened 2023-04-15
refactor: rename `build` option to `static`

pi0:
### ❓ Type of change

### 📚 Description

rename `build` option (unreleased) to `static` for better readability of the option purpose.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -25,7 +25,7 @@ export async function prepare(nitro: Nitro) {
   if (!nitro.options.noPublicDir) {
     await prepareDir(nitro.options.output.publicDir);
   }
-  if (nitro.options.build) {
+  if (!nitro.options.static) {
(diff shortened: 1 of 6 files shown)


danielroe:
I still think it's better to use `build` as that more precisely covers what will and won't happen when enabling the option. I think we might possibly have a preset that isn't fully static but doesn't want the normal nitro build to take place.

pi0:
I agree about your perspective way of thinking @danielroe. My main reason for making this change is because nitro normally is supposed to _build_ and this flag is similar to `dev` for a special running mode. We don't build because we are in static mode and i can't think of any other use cases that do not want to build. Also if we _infer_ static mode from `build: false` for other purposes like for `prepare` mode, it can make wrong results. We don't want static behavior adjustments for other modes.

In the end, we might introduce back the `build` flag also in the future if there were common usecase.

PR #1164, opened 2023-04-17
fix: dedup `unstorage` dependency and use subpath alias

pi0:
### ❓ Type of change

### 📚 Description

Similar to other core deps (h3, radix3, defu), unstorage should be deduped otherwise there might be missing chunks in final output (this is not root cause but a common issue and this PR is basically a stable workaround to the issue)

### 📝 Checklist


pi0:
Need to test more

danielroe:
Do you have an example reproduction?

pi0:
> Do you have an example reproduction?

Unofrountaly not a stable one but this has been reported many times, especially within the Nuxt ecosystem.

I am thinking to move it forward since this way of aliasing is better anyway by not pre-applying resolve conditions.

PR #1184, opened 2023-04-23
feat(cache): support cached event handlers with varies

### LINKED ISSUE #1031: Support cached event handlers with varies (custom headers)
Context: https://github.com/unjs/nitro/discussions/1023#discussioncomment-5226630

By default, we drop all incoming request headers from being passed to the cached event handler. This is to ensure cache is not leaked and users get the same response.

However, a valid usecase might be actually intentionally supporting multiple cached responses based on headers (language, variant, multi domain, etc). We might support an opt-in feature `varies: [host, x-domain]` that creates cache segments for each variant and using hashed values to map.

Related/Similar spec to inspire: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary

Botz:
### ❓ Type of change

### 📚 Description

Resolves #1031 
Adds an optional property varies to the cache options.
If varies option is present in the requests headers the value will be considered in key generation.
So if someone sets `varies: [ 'host' ]` example.com and exmaple.de will generate different cache entries

### 📝 Checklist


src/runtime/cache.ts
```diff
@@ -180,12 +181,15 @@ export function defineCachedEventHandler<T = any>(
       if (key) {
         return escapeKey(key);
       }
+      const varyParts =
+        opts.varies?.map((varyHeader) => event.node.req.headers[varyHeader]) ??
```
pi0, line 185: This would be nice if keep them as pair like `['host_foo.com','x-region_asia']` to avoid conflicting same values with different header name.

Botz: @pi0 Good points. Will change it.
Regarding the cache name, can we get problems with to long cache keys?

Botz:
@pi0 any updates on this?

pi0 (approved):

src/runtime/cache.ts
pi0: Good point (used `:{name:hash(value}[]` in last impl)

PR #1192, opened 2023-04-26
feat(vercel): add support for specifying edge regions

danielroe:
### ❓ Type of change

### 📚 Description

This allows specifying the regions for edge functions in Vercel, which can be beneficial to ensure they are running close to a DB (for example).

**More info**: https://vercel.com/docs/concepts/functions/edge-functions#edge-function-regions

### 📝 Checklist

### DIFF
--- a/src/presets/vercel.ts
+++ b/src/presets/vercel.ts
@@ -107,7 +107,7 @@ export const vercelEdge = defineNitroPreset({
       const functionConfig = {
         runtime: "edge",
         entrypoint: "index.mjs",
-        regions: nitro.options.vercel?.regions
+        regions: nitro.options.vercel?.regions,
       };
       await writeFile(
         functionConfigPath,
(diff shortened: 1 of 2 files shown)


ZainW:
Would it possible to also have the opt-in properties described [here](https://vercel.com/docs/concepts/functions/edge-functions#edge-function-regions)? that way we can mix and match as well?

no worries if its out of scope


```ts
export const config = {
  runtime: 'edge', regions: 'iad1'
}
```

leerob (approved):
Nice!

src/presets/vercel.ts
```diff
@@ -104,10 +104,13 @@ export const vercelEdge = defineNitroPreset({
...
         ".vc-config.json"
       );
-      const functionConfig = {
+      const functionConfig: Record<string, any> = {
         runtime: "edge",
         entrypoint: "index.mjs",
       };
+      if (nitro.options.vercel?.regions) {
+        functionConfig.regions = nitro.options.vercel.regions;
+      }
```
emeraldsanto, line 113: **suggestion:** Wouldn't assigning directly be the same?‏

```suggestion
      const functionConfig: Record<string, any> = {
        runtime: "edge",
        entrypoint: "index.mjs",
        regions: nitro.options.vercel?.regions
      };
```

pi0 (approved):

Issue #330, opened 2022-07-07
Bun preset


pi0:
https://bun.sh/

Hebilicious:
Implemented in #1203

PR #1203, opened 2023-05-03
feat: add `bun` preset

Eckhardt-D:
### 🔗 Linked issue

Addresses #330

### ❓ Type of change

### 📚 Description

Adds the Bun runtime as a build preset, including static asset serving / resolution.

### 📝 Checklist


Eckhardt-D:
@danielroe @pi0 Thinking aloud, while this does allow targetting Bun as a build output - there still is one issue in terms of dx - the 'Bun' global. Technically Bun is on a path to implement Node builtins for a 'just works' result - but technically should be able to use e.g. `Bun.file()` instead of `fs.readFile`.

This complicates things a little but my initial thought is a dedicated cli for each runtime executable e.g. `#!/usr/bin/env bun` and do `nitro-bun dev`. This will not work out of the box with rollup, but I see 2 paths:

1. Run the build process in Node still, somehow shimming the Bun global and only execute the handlers / server with Bun with e.g. an adapter
2. Use the Bun bundler instead of rollup in the build phase and run the server in Bun

Perhaps this warrants an issue? Because I do forsee that users would run into the issue of trying to use the Bun global in their handlers and build failing.

pi0:
Hi, dear @Eckhardt-D and thanks again for putting initial efforts on adding bun preset ❤️ 

As a short summary of my updates, i have made changes in order to leverage Bun's node compatibility layer as much as possible instead of unenv mocks. 

In next iterations, we might benchmark `Bun` API vs it's compatibility layer especially for static asset handling which is one of the Bun's biggest perf areas. But for now i decided to keep it as simple as possible and consistent with Node.

---

Follow-up ideas are also nice to try. Particularly being able to run `nitro build` within Bun itself using a dedicated CLI like `nitro-bun` 👍🏼 

I am also up to experiment Bun builder instead of rollup but probably for consistency it will force us having double efforts. (one idea would be converting all nitro plugins to [unplugin](https://github.com/unjs/unplugin) and add bun support. Plenty to experiment!)

PR #1212, opened 2023-05-05
feat: `future.nativeSWR`

pi0:
### 🔗 Linked issue

> Uses built-in SWR functionality (using caching layer and storage) for Netlify and Vercel presets instead of falling back to ISR behavior.

New `future` namespace will be used for potentially breaking but stable changes with an opt-in method until next major nitro versions. This way we can better collaborate with consumer frameworks such as Nuxt.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/presets/netlify.ts
+++ b/src/presets/netlify.ts
@@ -17,7 +17,11 @@ export const netlify = defineNitroPreset({
     },
   },
   hooks: {
-    "rollup:before": (nitro: Nitro) => deprecateSWR(nitro),
+    "rollup:before": (nitro: Nitro) => {
+      if (!nitro.options.future.nativeSWR) {
+        deprecateSWR(nitro);
+      }
+    },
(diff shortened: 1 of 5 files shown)


src/presets/netlify.ts
```diff
@@ -210,7 +214,7 @@ function deprecateSWR(nitro: Nitro) {
   }
   if (hasLegacyOptions) {
     console.warn(
-      "[nitro] Nitro now uses `isr` option to configure ISR behavior on Netlify. Backwards-compatible support for `static` and `swr` support with Builder Functions will be removed in the next major release."
+      "[nitro] Nitro now uses `isr` option to configure ISR behavior on Netlify. Backwards-compatible support for `static` and `swr` support with Builder Functions will be removed in the future versions."
```
danielroe, line 217: Might this change before the next major?

PR #1223, opened 2023-05-09
feat: support concurrency + interval when prerendering

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/15543
https://github.com/nuxt/nuxt/issues/4122

### ❓ Type of change

### 📚 Description

This adds support for `prerender.concurrency` and `prerender.interval`. I expect we might want to iterate on the implementation (it's late!) but basically this keeps the queue full (so it's not batched, which was a problem previously - see https://github.com/nuxt/nuxt/issues/4122).

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -72,6 +72,8 @@ const NitroDefaults: NitroConfig = {
   errorHandler: "#internal/nitro/error",
   routeRules: {},
   prerender: {
+    concurrency: 1,
+    interval: 0,
     crawlLinks: false,
     ignore: [],
     routes: [],
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -85,10 +85,11 @@ export async function prerender(nitro: Nitro) {
 
   // Start prerendering
   const generatedRoutes = new Set();
+  const skippedRoutes = new Set();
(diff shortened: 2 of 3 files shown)


pi0 (approved):
Awesome thanks!

I haven't tested the functionality enough but i guess would be much easier to test behavior on edge against real projects.

Also we might increase default concurrency once ensure it is stable feature wdyt?

pi0:
Sure thing. We could also add a `future.concurrentPrerendering` flag for easy opt-in (both by users and nuxt framework)

jmurphy-dev:
wondering if this could be documented here:
https://nitro.unjs.io/config#prerender

Screenshot 2023-08-22 at 5 41 59 PM

pi0:
@jmurphy-dev Where do you wish the docs to be documented? (it is already published)

jmurphy-dev:
@pi0 Just to make sure I am understanding correctly passing nuxt 3

```
prerender: {
    concurrency: 250,
    interval: 100
}
```

is the same as nuxt2

```
routes: {
    concurrency: 250,
    interval: 100
}
```

Issue #1235, opened 2023-05-14
Cloudflare module and service workers API request un-necessary KV reads


Hebilicious:
### Describe the feature

When requesting with cloudflare module or service, some un-necessary KV reads are being performed. 
I made a PR to check for the public asset URL first.

### Additional information

- [X] Would you be willing to help implement this feature?

Issue #1001, opened 2023-02-27
Cache Control for 404 response


oleghalin:
### Environment

Cloudflare Workers (i think all envs have same logic)

### Reproduction

Try to access localhost:3000/_nuxt/test.css and see Cache Control headers. 

### Describe the bug

I think nitro shouldnt cache 404's errors by default or atleast with different rules. For now we are facing issue when user visited page after deploy and 404 response was cached in browser.

pi0:
Hi. Can you please share nuxt/nitro versions?

oleghalin:
It's latest 2.2.3

oleghalin:
@pi0 here is example, https://test-docus.vercel.app/_nuxt/test.css.


As you see cache-control exists here, while response status is 404
```
age: 3
cache-control: public, max-age=2592000, immutable
content-encoding: br
content-type: text/html;charset=utf-8
date: Wed, 01 Mar 2023 13:11:00 GMT
server: Vercel
```

pi0:
Hi. I'm not sure if the deployment (cannot see source of open it anymore) were depending on cache headers generated by vercel edge or nitro so please feel free to open new issue with reproduction if facing similar issues.

PR #1236, opened 2023-05-14
perf(cloudflare): check for asset url

Hebilicious:
Remove KV access when an API route is called.

### 🔗 Linked issue

Fix https://github.com/unjs/nitro/issues/1001
Resolves https://github.com/unjs/nitro/issues/1235

### ❓ Type of change

### 📝 Checklist


Hebilicious:
I'm not 100% sure this works yet, and it doesn't look like the test case are very useful there. Extending them.

This was useful to debug :
```ts
  const list = await env.__STATIC_CONTENT.list();
  console.log("GET ASSET FROM KV", list.keys.map((k) => k.name), { publicAssetBases });
```

Should we add e2e tests ? Especially Nuxt ones. IE deploy nuxt project with nitro + wrangler and hit the endpoints.

pi0:
Can you please explain more what routes are not being matched?

Hebilicious:
Sure ! For the Cloudflare Module for example : 

Running the tests with isInPublicAssets
image

Running the tests with isInKv
image

pi0:
Good point about also checking `.html` suffixes. Since we are checking for public asset existence in other presets, it would be much better if we could merge `isInKV` functionality into `isPublicAssetURL`. Otherwise seems really good improvements 💯

Hebilicious:
> Good point about also checking `.html` suffixes. Since we are checking for public asset existence in other presets, it would be much better if we could merge `isInKV` functionality into `isPublicAssetURL`. Otherwise seems really good improvements 💯

This was done in the later commits. Reminder to review again 🙏🏽

pi0:
Thanks for changes. As an update, i still need to revisit this issue while it is important fix and i value your efforts, i am still not convinced why we are introducing a new util like this so need to check it more in depth (i might be wrong as well!)

Hebilicious:
> Thanks for changes. As an update, i still need to revisit this issue while it is important fix and i value your efforts, i am still not convinced why we are introducing a new util like this so need to check it more in depth (i might be wrong as well!)

Now that look more at it, this PR is more of a fix than a perf, as currently there's a bug where some matched assets aren't returned by KV, and some non existing assets are being accessed.

zsilbi:
This would be a very nice improvement!
I found this as I was investigating why my noop endpoint latency is so unpredictable. (bottom)

kv_asset

pi0:
With [workers static assets](https://developers.cloudflare.com/workers/static-assets/) we won't need to depend on binding anymore.

PR #1241, opened 2023-05-16
feat: add ability to use nested routers from ServerDir files

Morgbn:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/18571

### ❓ Type of change

### 📚 Description

These changes allow routers to be added to the h3 app using the same file-based routing system present in nitro.
Typically, this makes it possible to have one router per file.
For example, a todo file with GET /, POST /, PUT /:id, DELETE /:id ...
```js
// api/{todo}.ts
const db = ['todo 1', 'todo 2']
export default createRouter()
  .get('/', eventHandler(() => db))
  .post('/', eventHandler(async (event) => {
    const { todo } = await readBody(event)
    db.push(todo)
    return todo
  }))
  .put('/:id', …)
```

In this modification, I use the syntax _{route-name}_.ts to determine that it's a router that should be added (not a function handler). Does this make sense to you? I was hesitating with _route-name_.use.ts or _route-name_.router.ts, but the dot is used to specify HTTP request method, so it might be a bit confusing.

Resolves nuxt/nuxt#18571

### 📝 Checklist

### DIFF
--- a/docs/content/1.guide/3.routing.md
+++ b/docs/content/1.guide/3.routing.md
@@ -99,6 +99,20 @@ Check out [h3 JSDocs](https://www.jsdocs.io/package/h3#package-index-functions) 
 export default eventHandler(event => `Default page`)
 ```
 
+### Using a Nested Router
+
+```js
+// api/{todo}.ts
+const db = ['todo 1', 'todo 2']
+export default createRouter()
+  .get('/', eventHandler(() => db))
+  .post('/', eventHandler(async (event) => {
+    const { todo } = await readBody(event)
+    db.push(todo)
+    return todo
+  }))
+```
+
 ## Route Rules
 
 Nitro allows you to add logic at the top-level of your configuration, useful for redirecting, proxying, caching and adding headers to routes.
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -83,6 +83,8 @@ function createNitroApp(): NitroApp {
         "/"
       );
       h3App.use(middlewareBase, handler);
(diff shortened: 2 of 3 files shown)


pi0:
Hi, and thanks for the PR! I think it would be amazing that we support multiple HTTP methods per route for simplicity. But probably using a full router is adding too much overhead. Also it makes it harder to apply route rules, generated nested types, etc.

I am working on a new format for h3 event handlers with object syntax let's see if we can introduce a simpler and more lightweight alternative approach.

Morgbn:
Yes, that would be very convenient!
The main goal was to allow the use of a nested router in nuxt,
but I understand about the rules/types, can't wait to see this new format!

In the meantime I'll look for another way to fix https://github.com/nuxt/nuxt/issues/18571

mubaidr:
So is this feature supported or deprecated? I can't find it in documentation.

Hebilicious:
> So is this feature supported or deprecated? I can't find it in documentation.

This feature is not implemented. 

I implement something interesting in [server-block-nuxt](https://github.com/Hebilicious/server-block-nuxt#-usage), which gives you a syntax to define multiple request handlers in a route : 


```ts
const message = "Hello World!!!"
const bye = "bye!"
export const GET = defineEventHandler(() =>({ message }))
export const POST = defineEventHandler(() =>({ message: bye }))
```

We could implement something like this in Nitro, @pi0 what do you think ?

mubaidr:
Currently, defining routes per request per file is very tedious. 

Given that we have typed `fetch`, this I believe, will greatly enhance full stack code management.

PR #1273, opened 2023-05-31
feat: add handler file path

arashsheyda:
### ❓ Type of change

### 📚 Description

it would add handler path to `#internal/nitro/virtual/server-handlers` handlers so we can access the paths easily.

### 📝 Checklist


pi0:
I think this is a nice idea but probably we only need this for development. Adding to production exposes build environment to the bundle and is not safe.

arashsheyda:
@pi0 Oh, yes I didn't think about that thanks👌

PR #1274, opened 2023-05-31
fix(aws-lambda,netlify): base64 encode binary responses 

harlan-zw:
### 🔗 Linked issue

See https://nuxt-og-image-playground-netlify.netlify.app/api/resvg.png

<details>
<summary>
Code
</summary>

```ts
import { defineEventHandler, setHeader } from 'h3'
import { Resvg } from '@resvg/resvg-js'

export default defineEventHandler(async (e) => {
    const resvgJS = new Resvg('<svg xmlns="http://www.w3.org/2000/svg" version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:svgjs="http://svgjs.dev/svgjs" viewBox="0 0 700 700" width="700" height="700"><defs><linearGradient gradientTransform="rotate(150, 0.5, 0.5)" x1="50%" y1="0%" x2="50%" y2="100%" id="ffflux-gradient"><stop stop-color="hsl(315, 100%, 72%)" stop-opacity="1" offset="0%"></stop><stop stop-color="hsl(227, 100%, 50%)" stop-opacity="1" offset="100%"></stop></linearGradient><filter id="ffflux-filter" x="-20%" y="-20%" width="140%" height="140%" filterUnits="objectBoundingBox" primitiveUnits="userSpaceOnUse" color-interpolation-filters="sRGB">\n' +
        '  <feTurbulence type="fractalNoise" baseFrequency="0.005 0.003" numOctaves="2" seed="2" stitchTiles="stitch" x="0%" y="0%" width="100%" height="100%" result="turbulence"></feTurbulence>\n' +
        '  <feGaussianBlur stdDeviation="20 0" x="0%" y="0%" width="100%" height="100%" in="turbulence" edgeMode="duplicate" result="blur"></feGaussianBlur>\n' +
        '  <feBlend mode="color-dodge" x="0%" y="0%" width="100%" height="100%" in="SourceGraphic" in2="blur" result="blend"></feBlend>\n' +
        '  \n' +
        '</filter></defs><rect width="700" height="700" fill="url(#ffflux-gradient)" filter="url(#ffflux-filter)"></rect></svg>', {
        logLevel: 'trace',
        fitTo: {
            mode: 'width',
            value: 1200,
        },
    })
    setHeader(e, 'Content-Type', 'image/png')
    return resvgJS.render().asPng()
})
```

</details>

### ❓ Type of change

### 📚 Description

Netlify uses AWS Lambda. AWS Lambda requires sending buffers as strings (they are re-encoded via a proxy afaik). 

Nitro currently converts all buffers to default `toString()`. This breaks images, as they need to be sent with base64 encoding (see https://aws.amazon.com/blogs/compute/handling-binary-data-using-amazon-api-gateway-http-apis/), with a special flag `isBase64Encoded`.

The fix is to check if we're working with a buffer and the response is an image, then use base 64 encoding. This may affect other response types though.

### 📝 Checklist


src/runtime/entries/netlify-lambda.ts
```diff
@@ -30,9 +31,20 @@ export async function lambda(
     body: event.body, // TODO: handle event.isBase64Encoded
   });
 
+  const headers = normalizeOutgoingHeaders(r.headers);
+  // image buffers must be base64 encoded
+  if (Buffer.isBuffer(r.body) && headers["content-type"].startsWith("image/")) {
```
harlan-zw, line 36: this header check is hacky, looking for a better alternative

src/runtime/entries/netlify-lambda.ts
```diff
@@ -1,4 +1,5 @@
 import "#internal/nitro/virtual/polyfill";
+import { Buffer } from "node:buffer";
```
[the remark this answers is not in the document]
pi0: You can directly depend on global Buffer without importing it..

src/runtime/entries/netlify-lambda.ts
Hebilicious: If we trust mime.getType to return the correct type, this should be fine for images, as all images types should start with image. 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types.
However we should do this for all [Binary media](https://docs.aws.amazon.com/apigateway/latest/developerguide/lambda-proxy-binary-media.html)
As always with AWS, the list of supported media types isn't documented anywhere, but chatGPT gives me this:

image/png
image/jpeg
image/gif
application/octet-stream
application/pdf
application/x-zip-compressed
audio/mpeg
audio/mp4
video/mp4
application/zip
audio/webm
video/webm

Which sorts of make sense... But I think a cleaner way to handle all possible use cases without relying on the header would be to use https://www.npmjs.com/package/isbinaryfile

Hebilicious (approved):
If we're just interested in image support this is fine imo.

Hebilicious (commented):
Now that I review this again, I think this needs to be slightly refactored to work with cookies too, ie extract the logic into a `normalizeBody` method. 
It also needs to be applied to stormkit which uses lambda.

Hebilicious (commented):
LGTM

src/runtime/utils.lambda.ts
```diff
@@ -1,3 +1,4 @@
+import { isUtf8, Buffer } from "node:buffer";
```
pi0, line 1: `isUTF8` is quite recent. We might need to conditionally use it  https://nodejs.org/api/buffer.html#bufferisutf8input

harlan-zw: The current issue is that the JSON is a buffer as well, so without the utf8 check the JSON response is base64 encoded.

So we need something that works prior to node 18. 

I've implemented a hacky (?) fallback solution but would it make sense to add a polyfill to unenv for it using https://github.com/hcodes/isutf8/blob/master/src/index.ts make sense?

harlan-zw: Or do we just use the package https://www.npmjs.com/package/isbinaryfile

pi0: outgoing body wouldn't be JSON. h3 already handles that. I think we can remove the checks altogether. (body is either string or buffer at that stage) Can you please double check?

harlan-zw: Without the checks, 4 of the tests fail with this type of response

```
 FAIL  test/presets/aws-lambda.test.ts > nitro:preset:aws-lambda > handles errors
AssertionError: expected 'SGV5IEFQSQ==' to contain 'Hey API'
 ❯ test/tests.ts:249:29
    247|     expect(status).toBe(503);
    248|     const { data: heyData } = await callHandler({ url: "/api/hey" });
    249|     expect(heyData).to.have.string("Hey API");
       |                             ^
    250|   });
```

When inspecting the output using the following code

```ts
if (Buffer.isBuffer(body)) {
    console.log(typeof body, body.toString())
 // ...
```

The following is shown

```
object {"message":"Hello API"}
object Hey API
object Works!
```

Not sure if this is a testing artifact? As I would have thought JSON would be returned as a string as well.

pi0: Yeat it seems localCall already wraps to Buffer. Thinking to somehow improve that area to reduce workloads here.

pi0: (still working on it between meetings) Exciting finding after digging some holes! The two `/api/hey` and `/api/hello` fail because, during tests, we prerender those two, and then when trying to fetch, serve static middleware handles the response and sends the response as a buffer instead!) i am thinking of some root cause fixes to follow up.

pi0: I think safest would be to check about known text types for Buffer edge case. Later we might also improve serve static to avoid buffer conversion for text types.

PR #1302, opened 2023-06-12
Revert "types: support type generic for `useRuntimeConfig`"

Hebilicious:
Reverts unjs/nitro#1284

This breaks the CI and and the build command.


pi0:
Thanks for the quick PR. I kinda missed the CI from the previous PR. Types fixing in #1311. 

Allowing on-demand type overrides might be useful for some cases.

PR #1312, opened 2023-06-15
feat(cloudflare-pages): add config to customize generated `_routes.json`

### LINKED ISSUE #1291: Cloudflare Pages: Don’t overwrite _routes.json if present in the public folder
### Describe the feature

Auto-generating a `_routes.json` file for Cloudflare Pages is great, but there are some cases where it makes sense to just manually write a _routes.json
For example, if you have more than 100 static routes, or if an API path that is available during pre-rendering isn’t available on the server (for example, IPX/nuxt image)
Before #965 I could just put a `_routes.json` in my public folder and it would be copied and deployed, but after #965 it would overwrite my manual `_routes.json` with the auto-generated one. So, it would be nice if Nitro didn’t overwrite manual `_routes.json` files.
I’ve prepared a fix that passes all the tests and I can open a PR if this sounds good.

### Additional information

AWBroch:
### 🔗 Linked issue

Resolves #1291 

### ❓ Type of change

### 📚 Description

Resolves #1291, adds test for it and updates documentation. (Is the documentation change necessary, or should I take it out?)

### 📝 Checklist

### DIFF
--- a/test/fixture/nitro.config.ts
+++ b/test/fixture/nitro.config.ts
@@ -87,7 +87,7 @@ export default defineNitroConfig({
   cloudflare: {
     pages: {
       routes: {
-        include: ["/api/*", "/blog/*"],
+        include: ["/*", "/api/*", "/blog/*"],
         exclude: ["/blog/static/*"],
       },
     },


AWBroch:
Personally, I think having the override property in settings (`cloudflarePages: { routes: { include: [], exclude: [] } }`) makes sense. I could change the PR to do this, or would a discussion be better?

pi0:
Sure feel free to directly make changes in this PR.

AWBroch:
Alright, thanks, I think I it’s ready for a review

src/types/presets.ts
```diff
@@ -65,4 +67,23 @@ export interface PresetOptions {
       maxDuration: number;
     };
   };
+  cloudflarePages: {
```
Hebilicious, line 70: I would prefer if we used `cloudflare.pages.routes` instead of `cloudflarePages.routes`

Hebilicious (commented):
@pi0 Something similar is happening in the Azure PR #1344 , but this PR use a different approach.

- Azure PR merge the default nitro config with the user provided one, and gives priority to the user one
- This PR replaces the entire generated nitro config with user provided one.

I would go ahead even though this approach is different and try to collect user feedback to see which approach makes the most sense. To be extremely thorough we should offer both merge and replace, that being said I'm not sure if there would be users for both features.

NicholasDawson:
> @pi0 Something similar is happening in the Azure PR #1344 , but this PR use a different approach.
> 
> * Azure PR merge the default nitro config with the user provided one, and gives priority to the user one
> * This PR replaces the entire generated nitro config with user provided one.
> 
> I would go ahead even though this approach is different and try to collect user feedback to see which approach makes the most sense. To be extremely thorough we should offer both merge and replace, that being said I'm not sure if there would be users for both features.

Just to provide some input into this discussion since I implemented the azure version of this feature. The advantage of merging is that route entries are auto generated and updated on every build so the developer doesn't have to manually change these entries when they add new routes that are static.

With the case of merging, if the developer wants to overwrite or change a route they can add their own entry which will trump the autogenerated entry.

The only situation where this leaves something to be desired is if the developer wants to remove routes from the config that are autogenerated by nitro, but I can't see this happening from my experience.

AWBroch:
I could also add an `override` or `merge` option to the`cloudflare.pages` config, it wouldn’t be too complicated to have both. Personally, I’d rather have `override` as the default and it can be turned off with either `override: false` or `merge: true`. Thoughts?

AWBroch:
Alright, it’s ready for review

src/presets/cloudflare-pages.ts
```diff
@@ -56,56 +56,75 @@ export const cloudflarePagesStatic = defineNitroPreset({
...
     exclude: [],
   };
 
-  // Exclude public assets from hitting the worker
-  const explicitPublicAssets = nitro.options.publicAssets.filter(
-    (i) => !i.fallthrough
-  );
+  if (
+    nitro.options.cloudflare?.pages?.routes == null ||
+    nitro.options.cloudflare.pages.routes!.merge
```
Hebilicious, line 78:
```suggestion
    nitro.options.cloudflare?.pages?.routes?.merge
```

src/presets/cloudflare-pages.ts
```diff
@@ -56,56 +56,75 @@ export const cloudflarePagesStatic = defineNitroPreset({
...
+  ) {
+    routes.exclude.push(
+      ...(nitro.options.cloudflare?.pages?.routes?.exclude ?? [])
+    );
+    routes.include.push(
+      ...(nitro.options.cloudflare?.pages?.routes?.exclude ?? [])
```
Hebilicious, line 84:
```suggestion
      ...(nitro.options.cloudflare?.pages?.routes?.include ?? [])
```

Issue #1314, opened 2023-06-16
Enable Node by default with Cloudflare Pages preset


Smef:
### Describe the feature

When deploying on Cloudflare Pages the `cloudflare-pages` preset has `node: false` as a configuration option, which is incorrect. CFP v1 uses Node 12 by default, with Node 18 as the default in the new v2 beta. You can also specify any specific Node version you want for your deployment, which is what you'd need to do to deploy on CFP. (https://developers.cloudflare.com/pages/platform/language-support-and-tools/)

I think that the default for a CFP preset should be `node: true`.

PR: #1315 

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
We do use Node.js for building nitro apps. More context https://github.com/unjs/nitro/pull/1315#issuecomment-1596087858

Issue #1171, opened 2023-04-19
Enable cloudflare Node.js compatibility layer


pi0:
Cloudflare supports a subset of [Node.js compatible imports](https://blog.cloudflare.com/workers-node-js-asynclocalstorage/). 

Unenv, has a good coverage for most of the Node.js polyfills and shims. But by replacing the supported subset we can:
- Reduce bundle size as runtime provides them
- Leverage `node:internal` native features are impossible to add into userland such as async_hooks

There are two concerns:
- Cloudflare shims are not as complete as unenv ones
- We need to make sure `wrangler.toml` has enabled compatibility flag and otherwise fail or warn the build to ask users doing this based on usage.

Hebilicious:
We need to add a cloudflare preset to `unenv` first.

pi0:
Related https://github.com/unjs/unenv/pull/126

luc122c:
Hopefully this shoud resolve https://github.com/cloudflare/workers-sdk/issues/2081

Hebilicious:
> Hopefully this shoud resolve [cloudflare/workers-sdk#2081](https://github.com/cloudflare/workers-sdk/issues/2081)
I can't open your repo link https://github.com/cloudflare/workers-sdk/issues/2081#issuecomment-1328367369.
I can see we don't have a tracking repo on our side though, ideally try to pinpoint if its a Nuxt/Nitro/Nuxt-security issue and open an issue there 🙏🏽

andrevandal:
great improvement!

MickL:
Probably there are more important todos atm, but I just found this issue and wanted to say that this enhancement would be AMAZING :)

Hebilicious:
@pi0 Can this be closed now that cloudflare uses unenv https://github.com/cloudflare/workers-sdk/tree/main/packages/unenv-preset ? Or is there something in Nitro that needs to be updated ?

pi0:
we don't use that preset but yeah can be closed. See [2.11](https://github.com/nitrojs/nitro/releases/tag/v2.11.0)

```ts
// Nitro
export default defineNitroConfig({
  compatibilityDate: "2025-03-01",
  cloudflare: { nodeCompat: true, deployConfig: true }
});

// Nuxt
export default defineNuxtConfig({
  compatibilityDate: "2025-03-01",
  nitro: { cloudflare: { nodeCompat: true, deployConfig: true } }
});
```

PR #1315, opened 2023-06-16
feat: set `node: true` on cloudflare-pages preset

Smef:
### 🔗 Linked issue

#1314 

### ❓ Type of change

### 📚 Description

Cloudflare pages supports node by default, but the `cloudflare-pages.ts` preset has this option set to `false` as it is inherited from the base preset. This PR adds an explicit `node: true` to the cloudflare-pages preset.

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare-pages.ts
+++ b/src/presets/cloudflare-pages.ts
@@ -22,6 +22,7 @@ export const cloudflarePages = defineNitroPreset({
     // https://github.com/unjs/nitro/pull/933
     _mime: "mime/index.js",
   },
+  node: true,
   rollupConfig: {
     output: {
       entryFileNames: "_worker.js",


Hebilicious:
@Smef I am a little bit confused by this propose change. 

You're saying in the issue that CFP supports node, but this is for the build, not for the runtime. The runtime isn't node, it's [workerd](https://github.com/cloudflare/workerd), which has a nodejs_compat optional [flag](https://developers.cloudflare.com/workers/platform/nodejs-compatibility/#enable-nodejs-from-the-cloudflare-dashboard), but it's not the same as being a node environment.

Smef:
I’m talking about the Cloudflare Pages build environment, which is different than what you linked to. https://developers.cloudflare.com/pages/platform/language-support-and-tools/

It looks like your link was to Cloudflare Workers info, which isn’t the same thing. It’s quite possible that this PR is for the wrong preset, but the issue this is for is related to the CFP build environment which is full Node. Am I misunderstanding this preset?

Hebilicious:
> I’m talking about the Cloudflare Pages build environment, which is different than what you linked to. [developers.cloudflare.com/pages/platform/language-support-and-tools](https://developers.cloudflare.com/pages/platform/language-support-and-tools/)
> 
> It looks like your link was to Cloudflare Workers info, which isn’t the same thing. It’s quite possible that this PR is for the wrong preset, but the issue this is for is related to the CFP build environment which is full Node. Am I misunderstanding this preset?

You are mis-understanding Cloudflare. Both workers and pages are built on the same technology, workerd, which is a Javascript runtime that is different from node.js. Cloudflare Pages CI (ie the thing that hooks to a github repo and automatically rebuilds your website) uses Node for your builds, but *not* the runtime.

Smef:
Maybe you can point me in the right direction if this isn't it, or if this isn't the correct preset configuration option, but the CFB build environment is what I'm talking about, and it runs actual Node. The issue is that currently when running in the Cloudflare Pages build environment, either v1 or the v2 beta (as described [here](https://developers.cloudflare.com/pages/platform/language-support-and-tools/)) and deploying a static site, Nitro incorrectly things that Node isn't available. 

The build environment and build process give messages like 
```
02:04:51.263 | Attempting node version '18' from .nvmrc
-- | --
02:04:52.312 | v18.16.0 is already installed.
02:04:53.762 | Now using node v18.16.0 (npm v9.5.1)
```

It runs the `npx` commands here as well as part of the one-time build process. In this environment it appears to be Node, and not workerd, which matches the documentation for the platform.

Is the issue here that this preset configuration is not for the CFP build environment, but is for the runtime environment, and this PR is just for the wrong file? Again, the environment I'm trying to address here is definitely running Node and only during a single-run build step, not any sort of deployed runtime environment, and Nitro is guessing the node status incorrectly (see [this issue for Nuxt/Image and IPX](https://github.com/nuxt/image/issues/866#issuecomment-1588284530) as an example.)

Hebilicious:
> Maybe you can point me in the right direction if this isn't it, or if this isn't the correct preset configuration option, but the CFB build environment is what I'm talking about, and it runs actual Node. The issue is that currently when running in the Cloudflare Pages build environment, either v1 or the v2 beta (as described [here](https://developers.cloudflare.com/pages/platform/language-support-and-tools/)) and deploying a static site, Nitro incorrectly things that Node isn't available.
> 
> The build environment and build process give messages like
> 
> ```
> 02:04:51.263 | Attempting node version '18' from .nvmrc
> -- | --
> 02:04:52.312 | v18.16.0 is already installed.
> 02:04:53.762 | Now using node v18.16.0 (npm v9.5.1)
> ```
> 
> It runs the `npx` commands here as well as part of the one-time build process. In this environment it appears to be Node, and not workerd, which matches the documentation for the platform.
> 
> Is the issue here that this preset configuration is not for the CFP build environment, but is for the runtime environment, and this PR is just for the wrong file? Again, the environment I'm trying to address here is definitely running Node and only during a single-run build step, not any sort of deployed runtime environment, and Nitro is guessing the node status incorrectly (see [this issue for Nuxt/Image and IPX](https://github.com/nuxt/image/issues/866#issuecomment-1588284530) as an example.)

My apologies, now that I see the Image issue I see your problem. 

Looking into it, this node: true flag is for the rollup configuration, and it looks like it does affect the final bundle. So I'm not sure this is a sane default for all builds... 

Given that Nitro can only be built with Node, this issue shouldn't happen in the first place though.

Smef:
As a test of this, running `node -v` `npm -v` and `which node` commands in the build environment gives the following results:

```
17:02:29.619	v18.16.0
17:02:30.023	9.7.1
17:02:30.037	/opt/buildhome/.nvm/versions/node/v18.16.0/bin/node
```
Nitro definitely seems to be detecting (or assuming?) this incorrectly, and reading `nitro.options.node` returns false in this environment, which doesn't seem to be correct.

Smef:
Ah, so is your thought here that this just isn't the right fix for this issue, and the problem needs to be addressed elsewhere?

Hebilicious:
> Ah, so is your thought here that this just isn't the right fix for this issue, and the problem needs to be addressed elsewhere?

Yes, this could cause a bunch of others issues as doing this disable a lot of the unenv mocks. However I'm very often getting issues with this part of the code anyways, see here https://github.com/nuxt/nuxt/issues/21619

pi0:
Hi. Thanks for the PR. Following up conversation, @Hebilicious is correct. Nuxt nitro always uses Node.js for building phase. The `node: true` flag toggles unenv polyfill levels which are needed to add compatibility layer to cloudflare workers/pages which both are _not_ a Node.js environment.

A related matter, cloudflare workers, has some native polyfills for _some_ node.js imports which we need to enable once the _compatibility flag_ is enabled by users (which is not by default). https://github.com/unjs/nitro/issues/1171

PR #1329, opened 2023-06-21
fix: generate types for #imports

danielroe:
### ❓ Type of change

### 📚 Description

With new support for importing from `#imports`, we want to ensure we have type support for it.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -97,9 +97,11 @@ export async function writeTypes(nitro: Nitro) {
   }
 
   let autoImportedTypes: string[] = [];
+  let autoImportExports: string
 
   if (nitro.unimport) {
     await nitro.unimport.init();
+    autoImportExports = await nitro.unimport.toExports(typesDir).then(r => r.replace(/#internal\/nitro/g, runtimeDir))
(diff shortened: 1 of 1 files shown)


src/build.ts
```diff
@@ -97,9 +97,13 @@ export async function writeTypes(nitro: Nitro) {
   }
 
   let autoImportedTypes: string[] = [];
+  let autoImportExports: string;
 
   if (nitro.unimport) {
     await nitro.unimport.init();
+    autoImportExports = await nitro.unimport
+      .toExports(typesDir)
+      .then((r) => r.replace(/#internal\/nitro/g, runtimeDir));
```
danielroe, line 106: There must be a better solution for this 🤔

pi0, line 106: Can't we keep alias map?

pi0: Haha just added same comment. What is issue?

danielroe: The issue is that `#internal/nitro` type isn't exposed to users by default - the types of all of these imports will be `any` unless we resolve it

PR #1333, opened 2023-06-21
types: avoid overriding app runtime config namespace

danielroe:
### ❓ Type of change

### 📚 Description

Currently it's not possible to augment `runtimeConfig.app` whilst also using Nitro types for runtime config.

This PR makes it possible to augment `app`. It also exposes the `nitro` type. (I can also make augmentation disallowed for this if that is desirable.)

### 📝 Checklist

### DIFF
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -31,7 +31,20 @@ export type NitroDynamicConfig = Pick<
   "runtimeConfig" | "routeRules"
 >;
 
+export interface NitroRuntimeConfigApp {
+  baseURL: string;
+  [key: string]: unknown;
+}
+
+export interface NitroRuntimeConfigNitro {
+  routeRules?: {
+    [path: string]: NitroRouteConfig;
+  }
+}
+
 export interface NitroRuntimeConfig {
+  app: NitroRuntimeConfigApp;
+  nitro: NitroRuntimeConfigNitro;
   [key: string]: unknown;
 }
 
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -39,7 +39,7 @@ export interface NitroRuntimeConfigApp {
 export interface NitroRuntimeConfigNitro {
   routeRules?: {
     [path: string]: NitroRouteConfig;
-  }
+  };
 }
 
 export interface NitroRuntimeConfig {
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -43,7 +43,7 @@ export interface NitroRuntimeConfig {
       [path: string]: NitroRouteConfig;
     };
   };
-  [key: string]: unknown;
+  [key: string]: any;
 }
 
 export interface Nitro {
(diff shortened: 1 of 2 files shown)


src/types/nitro.ts
```diff
@@ -31,11 +31,21 @@ export type NitroDynamicConfig = Pick<
...
 
-export interface NitroRuntimeConfig {
-  app: {
-    baseURL: string;
+export interface NitroRuntimeConfigApp {
+  baseURL: string;
+  [key: string]: unknown;
+}
+
+export interface NitroRuntimeConfigNitro {
```
pi0, line 39: Do you think we need to augment this too?

src/types/nitro.ts
```diff
@@ -31,11 +31,21 @@ export type NitroDynamicConfig = Pick<
   "runtimeConfig" | "routeRules"
 >;
 
-export interface NitroRuntimeConfig {
-  app: {
-    baseURL: string;
+export interface NitroRuntimeConfigApp {
```
pi0, line 34: Maybe NitroAppRuntimeConfig? (I am fine with current one too if you prefer)

danielroe: Happy either way!

src/types/nitro.ts
danielroe: No, I think it is just good for internal type safety. We could even type it in a way that we _forbid_ augmenting it if you want, but leave possibility open for ourselves later.

pi0: Yeah for now let’s keep it inlined I don’t see a reason to externally extend nitro namespace

src/types/nitro.ts
```diff
@@ -31,11 +31,19 @@ export type NitroDynamicConfig = Pick<
   "runtimeConfig" | "routeRules"
 >;
 
+export interface NitroRuntimeConfigApp {
+  baseURL: string;
+  [key: string]: unknown;
```
pi0, line 36: Untyped keys were any since before 2.5. Changing any to unknown (at least for me) is a breaking change. When users where depending on any type it will basically break their type checks even if safer or better. Can we override it for nuxt?

pi0: Seems Nuxt itself is still any 

image

image

pi0, line 36:
```suggestion
  [key: string]: any;
```

src/types/nitro.ts
```diff
@@ -31,11 +31,19 @@ export type NitroDynamicConfig = Pick<
...
-  app: {
-    baseURL: string;
+  app: NitroRuntimeConfigApp;
+  nitro: {
+    routeRules?: {
+      [path: string]: NitroRouteConfig;
+    };
   };
-  [key: string]: any;
+  [key: string]: unknown;
```
pi0, line 46:
```suggestion
  [key: string]: any;
```

src/types/nitro.ts
pi0: I have made change to make PR mergeable. Would be happy to discuss about next steps plan of adopting unknown...

pi0 (approved):

Issue #1087, opened 2023-03-24
Azure preset: Make staticwebapp.config.json optional


jegardiner:
### Describe the feature

The Azure preset currently regenerates the staticwebapp.config.json file on build. Overwriting any additional configuration that you may have added to this file, such as route restrictions using Azure's built-in authentication.

[You can see the code in question here.](https://github.com/unjs/nitro/blob/main/src/presets/azure.ts#L134-L136)

For example, here is a staticwebapp.config.json _before_ `npm run build`:

```
{
  "platform": {
    "apiRuntime": "node:16"
  },
  "routes": [
    {
      "route": "/index.html",
      "redirect": "/"
    },
    {
      "route": "/",
      "rewrite": "/api/server"
    },
    {
      "route": "/admin",
      "allowedRoles": ["authenticated"]
    },
    {
      "route": "/logout",
      "redirect": "/.auth/logout"
    },
  ],
  "responseOverrides": {
    "401": {
      "statusCode": 302,
      "redirect": "/.auth/login/aad"
    }
  },
  "navigationFallback": {
    "rewrite": "/api/server"
  }
}
```

And the same file _after_ `npm run build`:

```
{
  "platform": {
    "apiRuntime": "node:16"
  },
  "routes": [
    {
      "route": "/index.html",
      "redirect": "/"
    },
    {
      "route": "/",
      "rewrite": "/api/server"
    }
  ],
  "navigationFallback": {
    "rewrite": "/api/server"
  }
}
```

Unless there is already a way to add this supplemental configuration so that Nitro will include it, would it be possible to add a configuration option to toggle whether the staticwebapp.config.json should be generated during build.

Until then, I'll be using a hook to replace the file after the build completes.

### Additional information

- [X] Would you be willing to help implement this feature?

tobiasdiez:
Related: https://github.com/unjs/nitro/discussions/420

danielroe:
I agree. We currently merge Netlify presets and I think this would be doable here too...

tobiasdiez:
[...]

PR #1344, opened 2023-06-25
feat(azure): support custom configuration

NicholasDawson:
### 🔗 Linked issue

#1087 

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)
- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

I, along with many others deploying a Nitro server to Azure Static Web Apps, need to customize the `staticwebapp.config.json` file beyond what Nitro generates. Currently, Nitro overwrites the entire file on every compilation. It would be better if Nitro extended an existing configuration if present not overwrite the whole file.

I added thorough documentation to explain how the system will now load and utilize the configuration. Please reference that to learn a little more about my implementation.

#### Implementation Notes
I put a lot of thought into this implementation as there wasn't necessarily a right or wrong way to handle this. I didn't originally want to have a separate file `custom.staticwebapp.config.json`, but the reason I had to add it was that if Nitro and the user are writing to the same file there is no way to keep track of what needs to take priority or be overwritten. Especially for routes, if Nitro generated the config for pre-rendered routes, but then the user deletes that pre-rendered route, it would be expected that the config removes that next time it compiles. If I tried to merge changes into the `staticwebapp.config.json` then Nitro would essentially lose the ability to remove lines from the configuration, only add them.

Therefore it was necessary to introduce the convention of having the user add their own configuration to a file I called `custom.staticwebapp.config.json` to provide a clear distinction between what the user wants to add and what the final generated result (`staticwebapp.config.json`) is.

### 📝 Checklist


NicholasDawson:
Can I please have an update on this PR? It has been 3 weeks, I tried really hard to add sufficient testing and documentation to make this PR easy to review and accept. I know you all must be very busy and I really appreciate your work, I would just appreciate an update on this PR.

Hebilicious:
> @Hebilicious Can I please have an update on this PR? It has been 3 weeks, I tried really hard to add sufficient testing and documentation to make this PR easy to review and accept. I know you all must be very busy and I really appreciate your work, I would just appreciate an update on this PR.

Hi @NicholasDawson , thank you for your hard work on this PR. I think these are all great changes that are well documented, and they will make the life of azure users easier.
@pi0 will need to review it and approve it before it gets merged.

docs/content/2.deploy/providers/azure.md
```diff
@@ -25,6 +25,24 @@ NITRO_PRESET=azure yarn build
...
+
+Nitro will automatically add the following properties based on the following criteria:
+| Property | Criteria | Default |
+| --- | --- | --- |
+| **[platform.apiRuntime](https://learn.microsoft.com/en-us/azure/static-web-apps/configuration#platform)** | Will automatically set to `node:16` or `node:14` depending on your package configuration. | `node:16` |
+| **[navigationFallback.rewrite](https://learn.microsoft.com/en-us/azure/static-web-apps/configuration#fallback-routes)** | Is always `/api/server` | `/api/server` |
+| **[routes](https://learn.microsoft.com/en-us/azure/static-web-apps/configuration#routes)** | All prerendered routes are added. Additionally, if you do not have an `index.html` file an empty one is created for you for compatibility purposes and also requests to `/index.html` are redirected to the root directory which is handled by `/api/server`.  | `[]` |
+
+### Custom Configuration
+If you need to add additional properties to your configuration or alter the configuration that Nitro generates every build you need to create a `custom.staticwebapp.config.json` file in your project root.
```
Hebilicious, line 42: Can be simplified : 
You can alter the Nitro generated configuration with a `custom.staticwebapp.config.json` file in your project root.

docs/content/2.deploy/providers/azure.md
```diff
@@ -25,6 +25,24 @@ NITRO_PRESET=azure yarn build
...
+
+Nitro will override the properties it generates with the properties you specify in this custom configuration file, with the only exception being `routes`. For `routes`, Nitro will add your custom routes to the beginning of the array and put the routes it generates at the end. In the case of a conflict (determined if an object has the same `route` property) the custom route will take priority over the generated one.
```
Hebilicious, line 44: We can simplify this : 

Custom routes will be added and matched first. In the case of a conflict (determined if an object has the same `route` property), custom routes will override generated ones.

src/presets/azure.ts
```diff
@@ -41,14 +41,31 @@ async function writeRoutes(nitro: Nitro) {
     }
   }
 
+  // Attempt to load custom config
+  let customConfig = {};
+  try {
+    customConfig = JSON.parse(
```
Hebilicious, line 47: This can fail silently if JSON.parse fails (invalid json). 
Maybe we should actually throw an error if JSON.parse fails ? 
One way to do it would be to check if the file exist first.
 
`const customConfig = exists ? JSON.parse() : {}`

src/presets/azure.ts
```diff
@@ -99,6 +116,26 @@ async function writeRoutes(nitro: Nitro) {
     });
   }
 
+  // Prepend custom routes to the beginning of the routes array and override if they exist
+  if ("routes" in customConfig && Array.isArray(customConfig.routes)) {
+    // We iterate through the reverse so the order in the custom config is persisted
+    for (const customRoute of customConfig.routes.reverse()) {
+      let existingRouteMatchIndex = -1; // We want to record the index so we can override if it does exist
+      const routeExistsAlready = config.routes.some((element, index) => {
```
Hebilicious, line 124: could use `config.routes.findIndex` instead of some to simplify the logic here.

test/fixture/custom.staticwebapp.config.json
```diff
@@ -0,0 +1,38 @@
+{
```
Hebilicious, line 1: Why do we need this file if we attempt to delete it when the tests runs ?

src/presets/azure.ts
```diff
@@ -99,6 +120,24 @@ async function writeRoutes(nitro: Nitro) {
...
 
+  // Prepend custom routes to the beginning of the routes array and override if they exist
+  if ("routes" in customConfig && Array.isArray(customConfig.routes)) {
+    // We iterate through the reverse so the order in the custom config is persisted
+    for (const customRoute of customConfig.routes.reverse()) {
+      const existingRouteMatchIndex = config.routes.findIndex(
+        (value) => value.route === customRoute.route
+      );
+
+      if (existingRouteMatchIndex === -1) {
```
Hebilicious, line 131:
```suggestion
      if (existingRouteMatchIndex) {
```

NicholasDawson:
> I would get rid of the try catch block entirely. Otherwise this looks good to me.

@Hebilicious All set.

NicholasDawson:
@Hebilicious do you know if this PR is ready to be merged, or does @pi0 need to review as well? Just wondering where it stands.

Hebilicious:
Pi is the code owner and has to review and merge all non documentation PRs.

pi0 (approved):
Thanks so much for the PR ❤️  (and sorry it took long to review!)

I have made a couple of reactors to support via `nitro.azure.config` instead to avoid introducing second file and be consistent with other presets i hope you are happy with it.

PR #1347, opened 2023-06-25
fix: allow configuring all esbuild transform 

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/21756

### ❓ Type of change

### 📚 Description

Currently it's not possible to configure the esbuild plugin more generally. (The `tsconfig` option is completely ignored, for example.) This is a small change to enable configuring the options passed to transform. I've tried to keep behaviour as close to current as possible but we may want instead to rework this plugin.

### 📝 Checklist


pi0 (approved):
LGTM. 

Note: While this fix/feature allows to override `tsconfig` and `target` directly, this is not something i would recommend or advice or that we might always support for users to set and should be determined by nuxt/nitro framework and presets.

PR #1351, opened 2023-06-27
feat: experimental database layer

pi0:
### ❓ Type of change

### 📚 Description

This PR adds support for experimental Database support with **zero configuration**, SQLite and multi-provider support powered by [unjs/db0](https://db0.unjs.io/).

It is also possible to integrate this layer with extra ORMS such as [drizzle-orm](https://github.com/drizzle-team/drizzle-orm) with built-in connectors from db0.

**Example usage:***

Enable experimental flag:

```ts
  experimental: {
    database: true,
  },
});
```

For nuxt:

```ts
  nitro: {
    experimental: {
        database: true,
      },
  }
});
```

Install `better-sqlite3`:

```sh
pnpm i -D better-sqlite3
npm i -D better-sqlite3
yarn add -D better-sqlite3
```

And then, voila!

```ts
export default defineEventHandler(async () => {
  const db = useDatabase();

  // Create users table
  await db.sql`DROP TABLE IF EXISTS users`;
  await db.sql`CREATE TABLE IF NOT EXISTS users ("id" TEXT PRIMARY KEY, "firstName" TEXT, "lastName" TEXT, "email" TEXT)`;

  // Add a new user
  const userId = String(Math.round(Math.random() * 10_000));
  await db.sql`INSERT INTO users VALUES (${userId}, 'John', 'Doe', '')`;

  // Query for users
  const { rows } = await db.sql`SELECT * FROM users WHERE id = ${userId}`;

  return {
    rows,
  };
});
```

TODO:

- [x] Add basic test
- [x] Support configurable connector
- [x] Opensource DB0
- [x] Add to the docs

Next steps:

- Enable `useDB` without the flag once db0 is stable
- Built-in Schema and Migration support
- Auto-detect database connectors (vercel postgres and cloudflare D1)

### 📝 Checklist


cayter:
In the case of multiple databases, how do we specify which one to use? Would it be better to support key-value or map? For example, this is how Rails supports [multiple databases](https://guides.rubyonrails.org/active_record_multiple_databases.html#setting-up-your-application).

cayter:
I'm new to how Nitro/Nuxt works but I have 2 questions:

- In the case of multiple databases, how do we specify which one to use? Would it be better to support key-value or map? For example, this is how the famous web frameworks supports multiple databases:
  - [Rails](https://guides.rubyonrails.org/active_record_multiple_databases.html#setting-up-your-application)
  - [Django](https://docs.djangoproject.com/en/4.2/topics/db/multi-db/)

- Does [this](https://github.com/unjs/nitro/pull/1351/files#diff-52f266496020479959b18883f1912b939561cf9a07e8454d13816074401f4d40) take care of persistent DB instance to avoid leaking DB connections between HMRs during local development? [This](https://www.prisma.io/docs/guides/other/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices) is a workaround we have to enable during local development when working with Prisma/Drizzle ORM in NextJS/Remix.

pi0:
> In the case of multiple databases, how do we specify which one to use? Would it be better to support key-value or map? 

I am thinking of `useDatabase(name?)` for multiple connections support

> Does [this](https://github.com/unjs/nitro/pull/1351/files#diff-52f266496020479959b18883f1912b939561cf9a07e8454d13816074401f4d40) take care of persistent DB instance to avoid leaking DB connections between HMRs during local development

Once we migrate to DB0, we can support reconnect support. As you mentioned, HMR is the tricky part because any (runtime) logic will be restarted. Another thing we have in plan, is a HTTTP driver this way we can initialize DBs in the builder namespace and expose it via HTTP (or another RPC channel) to the runtime to handle HMR.

cayter:
> > In the case of multiple databases, how do we specify which one to use? Would it be better to support key-value or map?
> 
> I am thinking of `useDatabase(name?)` for multiple connections support
> 
> > Does [this](https://github.com/unjs/nitro/pull/1351/files#diff-52f266496020479959b18883f1912b939561cf9a07e8454d13816074401f4d40) take care of persistent DB instance to avoid leaking DB connections between HMRs during local development
> 
> Once we migrate to DB0, we can support reconnect support. As you mentioned, HMR is the tricky part because any (runtime) logic will be restarted. Another thing we have in plan, is a HTTTP driver this way we can initialize DBs in the builder namespace and expose it via HTTP (or another RPC channel) to the runtime to handle HMR.

Got it, thanks for the quick reply!

imcm7:
I don't know what is `[drizzle-orm](https://github.com/drizzle-team/drizzle-orm)`, people use sequalize, typeorm ad mikro-orm.

husayt:
would it be possible to use with other orms, notably prisma

pi0:
It is on the track. As long as an ORM allow implementing external providers and compatible with SQLite dialects i don't see any reason we couldn't support any of them

 (this PR is BTW placeholder actual work is under progress in a currently private repo)

vinayakkulkarni:
> It is on the track. As long as an ORM allow implementing external providers and compatible with SQLite dialects i don't see any reason we couldn't support any of them
> 
> (this PR is BTW placeholder actual work is under progress in a currently private repo)

so basically, instead of using [all the code here](https://github.com/vinayakkulkarni/vinayakkulkarni.dev/blob/main/server/utils/db.ts), i could directly use `sql` from `useDatabase()` 

does useDatabase enable us to use different db servers? postgres for example? and passing args for db server like i've put in the codebase

PR #1352, opened 2023-06-27
feat(storage): default data storage for `node` and dev presets

pi0:
### ❓ Type of change

### 📚 Description

The storage interface from `useStorage` is by default all in memory. In order to write stateful KV logic, users therefore require to configure `storage` mountpoints.

This PR adds a default `data` mountpoint using fs driver for all node based presets and development mode. (only if not configured by user) This way data will be persisted in `.data/kv` by default only for `data:` .

```ts
  const kvStorage = useStorage("data");
  const counter = (Number(await kvStorage.getItem("counter")) || 0) + 1;
  await kvStorage.setItem("counter", counter);
  return {
    counter,
  };
});
``` 

Next steps:
- Support auto config storage for other worker presets when env/binding exits
- Document this feature + `.data` dir and gitignore
- Use data storage for cache as well (or consider persisting `.nitro`/`.nuxt` partially)

### 📝 Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ jspm_packages
 # Nitro
 nitro.d.ts
 .nitro
+.data
 
 package-lock.json
 
--- a/playground/routes/index.ts
+++ b/playground/routes/index.ts
@@ -1,4 +1,8 @@
-// eslint-disable-next-line require-await
 export default eventHandler(async (event) => {
-  return {};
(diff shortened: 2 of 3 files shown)


4F2E4A2E:
Could this be related? **Default Persistent Data Storage Import is broken https://github.com/nuxt/nuxt/issues/23484**

PR #1375, opened 2023-07-02
docs: add cdn docs for multi deployment handling

Hebilicious:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/20950
Original PR by @edwh #1300 

### ❓ Type of change

### 📚 Description

When using Nuxt on Netlify, deploying a new version renders chunk names in existing clients invalid because of Netlify's atomic deploys.  That causes errors if the client later tries to fetch new chunk files, either when loading async components or new routes.   This documentation PR explains how to avoid that.

It may also apply to Vercel but I don't use that, so I've not looked into or tested the corresponding change there.

### 📝 Checklist

### DIFF
--- a/docs/content/2.deploy/providers/netlify.md
+++ b/docs/content/2.deploy/providers/netlify.md
@@ -38,7 +38,7 @@ On-demand Builders are serverless functions used to generate web content as need
 ## Handling atomic deploys
 
 Netlify uses atomic deploys, which means that all old chunk files will become invalid when a deploy occurs.
-This causes errors when existing clients try to load chunk files which no longer exists.
+This causes errors when existing clients try to load chunk files which no longer exist.
 
 Nuxt, for example, has default behaviour to reload the page in some circumstances, but this is not always
 desirable and it is possible to do better by serving Netlify's permanent link to each deploy, which remains valid.
--- a/docs/content/2.deploy/providers/netlify.md
+++ b/docs/content/2.deploy/providers/netlify.md
@@ -57,3 +57,11 @@ Then in the `_redirects` file which Netlify will pick up, proxy it:
 
 Proxying is better than serving the files directly because some security software (e.g. Streamline3)
 will block scripts served from a different domain.
+
+For example, you might be using the custom domain `https://example.com`, and your Netlify account might be
+`example.netlify.app`. In this case, `process.env.DEPLOY_URL` might be something like `https://5b243e66dd6a547b4fee73ae--example.netlify.app`.
+
+Setting the `cdnURL` above will cause Nuxt to load scripts from `https://example.com/netlify/5b243e66dd6a547b4fee73ae--example.netlify.app/*`,
+a non-existent directory within the custom domain.
+[Netlify will proxy](https://docs.netlify.com/routing/redirects/rewrites-proxies/) this back to files under the original `DEPLOY_URL`
+address without the HTTP client being aware of the domain change.
--- a/docs/content/2.deploy/providers/netlify.md
+++ b/docs/content/2.deploy/providers/netlify.md
@@ -35,3 +35,25 @@ Nitro output can directly run the server at the edge. Closer to your users.
 
 On-demand Builders are serverless functions used to generate web content as needed that’s automatically cached on Netlify’s Edge CDN. They enable you to build pages for your site when a user visits them for the first time and then cache them at the edge for subsequent visits.  ([Read More](https://docs.netlify.com/configure-builds/on-demand-builders/))
 
+## Handling atomic deploys
+
+Netlify uses atomic deploys, which means that all old chunk files will become invalid when a deploy occurs.
+This causes errors when existing clients try to load chunk files which no longer exists.
+
+Nuxt, for example, has default behaviour to reload the page in some circumstances, but this is not always
+desirable and it is possible to do better by serving Netlify's permanent link to each deploy, which remains valid.
+
+For example in Nuxt, add this to `nuxt.config.ts`
+
+    $production: {
+        app: {
+            cdnURL: '/netlify/' + process.env.DEPLOY_URL?.replace('https://', ''),
+        },
+    },
+
+Then in the `_redirects` file which Netlify will pick up, proxy it:
+
+    /netlify/* https://:splat 200!
+
+Proxying is better than serving the files directly because some security software (e.g. Streamline3)
+will block scripts served from a different domain.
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for reworking this PR. 

This is an intelligent solution with Netlify and Nuxt! But essentially all providers inherit this kind of limitation (unless a custom CDN which is able to serve older deployment chunks is configured). Also, this doc is meant to be generic. `cdnURL` is only used by Nuxt not nitro.

In order to move this forward, we should:

- Create a generic page in the deploy section explaining CDN deployment and limitations
- Use a generic example with route rules and an example global env 
- Use a use case example for Nuxt + Netlify

Hebilicious:
> Thanks for reworking this PR.
> 
> This is an intelligent solution with Netlify and Nuxt! But essentially all providers inherit this kind of limitation (unless a custom CDN which is able to serve older deployment chunks is configured). Also, this doc is meant to be generic. `cdnURL` is only used by Nuxt not nitro.
> 
> In order to move this forward, we should:
> 
> * Create a generic page in the deploy section explaining CDN deployment and limitations
> * Use a generic example with route rules and an example global env
> * Use a use case example for Nuxt + Netlify

Given the proposed solution is Netlify specific, having it underneath the Netlify provider makes sense to me, as that's where I'd look first for information about my deployment. 

For a generic CDN deployment limitations, given that they appear to be Nuxt specific, I don't know what we can write here that would be useful for standalone Nitro users. That being said, I made a proposal here in the Nuxt docs https://github.com/nuxt/nuxt/issues/21930 that could allow us to deal with this issue.

Overall I'm worried we are being too conservative with the docs, and we are forcing new users to google around and read PRs/issues to do something.

pi0:
Soluton is Nuxt specific also :) I'm not sure it is mergable and also sure it can be improved in a generic way :)

edwh:
This will sound grumpy and ungrateful, which I'm really not - I'm a long-term devoted Nuxt user.  

Shouldn't we be aiming to have a deployment experience which is as slick as the developer experience of making a code change?  If it isn't, can we really say that we fully support a particular deployment environment?

Unless I'm missing stuff, which I may be, where we're at is that on many hosting setups a new release will trigger JS errors for a significant fraction of users.  That will at best result (in the Nuxt case) in the page reloading unexpectedly under their feet, and at worst in JS errors which cause the site to stop working in a way the user cannot be expected to understand or realise how to resolve.

At the very least, I would say that every provider where this kind of chunking issue exists should be very clearly flagged on https://nitro.unjs.io/deploy as something a developer has to resolve for themselves.  In some cases that's an absolutely killer issue for a production system.  And we're all in favour of continuous deployment, right?

A generic solution would be great, but even per-provider workarounds are better than leaving people to stumble across it themselves and then navigate through the multiple PRs/issues about it.

danielroe:
@edwh As an aside, we do have https://github.com/nuxt/nuxt/issues/21370 downstream to provide a solution to this at a different level within Nuxt itself, and I'm hoping to land that in 3.7. I've also spoken to @pi0 about providing 'skew protection' within Nitro. So I don't think this is a question about not providing a solution.

edwh:
@danielroe Yes, I've noticed that.  But it still has a feeling of "acceptably robust deployment is left as an exercise for the reader".  

I'm very aware that I risk annoying the very people who kindly spend their lives developing the stuff I'm relying on and not doing a good enough job of contributing to.  

But it seems to me that we could do with a bit more focus on the deployment experience.

pi0:
Hi, dear @edwh. I understand what you are saying a **super sorry** for the confusion of discussions happening here.

The deployment issue with SPA applications not finding old deployment chunks has **always been a case** for Nuxt (and other similar SSR frameworks) because Deployment providers such as Vercel and Netlify only serve the last deployment js chunks from the main domain. This issue is not new or not related to Nitro's scope (which is purely a _server builder). 

As @danielroe mentioned, we are thinking about Nuxt-specific solutions such as using a manifest to auto-reload applications when a new version is detected or using a service worker.

Yet, the solution in this PR is also a nice idea to leverage CDN subdomains for serving multiple build chunks but we need to put in the right place of documentation. It is more than a workaround than officially advisable solution.

edwh:
@pi0 Thanks for the explanation.  I do appreciate you taking the time.  As @Hebilicious alluded to, there's an issue for Nuxt users (such as myself) shuttling between Nuxt and Vite docs without fully understanding the scope of each.  

I understand it's a long-standing problem, but that is probably all the more reason to document it until it is resolved.  

The Nuxt docs deployment page takes you straight to the Vite docs, so if this is outside the scope of the Vite docs, then could it be flagged on https://nuxt.com/docs/getting-started/deployment#supported-hosting-providers ?

If there are providers which don't have this issue, then for some people that will be a very clear reason to choose one provider over another.  This chunk stuff is an issue which many people will only start to see biting them once they go live - at which point it is then a big headache with no trivial solution which causes real users real grief.

Hebilicious:
> @pi0 Thanks for the explanation. I do appreciate you taking the time. As @Hebilicious alluded to, there's an issue for Nuxt users (such as myself) shuttling between Nuxt and Vite docs without fully understanding the scope of each.
> 
> I understand it's a long-standing problem, but that is probably all the more reason to document it until it is resolved.
> 
> The Nuxt docs deployment page takes you straight to the Vite docs, so if this is outside the scope of the Vite docs, then could it be flagged on [nuxt.com/docs/getting-started/deployment#supported-hosting-providers](https://nuxt.com/docs/getting-started/deployment#supported-hosting-providers) ?
> 
> If there are providers which don't have this issue, then for some people that will be a very clear reason to choose one provider over another. This chunk stuff is an issue which many people will only start to see biting them once they go live - at which point it is then a big headache with no trivial solution which causes real users real grief.

I would be happy to close this and move these instructions to the Nuxt page that you linked. @danielroe what do you think ?

pi0:
`/deployment` page in **nuxt docs** could probably be a better place to target this _workaround_ for netlify deployment (it is up to @danielroe to decide if such a workaround makes sense to be included there too or not.)

We are working on building manifest / auto-reload which solves the SPA deployment issue for all targets which requires almost no extra configuration per provider. 

Also, we are working to find a way to use the remote source feature of Docus to cross-link some shared docs reducing the requirement of moving between docs back and forth.

I am closing this PR since it really not making sense to be included this way in nitro docs. (Still open to include if you can satisfy requirements in https://github.com/unjs/nitro/pull/1375/files). 


Thanks for your contribution ❤️

pi0:
This PR in nuxt that implements initial manifest integration: https://github.com/unjs/nitro/pull/1375#issuecomment-1617753461 (to be paired with some nitro stuff) and ofc your original issue https://github.com/nuxt/nuxt/issues/21370 👍🏼

PR #1377, opened 2023-07-02
feat(firebase): support renaming exported server function

### LINKED ISSUE #266: [firebase] support renaming functions
After successfully deploying Nuxt 3 to Firebase using the Nitro Firebase build preset, I noticed that the Firebase Function name generated through `nuxi build` is 'server' by default. 

Firebase allows its users to host multiple sites and create multiple cloud functions within a Firebase project, however, each function must be uniquely named.

Because of the build output not being customizable, to support multiple Nuxt 3 sites in the same Firebase project, I am forced to add a 'hacky' build step to update the `.output/server/index.mjs` file (created by Nuxt) to rename the exported function from 'server' to a custom function name.

I have not been able to find a way to customize the function name in the build output, and it seems this is not yet supported by Nitro, and therefore is not exposed in the Nuxt config either. 

I would like to request this as a feature, possibly something that could be added to the `nitro.config.js` file (and ultimately the `nuxt.config.js` file).

Example:

```
  preset: 'firebase',
  firebase: {
    serverFunctionName: 'my-custom-function-name'
  }
})
```

If this is already supported in any way, and I missed it, let me know. 
If you have questions or need more input, please let me know. 

Thanks, Nitro team,

AlexLavoie42:
### 🔗 Linked issue

#266  

### ❓ Type of change

### 📚 Description

Adds a new option to nitro config that allows giving a custom name to deployed firebase functions.

This is done using rollup replace. I changed the exported name for the server in the firebase entry to `__firebaseServerFunctionName__` to ensure it is unique to the firebase server export. The placeholder name is then replaced with `firebase.serverFunctionName` in the config, which is given a default value of `"server"`.

This change gives an intuitive way to deploy multiple servers to the same firebase project without having them be overwritten. The workaround given in #266 is not very intuitive, and could potentially break other deployments in the future.

I would also recommend adding the same option to V2 functions in #1142, since there doesn't appear to be a way to change the deployed function name in the V2 [HttpOptions](https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.https.httpsoptions) (I don't think [`labels`](https://firebase.google.com/docs/reference/functions/2nd-gen/node/firebase-functions.https.httpsoptions.md#httpshttpsoptionslabels) does this, but correct me if I am wrong)

Resolves #266

### 📝 Checklist

I have tested building my own project and the function renaming appears to work as expected. The only difference is that `export { s as server } from './chunks/nitro/firebase.mjs'` becomes `export { _ as server } from './chunks/nitro/firebase.mjs'` since the export now starts with `_` before the placeholder gets replaced. I don't think this will cause any issues, but I am not 100% sure.

### DIFF
--- a/docs/content/2.deploy/providers/firebase.md
+++ b/docs/content/2.deploy/providers/firebase.md
@@ -244,3 +244,33 @@ You may be warned that other cloud functions will be deleted when you deploy you
 ```bash
 firebase deploy --only functions:server,hosting
 ```
+
+## Advanced
+
+### Renaming Function
+
+When deploying multiple apps within the same Firebase project, you must give your server a unique name in order to avoid overwriting
+your functions.
+
+You can specify a new name for the deployed Firebase function in your configuration:
+
+::code-group
+```ts [nitro.config.ts]
+import { defineNitroConfig } from 'nitropack/config'
+
+export default defineNitroConfig({
+  firebase: {
+    serverFunctionName: "<new_function_name>"
+  }
+})
+```
+```ts [nuxt.config.ts]
+export default defineNuxtConfig({
+  nitro: {
+    firebase: {
+      serverFunctionName: "<new_function_name>"
+    }
+  }
+})
+```
+::
--- a/src/presets/firebase.ts
+++ b/src/presets/firebase.ts
@@ -1,6 +1,7 @@
 import { existsSync } from "node:fs";
-import { join, relative } from "pathe";
+import { join, relative, basename } from "pathe";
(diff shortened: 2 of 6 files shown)


Hebilicious:
This is looking good thanks ! Let's wait until https://github.com/unjs/nitro/pull/1142 is merged before moving forward with that one.

pi0 (approved):
Thanks for the PR 💚 I have rebased it and updated implementation to match the latest requirements

PR #1381, opened 2023-07-03
fix: remove fsevents packages to fix firebase build on mac os

### LINKED ISSUE #1367: [firebase] firebase builds can break on macos
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.2.0`
- Nuxt Version:     `0.0.0`
- Nitro Version:    `2.5.2`
- Package Manager:  `pnpm@8.6.3`
- Builder:          `webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/unjs/nitro-deploys

### Describe the bug

If one dependency uses fsevents, running `NITRO_PRESET=firebase npm run build` will create an un-deployable project.

To fix it, the `.output/server/package.json` and `node_modules` directories needs to be altered manually to remove the `fsevents` dependency, which is macos only and will break cloudbuild. 

This should happen on macos only, windows and linux should be fine.

### Additional context

We can fix it by filtering out fsevents in the firebase preset, since we should be targeting linux and not for darwin, regardless of the environment `firebase deploy` is being run in.

### Logs

```sh
i  functions: creating Node.js 18 (2nd Gen) function server(us-central1)...
Build failed with status: FAILURE and message: npm ERR! code EBADPLATFORM
npm ERR! notsup Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin"} (current: {"os":"linux","arch":"x64"})
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  undefined
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64

npm ERR! A complete log of this run can be found in:
npm ERR!     /www-data-home/.npm/_logs/2023-07-01T17_56_37_753Z-debug-0.log. For more details see the logs at https://console.cloud.google.com/cloud-build/builds;region=us-central1/3622450c-555e-4490-9d09-9e5dfa875243?project=416452184607.

Functions deploy had errors with the following functions:
        server(us-central1)
i  functions: cleaning up build files...

Error: There was an error deploying functions
```

JamieCurnow:
### 🔗 Linked issue

#1367 

### ❓ Type of change

### 📚 Description

In the firebase preset file, globby is used to find all required packages to generate a package.json file for firebase functions to use. On a mac device, the packages include fsevents so this was being added to the package.json. fsevents does not work on linux, so when the code landed on google cloud to be built, it was failing as fsevents could not be installed. I have just filtered out the fsevents package so it is not added to the generated package.json. This affects both firebase preset and the new firebase-v2 preset.

Resolves #1367 

### 📝 Checklist

- [ ] I have updated the documentation accordingly - N/A


JamieCurnow:
@Hebilicious PR here for the firebase mac os issue 👍

Hebilicious (approved):

Hebilicious:
@JamieCurnow Amazing thanks 🙏🏽

src/presets/firebase.ts
```diff
@@ -51,9 +51,14 @@ async function writeRoutes(nitro: Nitro) {
...
   const _require = createRequire(import.meta.url);
 
-  const jsons = await globby(
+  const allJsons = await globby(
     join(nitro.options.output.serverDir, "node_modules/**/package.json")
   );
+
+  // fsevents is a macOS specific package that breaks in google cloud build, so we filter-it ot
+  // see: https://github.com/unjs/nitro/issues/1367
+  const jsons = allJsons.filter((json) => !json.includes("fsevents"));
```
posva, line 60: Would it make more sense to check for something more specific like `"name": "fsevents"`

JamieCurnow: @posva granted, `jsons` is not a great name for this const, but if you follow the code, it comes from globby:

```ts
const jsons = await globby(
    join(nitro.options.output.serverDir, "node_modules/**/package.json")
  );
```

globby is going to return an array of file paths like:

```ts
[
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/anymatch/package.json",
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/binary-extensions/package.json",
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/buffer-equal-constant-time/package.json",
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/braces/package.json",
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/buffer-from/package.json",
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/chokidar/package.json",
  "/Users/USER_NAME/PROJECT_PATH/.output/server/node_modules/fsevents/package.json",
  // ... etc
]
```

So this is why we're using an `includes` function to strip out any globby paths that have a match for fsevents.

We could use some regex or a more specific string in includes here to make sure that for instance if a user's home dir on a mac is `/Users/fsevents` because thats what they used when they set up their mac...?

I'm not sure how specific we want to be - seems like a very rare case...

I guess we could check for `"fsevents/package.json"` and filter those out to be more specific?

JamieCurnow: Or we could move the check down into the reduce call after the require:

```ts
const dependencies = jsons.reduce((obj, packageJson) => {
  const dirname = packageJson.slice(prefixLength, -suffixLength);
  if (!dirname.includes("node_modules")) {
    const packageJsonObject = _require(packageJson)
    if (packageJsonObject.name !== "fsevents") {
      obj[dirname] = _require(packageJson).version;
    }
  }
  return obj;
}, {} as Record<string, string>);
```

What do you think @Hebilicious ?

posva: In that case you can exclude directly using the options of the globby function then

Eckhardt-D:
@Hebilicious Is there any progress here? This is affecting Nuxt deploys on Firebase too if e.g. `nuxt-simple-sitemap` is used.

pi0:
Hi. Thanks for PR again and sorry it took long. It should be solved by https://github.com/unjs/nitro/pull/1610

PR #1384, opened 2023-07-03
fix!: use new 'bundler' module resolution

danielroe:
### ❓ Type of change

### 📚 Description

This likely needs to be released carefully, but we should eventually move to new 'Bundler' module resolution, which supports types in subpath imports and more exactly models how rollup will resolve imports.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -210,7 +210,7 @@ declare module 'nitropack' {
         strict: nitro.options.typescript.strict,
         target: "ESNext",
         module: "ESNext",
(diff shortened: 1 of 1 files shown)


pi0:
Shall we release it as new "Experimental" flag option?

danielroe:
Makes sense :+1:

Do you have preference for API name?

pi0:
You speak native English so i trust you on this :D Since we might end-up supporting it as a Nuxt flag too, probably `experimental.` namespace is best.

PR #1385, opened 2023-07-03
docs: Documenting the workaround for firebase v1 function config changes

### LINKED ISSUE #1369: Allow customizing firebase regions
```typescript
// https://nuxt.com/docs/api/configuration/nuxt-config
  /* Other config options */
  nitro: {
    preset: "firebase",
    replace: {
      "functions.https.onRequest": "functions.region('europe-west2').https.onRequest",
    },
  },
});
```

This is little hacky, but better than nothing. 
After we had support for regions with firebase v2 we can add support for v1.

_Originally posted by @luc122c in https://github.com/unjs/nitro/issues/90#issuecomment-1430548908_

JamieCurnow:
### 🔗 Linked issue

#1369 

### ❓ Type of change

### 📚 Description

Documents the workaround for changing `region` and `runWith` options for firebase functions v1 using rollup `replace`.

Resolves #1369 

### 📝 Checklist

### DIFF
--- a/docs/content/2.deploy/providers/firebase.md
+++ b/docs/content/2.deploy/providers/firebase.md
@@ -90,3 +90,72 @@ Deploying to Firebase Hosting is a simple matter of just running the `firebase d
 NITRO_PRESET=firebase yarn build
 firebase deploy
 ```
+
+## Changing cloud function settings
+
+Using the firebase v1 preset, there is currently no configuration to be able to change the region that the functions are deployed to. In order to change the region and any other firebase functions options, you will need to use the `replace` key in the nitro config. Here you will be able to change the `region` as well as `runWith` options for the functions.
+
+### Changing the region
+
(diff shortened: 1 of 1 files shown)


JamieCurnow:
@Hebilicious Here are some docs for the firebase v1 changing region etc. issues. We've been doing this in a production app and it works really well, though definitely feels hacky/work-aroundy.

pi0:
Thanks for making this pull request for providing instructuions @JamieCurnow (and sorry took long to review).

I think supporting firebase v1 with custom regions is a nice idea but instead of making a documentation for workaround like this, we could directly introduce it as a feature in preset itself to add replace lines and read region from firebase configuration.

Therefore probably adding hacks like this to the docs is not safest approach. 

I will be keep tracking on this via #90 also together with @posva that is going to help us on firebase side.

PR #1388, opened 2023-07-03
feat: support regexps in external options

danielroe:
### ❓ Type of change

### 📚 Description

I'm hoping to move more inlining from vite server build -> nitro build to avoid duplication of libraries. (e.g. h3 createError, some vue dependencies, anything shared between vite + nitro builds, etc.)

Part of that means being able to respect Nuxt's `build.transpile` which may include RegExps, and the best way would be to be able to pass those options on to nitro externals options.

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/externals-legacy.ts
+++ b/src/rollup/plugins/externals-legacy.ts
@@ -8,8 +8,8 @@ import semver from "semver";
 import { isDirectory, retry } from "../../utils";
 
 export interface NodeExternalsOptions {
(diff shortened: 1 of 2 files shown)


pi0 (commented):
This looks a nice addition. Shall we support functions in same iteration as well? might help to do some early refactors while was looking at code like making shared inline util for filter and normalize.

pi0 (approved):
💯

PR #1401, opened 2023-07-06
feat: support `exportConditions` and add worker default conditions

### LINKED ISSUE #83: Allow extending export conditions
https://github.com/nuxt/nuxt.js/issues/13138

> @danielroe  https://nodejs.org/api/packages.html#resolving-user-conditions
> currently probably only production and development fall into the category of conditions we'd support out-of-the-box, but the capacity to add more or customise them seems useful

Currently, we do use `default module node import` and I think it is a viable default but allowing to extend woulnd't be a bad idea either

### LINKED ISSUE #1123: prefer `web` over `node` export condition for non-node targets
When resolving dependencies (both bundled and externalized), we also add a `node` export condition (like [this](https://github.com/unjs/nitro/blob/ea5ea881a7bfa74def754d0a7120be4bc5b3ba7d/src/rollup/config.ts#L404)). 

Doing this, basically breaks dual build for packages like `uncrypto` because they assume node is for Node.js only but it is also necessary to use `node` condition always because (sadly) many libraries in npm ecosystem still consider `node ==== server`. We need to be cautious how to approach this but also eventually push the ecosystem and common libs to use a better condition

Example for vue-router conditions:

image

### LINKED ISSUE #1371: rollup should respect  the preset while reading package.json exports
### Motivation

When building for worker runtimes such as cloudflare/vercel-edge, there's an issue where the rollup plugin node resolve will follow the default priority which is good for Node, but not for these environments, which means in these scenarios many dependencies needs to be alias manually to their browser version.

### Describe the feature

Depending on the preset, we should alter rollup node-resolve `exportConditions` order priority.

For example for Cloudflare like environments, we could try `workerd > browser > import > default >  module > deno > node`.
For deno we could try `deno > browser > import >   default > workerd > module > node`

It looks like there's a lot of edge cases and no convention defined, ie vercel uses `edge-light`, libsql use `web` and `http` ...
We could run into an issue where we don't have the correct export no matter how much we try, hence why we should give the ability to customise this through Nitro options, as well as document it.

https://github.com/rollup/plugins/blob/master/packages/node-resolve/README.md
https://github.com/unjs/nitro/blob/main/src/rollup/config.ts

Edit : There's some convention defined there https://runtime-keys.proposal.wintercg.org/ , but no guarantee that libraries authors respect it. 

### Additional information

Hebilicious:
### 🔗 Linked issue

Resolves #1371 
Resolves #83
Resolves #1123
Resolves https://github.com/nuxt/nuxt/issues/21571

### ❓ Type of change

### 📚 Description

This works locally, I tested it against https://github.com/Hebilicious/authjs-nuxt/ which previously needed
```
      "jose": resolve(__dirname, "../node_modules/jose/dist/browser/index.js"),
      "@panva/hkdf": resolve(__dirname, "../node_modules/@panva/hkdf/dist/web/index.js"),
```

 And now it works 🎉     
 
image

### 📝 Checklist


danielroe (commented):
Rather than hard coding details about presets in the rollup config, perhaps we could have the export conditions set in the options and specifically in each preset?

That would also make it possible for users to configure this, and keep the core rollup configuration generic.

Hebilicious:
> Rather than hard coding details about presets in the rollup config, perhaps we could have the export conditions set in the options and specifically in each preset?
> 
> That would also make it possible for users to configure this, and keep the core rollup configuration generic.

@pi0 Regarding  `nitro.options.dev ? "development" : "production"`, if I'm not mistaken, this should always be production, except for the "nitro-dev" preset right ? If that's the case I think @danielroe refactor suggestion make things tidy as we can have a static `exportConditions` for each preset.

src/options.ts
```diff
@@ -458,3 +464,43 @@ export function normalizeRouteRules(
...
 }
+
+function _resolveExportConditions(
+  conditions: string[] = [],
+  opts: { dev: boolean; node: boolean }
+) {
+  const resolvedConditions: string[] = [];
+
+  // 1. Add dev or production
+  resolvedConditions.push(opts.dev ? "development" : "production");
```
Hebilicious, line 475: I think we should add development but not production @pi0

pi0: It is valid https://nodejs.org/api/packages.html#packages_community_conditions_definitions

pi0: Usecase is also for nested conditions. `exports: { production: { node, import, worker }, development: { node, worker, import }`. Conditions are checked from first to last and can be combined

Hebilicious: Sorry, I meant we should add it for node, not for worker-like environments. But I haven't seen this key used too much in the wild, so this should be fine.

PR #1403, opened 2023-07-07
docs: add nitro plugins examples

Hebilicious:
### ❓ Type of change

### 📝 Checklist


docs/content/1.guide/7.plugins.md
```diff
@@ -27,6 +27,38 @@ export default defineNitroPlugin((nitroApp) => {
 })
 ```
 
+**Example:** Development Plugin
+
+You can use plugins to run development tasks, such as database migrations.
+
+```ts
+import { migrate } from 'drizzle-orm/better-sqlite3/migrator'
```
pi0, line 35: Let's remove migration as an example. it is not best usage of plugins and i am thinking to introduce a better way for them.

docs/content/1.guide/7.plugins.md
```diff
@@ -27,6 +27,38 @@ export default defineNitroPlugin((nitroApp) => {
...
+You can use plugins to modify the response or run tasks when Nitro is closed.
+
+```ts
+export default defineNitroPlugin((nitro) => {
+  nitro.hooks.hookOnce("close", async () => {
+    // Will run when nitro is closed
+    await new Promise((resolve) => setTimeout(resolve, 500));
+    console.log("Task is done!");
+  });
+  nitro.hooks.hook('render:response', (response) => {
```
pi0, line 55: Hooks are good example. Maybe we can split it into two examples graceful shutdown and modifying response?

docs/content/1.guide/7.plugins.md
pi0: As said before, we are working on a better solution which is not plugins. There is no best place to advice today.

Hebilicious: Removed it

docs/content/1.guide/7.plugins.md
```diff
@@ -27,6 +27,34 @@ export default defineNitroPlugin((nitroApp) => {
...
+export default defineNitroPlugin((nitro) => {
+
+  nitro.hooks.hook('render:response', (response) => {
+    // Modify the renderer response here
+    console.log(response)
+  })
+})
+```
+
+**Example:** Graceful Shutdown
```
pi0, line 44: This is more generic example to be first imo

docs/content/1.guide/7.plugins.md
```diff
@@ -27,6 +27,34 @@ export default defineNitroPlugin((nitroApp) => {
 })
 ```
 
+**Example:** Response Modification
+
+You can use plugins to register a hook that modifies the renderer response.
```
pi0, line 32: We might link to the `renderer` option here and add a note it only works for when using the feature and Nuxt SSR not for all responses.

docs/content/1.guide/7.plugins.md
```diff
@@ -27,6 +27,34 @@ export default defineNitroPlugin((nitroApp) => {
...
+
+```ts
+export default defineNitroPlugin((nitro) => {
+
+  nitro.hooks.hook('render:response', (response) => {
+    // Modify the renderer response here
+    console.log(response)
```
pi0, line 39: Maybe something better than logging since example is showcasing "modify"?

docs/content/1.guide/7.plugins.md
Hebilicious: I accidentally push to the wrong branch https://github.com/unjs/nitro/compare/main...Hebilicious:nitro:docs/nitro-plugins

PR #527, opened 2022-09-21
feat: resolve actual subpaths before guessing

danielroe:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This resolves actual subpaths rather than guessing, so (for example) `node-fetch-native/dist/polyfill.mjs` can correctly be resolved to `node-fetch-native/polyfill` and back again.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1429, opened 2023-07-14
fix(externals): absolute paths are inlined

pi0:
### ❓ Type of change

### 📚 Description

Discovered while trying to make reproduction tests for #527.

Normally importing a library with a subpath like `@fixture/nitro-utils/extra` correctly creates an externals chunk in `.output/server/node_modules/@fixture/nitro-utils/extra.mjs`. But we wrongly inline if this path is absolute for any reason (like using aliases). 

This PR fixes this by using a default scoring assigned to include/exclude patterns based on their length (function patterns have highest priority always). When an inline rule conflicts with an external, it will be only applied if is more precise and also by adding all `node_modules` as externals by default.

### 📝 Checklist


danielroe (approved):

danielroe:
Let's merge and test for a while - like you, I think this could be risky.

PR #1437, opened 2023-07-15
fix(rollup): Pin rollup to 3.26.1 temporarily

AaronDewes:
### 🔗 Linked issue

#1432

### ❓ Type of change

### 📚 Description

Rollup 3.26.2 contained exactly one commit, and that commit broke Deno builds.
As a workaround, this pins rollup to an older version until the upstream bug (https://github.com/rollup/rollup/issues/5067) is fixed.

### 📝 Checklist

### DIFF
--- a/docs/content/1.guide/community/9.contributing.md
+++ b/docs/content/1.guide/community/9.contributing.md
@@ -2,7 +2,7 @@
 
 Thank you for wanting to contribute to Nitro 💛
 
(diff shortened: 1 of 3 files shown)


pi0:
Hi. Thanks for the PR. I think we should probably help to fix the upstream issue in rollup and avoid pinning especially since it seems is only affecting one preset only. In the meantime, you can use `resolutions` field in `package.json` to pin the rollup dependency. Let's keep track in #1432.

PR #1438, opened 2023-07-15
fix(deno-deploy): treat all `https://` modules as external

AaronDewes:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1432

### ❓ Type of change

### 📚 Description

Rollup 3.26.2 (accidentally?) made a breaking change in the handling of `manualChunks`. This updates rollup to a fixed version, and also changes the config so that all `https://` modules in Deno are treated external.

### 📝 Checklist

### DIFF
--- a/src/presets/deno-deploy.ts
+++ b/src/presets/deno-deploy.ts
@@ -16,14 +16,10 @@ export const denoDeploy = defineNitroPreset({
   },
   rollupConfig: {
     preserveEntrySignatures: false,
-    external: ["https://deno.land/std/http/server.ts"],
+    external: (id) => id.startsWith("https://"),
     output: {
       entryFileNames: "index.ts",
-      manualChunks: (id) => {
-        if (id !== "https://deno.land/std/http/server.ts") {
-          return "index";
-        }
-      },
+      manualChunks: (id) => "index",
       format: "esm",
     },
   },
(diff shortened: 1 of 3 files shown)


src/presets/deno-deploy.ts
```diff
@@ -19,7 +19,9 @@ export const denoDeploy = defineNitroPreset({
     external: ["https://deno.land/std/http/server.ts"],
     output: {
       entryFileNames: "index.ts",
-      manualChunks: () => "index",
+      manualChunks: (id) => {
+        if (id !== "https://deno.land/std/http/server.ts") { return "index"; }
```
pi0, line 23: Would it make sense that we check if id starting with http(s) protocol?

AaronDewes: Probably. There is an upstream fix coming though, so I'm not sure if this PR makes sense: https://github.com/rollup/rollup/issues/5067#issuecomment-1636827568, maybe it's better to wait for upstream.

AaronDewes: Actually, it wouldn't, because only `"https://deno.land/std/http/server.ts"` is external right now. But that could probably also be fixed, imports from URLs probably fail right now...

Hebilicious: I agree we should probably match id that start with `https://`. Let's wait for https://github.com/rollup/rollup/pull/5068

pi0:
@AaronDewes Can you please confirm if https://github.com/rollup/rollup/releases/tag/v3.26.3 is resolving our issue?

AaronDewes:
Yes, it is. I'll update this PR to include an update to rollup. I also changed it so Nitro now treats all "https://" imports as external (as you suggested), so this PR is still relevant.

pi0 (approved):
LGTM. Thanks!

PR #1439, opened 2023-07-15
chore(starter): add `.npmrc`

Barbapapazes:
### ❓ Type of change

### 📚 Description

In documentation, you speak a lot about using `nitro` with `pnpm`. But when installing deps with `pnpm`, `h3` utils are just any since `h3` is not a first party dep and user must manually install it or add a `.npmrc`. I think that added, by default this file, could help.

### 📝 Checklist

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+shamefully-hoist=true
+strict-peer-dependencies=false


Hebilicious:
In the main branch, h3 is in the dependencies.

I see you made this PR against the starter branch though.
Are you trying to make this change for the presets starters? We have this repo here https://github.com/unjs/nitro-preset-starter

Barbapapazes:
> In the main branch, h3 is in the dependencies.
> 
> I see you made this PR against the starter branch though.
> Are you trying to make this change for the presets starters? We have this repo here https://github.com/unjs/nitro-preset-starter

I want to make this PR to the starter, the one that is downloaded when doing npx giget nitro

.npmrc
```diff
@@ -0,0 +1,2 @@
+shamefully-hoist=true
```
pi0, line 1: Nitro is compatible with pnpm hoist. And we shall not enable this flag to make sure it keeps compatible.

pi0:
Thanks for the pull request. The top level `.npmrc` is not for starter template but for repository itself and for the reason of making sure being pnpm compatible, we should not enable this flag.

Adding such npmrc to the starter template might be a good idea in the meantime until solving "type issue" for sub dependencies.

pi0:
Sorry my bad! Let's have this fix for starter until resolving type issue.

PR #1440, opened 2023-07-16
chore(starter): add prepare command to starter

Barbapapazes:
### ❓ Type of change

### 📚 Description

Adding a prepare script could help user to have generated types right after an installation.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "private": true,
   "scripts": {
+    "prepare": "npx nitropack prepare",
     "dev": "npx nitropack dev",
     "build": "npx nitropack build",
     "preview": "node .output/server/index.mjs"


pi0:
I think it is a good idea to add prepare to the starter templates but this `package.json` is not for starter it is main `nitro` package which can be used also as a library (for Nuxt).

Barbapapazes:
> I think it is a good idea to add prepare to the starter templates but this `package.json` is not for starter it is main `nitro` package which can be used also as a library (for Nuxt).

Do I have to make this PR somewhere else? Not sure to have fully understand where to find the starter 🤔

pi0:
Sorry for confusions. Starter is kinda hidden right now! It is in the same repo but in `starter` branch (https://github.com/unjs/nitro/tree/starter). You need to switch to this branch and then make a pull request against it.

pi0:
Sorry my bad! You made it right 👍🏼

PR #1443, opened 2023-07-16
docs: use h3 composable

Barbapapazes:
### ❓ Type of change

### 📚 Description

Instead of deep diving into the event, we could use an h3 composable, like we do with `readBody` above.

### 📝 Checklist


pi0:
Thanks for the PR dear @Barbapapazes. I think at this point, since we are working on typing event object directly, it is safest to keep suggesting to depend on main `event` as much as possible to reduce the chance of recommending something _might_ be deprecated or less type friendly in the future.

PR #1452, opened 2023-07-17
fix: split cookie headers 

### LINKED ISSUE #988: Set Cookie merged into one header (Cloudflare Workers)

Hebilicious:
- feat: handler cookie headers for Response
- chore: flag unhandled cookies

### 🔗 Linked issue

Closes https://github.com/unjs/nitro/pull/989
Fix #988 

### ❓ Type of change

### 📚 Description

Continuing the work done by @oleghalin and applying it more generally to every presets that needs it.
Note that the culprit for this would be `unenv`, but fixing this in Nitro allows us to not add any dependency to `unenv`.

### 📝 Checklist

### DIFF
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -7,6 +7,7 @@ import {
   Router,
   toNodeListener,
   fetchWithEvent,
+  splitCookiesString,
 } from "h3";
 import { createFetch, Headers } from "ofetch";
 import destr from "destr";
@@ -18,7 +19,7 @@ import { createHooks, Hookable } from "hookable";
 import type { NitroRuntimeHooks } from "./types";
 import { useRuntimeConfig } from "./config";
 import { cachedEventHandler } from "./cache";
-import { withNormalizedHeaders } from "./utils";
+import { stringifyHeaders } from "./utils";
 import { createRouteRulesHandler, getRouteRulesForPath } from "./route-rules";
 import type { $Fetch, NitroFetchRequest } from "nitropack";
 import { plugins } from "#internal/nitro/virtual/plugins";
@@ -135,3 +136,28 @@ function createNitroApp(): NitroApp {
 export const nitroApp: NitroApp = createNitroApp();
 
 export const useNitroApp = () => nitroApp;
+
+const normalizeOutgoingFetchHeaders = (headers: Headers) => {
+  const outgoingHeaders = new Headers();
+  for (const [name, header] of headers) {
+    if (name === "set-cookie") {
+      for (const cookie of splitCookiesString(stringifyHeaders(header))) {
+        outgoingHeaders.append("set-cookie", cookie);
+      }
+    } else {
+      outgoingHeaders.set(name, stringifyHeaders(header));
+    }
+  }
+  return outgoingHeaders;
+};
+
+function withNormalizedHeaders(fetch: typeof globalThis.fetch) {
+  return async (...args: Parameters<typeof fetch>) => {
+    const r = await fetch(...args);
+    return new Response(r.body, {
+      headers: normalizeOutgoingFetchHeaders(r.headers),
+      status: r.status,
+      statusText: r.statusText,
+    });
+  };
+}
--- a/src/runtime/app.ts
+++ b/src/runtime/app.ts
@@ -7,7 +7,6 @@ import {
   Router,
   toNodeListener,
   fetchWithEvent,
-  splitCookiesString,
 } from "h3";
 import { createFetch, Headers } from "ofetch";
 import destr from "destr";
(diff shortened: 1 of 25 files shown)


pi0:
Thanks. I think we can add fix to a wrapper of localFetch util directly so that entries using it already have fix

Hebilicious:
> Thanks. I think we can add fix to a wrapper of localFetch util directly so that entries using it already have fix

We can do it like this  :
```ts
  function normalizedHeaders(fetch: typeof globalThis.fetch) {
    return async (...args: Parameters<typeof fetch>) => {
      const r = await fetch(...args);
      return new Response(r.body, {
        headers: normalizeOutgoingHeaders(r.headers),
        status: r.status,
        statusText: r.statusText,
      });
    };
  }

  // Create local fetch callers
  const localCall = createCall(toNodeListener(h3App) as any);
  const localFetch = normalizedHeaders(
    createLocalFetch(localCall, globalThis.fetch)
  );
  ```
I'm worried about the possible impact of changing `$fetch` for Nuxt, or something else besides the preset runtimes.
If the goal is to DRY we can add a `localNormalizedHeadersFetch` and call this in the presets.
I think I'll let you decide.

**Comment by pi0** [comment]
Yes, i still think wrapper method would be better idea.

**Comment by Hebilicious** [comment]
Is there any reason why cloudflare, lagon, denos, netlify-edge, service worker and vercel-edge presets are using localCall ? They are all returning a Response.

**Comment by pi0** [comment]
> Is there any reason why cloudflare, lagon, denos, netlify-edge, service worker and vercel-edge presets are using localCall ? They are all returning a Response.

They need to be upgraded feel free to include fix in same PR (basically i made that refactor back in time for other presets to prepare for this unified fix :D)

File: src/runtime/app.ts [hunk 1]
```diff
@@ -120,6 +125,7 @@ function createNitroApp(): NitroApp {
     router,
     localCall,
     localFetch,
+    localFetchWithNormalizedHeaders,
```
pi0, line 128: We should directly apply patch to `localFetch` if for (any) reason you think we need an escape hatch to disable the normalization of cookie header, it can be a custom fetch option however i think we should not need it

src/runtime/entries/stormkit.ts
```diff
@@ -34,6 +34,7 @@ export const handler: Handler<StormkitEvent, StormkitResult> = async function (
     body: event.body,
   });
 
+  // @todo handle cookies
```
pi0, line 37: We probably can remove this todos. `normalizeOutgoingHeaders` should be able to fix this

src/runtime/app.ts
Hebilicious: We are no longer returning the full Response, but a subset of it that contains body, headers, status and statusText. 
This matters for $fetch usage. 
Should we try to return everything and change localFetch itself.

Hebilicious: Actually nvm I think it should work

pi0: We can make a internal `_localFetch` variable only being passed to `createFetch` but i guess if standard is to split cookies, it is also nice to be applied to subsequent local `$fetch` calls..

What is more than body, headers, status, statusText available in a Response object?

src/runtime/entries/stormkit.ts
Hebilicious: I believe stormkit just wraps lambda so the cookies needs to be handled like in the `aws-lambda` preset

src/runtime/utils.ts
```diff
@@ -100,3 +102,35 @@ export function trapUnhandledNodeErrors() {
...
+        outgoingHeaders.append("set-cookie", cookie);
+      }
+    } else {
+      outgoingHeaders.set(name, joinIfArray(header));
+    }
+  }
+  return outgoingHeaders;
+}
+
+export function withNormalizedHeaders(fetch: typeof globalThis.fetch) {
```
pi0, line 127: This as a generic utility might be confusing to be seen (and named like this) here. Do you mind to move it inline to `app.ts`? 🙏🏼 (also we might end up moving fix finally to the unenv)

src/runtime/entries/stormkit.ts
Hebilicious: This needs to be done in netlify-lambda too

pi0: Sure if it is an easy thing to do we can make this PR a fix to normalize cookies for all presets...

Hebilicious:
@danielroe tagging you for the Azure changes

pi0:
Added some refactors and tests. While making tests discovered some issues. Netlify is fixed but we have Node (https://github.com/unjs/nitro/issues/1462) and Bun (https://github.com/unjs/nitro/issues/1461) to be fixed.

Probably on nitro-deploys with edge channel we can E2E test better on tests to.

pi0 (approved):

Issue #1396, opened 2023-07-05
`on:response` Hook


Hebilicious:
## Motivation 

There's a `render:response` hook to modify the rendered response implemented in the nitro rendere, but there isn't a `on:response` hook that can be called to modify the response before it is sent.

## Describe the feature

Add a `on:response` hook in Nitro, that would run after each runtime entry. 
A `defineNitroResponse(nitroApp, response)` can be used in each preset to help implementing.

```ts
const defineNitroResponse = async (nitro, response) => {
	await nitroApp.hooks.callHook("on:response", response);
return response
} 

Bun Example
```ts
  //...

    const response = await nitroApp.localFetch(url.pathname + url.search, {
      url: url.pathname + url.search,
      host: url.hostname,
      protocol: url.protocol,
      headers: request.headers,
      method: request.method,
      redirect: request.redirect,
      body,
    });
	
    return defineNitroResponse(nitroApp, response);
  },
});
```

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
That would be nice idea generally to be able to hook into responses. It might be little bit tricky since we need some better control in h3 (and use a configuration from nitro to hook into h3 response stage) and make sure response is not directly handled by `node.res.end` (cached handler uses a [dirty trick to make sure of this](https://github.com/unjs/nitro/blob/f0d0d9162e98556e62be17a5966d5731aa0399e5/src/runtime/cache.ts#L236)).

tobychidi:
If I wanted to collect and modify the API response body in a middle ware, how would I do that? I have an example of what could be done in Express JS. The res.send function was modified like so
```
    // Intercept response body
    const originalSend = res.send
    res.send = function sendOverWrite(body) {
      console.log(body)
      originalSend.call(this, body)
      this._treblleResponsebody = body
    }
```

```
fastify.addHook('onSend', (request, reply, payload, done) => {
    reply.payload = payload
    done()
  })
```

Is there a reason why such modification is needed? I was expecting `res.body`/`res.payload` to be a thing everywhere by default.

pi0:
Upstream issue for h3: https://github.com/unjs/h3/issues/447

kyng-cytro:
Any luck with this? If not, could you please share your current logging implementation?

Hebilicious:
> Any luck with this? If not, could you please share your current logging implementation?

Would something like this work for you ? 🤔 

```ts
[...]

PR #1453, opened 2023-07-17
feat: implement on:response hook

Hebilicious:
### 🔗 Linked issue

Resolves https://github.com/unjs/nitro/issues/1396

### ❓ Type of change

### 📚 Description

WIP 

There's a one issue with this implementation : 

`toNodeListener` isn't compatible with `defineNitroResponse`, so the presets would need to be refactored to use `localCall`, or h3s `toNodeListener` would need to be updated to accept a callback function that can modify the response.

For the hook API name we could use `extendResponse` instead of `modifyResponse`

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/aws-lambda.ts
+++ b/src/runtime/entries/aws-lambda.ts
@@ -9,6 +9,7 @@ import type {
 import "#internal/nitro/virtual/polyfill";
 import { withQuery } from "ufo";
 import { nitroApp } from "../app";
+import { defineNitroResponse } from "../utils";
 
 export async function handler(
   event: APIGatewayProxyEvent,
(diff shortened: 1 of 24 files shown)


pi0:
I don't think implemeting hook this way would be the best idea. Also as you have noticed, only worker presets using `localFetch` could benefit this and this is against nitro goals.

pi0:
We have `render:response` to modify the hook already to modify renderer responses. For genetic hook, if we ever gonna support it, it might be coming from h3. Let's keep the discussion under #1396.

pi0:
Upstream: https://github.com/unjs/h3/issues/447

PR #1459, opened 2023-07-19
fix(static): safe decode path

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/22071

### ❓ Type of change

### 📚 Description

In some cases (such as a malformed URL) we might encounter an error when checking whether the URL is a static asset. However, we can be assured it is _not_ a static asset and either throw a 404 if it has the static prefix, or simply early return from the handler and directly progress to the next handler, which can allow the user to correct the URL or throw their own error.

### 📝 Checklist

### DIFF
--- a/src/runtime/static.ts
+++ b/src/runtime/static.ts
@@ -1,5 +1,5 @@
 import { eventHandler, createError } from "h3";
-import { joinURL, withoutTrailingSlash, withLeadingSlash, parseURL } from "ufo";
+import { decodePath, joinURL, parseURL } from "ufo";
 import {
   getAsset,
   readAsset,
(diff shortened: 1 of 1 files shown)


pi0:
I think we can use `ufo.decodePath` instead which has built-in try/catch safety (https://github.com/unjs/ufo/blob/main/src/encoding.ts#L124)

src/runtime/static.ts
```diff
@@ -15,38 +15,41 @@ export default eventHandler((event) => {
...
+        const _asset = getAsset(_id);
+        if (_asset) {
+          asset = _asset;
+          id = _id;
+          break;
+        }
       }
     }
-  }
+  } catch {}
```
pi0, line 52: Why we need a try catch here with wide scope?

danielroe: It was so we could avoid further normalising of the id if it couldn't be resolved. I'll push a refactor with `decodePath`.

pi0: Thanks. Would be nice if we could limit the scope of try/catch this way to avoid hiding other errors.

pi0 (approved):
Thanks!

PR #1460, opened 2023-07-19
fix: add types for public assets data

danielroe:
### ❓ Type of change

### 📚 Description

We had an untyped `asset` which can safely produce types for. This PR improves internal type safety but no bug or issue was encountered with the code.

### 📝 Checklist

### DIFF
--- a/src/runtime/static.ts
+++ b/src/runtime/static.ts
@@ -4,6 +4,7 @@ import {
   getAsset,
   readAsset,
   isPublicAssetURL,
+  PublicAsset,
 } from "#internal/nitro/virtual/public-assets";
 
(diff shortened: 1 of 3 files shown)


src/runtime/static.ts
```diff
@@ -4,6 +4,7 @@ import {
   getAsset,
   readAsset,
   isPublicAssetURL,
+  PublicAsset,
```
Hebilicious, line 7:
```suggestion
 type PublicAsset,
```
Should this one be a type import ?

danielroe: Yes 👍

pi0 (approved):
Thanks 💯

Issue #1464, opened 2023-07-20
invalid `preview` command path when customizing `publicDir` with static output


xjccc:
### Environment

nuxt: 3.6.1
nitropack: 2.5.2
node: 16.18.0

### Reproduction

https://stackblitz.com/edit/nuxt-starter-pghlwa?file=nuxt.config.ts

### Describe the bug

1.  pnpm generate
2. view the output

node-server/public is not generate, the files under the node-server dir.

emmm, its should correction the generate path or the output log?
![log](https://github.com/unjs/nitro/assets/16478776/24ef951c-5b2c-4887-b296-6d8b54e767d2)

Hebilicious:
You can achieve what you want like this : 

```ts
export default defineNuxtConfig({
  nitro: {
    output: {
      dir: '~/node-server'
    }
  }
})
```

That being said, I can't think of an usecase where you'd want `output.publicDir` to be specified separately from `output.dir`

xjccc:
> You can achieve what you want like this :
> 
> ```ts
> export default defineNuxtConfig({
>   nitro: {
>     output: {
>       dir: '~/node-server'
>     }
>   }
> })
> ```
> 
> That being said, I can't think of an use case where you'd want `output.publicDir` to be specified separately from `output.dir`

oh, only use `output.dir` is the correct output

pi0:
Nitro reproduction: https://stackblitz.com/edit/github-vlunk8?file=nitro.config.ts&startScript=build  

I can confirm this edge case happens when trying to use static preset (nuxt generate)

pi0x:
Fixed in v3 — the build no longer prints a hardcoded `npx serve ./public`; it tells you to run `npx nitro preview`, and preview serves from the real output dir and publicDir even when both are customized (src/preview.ts).

_Triaged with AI assistance._

PR #1483, opened 2023-07-24
fix: correct custom public directory output log

xjccc:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1464

### ❓ Type of change

### 📚 Description

when use pnpm generate in nuxt
1. prerender(nitro)
2. build(nitro)

https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/nitro.ts#L383

i think its build twice in nuxt/nitro build:done
and reset the nitro config to build the nitro.json
the simple resolve if nitro.json exists used it?
its that right ?
thanks

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -1,4 +1,4 @@
-import { existsSync, promises as fsp } from "node:fs";
+import { existsSync, promises as fsp, readFileSync } from "node:fs";
 import { relative, resolve, join, dirname, isAbsolute } from "pathe";
 import { resolveAlias } from "pathe/utils";
 import * as rollup from "rollup";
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the PR dear @xjccc. I don't think preventing override manifest by reading it back will fix nitro issue at least (also if it does, it is not safe because all builds can optionally have a prerender stage).

What we want to fix, is to remove hardcoded preview command in [`src/presets/static.ts`](https://github.com/unjs/nitro/blob/ab57c75c3af27a2bb38b06315b0f63f95301707b/src/presets/static.ts#L13) (and other static presets).

PR #1489, opened 2023-07-26
feat: add `transparent` option to proxy

### LINKED ISSUE #1488: Transparently handle proxy errors
### Describe the feature

https://github.com/Aareksio/reproduction-nitro-proxy-error/tree/main

When the request passing through proxy returns non-successful status code, the error is thrown. I would love to opt-in for fully transparent proxy, forwarding reply exactly as is. See reproduction for current and desired state, as well as basic workaround using routes directory instead of `pageRules`.

```ts
// nitro.config.ts
  routeRules: {
    '/': { proxy: { to: 'https://github.com/404' } },
  },
});
```

```ts
// routes/404.ts
export default defineEventHandler((event) => {
  return proxyRequest(event, 'https://github.com/404')
})
```

### Additional information

Aareksio:
### 🔗 Linked issue

https://github.com/unjs/h3/issues/464

### ❓ Type of change

### 📚 Description

Currently [ofetch](https://github.com/unjs/ofetch/blob/main/src/fetch.ts#L208-L217) is transforming responses with status code between 400 and 600 to errors. This may not be the desired outcome, as it masks plenty of useful information received from proxy target. Therefore as the user, I may want to opt-out of the masking and receive raw response as-is.

Test with:
```ts
// nitro.config.ts
  routeRules: {
    '/404': {
      proxy: {
        to: 'https://github.com/404',
        transparent: true,
      }
    },
  }
});
```

Resolves unjs/h3#464

### 📝 Checklist

### DIFF
--- a/src/runtime/route-rules.ts
+++ b/src/runtime/route-rules.ts
@@ -54,11 +54,14 @@ export function createRouteRulesHandler() {
         }
       }
 
-      return proxyRequest(event, target, {
-        fetch: $fetch.raw as any,
-        ...routeRules.proxy,
-        ...transparentOptions
-      });
+      const options = defu(
+        {},
+        transparentOptions,
+        routeRules.proxy,
+        { fetch: $fetch.raw as any },
+      )
+
+      return proxyRequest(event, target, options);
     }
   });
 }
--- a/src/runtime/route-rules.ts
+++ b/src/runtime/route-rules.ts
@@ -46,9 +46,18 @@ export function createRouteRulesHandler() {
         const query = getQuery(event.path);
         target = withQuery(target, query);
       }
+
+      const transparentOptions: any = {}
+      if (routeRules.proxy.transparent) {
+        transparentOptions.fetchOptions = {
+          ignoreResponseError: true
+        }
+      }
+
       return proxyRequest(event, target, {
         fetch: $fetch.raw as any,
         ...routeRules.proxy,
+        ...transparentOptions
       });
     }
   });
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for your pull request. I think it makes sense that h3's proxy transparently gives errors as well. Let's move it to h3.

Aareksio:
I think you misattributed the issue. As explained in comment to https://github.com/unjs/h3/issues/464, error masking only happens:

- when using custom fetch: [source](https://github.com/unjs/nitro/blob/main/src/runtime/route-rules.ts#L50)
- and passing `ignoreResponseError` is not possible because h3 expects native fetch types, not ofetch: [source](https://github.com/unjs/h3/blob/main/src/utils/proxy.ts#L12)

h3's `proxyRequest` is transparent by default

---

On the other hand, unhandled exception is only happening when handler is created by nitro from `routeRules`. Otherwise you can easily handle it with `try ... catch`. Therefore both _parts_ of the initial issue touch only nitro part, not h3.

src/runtime/route-rules.ts
```diff
@@ -46,10 +46,22 @@ export function createRouteRulesHandler() {
...
       }
-      return proxyRequest(event, target, {
-        fetch: $fetch.raw as any,
-        ...routeRules.proxy,
-      });
+
+      const transparentOptions: any = {}
+      if (routeRules.proxy.transparent) {
+        transparentOptions.fetchOptions = {
+          ignoreResponseError: true
```
pi0, line 53: We can pass it by default from h3 assuming nitro and others might pass `$fetch.raw` as implementation as a fix

PR #1497, opened 2023-07-28
fix(openapi): merge handlers with same route and different method

### LINKED ISSUE #1350: Problem merging subpath routes when openapi.json is generated
### Environment

Nitro 2.5.1
Node.js 18.16.0
Nuxi 3.6.1 

### Reproduction

https://github.com/piscis/nuxt-open-api-test

### Describe the bug

There is a problem when a server route is used multiple times with same parameters but with different methods.

For example if I declare the following routes:

server-routes

Nuxt JS picks up the server routes correct:

nuxt-debugger

But if you look at the swagger UI and openai.json you see that only one of the routes are included:

**swagger UI**
swagger

**openapi.json**
```json


{
  "openapi": "3.0.0",
  "info": {
    "title": "Nitro Server Routes",
    "version": null
  },
  "servers": [
    {
      "url": "http://localhost:3000",
      "description": "Local Development Server",
      "variables": {}
    }
  ],
  "schemes": [
    "http"
  ],
  "paths": {
    "/api/foo/{bar}": {
      "put": {
        "tags": [
          "API Routes"
        ],
        "parameters": [
          {
            "name": "bar",
            "in": "path",
            "required": true
          }
        ],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    },
    "/api/foo": {
      "get": {
        "tags": [
          "API Routes"
        ],
        "parameters": [],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    },
    "/__nuxt_error": {
      "get": {
        "tags": [
          "Internal"
        ],
        "parameters": [],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    },
    "/_nitro/openapi.json": {
      "get": {
        "tags": [
          "Internal"
        ],
        "parameters": [],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    },
    "/_nitro/swagger": {
      "get": {
        "tags": [
          "Internal"
        ],
        "parameters": [],
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    }
  }
}

```

### Additional context

I noticed that changing the parameter part of the route for example from `/api/foo/[id].get.json` to `/api/foo/[idx].get.json` the route will appear because its not registered by a different route.

### Logs

_No response_

Leo-Bourbon:
### 🔗 Linked issue
#1350

### ❓ Type of change

### 📚 Description
In the openapi.ts file, the `paths` property is generated by using `Object.fromEntries` on the result of a map operation on `handlersMeta`. However, this causes unwanted behaviour : when there are sub-paths routes that only differ by their method, only the last path is present, as it has overridden the other paths information in the final `paths` property.

To fix this, I am now creating the paths object and merging the sub-paths using `Array.reduce` on `handlersMeta` instead. The operations inside this reduce are still the same as the ones inside the previous map implementation.
What has changed is that now instead of returning an array of the route and its path, I now check if the accumulated value, (the total paths object), has the current route as a property. If it does not, I add the route property and the corresponding route information. If it does, I merge the current route information with the information already present for this route.

Resolves #1350 
### 📝 Checklist

- [ ] I have updated the documentation accordingly. - N/A

### DIFF
--- a/src/runtime/routes/openapi.ts
+++ b/src/runtime/routes/openapi.ts
@@ -24,54 +24,72 @@ export default eventHandler((event) => {
       },
     ],
     schemes: ["http"],
-    // eslint-disable-next-line unicorn/no-array-reduce
-    paths: handlersMeta.reduce((paths: PathsObject, h) => {
-      const parameters: ParameterObject[] = [];
-
-      let anonymousCtr = 0;
-      const route = h.route
-        .replace(/:(\w+)/g, (_, name) => `{${name}}`)
-        .replace(/\/(\*)\//g, () => `/{param${++anonymousCtr}}/`)
-        .replace(/\*\*{/, "{")
-        .replace(/\/(\*\*)$/g, () => `/{*param${++anonymousCtr}}`);
-
-      const paramMatches = route.matchAll(/{(\*?\w+)}/g);
-      for (const match of paramMatches) {
-        const name = match[1];
-        if (!parameters.some((p) => p.name === name)) {
-          parameters.push({ name, in: "path", required: true });
-        }
-      }
-
-      const tags: string[] = [];
-      if (route.startsWith("/api/")) {
-        tags.push("API Routes");
-      } else if (route.startsWith("/_")) {
-        tags.push("Internal");
-      } else {
-        tags.push("App Routes");
-      }
-
-      const item: PathItemObject = {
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks!

PR #1499, opened 2023-07-28
docs: prefer snake_case for preset names

marshallswain:
### ❓ Type of change

### 📚 Description

This updates the preset for Cloudflare Pages to be named `cloudflare_pages` instead of `cloudflare-pages`.  This brings the name inline with the one on the https://nitro.unjs.io/deploy page and actually works when deploying to Cloudflare Pages.

I have to set `cloudflare_pages` in my `nuxt.config.ts` in order for this preset to work.  I assume that means that the preset should be underscored and not dashed like `cloudflare-pages`.  If I use `cloudflare-pages` instead of `cloudflare_pages`, the `dist` directory does not get created.

### 📝 Checklist

### DIFF
--- a/docs/content/2.deploy/0.index.md
+++ b/docs/content/2.deploy/0.index.md
@@ -20,7 +20,7 @@ When running Nitro in development mode, Nitro will always use a special preset c
 When deploying to the production using CI/CD, Nitro tries to automatically detect the provider environment and set the right one without any additional configuration. Currently, providers below can be auto-detected with zero config.
 
 - [azure](/deploy/providers/azure)
-- [cloudflare_pages](/deploy/providers/cloudflare#cloudflare_pages)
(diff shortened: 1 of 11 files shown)


Hebilicious:
`nitro.preset` set to `cloudflare-pages` works for me. Can you please provide a reproduction for where it doesn't ?

pi0:
Thanks for PR @marshallswain . As @Hebilicious pointed out, there is no difference in two casings because we use [camel case normalization](https://github.com/unjs/nitro/blob/ad5f353e1201557394c13809c57d0f6191ddcc45/src/options.ts#L167) using [unjs/scule](https://github.com/unjs/scule) and it handles both casings...

However i think as a general best practice, underscores are safer to be in docs 👍🏼  (refactoring to reflect other pages too)

pi0 (approved):

PR #1500, opened 2023-07-28
feat(firebase): add support for 2nd generation functions

posva:
### 🔗 Linked issue

Forked from https://github.com/unjs/nitro/pull/1142

Resolves https://github.com/unjs/nitro/issues/1141
Resolves https://github.com/unjs/nitro/issues/666
Resolves https://github.com/unjs/nitro/issues/90

This PR adds a `fireabse.gen` option instead of `firebase.v2` of the original PR to allow customizing **gen 2 and gen 1** functions. It also allows passing other options for the functions. This assumes you are only using one generation at the same time (which, outside of migrations should be the case).

### ❓ Type of change

### 📝 Checklist


docs/content/2.deploy/providers/firebase.md
```diff
@@ -8,9 +8,15 @@ Nitro supports [Firebase Hosting](https://firebase.google.com/docs/hosting) with
 
 **Note**: You need to be on the **Blaze plan** to use Nitro with Cloud Functions.
 
-If you don't already have a `firebase.json` in your root directory, Nitro will create one the first time you run it. In this file, you will need to replace `<your_project_id>` with the ID of your Firebase project.
+**Note**: This preset will deploy to firebase functions 1st gen by default. If you want to deploy to firebase functions 2nd gen, see the [instructions below](#).
```
posva, line 11:
```suggestion
**Note**: This preset will deploy to firebase functions 1st gen by default. If you want to deploy to firebase functions 2nd gen, see the [instructions below](#using 2nd-generation-firebase-functions).
```

docs/content/2.deploy/providers/firebase.md
```diff
@@ -8,9 +8,15 @@ Nitro supports [Firebase Hosting](https://firebase.google.com/docs/hosting) with
 
 **Note**: You need to be on the **Blaze plan** to use Nitro with Cloud Functions.
 
-If you don't already have a `firebase.json` in your root directory, Nitro will create one the first time you run it. In this file, you will need to replace `<your_project_id>` with the ID of your Firebase project.
+**Note**: This preset will deploy to firebase functions 1st gen by default. If you want to deploy to firebase functions 2nd gen, see the [instructions below](#using-2nd-generation-firebase-functions).
```
pi0, line 11: Note for myself: banner to be added about nitro-edge before merging, until next release

docs/content/2.deploy/providers/firebase.md
```diff
@@ -20,28 +26,44 @@ This file should then be committed to version control. You can also create a `.f
 }
 ```
 
-Then, just add Firebase dependencies to your project:
+#### Install Firebase dependencies
+
+Then, add Firebase dependencies to your project:
+
+yarn:
```
pi0, line 33: We shall use `::code-group` (feature or docus to display multi pms)

package.json
```diff
@@ -76,6 +76,7 @@
     "esbuild": "^0.18.16",
     "escape-string-regexp": "^5.0.0",
     "etag": "^1.8.1",
+    "firebase-functions": "^4.4.1",
```
posva, line 79: Should this be moved to devDependencies to have the correct types but still let the user install their own version of firebase-functions?

docs/content/2.deploy/providers/firebase.md
```diff
@@ -82,11 +104,86 @@ NITRO_PRESET=firebase yarn build
...
+
+You can set options for the firebase functions in your `nitro.config.ts` file:
+
+```ts
+import { defineNitroConfig } from 'nitropack/config'
+
+export default defineNitroConfig({
+  // Nitro options...
+  firebase: {
+    gen2: {
```
pi0, line 145: Just double checking, is `genX` a known term for firebase team/docs that we use here? (i love it btw!)

docs/content/2.deploy/providers/firebase.md
```diff
@@ -82,11 +104,86 @@ NITRO_PRESET=firebase yarn build
...
+You can also set 1st gen options in the `gen1` key.
+
+## Runtime
+
+Firebase uses the `engines.node` version in your `package.json` to determine which node version to use for your functions. If you want to use a different version of node for your functions, you should set the `engines.node` version in your `package.json` to the version you want to use:
+
+```json [package.json]
+{
+  "engines": {
+    "node": "18"
```
pi0, line 179: This field is also used for Node version of building nitro app itself. Can't we purely depend on `firebase.json` to configure runtime node? What is purpose of defined node version here? (if we really need, it should be _generated_ in `.output/package.json` and picked somehow by cli)

posva:
@JamieCurnow It was really hard for me to create a code review with all of this so I created a fork of your PR while keeping all the commits. Let me know what you think 🙂

docs/content/2.deploy/providers/firebase.md
pi0: We can also make types smart by supporting `generation: 1|2` to switch types for rest of the keys. This way we don't need to make env required.

src/presets/firebase.ts
```diff
@@ -8,15 +8,28 @@ import { writeFile } from "../utils";
...
+
+if (firebaseFunctionsGen !== "1" && firebaseFunctionsGen !== "2") {
+  throw new Error(
+    `FIREBASE_FUNCTIONS_GEN must be 1 or 2, got "${firebaseFunctionsGen}"`
+  );
+}
+
 export const firebase = defineNitroPreset({
-  entry: "#internal/nitro/entries/firebase",
+  entry: `#internal/nitro/entries/firebase-gen-${firebaseFunctionsGen}`,
```
pi0, line 21: Tip: We can read it from input config too. Just make preset a factory function!

package.json
pi0: Yes that seems good. (later with dedicated package, we can try to auto install however explicit dep is always better)

docs/content/2.deploy/providers/firebase.md
posva: > Just double checking, is genX a known term for firebase team/docs that we use here? (i love it btw!)

They actually use _nth gen_ like 1st gen but I thought that having the number instead of letters was easier to spot 😅 . Docs with terms: https://firebase.google.com/docs/functions/version-comparison

> We can also make types smart by supporting generation: 1|2 to switch types for rest of the keys. This way we don't need to make env required.

That's pretty cool but I don't know how to add it 😆

docs/content/2.deploy/providers/firebase.md
posva: It's not mandatory, I have projects where I don't set it. This part was already added by @JamieCurnow so they will probably know better

src/presets/firebase.ts
posva: aah, much better, thanks!

docs/content/2.deploy/providers/firebase.md
pi0: Good to know abouth nth! Example for selective type: https://github.com/unjs/unbuild/blob/main/src/types.ts#L41

docs/content/2.deploy/providers/firebase.md
pi0: Okay let's remove it then to simplify and then iterate over to somehow introduce if needed

src/presets/firebase.ts
posva: How do I read the nitro options from the factory function? It seems to not have any argument

docs/content/2.deploy/providers/firebase.md
JamieCurnow: @pi0 this is used by firebase to determine the runtime of node to run the cloud function in. It is required as far as I know / as far as I can tell from the docs

https://firebase.google.com/docs/functions/manage-functions?gen=2nd#set_nodejs_version

> The engines field is required; it must specify one of the supported Node.js versions in order for you to deploy and run functions

src/presets/firebase.ts
pi0: Oh right sorry i forgot presets are layers. We have an alternative. `entry` also supports special templates ([src](https://github.com/unjs/nitro/blob/fddf57e50ed4bd4863a23d80c86645bcea79b884/src/utils/index.ts#L72)). Passing `entry: #internal/nitro/entries/firebase-gen-{{ firebase.gen }}` should do the trick (please note that we need to define default gen in `options.ts` defaults if there is no value, template is not replaced). I will improve it afterwards

docs/content/2.deploy/providers/firebase.md
JamieCurnow: The nitro preset for firebase will default to node 18 and warn of incompatible versions

https://github.com/unjs/nitro/pull/1500/files/93d49aab552a9d22449d581045047da97213edb8#diff-10ef001537e49ef8e8387ba0cf14acbdbcef1cc58a1aebfad7e93e28073c1274R81

pi0: From docs:

> Set the version in the engines field in the package.json file that was created in your functions/ directory during initialization. 

I think we can generate `package.json` with desired engines field in `.output/` directory to specify (runtime) node.

src/presets/firebase.ts
Hebilicious: Can we set a default value in the preset directly ?

docs/content/2.deploy/providers/firebase.md
JamieCurnow: @pi0 Yea it definitely reads the generated package.json in the output directory. I guess we could introduce it as a preset option instead of reading it from the root package.json as it does now?

src/presets/firebase.ts
pi0: Please try. Yes we might if c12 merges correctly.

docs/content/2.deploy/providers/firebase.md
pi0: Preset option for runtime node seems nice idea 💯  (still, i am wondering is it a required option as posva mentioned seems not and do we want to overide firebase defaults? it has some benefits like knowing our build target..)

JamieCurnow: @pi0 it would make sense to me that it should be the same node version that nitro uses to build. Do we have access to that version at all?

JamieCurnow: I'm guessing a developer may experience issues if nitro builds on node 18 and firebase deploys on node 14 for example 😅

pi0: It doesn't matter what (latest) version of node we use. But it is important what transforms `esbuild` applies to the `.output` files as that matters and that would be the main benefit of knowing runtime node version (in `firebase.nodeVersion`) for possible syntax transforms and polyfills when needed.

package.json
JamieCurnow: @posva @pi0 I did try adding this as a dev dep, but came up against some build errors that were way above my head. See this comment in the original PR:
https://github.com/unjs/nitro/pull/1142#discussion_r1249622185

pi0: Strange. I think it can be a regular dependency for users too. We really need to try merging one pr and try end-to-end via edge channel to resolve such possible issues.

JamieCurnow:
@posva I've made a PR to your branch on your fork with the suggestions above:
https://github.com/posva/nitro/pull/1

Hopefully this is the best way to do this? 😅

pi0:
Thanks for quick PR @JamieCurnow and all followup ❤️ 

Let's hope we get out of this situation soon and keep improving preset directly on one main branch 😅 

PS: I have sent you an email regarding to join discord server to collaborate closely.

docs/content/2.deploy/providers/firebase.md
posva: I think it's better to keep this instruction. One can also set the `functions.runtime` option in `firebase.json` but that seems less documented and it's just more convenient setting it in package.json.

docs/content/2.deploy/providers/firebase.md
```diff
@@ -78,15 +96,102 @@ You can find more details in the [Firebase documentation](https://firebase.googl
 You can preview a local version of your site if you need to test things out without deploying.
 
 ```bash
-NITRO_PRESET=firebase yarn build
+NITRO_PRESET=firebase npm run build
```
posva, line 99: In some places I just used npm to run commands to not add to much noise about switching the package manager each time as people should know what they use anyway

src/types/presets.ts
```diff
@@ -52,6 +55,38 @@ export interface VercelBuildConfigV3 {
   }[];
 }
 
+interface FirebaseBaseOptions {
+  gen: 1 | 2;
+  /**
+   * Firebase functions node runtime version.
+   * @see https://cloud.google.com/functions/docs/concepts/nodejs-runtime
+   */
+  nodeVersion?: "20" | "18" | "16" | "14" | "12" | "10";
```
JamieCurnow, line 64: To go with @pi0 comment [here](https://github.com/unjs/nitro/pull/1500/files/a0b602dc2662ffc09e149d85eeb9c00a417cdd83#r1278988243)

```suggestion
  nodeVersion?: "20" | "18" | "16";
```

src/presets/firebase.ts
```diff
@@ -78,8 +69,8 @@ async function writeRoutes(nitro: Nitro) {
     return obj;
   }, {} as Record<string, string>);
 
-  let nodeVersion = "18";
-  const supportedNodeVersions = new Set(["18", "16", "14", "12", "10"]);
+  let nodeVersion: string = nitro.options.firebase?.nodeVersion || "18";
+  const supportedNodeVersions = new Set(["20", "18", "16", "14", "12", "10"]);
```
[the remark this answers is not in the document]
JamieCurnow:
```suggestion
  const supportedNodeVersions = new Set(["20", "18", "16"]);
```

JamieCurnow (changes requested):
@posva Just adding some of the suggestions from @pi0 here. I've realised that we need to also update the doc changes now to account for the `gen` selection and to note that the node version can be pinned via the new `nodeVersion` key. I'm happy to make these changes - may be easier to add me as a collaborator to your fork? Otherwise, let me know and I'll make a PR to your fork again if you want me to jump on this 🙏

docs/content/2.deploy/providers/firebase.md
```diff
@@ -20,33 +26,47 @@ This file should then be committed to version control. You can also create a `.f
...
-You may instead prefer to set up your project with the Firebase CLI, which will fetch your project ID for you, add required dependencies (see above) and even set up automated deployments via GitHub Actions.
+### Log into the firebase cli
 
-#### Install Firebase CLI globally
+Make sure you are authenticated with the firebase cli. Run this command and follow the prompts:
 
 ```bash
-# yarn
-yarn global add firebase-tools
+npx login
```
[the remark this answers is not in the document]
posva: oops, it was meant to be firebase-tools login

src/options.ts
```diff
@@ -119,6 +119,11 @@ const NitroDefaults: NitroConfig = {
   nodeModulesDirs: [],
   hooks: {},
   commands: {},
+
+  // Presets
+  firebase: {
+    gen: 1,
```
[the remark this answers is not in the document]
posva: totally, it would also mean the warning to let users know to update won't go off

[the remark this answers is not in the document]
posva: The question is will it work with the template in `entry`?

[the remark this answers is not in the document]
posva: ah, nvm: https://github.com/unjs/nitro/pull/1500#discussion_r1277907600
It has to be set, although I need a way to warn the user to set the `gen` property if they haven't set it 🤔

src/options.ts
```diff
@@ -122,7 +122,9 @@ const NitroDefaults: NitroConfig = {
 
   // Presets
   firebase: {
-    gen: 1,
+    // we need this defined here so it's picked up by the template in firebase's entry
+    // @ts-expect-error: internal value to detect if the user is defining a `gen` property
+    gen: 'default',
```
pi0: Maybe `latest` as default?

docs/content/2.deploy/providers/firebase.md
pi0: Build time Node.js version !== Runtime Node.js version, so no i don't think it is a good idea to document it this way.

src/options.ts
posva: I don't think `latest`, specifically, is a good idea is the default is gen 1. what about using `null` instead?

docs/content/2.deploy/providers/firebase.md
posva: In the case of firebase projects, it is, since you run emulators using a specific node of version locally.

pi0 (approved):

pi0:
Thanks you all for contributing on this preset ❤️ It looks amazing! To move forward i have made some minor refactors as well as docs update about development. Feel free to try it and make PRs directly into the main branch.

snsxn:
Hi!
I see in the Nitro docs the new 2nd gen functions, region and node version.
is it released?

Thx!!

pi0:
Hi dear @snsxn. you can try with edge channel: https://nitro.unjs.io/guide/getting-started#edge-release-channel

PR #1519, opened 2023-08-01
feat: add `prerender:config`, `prerender:init` and `prerender:done` hooks

harlan-zw:
### 🔗 Linked issue

Related https://github.com/nuxt/nuxt/issues/22255

### ❓ Type of change

### 📚 Description

An initial step to giving greater prerendering powers to integrations.

Some examples of what this unblocks:
- emit manifests using data collected during the prerender stage
- improved error reporting for routes that fail
- stubbing in modules that are only for the prerender stage

Out of scope of this PR but still to do be done:
- Close the Prerenderer when it finishes processing all routes
- Prenderer API exposed on hooks (generateRoute, queueFinished, etc)

### 📝 Checklist


src/prerender.ts
```diff
@@ -49,13 +49,16 @@ export async function prerender(nitro: Nitro) {
   // Build with prerender preset
   nitro.logger.info("Initializing prerenderer");
   nitro._prerenderedRoutes = [];
+  const prerendererConfig = { ...nitro.options._config };
```
pi0, line 52: I think would be nice to pass the merged object (with static, rootDir, preset, logLevel) to this object to allow full control within hook

danielroe (commented):
Beautiful feature idea. I agree with @pi0's comment, but otherwise 👌

Hebilicious (commented):
I love it!
Regarding the hook names, we use the before/done terminology for modules and build, should we use this as a convention ? Or maybe define a new one such as init/finished  or before/start ?

src/prerender.ts
```diff
@@ -248,7 +248,7 @@ export async function prerender(nitro: Nitro) {
     interval: nitro.options.prerender.interval,
   });
 
-  await nitro.hooks.callHook("prerenderer:finished", {
+  await nitro.hooks.callHook("prerenderer:close", {
```
pi0, line 251: `:done` hook would be better. By convention in nuxt and unjs we use `close` for disposal step when closing resources.

harlan-zw:
> I love it! Regarding the hook names, we use the before/done terminology for modules and build, should we use this as a convention ? Or maybe define a new one such as init/finished or before/start ?

Thanks :) I don't really like before/done as the context is around the prerender app itself so it seems less semantic.

I used `init` for consistency with the nuxt hook `nitro:init` and the log just before it.

I didn't really like `prerenderer:finished`, I've gone with `prerenderer:close`. Only slightly misleading as nothing is closed. This raises a slightly unrelated question though, should we be closing the prerenderer app once it is finished (after the hook)?

```ts
 await nitro.hooks.callHook("prerenderer:close", {
    nitro,
    prerenderer: nitroRenderer,
    erroredRoutes,
    generateRoute,
  });
  await nitroRenderer.close()
```

pi0:
Yes closing seems a good idea (it will automatically call close hook afterwards)

src/prerender.ts
```diff
@@ -245,6 +248,13 @@ export async function prerender(nitro: Nitro) {
     interval: nitro.options.prerender.interval,
   });
 
+  await nitro.hooks.callHook("prerenderer:done", {
+    nitro,
+    prerenderer: nitroRenderer,
+    erroredRoutes,
+    generateRoute,
```
pi0, line 255: Allowing `generateRoute` here is kinda tricky. Because we are declaring prerendering is "done" to "all" listeners and with final errors.

We might expose API in other different ways. One way would be to pass it to second arg of `prerender:init(prerenderer, { generateRoute }`.

I understand this is final goal of your PR and init hook might not be best fitting as well. Because of this i would suggest to discuss and introduce this in next PR (we could think of a prerender controller object being provided earlier)

src/prerender.ts
```diff
@@ -245,6 +248,13 @@ export async function prerender(nitro: Nitro) {
     interval: nitro.options.prerender.interval,
   });
 
+  await nitro.hooks.callHook("prerenderer:done", {
+    nitro,
+    prerenderer: nitroRenderer,
```
pi0, line 253: For consistency, we shall pass `prerenderer` as first arg. Also main `nitro` instance is already accessible we don't need to pass it here.

harlan-zw: I did think this, but then the `prerender:generate` hook expose a nitro instance as well so wasn't sure the convention.

src/prerender.ts
harlan-zw: Make sense if we're renaming it and exposing a prerender controller seems like a good DX

src/prerender.ts
harlan-zw: I added the `prerenderedRoutes` if we're going in this direction, I think it's better than accessing it off the nitro instance with the internal flag

src/prerender.ts
harlan-zw: To clarify, I don't think it's worth adding generateRoute on the init hook, better to just wait until the prerenderer controller is available

pi0: Let's do this then! (feel free to open a new discussion issue to not forget)

pi0 (approved):
LGTM 💯 (have to fix failing ci, it needs parent dir creation)

harlan-zw:
> LGTM 100 (have to fix failing ci, it needs parent dir creation)

To fix this properly it would need access to the main nitro instance (not the prerenderer) for the real output directories :thinking: A native `nitro:init` hook that exposes the nitro instance would be a way to solve this without having to add the nitro instance to every hook.

For now, I just removed the test.

pi0 (approved):
Thanks again for the PR. I have pushed some relevant refactors on top of it hope it is all fine to you.

Issue #1480, opened 2023-07-23
Possible pre-rendering memory leak


Velka-DEV:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.17.0`
- Nuxt Version:     `3.6.3`
- Nitro Version:    `2.5.2`
- Package Manager:  `pnpm@8.6.9`
- Builder:          `vite`
- User Config:      `hooks`, `nitro`, `runtimeConfig`, `css`, `app`, `postcss`, `experimental`, `devtools`, `extends`, `modules`, `googleFonts`, `swiper`, `schemaOrg`, `image`, `i18n`, `sitemap`
- Runtime Modules:  `@nuxtjs/google-fonts@3.0.1`, `nuxt-swiper@1.2.0`, `@nuxtjs/i18n@8.0.0-beta.10`, `@nuxt/image@1.0.0-rc.1`, `nuxt-schema-org@2.2.0`, `nuxt-simple-sitemap@3.0.9`
- Build Modules:    `-`

### Reproduction

Just need to pre-render high number of pages (in my case > 11k)

### Describe the bug

When generating an high number of pages, the prerender engine increase progressively to go out of memory quickly.

(PS: Running on a MacBook pro M2 Max 32Go RAM, process fully fill the RAM)

### Logs

```sh
<--- Last few GCs --->

[10962:0x158078000]   108040 ms: Mark-sweep 3536.2 (4131.9) -> 3519.4 (4129.9) MB, 73.6 / 0.0 ms  (average mu = 0.306, current mu = 0.303) allocation failure; scavenge might not succeed
[10962:0x158078000]   108148 ms: Mark-sweep 3539.0 (4132.7) -> 3521.6 (4130.7) MB, 76.2 / 0.0 ms  (average mu = 0.302, current mu = 0.297) allocation failure; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x100f466e4 node::Abort() [/usr/local/bin/node]
 2: 0x100f468c8 node::ModifyCodeGenerationFromStrings(v8::Local<v8::Context>, v8::Local<v8::Value>, bool) [/usr/local/bin/node]
 3: 0x10109d60c v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
[...]

PR #1529, opened 2023-08-04
perf: reduce _prerenderRoutes memory footprint

Hebilicious:
### 🔗 Linked issue

Related https://github.com/unjs/nitro/issues/1535
Related https://github.com/unjs/nitro/issues/1480
Related https://github.com/nuxt/nuxt/issues/22449

### ❓ Type of change

### 📚 Description

Based on @simkuns patch. Internal `_prerenderedRoutes` now only contains route and fileName to reduce the memory footprint

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -24,6 +24,7 @@ export async function prerender(nitro: Nitro) {
 
   // Initial list of routes to prerender
   const routes = new Set(nitro.options.prerender.routes);
+
   // Extend with static prerender route rules
   const prerenderRulePaths = Object.entries(nitro.options.routeRules)
     .filter(([path, options]) => options.prerender && !path.includes("*"))
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -7,7 +7,7 @@ import { defu } from "defu";
 import { createNitro } from "./nitro";
 import { build } from "./build";
 import type { Nitro, NitroRouteRules, PrerenderGenerateRoute } from "./types";
-import { writeFile } from "./utils";
+import { RouteSet, writeFile } from "./utils";
 import { compressPublicAssets } from "./compress";
 
 const allowedExtensions = new Set(["", ".json"]);
@@ -23,8 +23,7 @@ export async function prerender(nitro: Nitro) {
   }
 
   // Initial list of routes to prerender
-  const routes = new Set(nitro.options.prerender.routes);
-
+  const routes = RouteSet(nitro.options.prerender.routes);
   // Extend with static prerender route rules
   const prerenderRulePaths = Object.entries(nitro.options.routeRules)
     .filter(([path, options]) => options.prerender && !path.includes("*"))
@@ -39,7 +38,7 @@ export async function prerender(nitro: Nitro) {
   }
 
   // Allow extending prereneder routes
-  await nitro.hooks.callHook("prerender:routes", routes);
+  await nitro.hooks.callHook("prerender:routes", routes.getSet());
 
   // Skip if no prerender routes specified
   if (routes.size === 0) {
@@ -201,7 +200,10 @@ export async function prerender(nitro: Nitro) {
 
     const filePath = join(nitro.options.output.publicDir, _route.fileName);
     await writeFile(filePath, Buffer.from(_route.data));
-    nitro._prerenderedRoutes.push(_route);
+    nitro._prerenderedRoutes.push({
+      route: _route.route,
+      fileName: _route.fileName,
+    });
 
     // Crawl route links
     if (!_route.error && isImplicitHTML) {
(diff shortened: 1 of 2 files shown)


danielroe:
Nice! ❤️

parzychu:
Great! <3

src/prerender.ts
```diff
@@ -213,7 +215,11 @@ export async function prerender(nitro: Nitro) {
       );
       for (const _link of extractedLinks) {
         if (canPrerender(_link)) {
-          routes.add(_link);
+          if (_link.endsWith("_payload.json")) {
```
pi0, line 218: sorry but i cannot let this to be merged in nitro. please consider proposing framework agnostic solutions AND make sure there is a nitro minimal reproduction that makes sure it actually proves and solves an issue instead of hiding root causes.

pi0 (changes requested):
Requirements:
- Framework agnostic implementation
- Nitro minimal reproduction to make sure there is an issue with Nitro itself, not downstream framework (Nuxt) or modules. 

As a Nuxt team member and nitro maintainer, I tend to fix the root cause of issues not patching them around.

src/prerender.ts
```diff
@@ -201,7 +200,10 @@ export async function prerender(nitro: Nitro) {
 
     const filePath = join(nitro.options.output.publicDir, _route.fileName);
     await writeFile(filePath, Buffer.from(_route.data));
-    nitro._prerenderedRoutes.push(_route);
+    nitro._prerenderedRoutes.push({
```
danielroe, line 203: For me, this is the key contribution of this PR and it is a genuine memory leak in nitro. Caching _all_ results of all prerendered routes will of course cause OOM issues.

pi0: Yes, that alone can be a good fix 👍🏼 (not reordering `_payload.json`).

Hebilicious:
Thanks for the feedback @pi0 

I'm open to suggestions for a framework agnostic design. 
What if we accept a regex from the settings like`nitro.options.prerender.prependLinkRegex`?
This can then be used by Nuxt to achieve such a behaviour.

src/prerender.ts
danielroe: Agreed. Nuxt can hook into prerender hooks hopefully to do the same.

pi0:
(i am working on reproduction and alternative fix for content issue)

@Hebilicious Can you please explain more why we would need to reorder prerender routes?

Hebilicious:
> (i am working on reproduction and alternative fix for content issue)
> 
> @Hebilicious Can you please explain more why we would need to reorder prerender routes and how it relates to a possible fix?

This re-ordering was based on this comment that included a nice heap profile.
https://github.com/unjs/nitro/issues/1480#issuecomment-1664176680

However, I think its better to expose a hook (before we run `runParallel` so that frameworks (nuxt, modules ...) can inspect and re-order routes if they wish to do so.

For testing we could test the perf differences with https://github.com/nuxt/nuxt/pull/22465 and https://github.com/unjs/nitro/pull/1529 and see how much re-ordering would impact things.

src/prerender.ts
```diff
@@ -201,7 +201,10 @@ export async function prerender(nitro: Nitro) {
...
+      route: _route.route,
+      fileName: _route.fileName,
```
danielroe, line 206: We could add synchronous getters that reads from the fs to access the properties we are removing, to avoid breaking integrations that rely on this API.

pi0: Nice idea. Please see my comments in https://github.com/unjs/nitro/pull/1536, it makes sure data and contents are available until first read. I would be more than happy to add compatibility sync read if an integration needs contents after the extend hook (and write). Didn't add it initially to avoid unwanted performance overhead of sync operation (specially that it is a hidden getter and to avoid memory leaks, we cannot cache return value. it should be sync read each time accessing getter which is costly (and hidden!)

danielroe: Not sure whether to comment here or there, but I think it is indeed worth adding a getter (and a console log deprecating the properties or warning about hidden host of access?) until a future version as otherwise it is a breaking behaviour change.

pi0:
Thanks for PR and explanations @Hebilicious and pin pointing the exact issue @danielroe  i have made a better rework in https://github.com/unjs/nitro/pull/1536 which seems to fix (reproducable) issue.

@Hebilicious I would recommend to try nuxt edge + nitro edge with both fixes (LRU cache and #1536) and rerun any previous benchmarks again to make sure sorting with fixes makes a difference. If it does, we already have `prerender:routes(routes)` hook that get's called before parallel run and can be used to reorder. We might think of another hook too `prerender:extractedlinks` (or similar that is called inside `generateRoute`) if sorting of later discovered. I think generally would be a nice addition but again, please test.

src/prerender.ts
pi0: (feel free to open an issue to not forget), checking you are right `prerender:route` will be broken but i guess we can fix it by reordering where the hook is being called to avoid adding that workaround. This is only hook that is currently being called before disk write.

pi0: https://github.com/unjs/nitro/pull/1537 should fix it for that hook.

PR #1162, opened 2023-04-17
feat: experimental `/_nitro/openapi.json` and `/_nitro/swagger` endpoints for dev mode

pi0:
### 🔗 Linked issue

Resolves #3 
Resolved #4

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR exposes available routes and an initial meta using [Open API 3 standard](https://www.openapis.org/).

It exposes two endpoints for development:
- `/_nitro/openapi.json`: JSON response that can be integrated with external tools and setups such as Postman and Nuxt devtools to access available routes meta (/cc @antfu)
- `/_nitro/swagger`: A basic UI that makes local debugging easier

This feature opens the next steps in supporting generic and extendable route meta via a well-defined standard.

image

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1532, opened 2023-08-04
feat: generate InternalFetch signature

danielroe:
### ❓ Type of change

### 📚 Description

This PR reworks how our internal API routes are typed. This aims to do more of the work at build time rather than asking typescript to do it. Testing in our fixture showed type-checking speed improvement of up of 45%.

**Note**: The _reason_ that this is so much faster is that we use function overloads, which means that we do not yet support all the matching patterns that our existing type-only algorithm does. It matches _only one_ result so if you have multiple overlapping patterns, like `/api/test/**/thing` and `/api/test/bob/thing` then if you request `$fetch('/api/test/${someUnknownString}')` then we won't be able to give a union of the two return times as before.

It adds the following type-safety functionalities:

- [x] it will throw a type error if you make a request that needs a specific method without an options object with that method defined
- [x] it will throw a type error if you make a request that needs specific input types without those values
- [x] type error if you make a request that needs specific input types without an options object
- [x] when `$fetch.create` is called with a baseURL, the result is not strongly typed but just a normal external `$fetch`

Further enhancements needed:
- [ ] auto-completion when used with endpoints with typed inputs (this would be very desirable for better DX)

### 📝 Checklist


src/build.ts
```diff
@@ -135,6 +185,11 @@ export async function writeTypes(nitro: Nitro) {
   const routes = [
     "// Generated by nitro",
     "import type { Serialize, Simplify } from 'nitropack'",
+    "import type { EventHandler, HTTPMethod } from 'h3'",
+    "import type { FetchOptions } from 'ofetch'",
+    "type DefaultMethod = HTTPMethod | Lowercase<Exclude<HTTPMethod, 'PATCH'>>",
```
DrJume, line 190: `Exclude<HTTPMethod, 'PATCH'> | Lowercase<Exclude<HTTPMethod, 'PATCH'>>` ?

danielroe: I think we still want `PATCH` to be an allowable method.

src/build.ts
```diff
@@ -77,29 +77,79 @@ export async function build(nitro: Nitro) {
...
+    const methodType = mw.method
+      ? [mw.method.toUpperCase(), mw.method.toLowerCase()]
+          .filter((m) => m !== "patch")
+          .map((m) => `'${m}'`)
+          .join(" | ")
+      : defaultMethod;
+
+    // TODO: 1. Fine-tune matching algorithm?
+    // TODO: merge return types
+    // TODO: 2. infer returns when we provide typed input
```
DrJume, line 138: Partial Inference is only beginning to be worked on in 5.2 (https://github.com/microsoft/TypeScript/pull/26349). 

Currently, identity functions are used to get this behaviour, providing sth like a "type closure": 
- https://medium.hexlabs.io/partial-type-inference-f669ae24c44e
- https://medium.com/@nandin-borjigin/partial-type-argument-inference-in-typescript-and-workarounds-for-it-d7c772788b2e

danielroe: Yes, this may only be possible with the new object-syntax event handlers: https://github.com/unjs/h3/pull/485

manniL:
> **Note**: The _reason_ that this is so much faster is that we use function overloads, which means that we do not yet support all the matching patterns that our existing type-only algorithm does. It matches _only one_ result so if you have multiple overlapping patterns, like `/api/test/**/thing` and `/api/test/bob/thing` then if you request `$fetch('/api/test/${someUnknownString}')` then we won't be able to give a union of the two return times as before.

Will this be possible in the future with the function overloading approach?

septatrix:
Will it be possible with the way it is implemented in this PR to augment the OpenAPI documentation #1162?

ImBoop:
This PR has been outstanding as a draft for a year now, are there any updates you can share?

pi0:
This is taking longer than expected because we are working on a better standard ecosystem-wise and better testable solution. 

I have created a tracker issue: https://github.com/unjs/nitro/issues/2758

(closing this one mainly for triage, lets keep track via issue)

PR #1533, opened 2023-08-04
test: add `~` alias to fixture

danielroe:
### ❓ Type of change

### 📚 Description

In working on https://github.com/unjs/nitro/pull/1532, I noticed we have an api route that uses an alias that doesn't exist.

### 📝 Checklist

### DIFF
--- a/test/fixture/nitro.config.ts
+++ b/test/fixture/nitro.config.ts
@@ -1,6 +1,6 @@
+import { fileURLToPath } from "node:url"
 import { dirname } from "pathe"
 import { defineNitroConfig } from "../../src/config";
-import { fileURLToPath } from "node:url"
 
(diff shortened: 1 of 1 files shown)


pi0:
It is strange that `~` is not working... (will double check locally)

pi0:
Alias exists

https://github.com/unjs/nitro/blob/1e134c1f05f5bd646ef209c929ccc5db2cf5c881/src/rollup/config.ts#L331

Creating `testfixture/routes/index.ts`:

```ts
export { default } from "~/routes/modules";
```

`pnpm nitro dev test/fixture` works! (but types are broken).

danielroe:
the issue is that we don't resolve aliases in writing types. I have resolved the issue in the type PR.

Issue #1469, opened 2023-07-21
Retry failed routes during prerendering


rudolfbyker:
### Describe the feature

When prerendering a large number of routes, it would be very beneficial to be able to automatically retry failed routes. Routes may fail spuriously due to a myriad of reasons, especially temporary 5XX errors from APIs. If a single page out of a few thousand has this problem, everything needs to be prerendered from scratch again.

Proposed config change: Add something like `retries: number` here https://github.com/unjs/nitro/blob/14670c94d8f11f24525295f55ed427ac616fc643/src/types/nitro.ts#L252C8-L252C8 . 0 will give the current behaviour. 1 will give it two attempts before failing. -1 will let it retry each route indefinitely.

### Additional information

- [ ] Would you be willing to help implement this feature?

rudolfbyker:
Better config proposal: Add `tries: number` here https://github.com/unjs/nitro/blob/14670c94d8f11f24525295f55ed427ac616fc643/src/types/nitro.ts#L252C8-L252C8 . `1` will give the current behaviour. `2` will give it two attempts before failing. `0` will let it retry each route indefinitely.

I think this is less confusing than calling it `retries` and shifting everything by one.

PR #1534, opened 2023-08-04
feat(prerender): support `retry` and `retryDelay`

Hebilicious:
### 🔗 Linked issue

Resolves https://github.com/unjs/nitro/issues/1469

### ❓ Type of change

### 📚 Description

The prerender by default will fail if a page fails to render. This PR changes the default behaviour to 3 attempts before failure. 
This also adds 2 configuration options `nitro.options.prerender.tries` and `nitro.options.prerender.retryDelay` to change the behaviour. `tries` can be set to`Inifinity`if someones wants the pre-render to try endlessly.

For the default values, I've arbitrarily chose 2 retries and a 250ms retry delay.
### Additional thoughts

We could nest the options: `nitro.options.prerender.retry.attempts` and `nitro.options.prerender.retry.delay`.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -179,7 +179,6 @@ export async function prerender(nitro: Nitro) {
     // Data will be removed as soon as written to the disk
     let dataBuff: Buffer | undefined = Buffer.from(await res.arrayBuffer());
 
-    _route.data = await res.arrayBuffer();
     Object.defineProperty(_route, "contents", {
       get: () => {
         return dataBuff ? dataBuff.toString("utf8") : undefined;
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -172,8 +172,8 @@ export async function prerender(nitro: Nitro) {
       withBase(encodedRoute, nitro.options.baseURL),
       {
         headers: { "x-nitro-prerender": encodedRoute },
-        retry: nitro.options.prerender.retries ?? 2, // By default we retry 2 times
-        retryDelay: nitro.options.prerender.retryDelay ?? 250,
+        retry: nitro.options.prerender.retries ?? 2, // default is 2 retries
+        retryDelay: nitro.options.prerender.retryDelay ?? 250, // default is 250ms
       }
     );
     // Data will be removed as soon as written to the disk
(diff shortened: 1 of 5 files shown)


pi0:
Since we use ofetch, i am wondering if we can make use of [retry feature](https://github.com/unjs/ofetch#%EF%B8%8F-auto-retry). Also was thinking to call it `retry` / `retries` because it is additional attempts (and 0 means once as default). 0 tries has no meaning.

Hebilicious:
> Since we use ofetch, i am wondering if we can make use of [retry feature](https://github.com/unjs/ofetch#%EF%B8%8F-auto-retry). Also was thinking to call it `retry` / `retries` because it is additional attempts (and 0 means once as default). `tries: 0` has no meaning.

The ofetch implementation doesn't support a configurable delay. 
I will refactor to retries instead of tries.

pi0:
> The ofetch implementation doesn't support a configurable delay. Should I use it regardless ?

I am positive to include it upstream. PR also welcome 👍🏼

Hebilicious:
Included here https://github.com/unjs/ofetch/pull/262

rudolfbyker:
Thanks for working on this!

Just my two cents about `tries` vs `retries`: https://github.com/unjs/nitro/issues/1469#issuecomment-1659995986 Indeed, `tries: 0` does not make sense. That should be an error, just like `tries: -1` or `tries: "foo"`. However, it's still less confusing to have `tries: 1`, `tries: 2`, etc. than `retries: 0`, `retries: 1`, etc. If you feel this is nit-picking, just ignore me :)

Hebilicious:
@pi0 I've rebased this with latest ofetch 👍🏽

src/prerender.ts
```diff
@@ -164,13 +167,15 @@ export async function prerender(nitro: Nitro) {
...
     // Fetch the route
     const encodedRoute = encodeURI(route);
-    const res = await (localFetch(
+
+    const res = await localFetch<Response>(
       withBase(encodedRoute, nitro.options.baseURL),
       {
         headers: { "x-nitro-prerender": encodedRoute },
+        retry: nitro.options.prerender.retries ?? 2, // default is 2 retries
+        retryDelay: nitro.options.prerender.retryDelay ?? 250, // default is 250ms
```
warflash, line 176: Looks like the fallback retry amount and delay don't match the default values stated in the docs

Hebilicious: done

pi0 (approved):
❤️

PR #1554, opened 2023-08-09
perf: bundle runtime dependencies

pi0:
### ❓ Type of change

### 📚 Description

Nitro output uses file tracing for all node modules by default. It has many benefits such as supporting native modules and the best compatibility with minimized rollup involvement.

However, there are downsides with externals too at least for built-in and standard unjs packages that inlining is better.

- Issues with package hosting and multi-version handling
- we can optimize tree-shaking output up to file level instead of function exports 
- Extra data such as `package.json` files are copied over the output directory
- Extra startup overhead because of `node_modules` resolution
- Possible wrong export condition resolutions
- Inability to tree-shake with predictable platform injections (such as `AsyncLocalStorage`)

This PR uses a list of nitro built-in runtime dependencies and explicitly inlines them in the final output bundle.

TODO: This method however has a downside in that inlined sourcemaps add to the bundle overhead. Looking for a way to disable inline-sourcemaps only for node_modules .

### Benchmark Results

Measurement / Mode     | external deps | inlined deps
----------------|---------------|------------------
number of files | 120           | 13
disk size (without sourcemaps) | 840 KB         | 476 KB
start | 31.823 ms | 20ms
fetch | 14.18 ms | 14ms
start-to-fetch | 47.464ms | 35ms

Benchmark is not accurate and is made overnight!

Config:

```ts
  sourceMap: false,
  analyze: {
    filename: "./analyze.html",
  },
});
```

(stupid!) Timing bench:

```ts
console.time("start");
console.time("start-to-fetch");

await import("./b/server/index.mjs").then(async () => {
  console.timeEnd("start");
  console.time("fetch");
  await fetch("http://localhost:3000/").then((r) => {
    console.timeEnd("fetch");
    console.timeEnd("start-to-fetch");
  });
});
```

Node.js version: v18.16.0 on Macbook Air M2

### 📝 Checklist


src/deps.ts
```diff
@@ -0,0 +1,21 @@
...
+  "destr",
+  "hookable",
+  "iron-webcrypto",
+  "klona",
+  "node-fetch-native",
+  "ofetch",
+  "ohash",
+  "pathe",
+  "radix3",
+  "radix3",
```
danielroe, line 14:
```suggestion
```

danielroe (commented):
In future we might need an opt out for major versions somehow? (If user, or other library is using dependency.)

Or equally if one of these is a dependency of an external we might have an issue too.

pi0:
> In future we might need an opt out for major versions somehow? (If user, or other library is using dependency.) Or equally if one of these is a dependency of an external we might have an issue too.

Yes, i was thinking to support `externals.exclude` (and filter-out default inlines when user/framework specifically requires opt-out for a feature) + a flag to completely disable this behavior.

Regarding major versions (or any custom version by user), we still use rollup's search path, resolution, and aliases so overriding is possible by all means, however even totday, we require them (defined list) to be in Nitro-Expected semver-major as nitro internally uses all these deps. 

I want to make sure mainly it is good to you too for becoming the default behavior .

danielroe:
It is fine to me as default behaviour. I'm observing the potential issue that if a library in externals uses ufo/pathe then this will be included twice (once in node_modules and once bundled). If other major versions of libraries like ufo are included in externals that entirely addresses my first point.

pi0:
> if a library in externals uses ufo/pathe then this will be included twice (once in node_modules and once bundled).

That's a valid point if both inline and external use same semver/major of ufo. And was initial reason that we made all externals by default. I guess we can think of some ways to automatically "deoptimize" inlining however it involves traced file dependency "pre-analysis" and sub-dependencny deptimization.

I guess for default behavior, it should benefit most of users both Nitro and Nuxt with both bundle size but more importantly the stability of resolutions with rollup comparing to current externals.

danielroe:
Yes, I see this as a good change, too.

pi0:
Updated: Sourcemap size issue is (kinda) resolved by using a sourcemap-ignore plugin that scans emited sourcemap chunks and dynamically disables the ones that include node modules (reducing basic build from `2.1MB` to `454 kB`. 

I think there is way more area for improving this plugin. We can tree-shake content only for example. Considering we use lazy chunks for user code, this should be _relatively_ ok. This can be disabled with  `experimental.sourcemapIgnore = false` in case of any issues.

Issue #79, opened 2022-04-12
aws lambda-edge


pi0:
https://github.com/nuxt/framework/issues/23

tresko:
Hi, I wonder if you have any ETA for this preset?

WinterYukky:
@tresko I created the PR exactly at the moment you commented. LOL!

The current escape hatch is to wrap an existing preset for Lambda@Edge.
An example of wrapping a `aws-lambda` preset is shown in the following code.

```js
import { URLSearchParams } from "url";
import * as nitro from "./index.mjs";

export const handler = async (event) => {
  const request = event.Records[0].cf.request;
  const queryStringParameters = Object.fromEntries(
    new URLSearchParams(request.querystring).entries()
  );

  const response = await nitro.handler({
    path: request.uri,
    queryStringParameters,
    httpMethod: request.method,
    headers: normalizeIncomingHeaders(request.headers),
    body: request.body,
  });
  return {
    status: response.statusCode,
    headers: normalizeOutgoingHeaders(response.headers),
    body: response.body,
  };
};

function normalizeIncomingHeaders(headers) {
  return Object.fromEntries(
    Object.entries(headers).map(([key, keyValues]) => [
      key,
      keyValues.map((kv) => kv.value).join(","),
    ])
  );
}
function normalizeOutgoingHeaders(headers) {
  return Object.fromEntries(
    Object.entries(headers).map(([key, values]) => [
      key,
      values.split(",").map((value) => ({ value })),
    ])
  );
}
```

Saving this code as `wrapper.mjs` and setting the Lambda entry point as `wrapper.handler` should work. 
NOTE: require `nodejs_14_x` or `nodejs_16_x`

tresko:
@WinterYukky Wow, amazing! Thank you!

I saw on your PR that you are using Nuxt3 on lambda@edge. Am I correct? 

Can you share a wrapper example for Nuxt 3?

WinterYukky:
@tresko Yes. I'm using Nuxt3 on lambda@edge.

I have created and published a simple project for you.
It is a site that when accessed via `/<anime title>` calls Free's API and SSRs the results.
The AWS configuration is written by AWS CDK in the `cdk` directory.
I have included a few comments as you may not be familiar with the AWS CDK, but if you have any questions please ask.

I have included the deployment procedure at the end of `README.md` so you can try it if you like.

**ref:** https://github.com/WinterYukky/nuxt3-lambda-edge-example

[...]

PR #1075, opened 2023-03-21
Aws lambda edge

juankamilo:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Add AWS Lambda@Edge to the preset.

The current aws-lambda preset is not compatible with the Lambda@Edge
This option is one of the most cost-effective on cloud technologies, it is much faster and more effective than "API Gateway".
I take WinterYukky job as a base, focusing more on ".output/server/index.mjs" and removing the zero config functionalities as there are several CloudFormation templates to deploy, leaving a simpler option to implement inside your CI/CD.

Resolves #79 

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #240, opened 2022-05-15
feat(preset): add aws-lambda-edge preset

WinterYukky:
### 🔗 Linked issue

- #79 

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Add AWS Lambda@Edge to the preset.

The current `aws-lambda` preset is not compatible with the Lambda@Edge format and requires users to create their own wrapper for Lambda@Edge. This PR is needed to resolve this issue.  
It would also be very powerful if Lambda@Edge were available in Nitro. It will be quite important for projects that require SSR (e.g Nuxt3) as static assets (`.output/public`) can be retrieved from AWS S3 and the rest (`.output/server`) can be resolved by Lambda@Edge.

Resolves #79 

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

Issue #80, opened 2022-04-12
aws amplify 


pi0:
https://github.com/nuxt/nuxt.js/issues/11651

https://github.com/aws-amplify/amplify-hosting

PR #1557, opened 2023-08-09
feat: add edge support to aws-lambda presets

Hebilicious:
### 🔗 Linked issue

Resolves https://github.com/unjs/nitro/issues/79
Related https://github.com/unjs/nitro/pull/240
Related https://github.com/unjs/nitro/pull/1075

### ❓ Type of change

### 📚 Description

Continuing on the amazing work done in https://github.com/unjs/nitro/pull/240 and https://github.com/unjs/nitro/pull/1075, this PR implements lambda edge as an option of the lambda-preset, usable with `{ target: edge }` adds a CDK construct based on @WinterYukky work, and adds a SST construct.

If we want to provide something that is directly deployable (ie for `nitro-deploys`), the CDK is an interesting option, however looking closely at it, the CDK itself has a few drawbacks : 

- It is very verbose and requires 2 separated stacks for lambda and lambda-edge
- It adds a large amount of maintenance burden and AWS CDK overhead
- It is quite a low level tool and not straightforward to use
- We heavily get into the SST / Pulumi / SLS territory 

Supporting CDK options is not straight forward... We could provide a Lambda-single and a Lambda-edge stack, and mirror some of these options in the config under a `cdkOptions` key. This is quite limited and I would assume most users will replace this with their own stack in real world scenario.

I think the best course on action here would be to rely on the higher level constructs provided by SST, example here 
 https://github.com/jdevdevdev/nuxt-sst/blob/lambda-edge/stacks/NuxtSite.ts 

This PR therefore adds a SST implementation that extends the `SSRSite` constructs to compare with the CDK ones.
(Later we could contribute it upstream to simplify the preset)

For reference, there is also the possibility of using [Amplify](https://github.com/unjs/nitro/issues/80) with the [CDK](https://aws.amazon.com/blogs/mobile/deploying-a-static-website-with-aws-amplify-and-cdk/). 
As usual with AWS, this is poorly documented, and we should deal with this in another PR #80 

(PS: I would like to give credit to @WinterYukky, @juankamilo and @jdevdevdev as co-authors of this PR)

### 📝 Checklist


src/build.ts
```diff
@@ -359,6 +359,7 @@ async function _build(nitro: Nitro, rollupConfig: RollupConfig) {
   const buildInfo = {
     date: new Date(),
     preset: nitro.options.preset,
+    entry: nitro.options.entry.split("/").pop(),
```
Hebilicious, line 362: We have no way to know which entry a preset use, maybe this should be in another PR

src/utils/sst.ts
```diff
@@ -0,0 +1,188 @@
...
+  sstOptions,
+  sstDir,
+}: {
+  edge: boolean;
+  sstOptions: PresetOptions["awsLambda"]["sstOptions"];
+  sstDir: string;
+}) =>
+  /* ts */ `
+import { SSTConfig } from "sst";
+import { NitroApp } from "${sstDir}";
```
Hebilicious, line 53: This should probably be a relative path.

Hebilicious (commented):
Both the CDK and SST logic can be removed from this PR and included later. However for `nitro-deploy` we need one or the other.

src/build.ts
Hebilicious: Extracted in https://github.com/unjs/nitro/pull/1649

src/utils/cdk.ts
```diff
@@ -0,0 +1,266 @@
+import { resolve } from "pathe";
+import { Nitro } from "../types";
+import { writeFile } from "../utils";
+
+export async function generateCdkApp(nitro: Nitro) {
```
Hebilicious, line 5: This is way more verbose and error prone than SST, harder to tests and doesn't handle regular lambdas.

src/utils/sst.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (!fs.existsSync(serverDir) || !fs.existsSync(clientDir)) {
+      throw new Error(
+        'Build output inside ".output/" does not contain the "server" and "public" folders.'
+      );
+    }
+
+    super.validateBuildOutput();
+  }
+
+  protected createFunctionForRegional(): CdkFunction {
```
jdevdevdev, line 130: This template may need to be updated for the [changes that were made to SST ](https://github.com/sst/sst/blob/c61396b0a9197dcac59bc3ffd5bdbdc308b47f30/packages/sst/src/constructs/SsrSite.ts).

It has been updated in my [test repo.](https://github.com/jdevdevdev/nuxt-sst/blob/d4da6aad6313dd6e4e7a20dd59509fd4bfe6dd0a/stacks/NuxtSite.ts)

A faster way to merging might be to split SST integration and the lambda edge preset code into different PRs. Let the user build the deployment and add deployment guides / documentation on nitro.unjs.io / nuxt.com

PR #1559, opened 2023-08-09
fix: remove non-standard `module` export condition

### LINKED ISSUE #1558: export condition issues with packages using `module` condition and commonJS modules
Possible regression from https://github.com/unjs/nitro/pull/1401. (Edge releases before then do not exhibit same behaviour.)

This reproduces for lots of Nuxt projects in https://github.com/nuxt/ecosystem-ci, particularly found this one from [nuxt-modules/sanity](https://github.com/nuxt-modules/sanity).

Nitro-only reproduction: https://stackblitz.com/edit/github-81ghjr?file=routes/index.ts,package.json,nitro.config.ts

pi0:
### 🔗 Linked issue

resolves #1558

### ❓ Type of change

### 📚 Description

`module` is a non-standard export condition not supported by Node.js or any runtimes which was previously used as a top-level property in order to help bundlers pick the esm-like/browser builds. 

We disabled [top level `module` field long before](https://github.com/unjs/nitro/blob/7fa84084dc2531d8274d0bb035d4f4977428311c/src/rollup/config.ts#L425) because of this externals, new resolution conditions in #1401 added it back and makes issues with packages such as `ts-lib` and `@sanity/client` that are (wrongly!) using AND preferring this export condition over `import.

I also updated Bun conditions. Checking their docs, it seems they recently made same decition.

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -498,7 +498,7 @@ function _resolveExportConditions(
   }
 
   // 4. Add default conditions
-  resolvedConditions.push("import", "module", "default");
+  resolvedConditions.push("import", "require", "default");
 
   // Dedup with preserving order
   return resolvedConditions.filter(
--- a/src/options.ts
+++ b/src/options.ts
@@ -498,7 +498,7 @@ function _resolveExportConditions(
   }
 
   // 4. Add default conditions
-  resolvedConditions.push("import", "require", "default");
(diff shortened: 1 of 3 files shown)


danielroe:
Nice work 👌

PR #1563, opened 2023-08-11
ci: use `h3-nightly@latest` for edge releases

pi0:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/scripts/bump-edge.ts
+++ b/scripts/bump-edge.ts
@@ -3,6 +3,10 @@ import { execaCommand } from "execa";
 import { resolve } from "pathe";
 import { globby } from "globby";
 
+const nightlyPackages = {
+  h3: "h3-nightly",
+};
+
 async function loadPackage(dir: string) {
   const pkgPath = resolve(dir, "package.json");
   const data = JSON.parse(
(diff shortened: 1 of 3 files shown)


scripts/bump-edge.ts
```diff
@@ -117,6 +119,11 @@ async function main() {
       `${pkg.data.version}-${date}.${commit}`
     );
     workspace.rename(pkg.data.name, pkg.data.name + "-edge");
+    pkg.updateDeps((dep) => {
+      if (nightlyPackages[dep.name]) {
+        dep.range = "npm:" + nightlyPackages[dep.name] + "@latest";
```
danielroe, line 124: see my comment at https://github.com/nuxt/nuxt/pull/22593#discussion_r1291241882.

I'm not sure `latest` is safe here. I'd recommend a caret + version hash.

pi0: Also see my comment there! I think it should be safer than pinned version to avoid hoisting issues on edge.

pi0: Reposting from https://github.com/nuxt/nuxt/pull/22593#discussion_r1291262322, caret + hash is respected by package manager in a way that it won't look for any newer version + (sadly) `h3-nightly@^1` doesn't works with packages published with `-` version.

pi0:
Merging to test end-to-end with the package manager behavior + coupled with nuxt edge.

I think if it is introducing to much trouble, we shall revert it and keep using `h3` latest stable release. 1.8 was kinda exceptional release that taking too much rc(s).

PR #1565, opened 2023-08-11
feat(wasm): support output esm imports

antfu:
### ❓ Type of change

### 📝 Checklist


src/rollup/plugins/wasm-import.ts
```diff
@@ -0,0 +1,69 @@
...
+          .update(buffer)
+          .digest("hex")
+          .slice(0, 16);
+        const ext = extname(filepath);
+        const name = basename(filepath, ext);
+
+        const outputFileName = "wasm/[name]-[hash][extname]"
+          .replace(/\[hash]/g, hash)
+          .replace(/\[extname]/g, ext)
+          .replace(/\[name]/g, name);
```
pi0, line 39: refactor: Since it is an internal code, we might use string templates directly i guess. Also any reason extension is not added to the name?

src/rollup/plugins/wasm-import.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+    },
+    async generateBundle() {
+      await Promise.all(
+        Object.keys(copies).map(async (name) => {
+          const copy = copies[name];
+          await this.emitFile({
+            type: "asset",
+            source: copy.buffer,
+            name: "Rollup WASM Asset",
```
pi0, line 62: Is it fine that name is same for file emits?

src/rollup/config.ts
```diff
@@ -150,9 +151,13 @@ export const getRollupConfig = (nitro: Nitro): RollupConfig => {
...
   if (nitro.options.experimental.wasm) {
     const options = {
-      ...(nitro.options.experimental.wasm as RollupWasmOptions),
+      ...(nitro.options.experimental.wasm as WasmOptions),
     };
-    rollupConfig.plugins.push(wasmPlugin(options));
+    if (options.directImport) {
+      rollupConfig.plugins.push(wasmImport());
+    } else {
+      rollupConfig.plugins.push(wasmPlugin(options));
```
pi0, line 159: We might merge wasmPlugin into the wasm import to keep config (little bit) simplified.

src/rollup/config.ts
```diff
@@ -365,7 +370,8 @@ export const plugins = [
             return { id: _resolved, external: false };
           }
         }
-        if (!resolved || resolved.external) {
+        // @ts-expect-error cast
+        if (!resolved || (resolved.external && !resolved.__nitro_external__)) {
```
pi0, line 374: Is this `__nitro_external__` flag used other places?

src/presets/cloudflare.ts
```diff
@@ -22,5 +22,15 @@ export const cloudflare = defineNitroPreset({
         JSON.stringify({ lockfileVersion: 1 }, null, 2)
       );
     },
+    "rollup:before"(nitro) {
```
antfu, line 25: This hook seems to be too late. Do we have a way for preset to extend the config?

pi0: Yes, preset is basically config preset we can add `wasm: {directImport: true }` to the preset. We shall introduce top level `wasm` option btw.

src/rollup/config.ts
antfu: I forgot to update it, pushed

antfu:
Then bundling works, tho the nitro dev doesn't. And it seems to be a rather complex thing to do right, maybe we better keep the experimental state?

pi0:
Working example with wasm bundled mode (dev/prod):

```ts
export default defineLazyEventHandler(async () => {
  const { sum } = await import("./sum.wasm")
    .then((wasm) => wasm.default({}))
    .then((r) => r.instance.exports);

  return eventHandler((event) => {
    return sum(1, 2);
  });
});
```

pi0 (approved):
Thanks, LGTM. Let's keep it under the experimental flag until we can unify usage across.

PR #1576, opened 2023-08-15
fix: don't use `_originalPath`

danielroe:
### 🔗 Linked issue

Dropped in https://github.com/unjs/h3/pull/506

### ❓ Type of change

### 📚 Description

I think caching on `path` probably makes sense but we could also use `event.node.req.originalUrl || event.node.req.url ||` to match current behaviour.

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -207,7 +207,7 @@ export function defineCachedEventHandler<T = any>(
         return escapeKey(customKey);
       }
       // Auto-generated key
-      const _path = event._originalPath || event.path;
+      const _path = event.path;
       const _pathname =
         escapeKey(decodeURI(parseURL(_path).pathname)).slice(0, 16) || "index";
       const _hashedPath = `${_pathname}.${hash(_path)}`;


src/runtime/cache.ts
```diff
@@ -207,7 +207,7 @@ export function defineCachedEventHandler<T = any>(
         return escapeKey(customKey);
       }
       // Auto-generated key
-      const _path = event._originalPath || event.path;
+      const _path = event.path;
```
danielroe, line 210:
```suggestion
      const _path =
        event.node.req.originalUrl || event.node.req.url || event.path;
```

pi0: Seems more safe

pi0 (approved):
Thanks ltgm. two notes:

- We might later simplify caching strategy to use `event.path` only. Need to double check all places
- Will also have to check why tests didn't caught it.

danielroe:
I think tests did catch it: https://github.com/unjs/nitro/actions/runs/5860687198/job/15889277077

PR #1577, opened 2023-08-15
feat: enable `timing` in debug mode

danielroe:
### ❓ Type of change

### 📚 Description

Feels like this might be a good default when debugging.

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -32,6 +32,7 @@ import { nitroImports } from "./imports";
 const NitroDefaults: NitroConfig = {
   // General
   debug: isDebug,
+  timing: isDebug,
   logLevel: isTest ? 1 : 3,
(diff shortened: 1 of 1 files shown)


pi0 (approved):

src/options.ts
```diff
@@ -32,6 +32,7 @@ import { nitroImports } from "./imports";
 const NitroDefaults: NitroConfig = {
   // General
   debug: isDebug,
+  timing: isDebug,
```
pi0, line 35: We might as well have it in `nitro-dev` preset (reason is that i am going to support non node dev experimental presets and timing feature is not that stable but can do later...)

oumarbarry: Interesting ! Non node dev presets support !? 😏

Issue #1571, opened 2023-08-14
Add the ability to use `ScheduledEvent` from Cloudflare


Barbapapazes:
### Describe the feature

Hello,

I would like to know how I could use Scheduled Event, https://developers.cloudflare.com/workers/runtime-apis/scheduled-event, directly from Nitro.

### Additional information

- [ ] Would you be willing to help implement this feature?

pi0:
It is a good idea. I think we might support feature via a custom `cloudflare:scheduledEvent` event triggered from default handler and can be hooked with nitro plugins to unblock feature 👍🏼 

For general feature in nitro, for a while i had been thinking to introduce Tasks and Cron job features. This way we can universally support events/cronjobs and integrate it with each platform and also easily test them locally.

Hebilicious:
> ### Describe the feature
> Hello,
> 
> I would like to know how I could use Scheduled Event, [developers.cloudflare.com/workers/runtime-apis/scheduled-event](https://developers.cloudflare.com/workers/runtime-apis/scheduled-event), directly from Nitro.
> 
> ### Additional information
> * [ ]  Would you be willing to help implement this feature?

AFAIK the only way to do it would be a custom nitro preset, I wrote about this somewhere.

Hebilicious:
@pi0 Do you mind explaining what is wrong with a custom preset approach ?

Barbapapazes:
> > ### Describe the feature
> > Hello,
> > I would like to know how I could use Scheduled Event, [developers.cloudflare.com/workers/runtime-apis/scheduled-event](https://developers.cloudflare.com/workers/runtime-apis/scheduled-event), directly from Nitro.
> > ### Additional information
> > 
> > * [ ]   Would you be willing to help implement this feature?
> 
> AFAIK here's a workaround with custom nitro presets : [nitro.unjs.io/deploy/custom-presets](https://nitro.unjs.io/deploy/custom-presets) Add the scheduled function there and it should work

And how to support both a `fetch` and a `ScheduledEvent`? If I overwrite the `entry`, the `fetch` will not be generated.

pi0:
@Hebilicious Nitro is a "Platform Agnostic" and "Zero Config" framework to build web servers. The whole topic of this issue is to make API available not to workaround and use a custom preset! Even introducing hooks is a workaround I am proposing until making the platform agnostic feature.

@Barbapapazes We need to register both `fetch` and new `scheduled` event handlers from main cloudflare entrypoints (just an example below)

[...]

PR #1579, opened 2023-08-15
feat(cloudflare): add queue and scheduled support WIP

Hebilicious:
### 🔗 Linked issue

Related https://github.com/unjs/nitro/issues/1571

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/cloudflare-module.ts
+++ b/src/runtime/entries/cloudflare-module.ts
@@ -2,10 +2,8 @@
 import "#internal/nitro/virtual/polyfill";
 import { withoutBase } from "ufo";
 import type {
-  ExecutionContext,
-  MessageBatch,
-  ScheduledEvent,
   ExportedHandler,
+  Response as CFResponse,
 } from "@cloudflare/workers-types";
 import {
   getAssetFromKV,
@@ -23,12 +21,9 @@ interface CFModuleEnv {
   [key: string]: any;
 }
 
-export default {
-  async fetch(
-    request: Request, // CFRequest,
-    env: CFModuleEnv,
-    context: ExecutionContext
-  ) {
+export default <ExportedHandler>{
+  async fetch(cfRequest, env: CFModuleEnv, context) {
+    const request = cfRequest as unknown as Request;
     try {
       // https://github.com/cloudflare/kv-asset-handler#es-modules
       return await getAssetFromKV(
@@ -73,28 +68,20 @@ export default {
       method: request.method,
       headers: request.headers,
       body,
-    });
+    }) as unknown as CFResponse;
   },
   // https://developers.cloudflare.com/queues/get-started/#5-create-your-consumer-worker
-  async queue(
-    batch: MessageBatch,
-    env: CFModuleEnv,
-    context: ExecutionContext
-  ) {
-    return nitroApp.hooks.callHook("cloudflare-module:queue", {
+  async queue(batch, env: CFModuleEnv, context) {
+    await nitroApp.hooks.callHook("cloudflare-module:queue", {
       batch,
       env,
       context,
     });
   },
   // https://developers.cloudflare.com/workers/runtime-apis/scheduled-event/
-  async scheduled(
-    event: ScheduledEvent,
-    env: CFModuleEnv,
-    context: ExecutionContext
-  ) {
-    return nitroApp.hooks.callHook("cloudflare-module:scheduled", {
-      event,
+  async scheduled(controller, env: CFModuleEnv, context) {
+    await nitroApp.hooks.callHook("cloudflare-module:scheduled", {
+      controller,
       env,
       context,
     });
--- a/src/runtime/entries/cloudflare-module.ts
+++ b/src/runtime/entries/cloudflare-module.ts
@@ -1,9 +1,11 @@
+/* eslint-disable require-await */
 import "#internal/nitro/virtual/polyfill";
 import { withoutBase } from "ufo";
 import type {
   ExecutionContext,
   MessageBatch,
   ScheduledEvent,
+  ExportedHandler,
 } from "@cloudflare/workers-types";
 import {
   getAssetFromKV,
@@ -73,21 +75,29 @@ export default {
       body,
     });
   },
-  // eslint-disable-next-line require-await
+  // https://developers.cloudflare.com/queues/get-started/#5-create-your-consumer-worker
   async queue(
     batch: MessageBatch,
     env: CFModuleEnv,
     context: ExecutionContext
   ) {
-    return nitroApp.hooks.callHook("cloudflare:queue", batch, env, context);
+    return nitroApp.hooks.callHook("cloudflare-module:queue", {
+      batch,
+      env,
+      context,
+    });
   },
-  // eslint-disable-next-line require-await
+  // https://developers.cloudflare.com/workers/runtime-apis/scheduled-event/
   async scheduled(
     event: ScheduledEvent,
     env: CFModuleEnv,
     context: ExecutionContext
   ) {
-    return nitroApp.hooks.callHook("cloudflare:scheduled", event, env, context);
+    return nitroApp.hooks.callHook("cloudflare-module:scheduled", {
+      event,
+      env,
+      context,
+    });
   },
 };
 
--- a/src/runtime/entries/cloudflare-module.ts
+++ b/src/runtime/entries/cloudflare-module.ts
@@ -1,6 +1,10 @@
 import "#internal/nitro/virtual/polyfill";
 import { withoutBase } from "ufo";
-import type { ExecutionContext } from "@cloudflare/workers-types";
+import type {
+  ExecutionContext,
+  MessageBatch,
+  ScheduledEvent,
+} from "@cloudflare/workers-types";
 import {
   getAssetFromKV,
   mapRequestToAsset,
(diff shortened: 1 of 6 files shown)


src/runtime/entries/cloudflare-module.ts
```diff
@@ -69,6 +73,22 @@ export default {
       body,
     });
   },
+  // eslint-disable-next-line require-await
+  async queue(
```
pi0, line 77: Would be nice to add a short link to the platform docs above each

src/runtime/entries/cloudflare-module.ts
```diff
@@ -69,6 +73,22 @@ export default {
...
+    batch: MessageBatch,
+    env: CFModuleEnv,
+    context: ExecutionContext
+  ) {
+    return nitroApp.hooks.callHook("cloudflare:queue", batch, env, context);
```
pi0, line 82: To make sure forward/backward compatibility, would be nice to send them all as `{ batch, env, context }`

src/runtime/entries/cloudflare.ts
```diff
@@ -3,12 +3,21 @@ import {
...
 } from "@cloudflare/kv-asset-handler";
+import { addEventListener } from "@cloudflare/workers-types";
 import { withoutBase } from "ufo";
 import { requestHasBody } from "../utils";
 import { nitroApp } from "#internal/nitro/app";
 import { useRuntimeConfig } from "#internal/nitro";
 import { getPublicAssetMeta } from "#internal/nitro/virtual/public-assets";
 
+addEventListener("scheduled", (event) => {
+  return nitroApp.hooks.callHook("cloudflare:scheduled", event);
```
pi0, line 14: Can you make sure if the format is same or similar to workers?

src/runtime/entries/cloudflare-module.ts
Hebilicious: Will try.
My main issue with this is that from what I can see, `return nitroApp.hooks.callHook("test", () => 1)` doesn't return 1 ?
Is there any other API suggestion here ?

src/runtime/entries/cloudflare.ts
Hebilicious: They have different arguments, but typing doesn't work properly. Should we go with `cloudflare-module:scheduled` and `cloudflare:scheduled` instead?

Hebilicious: > Is the event type also different between them?

According to Typescript, they are the same for scheduled. But they aren't for queues.

src/runtime/entries/cloudflare-module.ts
Hebilicious: Right, since we are supposed to return something, should we do something like this ? 

```ts
    const hookResponse = { response: new Response() }
    await nitroApp.hooks.callHook("cloudflare-module:queue", {
      batch,
      env,
      context,
      hookResponse
    });
    return hookResponse.response
 ```

Comment (pi0) [reply src/runtime/entries/cloudflare-module.ts:82]: Ah for handling, yes we might pass a callback like `sendResponse` (from the context object)

File: src/runtime/entries/cloudflare.ts (hunk 3 above)
Comment (pi0) [reply src/runtime/entries/cloudflare.ts:14]: At least object syntax would make them more similar... This is a low-level hook anyway btw so i won't worry too much about differences. We can set an indicator key like `isWorker`/`isModule` if needed.

Comment (Hebilicious) [reply src/runtime/entries/cloudflare.ts:14]:
```ts
interface CloudflareModuleRuntimeHooks {
  "cloudflare:scheduled": (moduleArgs: {
    controller: ScheduledController;
    env: CloudflareModuleEnv;
    context: ExecutionContext;
  }) => any;
  "cloudflare:queue": (moduleArgs: {
    batch: MessageBatch;
    env: CloudflareModuleEnv;
    context: ExecutionContext;
  }) => any;
}

interface CloudflareSWRuntimeHooks {
  "cloudflare:scheduled": (event: ScheduledEvent) => any;
  "cloudflare:queue": (event: QueueEvent) => any;
}
```

This is their signature at the moment, they don't have any common properties. 

Note that `addEventEventListener` can accept a 3rd optional argument that can't be specified easily with hooks :

```ts
export interface EventTargetAddEventListenerOptions {
  capture?: boolean;
  passive?: boolean;
  once?: boolean;
  signal?: AbortSignal;
}
```

src/runtime/entries/cloudflare-module.ts
Hebilicious:
```ts
export default defineNitroPlugin((app) => {
  app.hooks.hook("cloudflare:scheduled", (event) => {
    event.waitUntil(new Promise((resolve) => resolve("scheduled:cloudflare")));
  });

  app.hooks.hook(
    "cloudflare-module:scheduled",
    ({ controller, env, context }) => {
      context.waitUntil(
        new Promise((resolve) => resolve("scheduled:cloudflare-module"))
      );
    }
  );

  app.hooks.hook("cloudflare:queue", (event) => {
    event.sendResponse = () => "queued:cloudflare";
  });

  app.hooks.hook("cloudflare-module:queue", ({ batch, env, context }) => {
    context.sendResponse = () => "queued:cloudflare-module";
  });
});
```
This is what it looks like when used.

trylovetom:
Hello @pi0 @Hebilicious

I noticed that the PR is currently in a draft status. May I ask if there are any follow-up plans or an expected release timeline for this?

I believe it brings tremendous value and can truly elevate the cloudflare project to another level.

pi0:
@trylovetom Surely it is in the roadmap for next minor version of Nitro. I expect to release alongside new Tasks feature so that you can easily test everything universally before deploying!

pi0:
Hi @Hebilicious and thanks again for this PR. More i think it makes more sense that we introduce this feature on top of (universal) nitro tasks (https://github.com/unjs/nitro/issues/1571#issuecomment-1677366301) instead of directly supporting provider-flavored APIs and hooks.

I am thinking of closing this draft to clarify our roadmap but i would happy to have your help to build this feature on top of nitro nitro tasks or revisit if that took longer than expected and you think it is critical to support this feature ahead of the time.

Hebilicious:
@pi0 My personal thoughts is that we need 2 features : 

- Nitro Tasks (similar to [Task scheduling in Laravel](https://laravel.com/docs/10.x/scheduling))
- Nitro Queues (similar to [Queues in Laravel](https://laravel.com/docs/10.x/queues))

Should I open 2 issues to track these as a follow-up ? 

I do think this is going to be a big chunk of work to make them platform agnostic and to integrate with all providers; while introducing platform specific hooks (maybe under an experimental flag?) is relatively fast and clean. 

Regarding whether it's critical or not, personally, if I need this, I would use a custom preset.
I think documenting how to make a custom custom preset to use these 2 features, is cleaner and easier, as there's not a need to learn any extra syntax and to figure out how to do something that is already well explained in the cloudflare docs. But I understand that we shouldn't recommend them either ... though call... Maybe we could add an experimental flag for custom presets ? 🤔

pi0:
No please wait for me to open POC PR for Tasks. We can discuss Queues in the next steps.

> I do think this is going to be a big chunk of work to make them platform agnostic and to integrate with all providers

This is all mission of Nitro to do heavy tasks like this. Don't worry we can make it happen!

> Regarding whether it's critical or not, personally, if I need this, I would use a custom preset.
I think documenting how to make a custom custom preset to use these 2 features, is cleaner and easier, as there's not a need to learn any extra syntax and to figure out how to do something that is already well explained in the cloudflare docs. But I understand that we shouldn't recommend them either ... though call... Maybe we could add an experimental flag for custom presets

Surely as i mentioned above and in linked issue, the initiatives behind introducing hooks was to speed up introducing it for cloudflare, which took longer than expected with your PR and also tasks is one of my main priorities to deliver ASAP.

Again, i wouldn't recommend using custom presets, for same reasons we don't recommend making a custom Vue setup for Nuxt applications. The job of Nitro, Nuxt and other meta frameworks, is to keep users away from boilerplate and make sure they are using best optimized practices. I made both Nuxt and Nitro fully extendable so we don't "block" users of doing something like this just like you can do today but neither don't advice them to do as is a dangerous territory in regards of promised stability. I hope you understand this concept.

Hebilicious:
> No please wait for me to open RFC/POC for Tasks. We can discuss Queues in the next steps. (could be on top as a module. Not sure if queue can be generic feature easily while tasks can)
> 
> > I do think this is going to be a big chunk of work to make them platform agnostic and to integrate with all providers
> 
> This is all mission of Nitro to do heavy tasks like this. Don't worry we can make it happen!
> 
> > Regarding whether it's critical or not, personally, if I need this, I would use a custom preset.
> > I think documenting how to make a custom custom preset to use these 2 features, is cleaner and easier, as there's not a need to learn any extra syntax and to figure out how to do something that is already well explained in the cloudflare docs. But I understand that we shouldn't recommend them either ... though call... Maybe we could add an experimental flag for custom presets
> 
> Surely as i mentioned above and in linked issue, the initiatives behind introducing hooks was to speed up introducing it for cloudflare, which took longer than expected with your PR and also tasks is one of my main priorities to deliver ASAP.
> 
> Again, i wouldn't recommend using custom presets, for same reasons we don't recommend making a custom Vue setup for Nuxt applications. The job of Nitro, Nuxt and other meta frameworks, is to keep users away from boilerplate and make sure they are using best optimized practices. I made both Nuxt and Nitro fully extendable so we don't "block" users of doing something like this just like you can do today but neither don't advice them to do as is a dangerous territory in regards of promised stability. I hope you understand this concept.

I understand this concept and I'm reiterating the custom preset point because this PR was closed.
In that regard, I am happy to move forward and re-open this PR (ie add docs), and get it merged if that implementation is satisfactory; or to drop it entirely if you don't think it's important.

PR #1588, opened 2023-08-16
fix: use decoded asset id to access prerendered asset header overrides

danielroe:
### ❓ Type of change

### 📚 Description

Add test - sorry!

### 📝 Checklist


test/fixture/api/param/[id].ts
```diff
@@ -1,3 +1,4 @@
 export default eventHandler((event) => {
+  setHeader(event, "Content-Type", "text/plain; charset=utf-16");
```
Comment (pi0) [inline test/fixture/api/param/[id].ts:2]: Maybe something other than plain (which is default)?

Comment (danielroe) [reply test/fixture/api/param/[id].ts:2]: I'm testing via the `utf-16` bit

pi0 (approved):

PR #945, opened 2023-02-15
feat: allow ignoring public assets with `ignore` options

motea927:
add ignorePublic option to ignore public assets with a pattern (glob patterns)

### 🔗 Linked issue

#843 

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Resolves #843 
Add `ignorePublic` to ignore public assets with a pattern (glob patterns), it is very useful and saves a lot of time for production.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1632, opened 2023-08-25
fix: pass relative ignore paths when scanning public assets

danielroe:
### ❓ Type of change

### 📚 Description

We now support ignoring public assets (https://github.com/unjs/nitro/pull/945) but at the moment we do not resolve/relativise paths, so an ignore pattern of `public/test.txt` would not match `test.txt` when scanning the `public/` directory. The issue is that it then becomes impossible to specify ignore patterns that _only_ apply in public directories, for example.

This is a **change in behaviour** but I think more intuitive.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -55,7 +55,7 @@ export async function copyPublicAssets(nitro: Nitro) {
         dot: true,
         ignore: nitro.options.ignore
           .map((p) =>
-            p.startsWith("*")
+            p.startsWith("*") || p.startsWith("!*")
               ? p
               : relative(srcDir, resolve(nitro.options.srcDir, p))
           )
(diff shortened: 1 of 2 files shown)


src/build.ts
```diff
@@ -53,7 +53,13 @@ export async function copyPublicAssets(nitro: Nitro) {
...
         absolute: false,
         dot: true,
-        ignore: nitro.options.ignore,
+        ignore: nitro.options.ignore
+          .map((p) =>
+            p.startsWith("*") || p.startsWith("!*")
+              ? p
+              : relative(srcDir, resolve(nitro.options.srcDir, p))
+          )
+          .filter((p) => !p.startsWith("../")),
```
pi0, line 62: This might make issues with multi-layer when relative path is outside of main `srcDir` but inside other layer's source dir... But let's try and see!

pi0 (approved):

PR #1637, opened 2023-08-26
fix: check if is obj before checking if is a event

brendonmatos:
### Issue

The problem appeared after updating from nuxt@3.6 -> 3.7

```
TypeError: Cannot use 'in' operator to search for '__is_event__' in 2
      at isEvent (....../node_modules/.pnpm/h3@1.8.0/node_modules/h3/dist/index.mjs:1419:25)
      at <anonymous>(....../node_modules/.pnpm/nitropack@2.6.1/node_modules/nitropack/dist/runtime/cache.mjs:88:18)
````

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -150,11 +150,13 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = unknown[]>(
     }
     const key = await (opts.getKey || getKey)(...args);
     const shouldInvalidateCache = opts.shouldInvalidateCache?.(...args);
+    const [firstArg] = args;
+    const firstArgIsEvent = firstArg && typeof firstArg === 'object' && isEvent(firstArg);
     const entry = await get(
       key,
       () => fn(...args),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/runtime/cache.ts [hunk 1]
```diff
@@ -150,11 +150,14 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = unknown[]>(
     }
     const key = await (opts.getKey || getKey)(...args);
     const shouldInvalidateCache = opts.shouldInvalidateCache?.(...args);
+    const [firstArg] = args;
+    const firstArgIsEvent =
```
danielroe, line 154: maybe this check could be moved to `isEvent` instead?

pi0:
Thanks for the notice and PR dear @brendonmatos ❤️ I think it is safer that we fix upstream `isEvent` instead: https://github.com/unjs/h3/pull/521

If you like to help make similar refactors in nitro for more code safety, it is more than welcome!

PR #1640, opened 2023-08-26
fix: use new h3 generics for `defineCachedEventHandler`

danielroe:
### ❓ Type of change

### 📚 Description

When updating h3 we missed out updating the generic types for `defineCachedEventHandler`. This PR implements it although note that this is not backwards compatible. (Happy to add an overload as we did in `h3` if we need to support previous generic pattern.)

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -6,7 +6,7 @@ import {
   EventHandler,
   isEvent,
 } from "h3";
-import type { H3Event } from "h3";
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for quick PR! Would be nice if we keep supporting first generic type (although i know it is a pain to do but people might have been depending on it..)

pi0:
One small note: request body should be never in cached handlers

(also headers when we add type for, only ones from varies option are available)

PR #1649, opened 2023-08-27
feat: add entry to build info

### LINKED ISSUE #1650: add entry information to build info
### Describe the feature

While working on https://github.com/unjs/nitro/pull/1557, I noticed that for presets that can have multiple entries, like firebase, there's no straightforward way to identify which one has been used.

### Additional information

Hebilicious:
### 🔗 Linked issue

resolves #1650 

### ❓ Type of change

### 📚 Description

While working on https://github.com/unjs/nitro/pull/1557, I noticed that for presets that can have multiple entries, like firebase, there's no straightforward way to identify which one has been used. This adds a reference to the preset used in the build info.

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -405,6 +405,7 @@ async function _build(nitro: Nitro, rollupConfig: RollupConfig) {
   const buildInfo = {
     date: new Date(),
     preset: nitro.options.preset,
+    entry: nitro.options.entry.split("/").pop(),
     commands: {
       preview: nitro.options.commands.preview,
       deploy: nitro.options.commands.deploy,


src/build.ts
```diff
@@ -405,6 +405,7 @@ async function _build(nitro: Nitro, rollupConfig: RollupConfig) {
   const buildInfo = {
     date: new Date(),
     preset: nitro.options.preset,
+    entry: nitro.options.entry.split("/").pop(),
```
pi0, line 408: This is the **source entry** name.

Hebilicious: Yes, this is useful for presets like firebase, where one preset can have multiple entries : 
ie looking at the build output, we can't know if `firebase-gen-1` or `firebase-gen-2` is being used.

pi0: I know that's a valid case but generally it doesn't seems to be best idea to add source entry to production info. It can be implemented per preset imo.

Hebilicious: That's a good point, what about renaming the preset if it's firebase ? Something like `firebase_${variant}` ?

pi0: Renaming the user-input preset can also lead to more confusing behavior.

pi0:
Thanks for the PR @Hebilicious. I guess i would prefer to use current method for firebase preset at least to keep the preset implementations simplicity and decoupled as much as possible.

While current approach for firebase gen handling is not perfect, i am happy to discuss about some other alternative ideas once we migrated repo to a monorepo. Also we can use virtual chunks for entry or simply support dynamic entry based on user input config (it was absence by the time of implementing that PR...)

PR #1672, opened 2023-08-30
docs: add instructions for bun

colinhacks:
Adds instructions for users of the Bun package manager (where relevant)


danielroe (approved):

pi0 (approved):
Thanks. With last changes i think we are almost ready to go 🔥

Issue #1394, opened 2023-07-04
Deployment Overview page 404


ChrisScrumping:
### Environment

https://nitro.unjs.io/

### Reproduction

Try going to https://nitro.unjs.io/deploy

### Describe the bug

When opening https://nitro.unjs.io/deploy or refreshing the page I get a 404 page not found.

### Additional context

Initially I discovered this by clicking a link to the page from here: https://nuxt.com/docs/getting-started/deployment#deployment

Hebilicious:
@Atinux I've seen this happening a few times in Nuxt/UnJS websites. Is this docus / Nuxt content related?

ChrisScrumping:
I have had it happen recently on a Nuxt app I am creating but it turned out to be a redirect issue on the server not Nuxt.

In this case though there does not appear to be any redirects happening, the server just returns a 404.

```curl -vv https://nitro.unjs.io/deploy/workers```

**Comment by JamieCurnow** [comment]
I started looking into this. I couldn't reproduce locally...

Interestingly, I can reproduce the bug on vercel:
https://jamiecurnow-nitro-nomc9cbxu-jamiecurnow.vercel.app/deploy/workers

But when I deploy to Netlify it works perfectly:
https://stately-donut-53286f.netlify.app/deploy/workers/

Both the builds are using `npm run build` and looking at the `dist` directory for the public assets. I'll do some more digging tomorrow on node versions etc...

**Comment by JamieCurnow** [comment]
@Hebilicious does nitro detect that it's in a vercel / netlify environment and apply build presets accordingly?

**Comment by JamieCurnow** [comment]
I think I've found the cause... When doing a vercel preset build, most paths in the `.vercel/output/config.json` do not have a trailing slash. BUT the paths that don't work *do* have a trailing slash...

```
{
  "version": 3,
  "overrides": {
    "index.html": {
      "path": ""
    },
    "guide/getting-started/index.html": {
      "path": "guide/getting-started"
    },
    // ...
    "deploy/workers/index.html": {
      "path": "deploy/workers/" // <-- trailing slash
    },
  }
}
```

I've tested this by adding the following to the nuxt config:

```
nitro: {
    vercel: {
      config: {
        version: 3,
        overrides: {
          "deploy/index.html": {
            path: "deploy",
          },
          "deploy/workers/index.html": {
            path: "deploy/workers",
          },
        },
      },
    },
  },
```
And it works! 🥳 

Now to try to figure out why those trailing slashes are being added....

[...]

Issue #607, opened 2022-10-21
Prerender with named html files


slyduda:
### Describe the feature

The prerenderer currently prefers prerendering in subfolders. While I can take the advice of writing named html files in the prerender route config, I would prefer to make this the togglable behavior with a bool in the nitro config (similar to how there was one in Nuxt v2).

After doing some searching this was the default behavior before [#166](https://github.com/unjs/nitro/issues/166). It would be nice to have a bool that toggles this in nitro config if it does not already exist.

### Additional information

- [x] Would you be willing to help implement this feature?

emretepedev:
Same issue here.

I don't want to create a folder for each page. This behavior must be changeable via config.

davidsandoz:
I would need this as well.

As already mentioned, this used to be the default before it got changed in cdcd35adf3e4f1d3902605f97559c0dcf6d0febf via #166.

In nuxt/nuxt#13886 (where the opposite was requested, before #166 got implemented), a workaround was mentioned in [this comment](https://github.com/nuxt/nuxt/issues/13886#issuecomment-1397320165) by @pi0:

> This is now the default behavior of Nuxt 3 in latest RC to use dist/[path]/index.html (unless path has an explicit .html extension)

And indeed, if we name a page component `foo.html.vue`, then we get the file `foo.html` in `.output/public/`.

However, I noticed a problem with this workaround. The generated file has references to `/foo.html/_payload.js`:
```html
<link rel="preload" as="fetch" crossorigin="anonymous" href="/foo.html/_payload.js">
```
```html
<script type="module">import p from "/foo.html/_payload.js";window.__NUXT__={...p,...({state:{},_errors:{},serverRendered:true,prerenderedAt:1681384634837,config:{public:{},app:{baseURL:"\u002F",buildAssetsDir:"\u002F_nuxt\u002F",cdnURL:""}}})}</script>
```

As `foo.html` is not a directory, it fails at retrieving the `_payload.js` file.

Is there a way to configure this properly? Or is this something that has to be fixed in nitro or Nuxt?

davidsandoz:
Additionally, if I have a page component with a parameter, e.g. `bar-[id].html.vue`, then the link crawler doesn't generate the page even though there is the following nuxt-link in `index.vue`:
```html
<nuxt-link to="bar-3.html">Bar 3</nuxt-link>
```
And there are no errors during generation.

While if I put `bar-3` instead of `bar-3.html`, I get the following error:
```
[...]

PR #1676, opened 2023-08-30
fix(prerender): allow disabling html sub-folders

danielroe:
### 🔗 Linked issue

Related: https://github.com/unjs/nitro/issues/1394, #607

### ❓ Type of change

### 📚 Description

This PR allows disabling prerendering in subfolders.

~~It also includes a potential fix for cloudflare pages/vercel presets, which I need to test. (Opening PR in the interim to get feedback on the API/implementation.)~~

I think safer to merge this first and treat cf/vercel preset issues separately.

### 📝 Checklist


src/prerender.ts
```diff
@@ -214,8 +214,12 @@ export async function prerender(nitro: Nitro) {
     const isImplicitHTML =
       !route.endsWith(".html") && contentType.includes("html");
     const routeWithIndex = route.endsWith("/") ? route + "index" : route;
+    const htmlPath =
+      route.endsWith("/") || nitro.options.experimental.prerenderIndexHTML
```
pi0, line 218: More thinking, we can directly introduce it as a `prerender` option as a fix since unblocks platform compatibility. (ideas welcome for naming!)

pi0:
@danielroe I think we can move forward to test on edge directly and revert if any issues before patch release 👍🏼

src/prerender.ts
danielroe: `prerender.forceSubfolder`? As we should render subfolder by default for an explicit trailing slash route, I think.

pi0: Force has some negative sense to me. Maybe `auto`? (also would be nice that `html` is included in key somehow to clarify behavior)

danielroe: I think the default is true (for using subfolders). how about `prerender.autoSubfolderIndex`?

pi0: `autoSubfolderIndex` seems nice!

Atinux:
FYI: https://github.com/unjs/nitro/commit/11a241244a166b0733b924d16ce5a32e1be89005

Would be nice to think of updating the docs when adding new options 👀

pi0:
Thanks! And yes we should automate it sometime soon similar to nuxt options...

Atinux:
I don't think we need to automate this as we can just better explain the configuration and it does not take much time to add in docs.

It's just a process we need to get used to.

pi0:
Without automation it is impossible to not forget and keep docs in sync (with version as well). I plan to do this for the upcming h3 docs as well...

Hebilicious:
> Without automation it is impossible to not forget and keep docs in sync (with version as well). I plan to do this for the upcming h3 docs as well...

Even if we list the existing options in the docs, we should write SDOC comments to explain what they do, a lot of the options are not documented anywhere currently.

pi0:
@Hebilicious yes true. that's the plan to use jsdocs (similar to nuxt) and automatically extract them to the docs kept in sync :)

PR #1683, opened 2023-09-01
fix(aws-lambda,netlify-lambda):  binary body v2 and cookies v1

Hebilicious:
### 🔗 Linked issue

Fix https://github.com/unjs/nitro/issues/852

### ❓ Type of change

### 📚 Description

The aws-lambda preset works with apigw v1 and v2. 
This PR fix the body type for apigw v2, and the cookies for apigw v1

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/netlify-lambda.ts
+++ b/src/runtime/entries/netlify-lambda.ts
@@ -36,16 +36,13 @@ export async function lambda(
   });
 
   const cookies = normalizeCookieHeader(String(r.headers["set-cookie"]));
-  const { body, bodyType } = await normalizeLambdaOutgoingBody(
-    r.body,
-    r.headers
-  );
+  const awsBody = await normalizeLambdaOutgoingBody(r.body, r.headers);
 
   return {
     statusCode: r.status,
     headers: normalizeLambdaOutgoingHeaders(r.headers, true),
-    body,
-    isBase64Encoded: bodyType === "binary",
+    body: awsBody.body,
+    isBase64Encoded: awsBody.type === "binary",
     ...(cookies.length > 0 && {
       multiValueHeaders: { "set-cookie": cookies },
     }),
--- a/src/runtime/entries/netlify-lambda.ts
+++ b/src/runtime/entries/netlify-lambda.ts
@@ -36,15 +36,18 @@ export async function lambda(
   });
 
   const cookies = normalizeCookieHeader(String(r.headers["set-cookie"]));
-  const outBody = await normalizeLambdaOutgoingBody(r.body, r.headers);
+  const { body, bodyType } = await normalizeLambdaOutgoingBody(
+    r.body,
+    r.headers
+  );
 
   return {
     statusCode: r.status,
     headers: normalizeLambdaOutgoingHeaders(r.headers, true),
-    body: outBody.body,
-    isBase64Encoded: outBody.type === "binary",
-    multiValueHeaders: {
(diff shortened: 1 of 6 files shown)


src/runtime/utils.lambda.ts
```diff
@@ -31,18 +31,18 @@ export function normalizeLambdaOutgoingHeaders(
 export async function normalizeLambdaOutgoingBody(
   body: BodyInit | ReadableStream | Buffer | Readable | Uint8Array,
   headers: Record<string, number | string | string[] | undefined>
-): Promise<{ type: "text" | "binary"; body: string }> {
+): Promise<{ bodyType: "text" | "binary"; body: string }> {
```
pi0, line 34: Was this change necessary?

pi0: I mainly refactored like this to avoid restructuring and having simple keys. I would appreciate if keep changes in PR minimal 🙏🏼 (one idea if you don't like long names, we can call variable `_body` or `awsBody` instead of `normalizedBody`.)

Hebilicious: Got it, updated everything to `awsBody` 👍🏽

src/runtime/entries/netlify-lambda.ts
```diff
@@ -36,15 +36,15 @@ export async function lambda(
...
     headers: normalizeLambdaOutgoingHeaders(r.headers, true),
-    body: outBody.body,
-    isBase64Encoded: outBody.type === "binary",
-    multiValueHeaders: {
-      ...(cookies.length > 0 ? { "set-cookie": cookies } : {}),
-    },
+    body: awsBody.body,
+    isBase64Encoded: awsBody.type === "binary",
+    ...(cookies.length > 0 && {
+      multiValueHeaders: { "set-cookie": cookies },
```
pi0, line 47: Just checking, was this (another) fix or refactor to avoid empty object?

test/presets/aws-lambda.test.ts
```diff
@@ -62,21 +57,23 @@ describe("nitro:preset:aws-lambda", async () => {
...
-        status: res.statusCode,
-        headers: resHeaders,
-      };
+      return makeResponse(res, "v2");
     };
   });
 });
+
+const makeResponse = (response: any, version: "v1" | "v2") => {
+  const headers = { ...response.headers };
```
pi0, line 66: Perf: We don't need to duplicate memory for headers if there is no `set-cookie` in it (nor the version checks logic). I wouldn't also make a copy even for `set-cookie` unless it is necessary to prevent a valid scenario where modifying original headers makes issues

pi0:
@Hebilicious Since planning for next patch release, what is blocker for this PR? Do you need to verify it against a deployment?

src/runtime/entries/netlify-lambda.ts
Hebilicious: Refactor due to the initial fix, I made the modification here too so that both `aws-lambda` and `netlify-lambda` have consistent behaviour.

Hebilicious:
> @Hebilicious Since planning for next patch release, what is blocker for this PR? Do you need to verify it against a deployment?

Was just in draft because WIP, it should be good to go now 👍🏽

EnzoAlbornoz:
It would be great if this is merged. I'm having issues with hosting IPX on my lambda function.

pi0 (approved):
Thanks (i haven't test it through but i trust you did)

PR #1697, opened 2023-09-05
Fix: make node-server.ts accept iisnode named pipe

tidan-16:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/22961

### ❓ Type of change

### 📚 Description

I have change the condition a little to node-server.ts. It just checks if the port number is a number or not and if the port is not a null and is not a number, it puts the port (string) inside the path variable instead, wich resolve the issue for iisnode. Iisnode uses Named pipe instead of port numbers. Please check issue more infos

### 📝 Checklist

### DIFF
--- a/src/runtime/entries/node-server.ts
+++ b/src/runtime/entries/node-server.ts
@@ -24,7 +24,7 @@ const host = process.env.NITRO_HOST || process.env.HOST;
 const path = process.env.NITRO_UNIX_SOCKET;
 
 // @ts-ignore
-const listener = server.listen(path ? { path } : { port, host }, (err) => {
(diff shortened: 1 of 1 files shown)


src/runtime/entries/node-server.ts
```diff
@@ -21,7 +21,8 @@ const port = (destr(process.env.NITRO_PORT || process.env.PORT) ||
   3000) as number;
 const host = process.env.NITRO_HOST || process.env.HOST;
 
-const path = process.env.NITRO_UNIX_SOCKET;
+const path =
+  port && Number.isNaN(Number(port)) ? port : process.env.NITRO_UNIX_SOCKET;
```
pi0, line 25: I think condition is reverse (when number is (not) NaN should be picked

tidan-16:
Any updated?

PR #1708, opened 2023-09-08
fix: sort dependencies in `.output/package.json`

jenjen75:
### ❓ Type of change

### 📚 Description

It is critical to keep `package.json` consistent on each building. Some CI/CD, like Docker, shines when caching layer is defined to prevent a heavy-low compilation time. The problem is at each building, `package.json` is inconsistent (version 2.6.3).

In the generation, `sort` and `localeCompare` are called, but the comparison is evaluated only for the first char of each key.
This PR solve the problem, to compare each string instead of each char.

Thank you

### 📝 Checklist


pi0 (approved):
Thanks!

PR #1753, opened 2023-09-23
fix(azure): fix cookie format normalization

tobiasdiez:
### ❓ Type of change

### 📚 Description

Cookies set by the users were dropped when the app is deployed to Azure. Reason is that azure expects the cookie to be of the form `{ name: 'key', value: 'val }` but nitro was passing it in the form `{ key: 'val' }`.
This is fixed, along with a few other improvements. (Refs https://github.com/unjs/nitro/pull/1452)

While working on this, I found a bug in the cookie parser: https://github.com/unjs/cookie-es/issues/17

### 📝 Checklist


pi0 (approved):
Thanks for this nice PR and also unit tests 💯 

I have added few refactors to have basic httponly/secure support, one typo in `sameSite` case and also more generic code style.

If something was bad, feel free to amend via a new PR.

tobiasdiez:
Thanks @pi0, your changes are nice improvements!

Btw, I think it would make sense to put this cookie parsing actually in h3, and use it to internally store the cookies as nice typed objects. What happens right now if you set a cookie is the following: set cookie as object > serialize cookie to string > now here in the azure handler: parse cookie again to get it back as an object. This whole serialization and parsing business can be skipped for azure and a few other frameworks. As a bonus, you get a nicer api in h3 to handle cookies.

pi0:
For now, i think we can also expose it from `cookie-es`. It was initially for compatibility but seems the API is also not that nice and could be better. PR is welcome 👍🏼

tobiasdiez:
Makes sense. I don't have the time right now, but added it to https://github.com/unjs/h3/issues/509 and https://github.com/unjs/cookie-es/issues/17 so we keep track of it.

pi0:
Perfect! Thanks for creating issues.

PR #1755, opened 2023-09-23
fix(renderer): allow all h3 handled body types

danielroe:
### ❓ Type of change

### 📚 Description

This updates the handling of render response to allow directly returning a stream as well as a string.

Feedback welcome on concept. We could alternatively create a new helper that could handle other common stream-related issues.

### 📝 Checklist

### DIFF
--- a/src/runtime/renderer.ts
+++ b/src/runtime/renderer.ts
@@ -6,7 +6,7 @@ import {
   setResponseHeader,
   setResponseHeaders,
   setResponseStatus,
-  isStream
+  isStream,
 } from "h3";
 import { useNitroApp } from "./app";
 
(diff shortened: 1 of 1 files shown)


src/runtime/renderer.ts
```diff
@@ -58,7 +59,7 @@ export function defineRenderHandler(handler: RenderHandler) {
     }
 
     // Send response body
-    return typeof response.body === "string"
+    return typeof response.body === "string" || isStream(response.body)
       ? response.body
       : JSON.stringify(response.body);
```
pi0, line 64: I am wondering do we still need this check still? Not sure why before, but we might directly always return `response.body` and reduce overhead of `isStream` check

PR #1789, opened 2023-10-05
feat: support modules to extend nitro builder

### LINKED ISSUE #1788: Nitro modules to extend core
### Describe the feature

Hello,

I'm currently asking myself how to share code between Nitro app.

Imagine, I'm building a server with auth on app A. I implement auth within code source A. Then, I'm building app B. I also need auth. How could I share code?

Currently, if I use Nuxt with Nitro, I can build a Nuxt module and modify directly the Nitro config to import some utils. If I use Nitro alone, it's not possible. In the same time, some features currently build for Nuxt could be only for Nitro and shared to other Nitro based framework. 

I could create a simple JS package but I will not be able to provide great DX like auto-import, to update the config, inject routes or more...

What do you think of modules for Nitro like in Nuxt?

### Additional information

Barbapapazes:
### 🔗 Linked issue

fix #1788

### ❓ Type of change

### 📚 Description

This PR introduce the concept of modules like in Nuxt.

- Support for function in the `nitro.config` file
- Support for function in the `modules` directory

This PR is a first implementation. depending of the usage, we will update the modules feature.

### 📝 Checklist


playground/nitro.config.ts
```diff
@@ -1 +1,14 @@
-export default defineNitroConfig({});
+import { defineEventHandler } from "h3";
+
+export default defineNitroConfig({
+  modules: [
+    (_, nitro) => {
```
pi0, line 5: Having resolved options first for Nuxt was mostly a backyard compatibility. For nitro we can simply accept `(nitro)`

src/module/define.ts
```diff
@@ -0,0 +1,22 @@
...
+export function defineNitroModule(definition: ModuleDefinition | NitroModule) {
+  if (typeof definition === 'function') { return defineNitroModule({ setup: definition }) }
+
+  const module = defu(definition, { meta: {} })
+  const options = {}
+
+  async function normalizedModule(this: any, inlineOptions: any, nitro: Nitro = useNitro()) {
+    const res = await module.setup?.call(null as any, options, nitro) ?? {}
+
+    if (res === false) {
```
pi0, line 14: This is not required for nitro

src/module/define.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ModuleDefinition, Nitro, NitroModule } from "../types";
+import { useNitro } from "../context";
+
+export function defineNitroModule(definition: ModuleDefinition | NitroModule) {
+  if (typeof definition === 'function') { return defineNitroModule({ setup: definition }) }
+
+  const module = defu(definition, { meta: {} })
+  const options = {}
+
+  async function normalizedModule(this: any, inlineOptions: any, nitro: Nitro = useNitro()) {
```
pi0, line 11: this and inline options is not required for nitro

src/module/define.ts
```diff
@@ -0,0 +1,22 @@
+import { defu } from "defu";
+import { ModuleDefinition, Nitro, NitroModule } from "../types";
```
pi0, line 2:
```suggestion
import type { ModuleDefinition, Nitro, NitroModule } from "../types";
```

src/module/install.ts
pi0, line 1: Install API is not required for nitro. For Nuxt we had to introduce it

src/context.ts
pi0, line 1: We might consider supporting multi instances for Nitro (also it is actually a case for prerenderer instance). Lets not introduce it

pi0:
Thanks for picking up on it and nice you could take inspiration from the Nuxt module.

If you would like to work on your PR to be landed, i have added few comments to minify changes. Many of APIs and consideration in Nuxt happened over time and some based on legacy requirements we don't need to Nitro...

Barbapapazes:
> Thanks for picking up on it and nice you could take inspiration from the Nuxt module.
> 
> If you would like to work on your PR to be landed, i have added few comments to minify changes. Many of APIs and consideration in Nuxt happened over time and some based on legacy requirements we don't need to Nitro...

Thanks! 🪄

I update this PR a little and allow modules to be specify in a `modules` folder. Before continuing, I would like to know if I should specify how end-user will write a module. In Nuxt, there is the object `meta` and the `nuxt-module-builder`. Should we have the same thing? How could we improve the module implementation to have a better usage?

pi0:
I expect as simple as this:

```ts
import type { NitroModule } from 'nitropack'
export default <NitroModule>{ async setup(nitro) {} }
```

Let's not keep ahead of ourselves. This feature is initially intended for advanced use cases not user-facing as Nuxt modules are until we can figure out use-cases :)

Barbapapazes:
> I expect as simple as this:
> 
> ```ts
> import type { NitroModule } from 'nitropack'
> export default <NitroModule>{ async setup(nitro) {} }
> ```
> 
> Let's not keep ahead of ourselves. This feature is initially intended for advanced use cases not user-facing as Nuxt modules are until we can figure out use-cases :)

So I think this PR is ready for a review!

src/build.ts
```diff
@@ -504,6 +504,8 @@ async function _watch(nitro: Nitro, rollupConfig: RollupConfig) {
     join(dir, "api"),
     join(dir, "routes"),
     join(dir, "middleware", GLOB_SCAN_PATTERN),
+    join(dir, "plugins"),
```
pi0, line 507: Please split unrelated changes to new PR

playground/nitro.config.ts
```diff
@@ -1 +1,14 @@
-export default defineNitroConfig({});
+import { defineEventHandler } from "h3";
+
+export default defineNitroConfig({
```
pi0, line 3: Please consider not pushing commits for playground changes. They are for local testing only

tsconfig.json
```diff
@@ -14,12 +17,28 @@
     "jsxFactory": "h",
     "jsxFragmentFactory": "Fragment",
     "paths": {
-      "nitropack": ["./src/index"],
-      "nitropack/config": ["./src/config"],
-      "#internal/nitro": ["./src/runtime/index"],
-      "#internal/nitro/*": ["./src/runtime/*"]
+      "nitropack": [
+        "./src/index"
```
pi0, line 21: Please consider moving unrelated changes like this + playground types to separate PRs to avoid blocking merges and allow more clear changelogs 🙏🏼

src/build.ts
Barbapapazes: Yes, you're right! https://github.com/unjs/nitro/pull/1800

src/module.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function resolveNitroModule(
+  mod: NitroModuleInput
+): Promise<NitroModule> {
+  if (typeof mod === "string") {
+    // @ts-ignore
+    globalThis.defineNitroModule =
+      // @ts-ignore
+      globalThis.defineNitroModule || defineNitroModule;
+    mod = (await import(mod).then((m) => m.default)) as NitroModule;
```
pi0, line 15: @Barbapapazes FYI, this works in nitro repo because in stup mode, jiti adds typescript support. We need explicit jiti if we wanted to support ts.

pi0 (approved):

PR #1805, opened 2023-10-07
fix(storage): validate and skip invalid mounts

### LINKED ISSUE #1804: Nitro fails to build when storage driver isn't given
### Environment

Nitro config (in nuxt.config.ts):
nitro:{
    storage:{
      'redis':{
      }
    }
  },
  Node version:
v18.16.0
Nitro version: latest (that is supported by nuxt)

### Reproduction

Putting this even on a blank project would error out the build process with the message of "Cannot find package 'undefined' imported from [nitro path]"

### Describe the bug

It appears as though it is trying to import a storage driver that doesn't exist, as it indeed was not specified, and exits out, stopping the entire build process. Moreover, since there is no proper error handling, the error message may be too confusing for new developers and wouldn't quite help them fix the issue, as they might think they're importing some faulty package.

### Additional context

_No response_

### Logs

_No response_

GalacticHypernova:
### 🔗 Linked issue

fix #1804
### ❓ Type of change

### 📚 Description

This change checks whether or not the driver was provided to the storage mount point, as `opts.driver` in the `createStorage` method is reading the driver. As a result, in previous versions, if no driver was provided, the build errors (which I have tested thoroughly) as it's trying to import a non-existing package with no proper error handling (non-existing due to the driver being "undefined"). This fixes that issue by not trying to import the driver if it wasn't specified, and in stead warning the user that the storage was not mounted, therefore allowing the build process to continue and succeed. In addition, this improves the DX as this error message provides more info than the previous default "Cannot find package 'undefined'", which may be confusing for new developers.
### 📝 Checklist


pi0 (approved):

pi0:
Thanks!

PR #1813, opened 2023-10-10
fix(cache): set cache item before returning response on first request

### LINKED ISSUE #1812: Cache item is not set before returning response
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      nitro
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Here is a reproduction in a Nuxt app that shows that running a cached request 3 times in a row **_sometimes_** returns different responses and **_sometimes_** not:
https://stackblitz.com/edit/nuxt-starter-blfmly

### Describe the bug

The first time a cached request is made, the handler should be awaiting the item to be set in the cache before returning a response, else the handler will be run multiple times (and re-cached multiple times) even when the value has already been fetched. I think in subsequent calls, this may not be necessary as the handler can return a stale value while renewing the cache in the background.

I believe this bug has to do with the introduction of the `waitUntil` utility here: 
https://github.com/unjs/nitro/blob/ae253596a847f764e275f7e0450ffdfcdcd8957e/src/runtime/cache.ts#L132C16-L132C16

### Additional context

_No response_

### Logs

_No response_

yassilah:
### 🔗 Linked issue

fixes #1812 

### ❓ Type of change

### 📚 Description

This PR fixes the linked issue; however, while writing my test I noticed there may be another issue with setting the `integrity` key in `defineCachedEventHandler` to an array of items as this will inevitably result in `entry.integrity === integrity` to be false... I changed it to be `hash([opts.integrity, handler])` and all other tests still pass but there may be something I'm not seeing here 😅

### 📝 Checklist


pi0:
Thanks for PR dear @yassilah. I will have to locally verify your reproduction from #1812. In the meantime, if there is another fix for integrity, feel free to split it into new PR for better changelog.

yassilah:
You're very welcome! You're right, i'll split it into 2 PRs 😊

pi0 (approved):
Thanks!

PR #1820, opened 2023-10-11
fix(cache): fix event handler integrity hash generation

yassilah:
### 🔗 Linked issue

In reference to: https://github.com/unjs/nitro/pull/1813#issuecomment-1757962340

### ❓ Type of change

### 📚 Description

This PR resolves an issue when setting cache integrity as an array.

### 📝 Checklist


yassilah:
> Thanks for PR dear @yassilah. I will have to locally verify your reproduction from #1812. In the meantime, if there is another fix for integrity, feel free to split it into new PR for better changelog.

Actually, taking a look at https://github.com/unjs/nitro/pull/1791 you're referencing, I don't think the change I made has any effect:
https://github.com/unjs/nitro/blob/ae253596a847f764e275f7e0450ffdfcdcd8957e/src/runtime/cache.ts#L54

You only allowed the `opts.integrity` to be value to check against, which is still the case in this PR. However, the issue came from this line, which makes `opts.integrity` an array:
https://github.com/unjs/nitro/blob/ae253596a847f764e275f7e0450ffdfcdcd8957e/src/runtime/cache.ts#L257

And that will necessarily fail here, since `entry.integrity` can only be a string:
https://github.com/unjs/nitro/blob/ae253596a847f764e275f7e0450ffdfcdcd8957e/src/runtime/cache.ts#L78

src/runtime/cache.ts
```diff
@@ -254,7 +254,7 @@ export function defineCachedEventHandler<
       return true;
     },
     group: opts.group || "nitro/handlers",
-    integrity: [opts.integrity, handler],
+    integrity: hash([opts.integrity, handler]),
```
pi0, line 257: (haven't locally checked the source code but we have to do something like this in order to allow completely overriding integrity - independent of handler's source hash)

```suggestion
    integrity: opts.integrity hash(handler),
```

yassilah: Ok, I see!

pi0 (approved):

Issue #1811, opened 2023-10-10
Netlify SWR New caching strategies


Rigo-m:
### Describe the feature

As of right now, routerules for netlify preset passes through netlify on demand builders, although this works flawlessly for pages without query parameters, it's not possible to cache or pass them onto nuxt pages.

This is solved via the new [netlify caching swr strategy](https://www.netlify.com/blog/swr-and-fine-grained-cache-control/)
I think this should be implemented in nitro.

### Additional information

- [X] Would you be willing to help implement this feature?

PR #1823, opened 2023-10-14
feat(netlify): added SWR implementation

Rigo-m:
### 🔗 Linked issue

resolves https://github.com/unjs/nitro/issues/1811

### ❓ Type of change

### 📚 Description

As stated in the issue, currently netlify "base" preset handles ISR by redirecting to either a normal function (no caching behaviour) or a builder function (cached on a per-URL basis). This approach has some downsides because of netlify ODB limitations (i.e.: no query parameters passed to the server, not so transparent cache control [specifically ttl parameter is quite hidden and not accessible directly])

This PR hopes to implement [new netlify caching swr strategy](https://www.netlify.com/blog/swr-and-fine-grained-cache-control/).

### A note on implementation:
I've arbitrarily implemented the SWR header when `isr` is set to true. This means that server response gets cached by up to 1 year in netlify edge cache servers, but gets revalidated each time. In contrast, setting a number in the `isr` routeRule defaults to the other caching method where page gets cached and not revalidated until cache expires.
Let me know if this behaviour is a good implementation of nitro routeRules API or needs to be changed.
### A note on documentation:
I'll document everything after changes gets approved

### 📝 Checklist

### DIFF
--- a/test/fixture/types.ts
+++ b/test/fixture/types.ts
@@ -1,6 +1,7 @@
 import { expectTypeOf } from "expect-type";
 import { describe, it } from "vitest";
 import { EventHandler, EventHandlerRequest, defineEventHandler } from "h3";
+// @ts-ignore
 import { $Fetch } from "../..";
 import { defineNitroConfig } from "../../src/config";
 
--- a/src/presets/netlify.ts
+++ b/src/presets/netlify.ts
@@ -169,29 +169,6 @@ async function writeHeaders(nitro: Nitro) {
     (a, b) => b[0].split(/\/(?!\*)/).length - a[0].split(/\/(?!\*)/).length
   );
 
-  // convert isr rules to header rules
-  for (const [path, routeRules] of rules.filter(
-    ([_, routeRules]) => routeRules.isr !== undefined
-  )) {
-    // create headers if missing
-    nitro.options.routeRules[path].headers ||= {};
-    // transform isr rules to SWR headers https://www.netlify.com/blog/swr-and-fine-grained-cache-control/
-    if (routeRules.isr === true || routeRules.isr === 0) {
-      // set SWR header
-      nitro.options.routeRules[path].headers["Cache-Control"] =
-        "public, max-age=0, must-revalidate"; // Tell browsers to always revalidate
-      nitro.options.routeRules[path].headers[
-        "Netlify-CDN-Cache-Control"
-      ] = `public, max-age=0, stale-while-revalidate=31536000`; // Tell Edge to cache asset for up to a year, but revalidate every call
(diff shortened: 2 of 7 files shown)


Rigo-m:
As stated above, this is not yet tested but is a first implementation straight out of my head. Need help with testing properly in a netlify environment

Rigo-m:
Was able to setup a test environment here https://nitro-deploy-swr-test.netlify.app/api/hello?querytest=test3 by forking nitro-deploys repository.

First attempt at implementing the feature was a failure, apparently Netlify disregard _headers file when setting cache control headers.

Manually using h3 `setHeader` inside of `defineEventHandler` worked.
I'm currently trying to set headers at the entry level, but for now unsuccessfully. Can you help me on that?

Rigo-m:
I've added headers to the generated localCall response instead of appending them to the incoming request event.
This works on my tests, not sure if this is the correct implementation because I didn't quite understood how request headers gets transformed/used by nitro localCall handling.

Gonna fix test suite and mark this PR as ready for review

test/fixture/types.ts
```diff
@@ -1,6 +1,7 @@
 import { expectTypeOf } from "expect-type";
 import { describe, it } from "vitest";
 import { EventHandler, EventHandlerRequest, defineEventHandler } from "h3";
+// @ts-ignore
```
Rigo-m, line 4: Tests failed on me here, added ts-ignore. Didn't know if I had to run some sort of `prepare` to fix this issue. Remove if unnecessary

Rigo-m:
A note on `netlify-edge` preset: I've added the same headers to the response, did not test them for now. It's not clear from the article whether edge functions caching strategies works the same as in normal functions.

Rigo-m:
ATM, Cache headers behaviour is not as predictable as ODB behaviour. Meaning it doesn't return stale content as you would expect.
Opened an issue (here)[https://github.com/netlify-labs/cache-control-and-swr/issues/1]
 and a support ticket (here)[https://answers.netlify.com/t/netlify-new-swr-cache-policy-behaviour-is-not-clear/104747]
 
I'll keep you up to date. In the meantime, I'd appreciate a review on the code written :)

danielkellyio:
The reason the cache headers behavior isn't predictable seems to be based on different CDN nodes handling different requests:
https://answers.netlify.com/t/cannot-get-netlify-api-endpoint-caching-to-work-as-expected/105763/7

So I think the SWR support from Netlify isn't quite up to par yet.

pi0:
Thanks for your work on this @Rigo-m ❤️ Moving to #2406

Rigo-m:
Thanks for the heads up @pi0 💘

PR #1838, opened 2023-10-18
fix(cache): allow setting multiple `set-cookie` headers (bad practice)

### LINKED ISSUE #1837: cachedEventHandler overrides middleware cookies
### Environment

see codesandbox

### Reproduction

https://codesandbox.io/p/sandbox/interesting-cohen-pnq9y5

notice how / gets two set-cookie headers while /test only gets one

### Describe the bug

when a middleware and a route both define cookies and the  route cookie overrides the middleware cookie instead of just appending the header

### Additional context

the overriding happens here: https://github.com/unjs/nitro/blob/be6bb7ea136c464ab28eb8662eb630e11c7b0b68/src/runtime/cache.ts#L407
the setHeader should be changed to appendHeader (maybe only for the set-cookie case?)
h3 does it like this https://github.com/unjs/h3/blob/b7aca9614e7e8e8921fd54804463ce78fdb034f1/src/utils/response.ts#L341

if a consensus on a solution is reached, am open to PR it.

### Logs

_No response_

MiniDigger:
### 🔗 Linked issue

#1837

### ❓ Type of change

### 📚 Description

cookie headers can be repeated, adapted from 
https://github.com/unjs/h3/blob/b7aca9614e7e8e8921fd54804463ce78fdb034f1/src/utils/response.ts#L341

I tried writing a test case for this but stuff didnt work as expected (doesnt help that my work PC is kinda locked down and on windows...), but I tested via playground and in my original nuxt project

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -5,6 +5,7 @@ import {
   createEvent,
   EventHandler,
   isEvent,
+  splitCookiesString,
 } from "h3";
 import type { EventHandlerRequest, EventHandlerResponse, H3Event } from "h3";
 import { parseURL } from "ufo";
@@ -404,7 +405,12 @@ export function defineCachedEventHandler<
     // Send status and headers
     event.node.res.statusCode = response.code;
     for (const name in response.headers) {
-      event.node.res.setHeader(name, response.headers[name]);
+      const value = response.headers[name];
+      if (name === "set-cookie") {
+        event.node.res.appendHeader(name, splitCookiesString(value as string[]));
+      } else {
+        event.node.res.setHeader(name, value);
+      }
     }
 
     // Send body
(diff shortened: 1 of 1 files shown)


src/runtime/cache.ts
```diff
@@ -404,7 +405,15 @@ export function defineCachedEventHandler<
     // Send status and headers
     event.node.res.statusCode = response.code;
     for (const name in response.headers) {
-      event.node.res.setHeader(name, response.headers[name]);
+      const value = response.headers[name];
+      if (name === "set-cookie") {
+        event.node.res.appendHeader(
```
pi0, line 410:
```suggestion
        // TODO: Show warning and remove this header in the next major version of Nitro
        event.node.res.appendHeader(
```

MiniDigger:
Mmh, in my case the cookie is set based on the path, not the user, so caching it is fine, appreciate the heads-up tho!

Issue #1757, opened 2023-09-23
offline swagger/scalar ui


dargmuesli:
### Describe the feature

Elevating https://github.com/unjs/nitro/pull/1162/files#r1173842179 to an issue:

It would be great if `/_nitro/swagger` could make use of a locally available `swagger-ui-dist` instead of pulling it from `jsdeliver.net`.

### Additional information

- [ ] Would you be willing to help implement this feature?

pi0:
Makes sense to allow offline support i am wondering if we can support it via an optional dependency.

Also it is not definite but we might replace swagger UI with something from unjs ;)

tobychidi:
Yes. I would love something from UnJS

pi0:
pending this idea for now as we need to find a better solution that has no major install size issues and works with scalar endpoint too.

pi0:
Moving to https://github.com/unjs/openapi-renderer/issues/4 as nitro plans to eventually also migrate to this new library.

PR #1840, opened 2023-10-19
feat: use pre-bundled `swagger-ui-dist`

HigherOrderLogic:
### 🔗 Linked issue

Resolve https://github.com/unjs/nitro/issues/1757.

### ❓ Type of change

### 📚 Description

Use pre-bundled package `swagger-ui-dist` for offline usage.

### 📝 Checklist


septatrix:
Having this as a non-optional dependency is a drawback in terms of bundle size as not everyone will need this. I think the better way would be to allow customization of the URL/endpoint (or add a section to the documentation of how to add a UI route and remove the UI from the builtin routes altogether). In the end there are also alternative UIs like redoc and scalar which people want to use (and the [comment from @pi0](https://github.com/unjs/nitro/issues/1757#issuecomment-1735452600) that unjs *might* develop something similar in the future as well)

HigherOrderLogic:
Swagger route is only added in dev build, hence the production bundle size wont be affected.

septatrix:
> Swagger route is only added in dev build, hence the production bundle size wont be affected.

But the download size will. It an additional 10MB on top of what currently seems to be ~180MB for nuxt itself. This slows down installation, CIs, impact people with metered bandwidth etc. – all for something which many people might never use.

pi0:
Thanks for PR. Keeping this idea on hold since there are install size concerns + we are introducing new uis (`/_nitro/scalar`)

PR #1843, opened 2023-10-20
feat: support `framework` config

pi0:
### ❓ Type of change

### 📚 Description

This adds a new advanced `framework` configuration allowing meta-frameworks such as Nuxt to declare their meta-data that can be passed to the underlying presets or reflect nitro behavior when needed. The default is `framework: { name: 'nitro', version }` when using nitro CLI/API directly.

**Update:** Draft status is to check if we might need "multi" framework information (having version info about both nitro and nuxt for example)

### 📝 Checklist


danielroe (approved):
Nice idea! What are you thinking that presets will do with this info? I'm sure you've thought about it, but is there a danger that this ends up coupling Nitro presets too closely to upstream frameworks?

pi0:
I think the main reason would be displaying info in output meta. I also agree that there should be absolutely no differences when coupled with an upstream framework to have a platform behavior other than displaying.

PR #1870, opened 2023-10-30
feat: allow setting preset using `SERVER_PRESET` environment variable

### LINKED ISSUE #1860: Support `SERVER_PRESET` env to allow unbranded usage of Nitro
Today we allow setting a nitro production build preset by using either `NITRO_PRESET` environment variable (used in official [deployment docs](https://nitro.unjs.io/deploy/#changing-the-deployment-preset)), `nitro build --preset` or `preset` config in `nitro.config.ts`.

When it comes to Nitro usage via meta-frameworks such as [Nuxt 3](https://nuxt.com), or [Analog](https://analogjs.org/), I expect a consistent experience so that when users read Nitro [deploy docs](https://nitro.unjs.io/deploy/#changing-the-deployment-preset) they can just use same env and instructions for any framework powered by Nitro.

Using the `NITRO_` prefix is to explicitly tell users that they are configuring something related to Nitro and if something goes wrong, where they can refer to or search Google/docs about.

------

Now today, frameworks are starting to consider "debranded" alternative ways to set Nitro preset. I guess this has the benefit of hiding complexities from users by (meta-framework) abstraction and less confusing them to not bother about learning nitro, if more benefits, I am unaware.

Today, Analog [uses `BUILD_PRESET`](https://analogjs.org/docs/features/deployment/providers#setting-the-preset-manually) to set preset and Nuxt is discussing going with the same `BUILD_PRESET` approach (https://github.com/nuxt/nuxt/issues/23996).

I am personally opposed to this direction because I believe having explicit names can be much more helpful for users to trace where a feature comes from and learn more about it rather than hiding traces by debranding (and also that it hurts the ecosystem and Nitro recognition by intentionally _hiding_ it).

With all these, and above all, I believe in freedom of choice in OpenSource and this had been the mindset in the past when developing Nuxt 3 base (that today is led by @danielroe) and today most of UnJS projects and Nitro. 

This proposal is for the sake of ecosystem consistency. Now that it is proven that meta framework authors might prefer to advise for debranding, at least deciding to introduce a standard name in one place and allowing to adopt it might be a better choice. 

Also about naming, I am proposing `SERVER_PRESET` since my goal is to increase adopting this convention (when needed) in other places and APIs such as `server` dir, `SERVER_*` env variables, `server.config.ts` (just an idea)

pi0:
### 🔗 Linked issue

Resolves #1860

### ❓ Type of change

### 📚 Description

This allows setting nitro preset using `SERVER_PRESET` instead of `NITRO_PRESET`.

### 📝 Checklist

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -138,7 +138,9 @@ export async function loadOptions(
 ): Promise<NitroOptions> {
   // Preset
   let presetOverride =
-    (configOverrides.preset as string) || process.env.NITRO_PRESET;
+    (configOverrides.preset as string) ||
+    process.env.NITRO_PRESET ||
+    process.env.SERVER_PRESET;
   if (configOverrides.dev) {
     presetOverride = "nitro-dev";
   }


Atinux (approved):
Looks good to me 😊

PR #1884, opened 2023-11-04
docs(routing): add note about middleware returns

oof2win2:
Better describe middleware and how returning within it behaves

### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)

### 📚 Description

This improves on documentation. Before it stated that **they should not return anything**, which is untrue - returning from middleware returns the value as a response.

### 📝 Checklist

### DIFF
--- a/docs/content/1.guide/3.routing.md
+++ b/docs/content/1.guide/3.routing.md
@@ -172,6 +172,7 @@ nitro.config.ts
 ### Simple Middleware
 
 Middleware are defined exactly like route handlers with the only exception that they should not return anything.
+Returning from middleware behaves like returning from a request - the value will be returned as a response and further code will not be ran.
 
 ```ts [middleware/auth.ts]
 export default defineEventHandler((event) => {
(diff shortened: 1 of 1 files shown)


docs/content/1.guide/3.routing.md
```diff
@@ -181,7 +182,7 @@ export default defineEventHandler((event) => {
 ```
 
 ::alert
-Returning anything from a middleware will close the request and should be avoided!
+Returning anything from a middleware will close the request and should be avoided! Any returned value from middleware will be the response and further code will not be executed.
```
pi0, line 185:
```suggestion
Returning anything from a middleware will close the request and should be avoided! Any returned value from middleware will be the response and further code will not be executed however **this is not recommended to do!**
```

pi0 (approved):
Thanks!

oof2win2:
Why is this not recommended to do? Unauthenticated reauests can be ended before they reach other handlers with this.

pi0:
Yes, also this breaks the nitro request handling lifecycle (i am planning also to introduce `defineMiddlewareHandler` in h3 that essentially disallows return value as a convention)

oof2win2:
Oh okay. Is there any method that would allow for this, to just directly respond to unauthenticated requests with 401s in one place?

PR #1893, opened 2023-11-08
feat: allow setting `rewrite` option for devProxy

### LINKED ISSUE #1892: devProxy need rewrite option to rewrite request path
### Describe the feature

In my case, I'm putting all server requests under `/api` then I rewrite it and remove api path and dispatch the request to the actual server.

Here's how vite is does it and we need something similar
https://github.com/vitejs/vite/blob/987b8fadc176434b1b18c5dcf15a8cefe434e500/packages/vite/src/node/server/middlewares/proxy.ts#L130

### Additional information

gliheng:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1892

### ❓ Type of change

### 📚 Description

This PR adds a path rewrite option to devProxy config.
This is useful in case that user wants to change request path, before the request is sent to upstream server.

Resolves #1892 

### 📝 Checklist

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -5,7 +5,6 @@ import {
   App,
   createApp,
   eventHandler,
-  getRequestPath,
   fromNodeMiddleware,
   H3Error,
   H3Event,
@@ -171,8 +170,8 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
       eventHandler(async (event) => {
         const { rewrite } = opts;
         if (rewrite) {
-          const path = getRequestPath(event);
-          event.node.req.url = rewrite(path);
+          event.path = rewrite(event.path);
+          event.node.req.url = event.path;
         }
         await proxy.handle(event);
       })
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -5,6 +5,7 @@ import {
   App,
   createApp,
   eventHandler,
+  getRequestPath,
   fromNodeMiddleware,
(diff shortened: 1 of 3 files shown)


src/dev/server.ts
```diff
@@ -168,6 +169,11 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
     app.use(
       route,
       eventHandler(async (event) => {
+        const { rewrite } = opts;
+        if (rewrite) {
+          const path = getRequestPath(event);
+          event.node.req.url = rewrite(path);
```
pi0, line 175: We might need to rewrite `event.path` too (if have too).

Also consider we have a built-in option for http-proxy to rewrite path as far as i remember will check.

gliheng: I just checked nodejs doc. node.req is a IncomingMessage and path is not a property of it.

gliheng: Here's what I found on http-proxy: http-proxy-middleware has a pathRewrite option, but it's not on http-proxy.

gliheng:
event.path is also updated now

pi0:
Thanks! Sorry for delay. I am thinking to move this feature directly to httpxy.

gliheng:
OK, close this PR if you want.

pi0:
(sorry for delay, rewrite should be implemented in upstream but we might also type it in nitro + consider a serialzable rewrite for production, that's why keeping it open)

PR #1897, opened 2023-11-08
fix: generate route typed with resolve path and stripped extension

### LINKED ISSUE #1378: nitro-routes not generated properly for filename separated by dots like handler.get.ts
### Environment

nitropack 2.5.2
node 18.15.0

### Reproduction

No reproduction, bug is easy to reproduce.

### Describe the bug

I found the issue using a nuxt module after exposing an event handler from a file named `assets.get.ts`
```javascript
 addServerHandler({
    handler: resolve('./runtime/server/assets.get'),
    route: '/api/assets',
    method: 'get',
  })
```
The resulting `nitro-routes.d.ts` generated is
```javascript
// Generated by nitro
import type { Serialize, Simplify } from 'nitropack'
declare module 'nitropack' {
  type Awaited<T> = T extends PromiseLike<infer U> ? Awaited<U> : T
  interface InternalApi {
   ...
'/api/assets': {
      'get': Simplify<Serialize<Awaited<ReturnType<typeof import('../../../src/runtime/server/assets').default>>>>
    }
```
It's missing the `.get` from the import.

The fix is easy, I just renamed the file `assets-get.ts` but I wanted to keep the same naming convention in my nuxt module as in a Nuxt application using the `.get,.post,..` filename convention.

### Additional context

_No response_

### Logs

_No response_

pi0:
### 🔗 Linked issue

Resolves #1378

### ❓ Type of change

### 📚 Description

We were not resolving event handler paths for generating types and also removing any last `.*` segment. So if input is `~/foo.get`, generated type would be something like `../~/foo`.

This PR both resolves paths and strips only known extension from last (it is unnecessary but as another safety measure)

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -19,7 +19,12 @@ import {
 } from "mlly";
 import { generateFSTree } from "./utils/tree";
 import { getRollupConfig, RollupConfig } from "./rollup/config";
-import { prettyPath, writeFile, isDirectory } from "./utils";
+import {
+  prettyPath,
+  writeFile,
+  isDirectory,
+  resolvePath as resolveNitroPath,
+} from "./utils";
 import { GLOB_SCAN_PATTERN, scanHandlers } from "./scan";
 import type { Nitro } from "./types";
 import { runtimeDir } from "./dirs";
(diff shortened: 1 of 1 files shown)


danielroe (approved):
Nice

PR #1172, opened 2023-04-19
feat(cloudflare-module, cloudflare-pages): experimental dynamic imports

danielroe:
### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This adds support for dynamic imports with CF modules following https://github.com/cloudflare/workers-sdk/pull/2769. (Feature coming soon to CF Pages too: merged via https://github.com/unjs/nitro/pull/1531 by @dario-piotrowicz to this PR)

Experiment can be enabled by setting `NITRO_EXP_CLOUDFLARE_DYNAMIC_IMPORTS` environment variable.

Pending: https://github.com/cloudflare/workers-sdk/pull/3707

TODO:
- [ ] `__STATIC_CONTENT_MANIFEST` import support
- [x] update docs with `rules` info (and console warning)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1905, opened 2023-11-11
feat(cloudflare-pages, cloudflare-module): enable code splitting by default

dario-piotrowicz:
### ❓ Type of change

### 📚 Description

remove the NITRO_EXP_CLOUDFLARE_DYNAMIC_IMPORTS flag (added in https://github.com/unjs/nitro/pull/1172) used to enable code splitting with lazy loading for the cloudflare-pages and cloudflare_module presets and enable such behavior by default since that is not stable and working with up-to-date versions of wrangler

For the stability of this feature please have a look at these reproductions: https://github.com/dario-piotrowicz/nitro-nuxt-lazy-loading-test/tree/main

They show that things do work fine (both locally and in production) when using lazy loading :muscle: 

### 📝 Checklist

### DIFF
--- a/docs/content/2.deploy/20.providers/cloudflare.md
+++ b/docs/content/2.deploy/20.providers/cloudflare.md
@@ -22,6 +22,10 @@ compatibility_date = "2022-09-10"
 account_id = "<the account_id you obtained (optional)>"
 route = "<mainly useful when you want to setup custom domains (optional too)>"
 
+rules = [
+  { type = "ESModule", globs = ["**/*.js", "**/*.mjs"]},
+]
+
 [site]
 bucket = ".output/public"
 ```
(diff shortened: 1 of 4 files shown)


docs/content/2.deploy/20.providers/cloudflare.md
```diff
@@ -22,6 +22,10 @@ compatibility_date = "2022-09-10"
 account_id = "<the account_id you obtained (optional)>"
 route = "<mainly useful when you want to setup custom domains (optional too)>"
 
+rules = [
+  { type = "ESModule", globs = ["**/*.js", "**/*.mjs"]},
+]
+
```
dario-piotrowicz, line 28: I've added those here to simplify the documentation and not requiring people to set them only when using the cloudflare_module preset.

The the service worker output these are not going to do anything but they are not breaking anything either (as you can see from this [example](https://github.com/dario-piotrowicz/nitro-nuxt-lazy-loading-test/tree/main/nitro-no-lazy-loading-workers-test)), that's why I think it's simpler to just include them here, but let me know what you think, I don't mind moving it into the cloudflare_module preset section.

test/presets/cloudflare-pages.test.ts
```diff
@@ -12,7 +12,7 @@ describe("nitro:preset:cloudflare-pages", async () => {
   testNitro(ctx, () => {
     const mf = new Miniflare({
       modules: true,
-      scriptPath: resolve(ctx.outDir, "_worker.js"),
+      scriptPath: resolve(ctx.outDir, "_worker.js", "index.js"),
```
pi0, line 15: Do we always need a path like this as `.../_worker.js/index.js`? (double extension seems strange but fine if it is only allowed convention)

pi0 (approved):
Lets try on nightly channel!

pi0:
This is huge 🔥

test/presets/cloudflare-pages.test.ts
dario-piotrowicz: yeah it's the convention unfortunately :sweat: 

later this might be configurable but for now it has to be like this :sweat:

PR #1906, opened 2023-11-12
fix: initialize nitro app before plugins run

harlan-zw:
### ❓ Type of change

### 📚 Description

This is more of a proof-of-concept to highlight the issue and maybe offer a solution, feel free to close for your own solution.

The issue is that when a plugin wants to make an internal fetch, if that endpoint is using `useNitroApp`, then it will fail. This is because plugins are ran without the nitroApp context being available.

I've got around it previously using a timeout "hack" (https://github.com/harlan-zw/nuxt-link-checker/commit/646e6ff8721eaeb4ac4c56a3a0e03c0ce764f90e), but as has been pointed out, this isn't exactly reliable (https://github.com/harlan-zw/nuxt-simple-sitemap/issues/149#issuecomment-1806885881).

### 📝 Checklist


pi0:
Thanks for making PR. Although it is slightly a breaking/risky behavior change i was considering to do similar change also in order to allow async plugins. I will carefully review it soon to make sure we won't break anything before moving forward.

pi0:
(delaying to next minor)

pi0 (approved):

pi0:
Thanks you!!

PR #1914, opened 2023-11-15
fix(prerender): decode generated routes

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22660
closes https://github.com/nuxt/nuxt/pull/24320

### ❓ Type of change

### 📚 Description

When a link is rendered in HTML (e.g. by vue router) with URI components, Nitro currently does not decode these when requesting the URL, which can cause double encoding and a 404 from vue-router.

Instead we should likely decode the link at source, just as we do with headers.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -380,7 +380,7 @@ function extractLinks(
   if (crawlLinks) {
     _links.push(
       ...[...html.matchAll(LINK_REGEX)]
-        .map((m) => escapeHtml(m[1]))
+        .map((m) => decodeURIComponent(escapeHtml(m[1])))
         .filter((link) => allowedExtensions.has(getExtension(link)))
     );
   }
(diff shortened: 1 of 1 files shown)


pi0:
Can you please also add a test? 🙏🏼 (i am also not sure about this because we might wrongly decode/change original links. we could make it opt-in for nuxt as vue-router supports encoding for example but test case would be useful)

danielroe:
The source of the double encoding is from this line:

https://github.com/unjs/nitro/blob/466535abcce71bd6c39a7fd7a698da4bbd2f7059/src/prerender.ts#L168-L169

I think if we are forcibly going to encode the URI it makes sense to decode scanned links first.

danielroe:
Okay, pushed a change that means `route` is always decoded. This covers both scanned links and link headers, as well as ensuring that we write the decoded version to the file system. ✅

pi0 (approved):

PR #1921, opened 2023-11-17
fix(error): respect `accept: text/html` request header

### LINKED ISSUE #1919: The built-in error handler over-assumes json request type
### Environment

Affects latest version of nitro, the project I am working on has `nitropack@2.7.2`.

### Reproduction

[Stackblitz minimal example](https://stackblitz.com/edit/github-c7srhy?file=public%2Findex.html,routes%2Ferror.ts)

### Describe the bug

Preconditions:
- An error occurs during request handling (e.g. `throw createError({statusCode:500,statusMessage:'SomeError'})`)
- The default nitro error handler is used.
- The request is made with the Fetch API; but really: the request has `sec-fetch-mode` set to `cors`.

Then the request over-assumes that the request is JSON, and sending an `Accept` header set to `text/html` does not coerce the response to HTML, it remains JSON.

### Additional context

[The related check in `isJsonRequest`](https://github.com/unjs/nitro/blob/a3120b6acfcf67343e411e51107c45d6ccf432b8/src/runtime/utils.ts#L47C26-L47C40)

### Logs

```sh
Not relevant
```

passionate-bram:
Requests made with the Fetch API cause `sec-fetch-mode` to be set, leading to these requests to be identified as JSON requests. The default error handler uses isJsonRequest to render errors. If a request were to set the Accept header to "text/html" the response should be HTML as well.

### 🔗 Linked issue

#1919 

### ❓ Type of change

### 📚 Description

`isJsonRequest` is modified, its only usage (within nitro) is by the default event handler to determine when to serialize errors as HTML or JSON.

This change is needed because requests made with the Fetch API automatically get the `sec-fetch-mode=cors` header. This header is one of the conditions by which `isJsonRequest` function identifies JSON, but the header has no such meaning as far as the standard is concerned.

More importantly, the function does not provide any means to enforce an HTML response (for `isJsonRequest` to return false) when these fallback detections are used.

Resolves #1919 

### 📝 Checklist

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -40,15 +40,12 @@ export function hasReqHeader(event: H3Event, name: string, includes: string) {
 }
 
 export function isJsonRequest(event: H3Event) {
-  // First adhere to the "Accept" header, to match what the client expects to receive.
-  if (hasReqHeader(event, "accept", "application/json")) {
-    return true;
-  }
+  // If the client specifically requests HTML, then avoid classifying as JSON.
   if (hasReqHeader(event, "accept", "text/html")) {
     return false;
   }
-  // Fallback rules, when no explicit request for the content type is made.
   return (
+    hasReqHeader(event, "accept", "application/json") ||
     hasReqHeader(event, "user-agent", "curl/") ||
     hasReqHeader(event, "user-agent", "httpie/") ||
     hasReqHeader(event, "sec-fetch-mode", "cors") ||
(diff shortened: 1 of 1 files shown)


src/runtime/utils.ts
```diff
@@ -40,8 +40,15 @@ export function hasReqHeader(event: H3Event, name: string, includes: string) {
 }
 
 export function isJsonRequest(event: H3Event) {
+  // First adhere to the "Accept" header, to match what the client expects to receive.
+  if (hasReqHeader(event, "accept", "application/json")) {
+    return true;
+  }
+  if (hasReqHeader(event, "accept", "text/html")) {
+    return false;
```
pi0, line 48: I think addition of this condition should be enough and we can keep checking application/json in fallback as before. If it is set to HTML, we early return.

pi0 (approved):
Thanks!

PR #1944, opened 2023-11-22
feat(node-server): add env variables for timeout setting in node-server runtimes

lastmirage:
### ❓ Type of change

### 📚 Description

When deploy node runtime to behind AWS ELB, timeout missmatching can leads to 502 error.
For resolve this error, we have to set timeout values in HttpServer.

### 📝 Checklist


pi0:
Thanks for this PR. It makese sense to allow supporting more customizations for Node runtime server but adding more as direct env variables might be a tricky task to maintain.

I am pending PR to think little more what would be alternative.

unite4:
Hi @pi0, do you have an idea for a better solution than the one provided by @lastmirage?

joaoaraujo-hotmart:
@pi0 any chance to have an update here? Thanks!

vsergiu93:
Hey @pi0, first of all I want to thank you for your effort and time you put in to all of this, I know it is not easy, the web space became so complicated, we have now different runtimes, so many ways to deploy apps, so many infra variations, and I guess making sure all of this works is not an easy task, again thank you. 
I just want to ask you about this issue, do you guys plan to add something in the next major release that will allow us to configure this values, either how is proposed here or maybe in a different way?

pi0:
I am thinking to finally land it for v2 at least (but need to cleanup changes little bit before merge).

(this is what working on for Nitro v3: [spoiler](https://srvx.unjs.io/) - i hope to find a better standard away to set these regardless of runtime and also framework so it won't be nitro specific convention)

PR #1961, opened 2023-11-28
feat: migrate docs to ui-pro

Flosciante:


Barbapapazes:
Hello @Flosciante,

I'm Estéban, maintainer of unjs/website and leading marketing of UnJS.

Thank you so much for starting the migration from docus to nuxt ui pro.

I'd like to keep the documentations in line with the same idea and design, to have harmony within UnJS. Here are a few points for improvement:

- Update the font to use Nunito, https://unjs.io/design-kit#typography
- Update the gray to cool
- Update the header and the footer to have something more similar to new UnJS website

We could also update the tagline by adding "Nuxt, Analog and SolidStart".

What do you think?

> [!NOTE]
> Next version of UnJS website: https://feat-new-theme.unjs.pages.dev/

pi0:
> We could also update the tagline by adding "Nuxt, Analog and SolidStart".

Probably not.

pi0:
We need to rebase content although i think after moving to new unjs-docs CLI we might need less of changes as we can keep `# h1` headings.

PR #1962, opened 2023-11-28
fix(prerender): avoid detecting `.json` routes as implicit html

pi0:
### ❓ Type of change

### 📚 Description

On investigation, there are cases where a framework renderer, returns `text/html` responses for routes ending with `.json` extension. Content is most likely also wrong (it does not fix https://github.com/nuxt/nuxt/issues/24487 only generates `manifest.json` with HTML in it!) but at least avoids making wrong structure.

(root cause is that Nuxt ordering is different and first prerenders and then copies public assets)

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -220,7 +220,9 @@ export async function prerender(nitro: Nitro) {
     // Guess route type and populate fileName
     const contentType = res.headers.get("content-type") || "";
     const isImplicitHTML =
-      !route.endsWith(".html") && contentType.includes("html");
+      !route.endsWith(".html") &&
+      !route.endsWith(".json") &&
+      contentType.includes("html");
     const routeWithIndex = route.endsWith("/") ? route + "index" : route;
     const htmlPath =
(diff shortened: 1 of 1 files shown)


src/prerender.ts
```diff
@@ -220,7 +220,9 @@ export async function prerender(nitro: Nitro) {
     // Guess route type and populate fileName
     const contentType = res.headers.get("content-type") || "";
     const isImplicitHTML =
-      !route.endsWith(".html") && contentType.includes("html");
+      !route.endsWith(".html") &&
+      !route.endsWith(".json") &&
```
danielroe, line 224: We could also add `.xml` perhaps...

pi0: Yeah, we could make a list also (for like `.webappmanifest`) but consider this is mainly a hotfix for when content-type is also wrongly text/html...

danielroe: A safelist is probably not a bad idea given this is only for _implicit_ content type...

pi0:
https://github.com/unjs/nitro/pull/1963

Issue #1894, opened 2023-11-08
Caching fails when returning a Buffer (PNG image)


jschroeter:
### Environment

Node: 18.18
Nitro: latest

### Reproduction

1. open [Stackblitz](https://stackblitz.com/edit/nitro-cache-buffer?file=routes%2Fimage.ts,routes%2FcachedImage.ts)
2. open dev tools and disable cache in network tab
3. click the reload button in the preview multiple times

Expected: 2 logos should be visible

Actual result: the second logo using the cached route breaks when reloading in between 5s (since the cache max age is set to 5s)

### Use case
I do want to return a dynamically generated PNG image in a Nitro route. This works fine in general.

### Issue
I also would like to add caching using `cachedEventHandler`. But all requests that should return a cached response, return the serialised JSON instead of the binary data, e.g. `{"type": "Buffer", "data": [137,80,78,71,13,10,26,10,0,...`

### Additional context

I guess it's related to https://github.com/unjs/unstorage/issues/142 and the Nitro Cache should use `setItemRaw()`/`getItemRaw()` for the case of a Buffer?

bernhardberger:
I can confirm this issue. It breaks SWR when using sharp on a node server. Interestingly though this issue isn't present when using ISR (which supposedly as per docs is the same as SWR with just some cache headers added for Netlify and Vercel?!)

jryd:
We're experiencing this now also after swapping `defineEventHandler` for `defineCachedEventHandler`.

We run an API endpoint in Nuxt that proxies image requests to a protected/private source. We want to cache the response, but after using this we sporadically see the serialised JSON returned instead.

alexMugen:
I have exactly the same issue

alexMugen:
Any updates forthcoming?

alexMugen:
Any updates ?

f-lawe:
Experiencing the same issue here too. Is there a way to wrap up the response of `defineCachedEventHandler` so we can eventually process the serialized JSON into binary data when it occurs? The overall response won't be fully cached, but at least the proxied image will be.

f-lawe:
So, we implemented the workaround I mentioned earlier: we have a regular `defineEventHandler` endpoint that calls our image cached with `defineCachedEventHandler` and converts the serialized JSON back to a buffer. The overall response is not cached by the browser, however we reduced our response time and made it flat (we used to have huge variations, depending on server load),

That's not ideal, but that's an improvement, and it works!

[...]

PR #1969, opened 2023-11-30
fix: encode buffer responses from cached event handler

harlan-zw:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1894

### ❓ Type of change

### 📚 Description

Caching an endpoint that returns a buffer will break the endpoint. As an easy fix, we just base 64 encode whatever the result is and mark it as so.

### 📝 Checklist

### DIFF
--- a/src/runtime/cache.ts
+++ b/src/runtime/cache.ts
@@ -44,16 +44,6 @@ const defaultCacheOptions = {
   maxAge: 1,
 };
 
-function base64ToArray(base64: string) {
-  const str = atob(base64);
-  const bytes = new Uint8Array(str.length);
-  for (let i = 0; i < str.length; i++) {
-    // eslint-disable-next-line unicorn/prefer-code-point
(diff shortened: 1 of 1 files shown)


pi0:
As of discussion in #1894, I think we might use raw methods of unstorage (behind a flag) also unstorage should automatically handle base64 transformation for drivers not supporting natively binary data if it does not it is something to fix there.

harlan-zw:
> As of discussion in #1894, I think we might use raw methods of unstorage (behind a flag) also unstorage should automatically handle base64 transformation for drivers not supporting natively binary data if it does not it is something to fix there.

I agree, perhaps we can merge this as a temporary fix until that's worked on? Again, the current behavior is broken so kind of important.

pi0:
(was hoping to rebase with latest [undio](https://undio.unjs.io/) version to support at least all h3 binary types including streams. might delay to prioritize other reactors sorry for this delay it is an important issue)

PR #1976, opened 2023-12-01
feat(routeRules): allow wildcard redirects

### LINKED ISSUE #1748: Wildcard redirects
### Describe the feature

If the target URL ends with `/**`, treat it as wildcard redirect and append the subpath properly.

```ts
  routeRules: {
    '/single/': { redirect: { to: '/single-2/', statusCode: 301 } }, // Works 
    '/wildcard/**': { redirect: { to: '/single/', statusCode: 301 } }, // Works
    '/wildcard/**': { redirect: { to: '/wildcard-2/**', statusCode: 301 } }, // Does not work but would be nice
  },
```

`wildcard/a` -> `wildcard-2/a`

### Additional information

JaZo:
### 🔗 Linked issue

#1748

### ❓ Type of change

### 📚 Description

This PR adds support for wildcard redirects in routeRules:
```
  routeRules: {
    '/wildcard/**': { redirect: { to: '/wildcard-2/**', statusCode: 301 } },
  },
```

It also keeps query parameters, just like the proxy does.

Resolves #1748.

### 📝 Checklist


JaZo:
Ahh the presets, yes! Forgot about them as we don't use any of those. I'll see if I can add this, but that will probably take a while.

pi0:
Good point regarding platform support. It is not necessary however we can iterate and support it later in the future as mainly nitro server should be able to (universally) catch any platform-unsupported patteren.

manniL:
@pi0 I agree. Happy to add it later and let nitro handle it for now 👍🏻

Anoesj:
Nice! It would be fantastic if this could be merged soon. Adding some tests would probably be a good idea.

JaZo:
I've added tests and implemented Netlify and Vercel as these tests failed!

pi0 (approved):

pi0:
Thanks for quality changes ❤️

Issue #2007, opened 2023-12-13
`defineNitroModule` is not exported from the package


huang-julien:
### Describe the feature

Hi :wave:

currently working on nitro-applicationinsights, it seems that `defineNitroModule` isn't exported. So we can't do 
`import { defineNitroModule } from "nitropack"`

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
it is exported from `nitropack/kit` now (however undocumented). Stable way is to import `NitroModule` **type** this way avoid having dependency on nitropack and makes modules better ad-hoc.

PR #2008, opened 2023-12-13
feat: export `defineNitroModule` 

huang-julien:
### 🔗 Linked issue

resolve https://github.com/unjs/nitro/issues/2007

### ❓ Type of change

### 📚 Description
Hi :wave: this PR simply export `defineNitroModule` from the package

### 📝 Checklist


pi0:
Thanks for the pull request ❤️ 

 It was intentional that currently this util is not exposed to avoid introducing potential issues in community packages used in nested dependencies since unlike Nuxt, nitro doesn't have a module kit and directly importing from `nitropack` means your modules should have direct dependency on `nitropack` which is not desired. 

I guess we can export it but make sure to only mark nitropack as a `peerDependency` in your modules.

pi0:
BTW just checking, without importing `defineNitroModule` from anywhere, there should be global types and working `defineNitroModule` function doesn't it works ([src](https://github.com/unjs/nitro/blob/main/src/types/nitro.ts#L402) and [src](https://github.com/unjs/nitro/blob/main/src/module.ts#L17))

pi0:
Closing for triage. Please let me know if have any issues with `defineNitroModule` global types

PR #2017, opened 2023-12-21
feat(wasm): universal support with esm import syntax

pi0:
### ❓ Type of change

### 📚 Description

This PR unifies WASM support using Nitro's built-in wasm plugin and simplified ESM import syntax: (Following [ES Module Integration Proposal for WebAssembly](https://github.com/WebAssembly/esm-integration/tree/main/proposals/esm-integration))

Using dynamic imports:

```ts
export default defineLazyEventHandler(async () => {
  const { sum } = await import("~/wasm/sum.wasm");
  return eventHandler(() => {
    return `2+3=${sum(2, 3)}`;
  });
});
```

Using static imports: (1)

```ts
import { sum } from "~/wasm/sum.wasm";

export default eventHandler(() => {
  return `2+3=${sum(2, 3)}`;
});
```

- On Node.js and platforms yet not supporting native `.wasm` imports, Nitro by default creates a base64 inlined version (in a chunk) and instantiates the module
- On supported platforms (Cloudflare pages and modules), plugin will use native wasm import and copy the asset to output

(1) One of the main blockers, are worker targets that don't support top level await, combined with [rollup limitation with `inlineDynamicImports`](https://github.com/rollup/rollup/issues/4166) we cannot use top level await for `await WebAssembly.instantiate`. This mainly affects/breaks Vercel Edge target. Internally we support `wasm.lazy` flag that causes to avoid depending on top-level wait for maximise compatibility. It requires static import slight syntax change to work universally again: 

```ts
import _mod from "~/wasm/sum.wasm";

export default eventHandler(async () => {
  const { sum } = await _mod;
  return `2+3=${sum(2, 3)}`;
});
```

The value of `_mod` is a virtual Promise-like that lazily get's evaluated. Hopefully at some point we don't require this syntax change too.

### 📝 Checklist


pi0:
Merging to make it testable via edge-channel. Feedbacks more than welcome specially @antfu @harlan-zw since I know you have some good projects to verify against.

pi0:
Early checking against shikiji, seems some issues

antfu:
Looks awesome! Will give a try tmr

PR #1783, opened 2023-10-03
fix(iis-node): pass `PORT` as `NITRO_UNIX_SOCKET`

pi0:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22961 

(alternative rework of #1697 by @tidan-16)

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

`iis-node` runtime passes unix socket as `PORT` environment variable this makes issue for Nitro node listener (Note: this issue only related to `iisnode`, `iis` handler passes a valid port.)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2020, opened 2023-12-22
fix(iisnode): correctly escape backslashes for port fix

manniL:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/22961

### ❓ Type of change

### 📚 Description

Uses a more general fix than #1783 based on #1697

### 📝 Checklist

### DIFF
--- a/src/presets/iis.ts
+++ b/src/presets/iis.ts
@@ -30,7 +30,8 @@ export const iisNode = defineNitroPreset({
       await writeFile(
         resolve(nitro.options.output.dir, "index.js"),
         `
-        if (Number.isNaN(Number(process.env.PORT))) {
+        const port = process.env.PORT
+        if (port && Number.isNaN(Number(port))) {
           process.env.NITRO_UNIX_SOCKET = process.env.PORT
           delete process.env.PORT
         }
--- a/src/presets/iis.ts
+++ b/src/presets/iis.ts
@@ -30,7 +30,7 @@ export const iisNode = defineNitroPreset({
       await writeFile(
         resolve(nitro.options.output.dir, "index.js"),
         `
-          if (process.env.PORT.startsWith('\\\\')) {
+        if (process.env.PORT.startsWith('\\\\')) {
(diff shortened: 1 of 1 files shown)


src/presets/iis.ts
```diff
@@ -30,7 +30,8 @@ export const iisNode = defineNitroPreset({
       await writeFile(
         resolve(nitro.options.output.dir, "index.js"),
         `
-        if (process.env.PORT.startsWith('\\')) {
+        const port = process.env.PORT
+        if (port && Number.isNaN(Number(port))) {
```
pi0, line 34: I rather to have a specific check for `\` as it is a very specific behavior check and if another invalid value is provided we shall use explicit error

manniL: @pi0 No, I didn't yet. Only followed up because [people reported](https://github.com/nuxt/nuxt/issues/22961#issuecomment-1867941123) the previous fix didn't work.

pi0: It should normally work with `\` check. I have a working IIS setup in my home lab made basically for this issue! Aso wondering what is wrong with reports... Will try to check later

manniL: FYI: @tolgahanbeyazoglu confirmed that this workaround is working as expected. But I'd be happy with a more strict one either.

manniL: But there is even a better fix now :D

manniL:
Updated the PR to simply escape the slashes correctly.

before:

```ts

        if (process.env.PORT.startsWith('\')) {
          process.env.NITRO_UNIX_SOCKET = process.env.PORT
          delete process.env.PORT
        }
        import('./server/index.mjs');
```

With this PR:

```ts
        if (process.env.PORT.startsWith('\\')) {
          process.env.NITRO_UNIX_SOCKET = process.env.PORT
          delete process.env.PORT
        }
        import('./server/index.mjs'); 
  ```

pi0 (approved):

Issue #2029, opened 2023-12-25
Cannot declare a package module as module in the nitro config


huang-julien:
### Describe the feature

Hey :wave: i'm currently converting `nitro-applicationinsights` to a nitro module. 

Currently it doesn't seems to be possible to declare a package as a module within the config

```ts
export default defineNitroConfig({
  modules: [ 'nitro-applicationinsights']
})
```

### Additional information

- [X] Would you be willing to help implement this feature?

PR #2030, opened 2023-12-25
feat(modules): allow to declare packages as nitro modules

huang-julien:
### 🔗 Linked issue

solve https://github.com/unjs/nitro/issues/2029

### ❓ Type of change

### 📚 Description

Hi :wave: this PR allows to declare packages as modules using `mlly` resolvePath when resolving nitro modules

### 📝 Checklist

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -1,8 +1,6 @@
 import jiti from "jiti";
+import { resolvePath } from "mlly";
 import type { Nitro, NitroModule, NitroModuleInput } from "./types";
-import { normalize } from "path";
-import { pathToFileURL } from 'node:url'
-import { interopDefault, resolvePath } from "mlly"
 
 export function defineNitroModule(def: NitroModule) {
   return def;
@@ -21,7 +19,7 @@ export async function resolveNitroModule(
       globalThis.defineNitroModule || defineNitroModule;
 
     const _jiti = jiti(nitroOptions.rootDir, { interopDefault: true, esmResolve: true });
-    const _modPath = _jiti.resolve(await resolvePath(mod, {url: nitroOptions.rootDir, extensions: ['.mjs', '.mts', '.js', '.ts', '.cjs']}))
+    const _modPath = _jiti.resolve(await resolvePath(mod, { url: nitroOptions.rootDir, extensions: ['.mjs', '.mts', '.js', '.ts', '.cjs'] }))
     _url = _modPath;
     mod = _jiti(_modPath) as NitroModule;
   }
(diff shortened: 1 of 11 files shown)


src/module.ts
```diff
@@ -18,8 +18,16 @@ export function resolveNitroModule(
       // @ts-ignore
       globalThis.defineNitroModule || defineNitroModule;
 
-    const _jiti = jiti(nitroOptions.rootDir, { interopDefault: true });
-    const _modPath = _jiti.resolve(resolvePath(mod, nitroOptions));
```
pi0, line 22: `jiti.resolve` itself should be able to do the resolution (it uses mlly when in esm mode even!).

We might need to extend options part to pass `modulesDir` (which also contains `rootDir`)

test/e2e/basic.test.ts
```diff
@@ -0,0 +1,22 @@
+import { resolve } from "pathe";
```
pi0, line 1: I guess additional tests are unnecessary it seems mostly a specific bug as I tend to keep tests simple for now.

src/module.ts
huang-julien: The issue may come from `resolvePath` which don't use mlly nor jiti to resolve the path. So `jiti.resolve` is actually receiving an absolute path to resolve.
Maybe removing it and let `jiti.resolve` do all the work of `resolvePath` would also work 🤔

test/e2e/basic.test.ts
huang-julien: :+1: i'll revert it

src/module.ts
huang-julien: oh it does work by removing the `resolvePath` !

pi0: Yes, we also need to support aliases btw `~/modules/`, etc. Wondering why these two are incompatible..

huang-julien: there's an alias option in JITI. We could probably use it

pi0: was it finally a bug in pathe? If too hard using jiti aliases should be also good idea but if easy solvable would be also nice to fix `resolvePath` because it is used for other keys and I worry behavior can be different in long term.

huang-julien: Yes it was :+1:
I just tested this with a local link to the pr from pathe with this PR and everything seems to be solved
nitro -> link:../jiti -> link:../pathe

huang-julien:
jiti is now correctly resolving both alias and modules !

pi0 (approved):

PR #2039, opened 2023-12-29
chore: remove lagon

maximepvrt:
### ❓ Type of change

### 📝 Checklist


pi0:
Thanks for the PR. I think we can instead add a banner for lagon being discontinued (1 march 20204) and finally we should keep preset bits until next major version of nitro.
https://github.com/unjs/std-env/issues/105

maximepvrt:
@pi0 Lagon is only available in private alpha. It is no longer possible to deploy new applications. I think the doc can be deleted now. The binaries must, however, only be deleted when the service is effectively closed.

pi0:
resolved by 17f922aae1a550751a632f01de6d1d5f085a1778

PR #2045, opened 2024-01-02
Update compress.ts

thednp:
Add support for `font/woff2` compression. I will test this implementation.

### 🔗 Linked issue
None

### ❓ Type of change

### 📚 Description
Sending a MonaSans(137k) / HubotSans(170k) file over to the client is quite expense no matter how you want to put it, we're expecting a 50-60% reduction in that size. Their OTF variants (currently supported by Nitro plugins) are out of the question with their more than double size.

### 📝 Checklist

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -114,6 +114,7 @@ const COMPRESSIBLE_MIMES_RE = new Set([
   "font/opentype",
   "font/otf",
   "font/ttf",
+  "font/woff2",
   "image/svg+xml",
   "text/css",
   "text/csv",


manniL:
WOFF and WOFF2 are already compressed using Brotli. Compressing them again is not a good idea IMO

thednp:
So where should the woff files be located to get into compression pipeline?

manniL:
@thednp Nowhere, they are compressed by default already. No compression (again) needed as it won't save any bytes

thednp:
@manniL sorry to bother you guys, but I did some digging and found some interesting things you might wanna have a look at.
I modified the file at `node_modules/.pnpm/nitropack@2.8.1/node_modules/nitropack/dist/shared/nitro.4ea992bc.mjs` and added `  "font/woff",  "font/woff2",` and found the following:
* now my font IS included into the build compression pipeline
* the br file is exactly the same size as the original (137Kb), the gz file is 0.1Kb more for some strange reason
* see screenshot below

Screenshot from 2024-01-15 20-44-57

May I suggest that you download a Woff2 file and make some tests yourselves, perhaps there's some bug you don't know yet about.

Thanks

PR #2047, opened 2024-01-03
fix(dev): set correct x-forwarded-proto value

### LINKED ISSUE #1978: Incorrect value for X-Forwarded-Proto
### Environment

Irrelevant but Nitro 2.8.1 on Node 20.5

### Reproduction

Issue is visible in:

https://github.com/unjs/nitro/blob/06323b736f0d8adaa609d5a0348fa6b9e0db75e7/src/dev/server.ts#L203-L205

### Describe the bug

The X-Forwarded-Proto header *should* take a value in [`http`, `https`], and instead takes a value in [`IPv4`, `IPv6`]

With the current code, accessing a Nitro server with plaintext results in `X-Forwarded-Proto: IPv4` (for example)

This is not correct, and should instead give `X-Forwarded-Proto: http`

### Additional context

Since it is not a standardised header, the best source of documentation is your pick between:
a. The Mozilla docs about the de-facto way it is meant to work: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto#syntax
b. The spec for the `Forwarded` header which is based on it (and other X-Forwarded-* friends): https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded#proto
c. Nearly every (widespread) webserver in existence (CF: https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#x-forwarded-proto, AWS: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html#x-forwarded-proto, HAProxy: https://docs.haproxy.org/2.8/configuration.html#7.3.1-iif, ...)

### Logs

_No response_

Tristan971:
Should be 'https' when we're certain that *we* received an https request, and 'http' in every other case, unless set already

### 🔗 Linked issue

#1978 

### ❓ Type of change

### 📚 Description

Resolves #1978 

This is technically a breaking change, even if the previous behaviour was wrong and at odds with the rest of the industry.

I'm ambivalent on whether to provide some kind of legacy-behaviour flag here.

### 📝 Checklist

I presume it'd be good to document *somewhere* the headers added by Nitro, in which context they are added (or not), and at what point of the lifecycle (ie when can an end-user depend on it wrt middlewares etc).

Hoping you can suggest the right place to do so.

Also would have been ideal to be able to add at least a test for it, but I am a bit confused as to how Nitro tests work currently.

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -1,6 +1,7 @@
 import { Worker } from "node:worker_threads";
 import { existsSync, accessSync, promises as fsp } from "node:fs";
 import { writeFile } from "node:fs/promises";
+import { IncomingMessage } from "node:http";
 import { debounce } from "perfect-debounce";
 import {
   App,
@@ -192,7 +193,7 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
 
   // Main worker proxy
   const proxy = createProxy();
-  proxy.proxy.on("proxyReq", (proxyReq, req) => {
+  proxy.proxy.on("proxyReq", (proxyReq, req: IncomingMessage) => {
     const proxyRequestHeaders = proxyReq.getHeaders();
     if (req.socket.remoteAddress && !proxyRequestHeaders["x-forwarded-for"]) {
       proxyReq.setHeader("X-Forwarded-For", req.socket.remoteAddress);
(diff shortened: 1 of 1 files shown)


Tristan971:
While doing this MR I noticed a couple of additional issues, which I guess I will report separately

src/dev/server.ts
```diff
@@ -192,16 +193,18 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
...
     if (req.socket.remoteAddress && !proxyRequestHeaders["x-forwarded-for"]) {
       proxyReq.setHeader("X-Forwarded-For", req.socket.remoteAddress);
     }
     if (req.socket.remotePort && !proxyRequestHeaders["x-forwarded-port"]) {
       proxyReq.setHeader("X-Forwarded-Port", req.socket.remotePort);
     }
-    if (req.socket.remoteFamily && !proxyRequestHeaders["x-forwarded-proto"]) {
-      proxyReq.setHeader("X-Forwarded-Proto", req.socket.remoteFamily);
+    if (!proxyRequestHeaders["x-forwarded-proto"]) {
+      const url = new URL(req.url);
```
pi0, line 205: `req.url` is a relative URL like `/foo`, we cannot parse it this way to infer proto

pi0:
Thank you for PR. It is fixed via 29ddd948407322c67aace0336fdc65eed12bad20

PR #2083, opened 2024-01-16
docs: update the configuration page

Barbapapazes:


docs/content/1.guide/1.configuration.md
```diff
@@ -40,81 +38,82 @@ nitro.timing=true
...
-    helloThere: "foobar",
+    foo: "bar", // `bar` is the default value
   }
 })
 ```
 
 ```ts [nuxt.config.ts]
   runtimeConfig: {
-    helloThere: "foobar",
```
pi0, line 61: using `helloThere` were intentional to show multi part override via env (`NUXT_HELLO_THERE`) we might make it more realistic example like `apiToken` that still has two parts

Barbapapazes: You are right. I've update it.

docs/content/1.guide/1.configuration.md
```diff
@@ -4,9 +4,7 @@ icon: ri:settings-3-line
 
 # Configuration
 
-You can customize your Nitro server with single configuration file: `nitro.config.ts`.
-
-All deployment providers are built on the same options API.
+You can customize your Nitro server with single configuration file called `nitro.config.ts`. This file will be used by [deployement presets](/deploy) so you do not need to worry about them when configuring your server.
```
pi0, line 7: THe left senstence does not means that -- it meant that presets are **made with** same interface 😅 Let's remove it altogether for now seems more confusing

```suggestion
You can customize your Nitro server with a single configuration file `nitro.config.ts`.
```

docs/content/1.guide/1.configuration.md
```diff
@@ -40,81 +38,82 @@ nitro.timing=true
...
-You can now access the runtime config using `useRuntimeConfig(event)`.
-
-**Example:**
+You can now access the runtime config using `useRuntimeConfig(event)`.  Use `useRuntimeConfig(event)` within event handlers and utilities and **avoid** calling it in ambient global contexts. This could lead to unexpected behavior such sharing the same runtime config across different requests.
 
 ::code-group
 ```ts [api/example.get.ts (nitro)]
 export default defineEventHandler((event) => {
-  return useRuntimeConfig(event).helloThere // foobar
+  return useRuntimeConfig(event).apiToken // Return `dev_token`
```
pi0, line 68:
```suggestion
  return useRuntimeConfig(event).apiToken // Returns `dev_token`
```

docs/content/1.guide/1.configuration.md
```diff
@@ -40,81 +38,82 @@ nitro.timing=true
...
 
-You can universally access environment variables using `import.meta.env.TEST` or `process.env.TEST`.
-
-::alert{type="info"}
-**Note:** Consider writing any logic that depends on environment variables within event handlers and utilities and avoid accessing and caching them in ambient global contexts.
-::
+Re-start the development server, fetch the `/api/example` endpoint and you should see `123` as the response instead of `dev_token`.
 
-### Update runtime config using environment variables
+Do not froget taht you can still universally access environment variables using `import.meta.env` or `process.env` but avoid using them in ambiant global contexts to prevent unexpected behavior.
```
pi0, line 93:
```suggestion
Do not forget that you can still universally access environment variables using `import.meta.env` or `process.env` but avoid using them in ambiant global contexts to prevent unexpected behavior.
```

Issue #1785, opened 2023-10-03
Update dev dependency @azure/functions to v4


pi0:
https://github.com/unjs/nitro/pull/1767

itpropro:
Functions v4 uses a different function and binding declaration syntax so that functions are defined directly inside of the code instead of in the functions.json file. We should update the presets for azure and azure-functions to use the new Azure Functions syntax.

PR #2092, opened 2024-01-17
docs: add azure functions warn

GerryWilko:
### ❓ Type of change

This PR adds a warning to the Azure Functions page to clarify that support for this deployment type is not currently working reliably.

I got myself a little stuck in a hole trying to fix this issue for a service I wished to deploy to an Azure Function to later realise that it doesn't really work at present.

The work arounds listed in https://github.com/unjs/nitro-deploys/issues/48 may be helpful to some but I couldn't find a way to get my Nuxt 3 deployment working properly.

### 📝 Checklist

### DIFF
--- a/docs/content/2.deploy/20.providers/azure.md
+++ b/docs/content/2.deploy/20.providers/azure.md
@@ -74,6 +74,10 @@ If you are using runtimeConfig, you will likely want to configure the correspond
 
 ## Azure Functions
 
+::alert{type="warning"}
+There are several known issues with Azure Function deployments [#48](https://github.com/unjs/nitro-deploys/issues/48).
+::
+
 **Preset:** `azure_functions`
 
 **Note:** If you encounter any issues, please ensure you're using a Node.js 14+ runtime. You can find more information about [how to set the Node version in the Azure docs](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#setting-the-node-version).


pi0:
Thanks! Checking with @danielroe if we should deprecate preset fully. Because it was more of an experiment as I recall.

danielroe:
I think we should keep the preset. Similar to AWS Lambda, it's a lower level preset that might be used in advanced use cases but is almost certainly not the right preset to pick, standalone. (For example, it would lack the ability to serve static assets in a performance way.)

itpropro:
> cc @itpropro who is doing some work on updating the Azure functions presets IIRC

I am currently updating the preset to Azure Functions Runtime v4. We already discussed having a hint to that in the docs, but as there are still scenarios where it is working, it's hard to make a clear statement. 
I think we should merge this PR for now as it gives users a good indication of problems that currently exist, even though they are partly a result of changes at the Azure platform.

pi0:
> For example, it would lack the ability to serve static assets in a performance way

Indeed. More thinking we should put some presets like browser sw explicitly aside. And if there is no plan of active maintenance, I prefer to unsupport it at all. Having this state of half-borken presets is so frustrating honestly..

@itpropro Can you please kindly open an issue in this repo that we can mention it in docs as tracker for azure functions and perhaps what is exactly broken that needs update? (I can check it too but since you plan to take lead on it you probably know more context than I know currently)

itpropro:
> > For example, it would lack the ability to serve static assets in a performance way
> 
> @danielroe Indeed. More thinking we should put some presets like azure functions browser sw explicitly aside and not advice them. And if there is no plan of active maintenance, I prefer to unsupport it at all. Having this state of half-broken presets is so frustrating honestly..
> 
> @itpropro Can you please kindly open an issue in this repo that we can mention it in docs as a tracker for azure functions and perhaps what is exactly broken that needs update if v4 is absolutely necessary? (we might delay #1785 if there is a faster way of solving issues) (I can check it too but since you plan to take lead on it you probably know more context than I know currently)

I will open an issue for that. I think the goal should be to have a preset that uses the provider specific tooling to be the least maintenance possible. In the Azure case I would prefer to only focus on a simple Azure Functions preset and based on that a static web apps (which also uses Azure Functions by default) one. We should probably deprecate special features like generating a static web app config. That would compromise the idea of a zero config, but would enable us to still have a Azure preset that is maintainable.

danielroe:
related: https://github.com/nuxt/nuxt.com/pull/1483

cc: @thomasgauvin

thomasgauvin:
Thanks for the ping @danielroe. (I'm the PM on Azure Static Web Apps by the way).

While I can't speak to the issues that exist with Azure Functions, Nuxt SWA deployments seem to work quite well in my experience. I'm curious as to how SWA deployments work but Functions don't if SWA depends on them. Also, if there are any pressing SWA issues, would love to hear them so we can discuss.

Re deprecating SWA config, I think it's pretty essential to keep as you get the 'global serverless' server nature that SWA (and other similar hosts) offer. We aren't planning on making important changes to the config file any time soon, so maintenance on that could be considered relatively lower.  With Functions, you may encounter limitations around the routing, etc.

All that said, as a developer, and I'm a big fan of Nuxt deployments, it's one of the few server-side rendered frameworks that actually works quite well on SWA.

pi0:
Happy having you here dear @thomasgauvin. SWA is only preset I tried and at least inside emulator works well. Altough it seems there are some production deployment issues.

If you are on discord, would you mind to send me a DM (pi0) so I can invite you to a proper channel to keep in loop?

itpropro:
> Thanks for the ping @danielroe. (I'm the PM on Azure Static Web Apps by the way).
> 
> While I can't speak to the issues that exist with Azure Functions, Nuxt SWA deployments seem to work quite well in my experience. I'm curious as to how SWA deployments work but Functions don't if SWA depends on them. Also, if there are any pressing SWA issues, would love to hear them so we can discuss.
> 
> Re deprecating SWA config, I think it's pretty essential to keep as you get the 'global serverless' server nature that SWA (and other similar hosts) offer. We aren't planning on making important changes to the config file any time soon, so maintenance on that could be considered relatively lower. With Functions, you may encounter limitations around the routing, etc.
> 
> All that said, as a developer, and I'm a big fan of Nuxt deployments, it's one of the few server-side rendered frameworks that actually works quite well on SWA.

Hey @thomasgauvin,
thanks for the infos. There are several scenarios where the action works, but the SWA doesn’t render, where there are resolution problems with pnpm etc.. These are all solvable and I think this can be done with pretty low effort according to my test scenarios of the last few days.
Regarding the config, it’s just important that we provide some way of overwriting it for the user. Now we have several things “hardcoded”, like the platform/apiRuntime property. 
I think providing a base configuration that works out of the box plus the possibility to overwrite should work.
I have been working with SWAs since the cli was not not even existent and collected a comprehensive list of scenarios and edge cases of configurations that I currently test against and also had some customers run into problems, so providing a consistently working baseline config that is still maintainable is my main goal.

pi0:
ideally, similar to most of mainstream providers i like to see one preferred preset for azure deployment and it remains zero config. Currently we (sadly) always override a top level config which is not desired. It should be merged into `.output` for users.

GerryWilko:
> Thanks for the ping @danielroe. (I'm the PM on Azure Static Web Apps by the way).
> 
> While I can't speak to the issues that exist with Azure Functions, Nuxt SWA deployments seem to work quite well in my experience. I'm curious as to how SWA deployments work but Functions don't if SWA depends on them. Also, if there are any pressing SWA issues, would love to hear them so we can discuss.
> 
> Re deprecating SWA config, I think it's pretty essential to keep as you get the 'global serverless' server nature that SWA (and other similar hosts) offer. We aren't planning on making important changes to the config file any time soon, so maintenance on that could be considered relatively lower. With Functions, you may encounter limitations around the routing, etc.
> 
> All that said, as a developer, and I'm a big fan of Nuxt deployments, it's one of the few server-side rendered frameworks that actually works quite well on SWA.

@thomasgauvin Just FYI I think there may be some issues with Nuxt SWA deployments too. I had deployed this site https://signals.ignite.systems/ a few days ago and the routes seem broken. I didn't manage to debug this much just yet but the same deployment worked fine with preset `vercel` and local dev so it may not be quite as smooth on SWA just yet. I'm hoping to get back to take a closer look at what is going wrong next week.

docs/content/2.deploy/20.providers/azure.md
```diff
@@ -74,6 +74,10 @@ If you are using runtimeConfig, you will likely want to configure the correspond
 
 ## Azure Functions
 
+::alert{type="warning"}
+There are several known issues with Azure Function deployments [#48](https://github.com/unjs/nitro-deploys/issues/48).
```
pi0, line 78:
```suggestion
There are several known issues with Azure Function deployments. Please see [unjs/nitro#2114](https://github.com/unjs/nitro/issues/2114).
```

GerryWilko:
It seems that both Azure presets actually don't work. I been trying to debug why all morning and got no where really.

A simple docus site is not working:
```bash
npx nuxi@latest init -t themes/docus
cd docs
NITRO_PRESET=azure pnpm build
npx @azure/static-web-apps-cli deploy .output/public --api-location .output/server --deployment-token your-token-here --env production
```

Does not work. It appears that some static assets work but others dont and it also seems that on SSR all routes return 404 then do seem to match on hydration.

Example site: https://signals.ignite.systems

@pi0 should I raise a PR with a similar warning on the `azure` preset?

pi0:
@GerryWilko can you please mirror your message in #2114 to keep track of context maybe possibly little more info.

I am busy but will do my best to investigate status of azure as soon as I can. Sorry for this all.

GerryWilko:
Of course, thank you @pi0 absolutely no need to apologise. I'm still trying to see whats wrong. I'm currently putting together a simple reproduction.

To be honest I feel this is made much much harder due to the black box nature of the Azure Static Web App portal tools and the really difficult to use swa cli.

https://learn.microsoft.com/en-us/azure/static-web-apps/deploy-nuxtjs

The docs here and on Next.js imply the config seems to be "magic" and its not really documented how it works. I started looking at how SSR on Next.JS works with SWA to see if I could try fix it. 🤷

pi0:
Code to all magic we do for Nitro/Nuxt:

azure swa:
- https://github.com/unjs/nitro/blob/main/src/runtime/entries/azure.ts
- https://github.com/unjs/nitro/blob/main/src/presets/azure.ts

azure fns:
-  https://github.com/unjs/nitro/blob/main/src/runtime/entries/azure-functions.ts
- https://github.com/unjs/nitro/blob/main/src/presets/azure-functions.ts

(although let's mention it is not uncommon that deployment providers add special bits in their builder layer specifically for Next.js -- sadly. I hope it is not a case for azure only a guess since you have mentioned it works for Next)

GerryWilko:
@pi0 I believe I may have found where the issue lies. I have put an explainer of what I have found on the issue you mentioned but thought it best to provide a TLDR here too. https://github.com/unjs/nitro/issues/2114#issuecomment-1908502510

TLDR: You cannot current use Azure Static Web Apps, with any framework it seems, when using a package manager other than `npm`. You also have limited control over the builder which may make it infeasible to use for some users.

I believe this should be called out on the Nitro presets docs pages?

Issue #938, opened 2023-02-10
$fetch type safety for query and body parameters


ozum:
### Describe the feature

$fetch provides type safety for return types which is great. It would be greater if it **optionally** checks types for `query` and `body` parameters for internal API requests.

Below is a rough proposal:
1. Server routes optionally export `QuerySchema` and `BodySchema`.
  -> _Developer's responsibility_
2. Generate necessary types for those routes `/.nuxt/types/nitro.d.ts`.
  -> Below is an example.
3. Add types to `/node_modules/nitropack/dist/index.d.ts`
  -> Below is a proposal.

**/server/api/product.units.ts**

```ts
export default defineEventHandler((event) => {
  const query = getQuery(event)
  const body = await readBody(event)
})

export interface QuerySchema {
  name: string;
  id: number;
}

export interface BodySchema {
  content: string
}
```

**/.nuxt/types/nitro.d.ts**

```js
// It would be better if `InternalApi` and the proposed `InternalApiQuerySchema` and `InternalApiQuerySchema`
// are merged into one interface.
// However separated interfaces are easier to implement for re-using the current code base.

declare module 'nitropack' {
  interface InternalApi {
    '/api/units': {
      'get': Awaited<ReturnType<typeof import('../../server/api/units.get').default>>
    }
  }

  interface InternalApiQuerySchema {
    "/api/units": {
      get: import("../../server//api/units.get").QuerySchema;
    };
  }

  interface InternalApiBodySchema {
    "/api/units": {
      get: import("../../server//api/units.get").BodySchema;
    };
  }
}
```

**/node_modules/nitropack/dist/index.d.ts**

```ts
// ─── Added ───────────────────────────────────────────────────────────────────
type RequestSchema<Base, R extends NitroFetchRequest, M extends AvailableRouterMethod<R> = AvailableRouterMethod<R>> = R extends keyof Base
  ? M extends keyof Base[R]
    ? Base[R][M]
    : never
  : never;

// ─── Modified ────────────────────────────────────────────────────────────────
// Added `query` and `body`
interface NitroFetchOptions<R extends NitroFetchRequest, M extends AvailableRouterMethod<R> = AvailableRouterMethod<R>>
  extends Omit<FetchOptions, "query" | "body"> {
  method?: Uppercase<M> | M;
  query?: RequestSchema<InternalApiQuerySchema, R, M>;
  body?: RequestSchema<InternalApiBodySchema, R, M>;
}
```

# Problems I stumbled upon:
[...]

PR #2102, opened 2024-01-20
feat: experimental `defineRouteMeta`

eriksLapins:
### ❓ Type of change

### 📚 Description

I added the option to change the default provided schema directly in the server route file with defineRouteMeta, which can take in any key: value pair, with one defined in this PR - openAPI operations object, which will amend the default schema with types inferred from OperationsObject in openapi-typescript

This solves the issue that we can get better looking openapi documentation with better support

Possibly resolves the problem shown in discussion #2056 

### 📝 Checklist


eriksLapins:
Also, I wanted to add a test for the defineOpenAPISchema function, but it errors out with
`Failed to load url #internal/nitro/virtual/server-handlers` if I load in the function as below

or `Failed to load url #internal/nitro` if I load the added function from #internal/nitro/routes/openapi

My idea was to add an additional preset with something like:
```
import { describe, expect, it } from "vitest";
import { defineOpenAPISchema } from "../../src/runtime/routes/openapi";
import { useStorage } from "#internal/nitro";


describe("nitro:openapi:schema-definition", () => {
    it("stores the result in useStorage", async () => {
        defineOpenAPISchema({
            routeBase: '/api/test',
            method: 'get',
            schema: {
                tags: ["test"]
            }
        })

        const storage = await useStorage("/api/test").getItem('openapi-get');

        expect(storage).toBe({
            routeBase: '/api/test',
            method: 'get',
            schema: {
                tags: ["test"]
            }
        })
    })
})
```

examples/api-routes/api/test.get.ts
```diff
@@ -1 +1,14 @@
+import { defineOpenAPISchema } from "#internal/nitro/routes/openapi";
+
 export default defineEventHandler(() => "Test get handler");
+
+defineOpenAPISchema({
```
pi0, line 5: We need to generalize util to `defineRouteMeta({ openAPI: { ... } }`)

examples/api-routes/api/test.post.ts
```diff
@@ -1 +1,28 @@
+import { defineOpenAPISchema } from "#internal/nitro/routes/openapi";
+
 export default defineEventHandler(() => "Test post handler");
+
+defineOpenAPISchema({
+    method: 'post',
+    routeBase: '/api/test',
```
pi0, line 7: Some props such as method, and url, can be auto-generated already without user needing them

src/runtime/routes/openapi.ts
```diff
@@ -96,3 +115,9 @@ function defaultTags(route: string) {
 
   return tags;
 }
+
+export function defineOpenAPISchema(schema: NitroOpenapiSchema) {
+  useStorage(schema.routeBase).setItem(`openapi-${schema.method}`, schema);
```
pi0, line 120: This is a really interesting approach however considering overhead and limitations (dev<>build communication) I think we have to use static code extraction for route meta data.

eriksLapins: I changed the approach to parsing it at build time, please check if this version seems to be on the right path at least

eriksLapins: @pi0 I redid the parse to build time, could you please check if this is better? Got some inspiration from [(Nuxt's extract route rules)](https://github.com/nuxt/nuxt/blob/95a521376614d9cf605ca16fefe80d6c216bd906/packages/nuxt/src/pages/route-rules.ts#L25)

src/scan.ts
```diff
@@ -122,3 +134,48 @@ async function scanDir(
     })
     .sort((a, b) => a.path.localeCompare(b.path));
 }
+
+export async function scanRouteMeta(
+  fullPath: string
+): Promise<ServerRouteMeta> {
+  const file = fs.readFileSync(fullPath, { encoding: "utf8", flag: "r" });
```
septatrix, line 141: Have you investigated how this performs for large projects with many routes and big files?

septatrix (commented):
Definitely looks like a nice improvement. I am still mostly looking forward to automatically inferring the type (see #938) but this is a good compromise for the time being.

What I do not like is that is seems to only support object literals (based on reading the code, please correct me if I am wrong). No real evaluation for the route metadata is possible. This can and will confuse a lot of people. My first though was that this could be used with zod to generate zome of the types automatically but this does not seem feasible.

src/scan.ts
pi0: (this is updated now)

pi0:
Thanks for your initiatives on this dear @eriksLapins. I have made some refactors in the way it works to experiment alternative approach for Nitro. 

The meta is scanned using a rollup plugin now to allow HMR and uses a native parser of rollup + AST to Object transform and is only enabled with an experimental open API flag.

Atinux:
How can we share the validation (let's say I am using zod) between the route meta and the `readValidatedBody` for example?

pi0:
It is experimental and not for validation right now.

rmarquet21:
@pi0 Any idea when this will be in production and not just on the "nightly" version?

chenyuncai:
Currentlly it cannot collect nested router endpoint, can we hack the router difinition and get all defined api?

pi0:
2.10 release no ETA. Also mind you, this API is likely to completely change into event handler object format nothing to rely on.

PR #2110, opened 2024-01-23
fix: only mock consola in production

### LINKED ISSUE #1991: use full build of consola for development time
Context: https://github.com/unjs/nitro/issues/1761

We automatically use lighter build of consola for production but for development, we can keep using fancy logger.

cany748:
### 🔗 Linked issue

#1991

### ❓ Type of change

### 📚 Description

Resolves #1991 

This change uses the full build consola in development mode. It might be confusing if the behavior in development mode and production mode is different, maybe it would be worth writing about this in the documentation?

### 📝 Checklist


pi0 (approved):
Thanks!

PR #2116, opened 2024-01-24
docs: wrong example in Catch all params in Routing page. Fix it.

jamaluddinrumi:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/1.guide/3.routing.md
+++ b/docs/content/1.guide/3.routing.md
@@ -86,9 +86,9 @@ You can capture all the remaining parts of a URL using `[...<param>]` syntax. Th
 
 ```ts [/routes/hello/[...name\\].ts]
 export default defineEventHandler(event => {
-  const name = getRouterParam(event, 'name')
+  const params = getRouterParams(event)
 
-  return `Hello ${name}!`
+  return `Hello ${params}!`
 })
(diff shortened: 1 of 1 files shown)


manniL (changes requested):
image

To me, the current version looks totally fine (and also safer if you have nested routes, e.g. `/[some-param]/abc/[...catch].ts`)

pi0:
Yeah there is no difference. Only we might have two examples to show both utils.

@manniL what do you mean it is safer?

jamaluddinrumi:
> image
> 
> To me, the current version looks totally fine (and also safer if you have nested routes, e.g. `/[some-param]/abc/[...catch].ts`)

ah ya, i miss understood because it's turned out that it's for `/routes/hello/[...name].ts` route.

i didn't realized that because it's hidden when our mouse hover it.

Artboard

do i need to close this pr by myself right now?

pi0:
We need to escape [ ] with \ in markdown...

jamaluddinrumi:
in my personal opinion, it's better to <ins>keep file name stay appears</ins> in code example so the newbies like me don't lose track.

jamaluddinrumi:
like what tailwind does

Artboard

Issue #1833, opened 2023-10-17
Many files in server assets fails to build the app with error: JavaScript heap out of memory


ManasMadrecha:
### Environment

Nuxt project info:                                                                                                                            4:49:34 pm

------------------------------
- Operating System: Windows_NT
- Node Version:     v20.8.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@10.1.0
- Builder:          -
- User Config:      css, devtools, experimental, nitro, modules, postcss, pwa, routeRules, runtimeConfig
- Runtime Modules:  @vueuse/nuxt@10.4.1, @bg-dev/nuxt-naiveui@1.4.7, @pinia/nuxt@0.4.11, @pinia-plugin-persistedstate/nuxt@1.1.2, @vite-pwa/nuxt@0.1.1    
- Build Modules:    -
------------------------------

### Reproduction

- I have a Nuxt 3 app. 
- I added 3500 JSON files of 300 KB each in `/server/assets` folder

### Describe the bug

No issues in dev (`npm run dev`). The app starts up properly and runs properly.

But on `npm run build`, the terminal shows `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`. This error is shown after `ℹ Building Nitro Server (preset: node-server)`. 

If I remove these 3500 JSON files, the error is not there. Also, I tried keeping just 1 file; the error goes away. So, the error exists only when there are too many files. 

### Additional context

This error exists even after trying different presets, i.e., firebase or the default node-server.

I think it would be better to add some wait time after this line so as to not cause memory error. 

https://github.com/unjs/nitro/blob/be6bb7ea136c464ab28eb8662eb630e11c7b0b68/src/rollup/plugins/server-assets.ts#L50

### Logs

```sh
✔ Server built in 15417ms                                                                                                                    4:28:26 pm
✔ Generated public .output/public                                                                                                      nitro 4:28:26 pm

PWA v0.16.5
mode      generateSW
precache  86 entries (1942.23 KiB)
files generated
  .output\public\sw.js
  .output\public\workbox-fa446783.js
ℹ Building Nitro Server (preset: node-server)                                                                                          nitro 4:28:32 pm

<--- Last few GCs --->

[...]

PR #2145, opened 2024-02-11
perf: limit open files in generateFSTree

marvin-j97:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1833

### ❓ Type of change

### 📚 Description

generateFSTree opens up to N files at the same time. Operating systems limit the amount of open files per process (linux defaults to 1024). Also, awaiting many Promises adds overhead, especially when a function is CPU-bound (gzipping for example). Collecting files has been extracted into `collectFiles`, which uses globby & tiny-async-pool to process 8 (subject to change) files at a time.

Should limit the amount of open files to fix https://github.com/unjs/nitro/issues/1833

### 📝 Checklist


pi0:
Thanks for this PR. Quickly looking it seems nice changes. I might delay it to next releases to test better.

Ingramz:
While the fix here mainly concerns memory exhaustion caused by too many open promises at once, I propose some further optimizations to additionally work around possible performance issues caused by processing larger files.

It seems unnecessary to read entire files into memory just to calculate gzip size on them. `gzip-size` already contains methods to do this on a path directly, which internally uses streams for reduced memory usage. `gzipSize(src)` could be in this case replaced by [`gzipSizeFromFile(path)`](https://www.npmjs.com/package/gzip-size#gzipsizefromfilepath-options). By using this method instead, the entire file read process can be eliminated in cases when gzip size is not required.

Obtaining the uncompressed file size could be done through `(await fsp.stat(path)).size` instead.

Also Node v19.4.0, v18.14.0 introduced [`os.availableParallelism()`](https://nodejs.org/api/os.html#osavailableparallelism), this however isn't available in older versions still supported by nitro and I think falling back to a sane number like 8 or `os.cpus().length` on such versions could work. Something like `const parallelism = os.availableParallelism?.() ?? os.cpus().length`, perhaps with a comment to remove the optional & fallback after mentioned nodejs versions become minimum required by nitro.

katywings:
@marvin-j97 Did you close this PR by removing your repo deliberately 😅? 

@pi0 should we create a new one 🤔?

marvin-j97:
> katywings

Damnit I think I did. I cleaned up my repos because I have way too many and GitHub doesn't show you you have open PRs :skull:

marvin-j97:
> @marvin-j97 Ohh I see 🙈, unlucky 😅! Do you wanna create a new PR or would you prefer if someone else would take care of this?

https://github.com/unjs/nitro/pull/2458

PR #2152, opened 2024-02-12
fix: handle optional `routeHandler.route` in rollup chunk

antfu:
### ❓ Type of change

### 📚 Description

Was trying to use the Nitro edge on https://github.com/antfu/atinotes and got this error:

image

Middleware handler can have undefined `.route` value, where here we forgot to check
### 📝 Checklist

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -87,7 +87,7 @@ export const getRollupConfig = (nitro: Nitro): RollupConfig => {
     const routeHandler =
       nitro.options.handlers.find((h) => id.startsWith(h.handler as string)) ||
       nitro.scannedHandlers.find((h) => id.startsWith(h.handler as string));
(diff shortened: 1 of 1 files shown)


src/rollup/config.ts
```diff
@@ -87,7 +87,7 @@ export const getRollupConfig = (nitro: Nitro): RollupConfig => {
     const routeHandler =
       nitro.options.handlers.find((h) => id.startsWith(h.handler as string)) ||
       nitro.scannedHandlers.find((h) => id.startsWith(h.handler as string));
-    if (routeHandler) {
+    if (routeHandler?.route) {
```
pi0, line 90: Maybe as fallback, we can use `chunks/middleware`?

pi0: we should fix it later

PR #2207, opened 2024-03-02
fix: import functions from `h3` with relative path

Louis9902:
### ❓ Type of change

### 📚 Description

When I use nuxt with pnpm **without** shamefully-host, I noticed that I can't import functions like `defineEventHandler` from `#imports`. When I inspected the generated types in the `nitro-imports.d.ts` I found why this is. The import in this file is not relative into the node modules store. It is a named re-export like `export { defineEventHandler, ... } from 'h3'`. This re-export will not work because of the not hoisted `h3` package. Every other export/import in this file already used a relative import, so i figured this might be a problem.

With this change, the already existing import resolution map is used to resolve the named import to a relative path. If this named import was by design, I'm interested why, but feel free to just close this PR.

### 📝 Checklist

- [ ] I have linked an issue or discussion. (not applicable)
- [ ] I have updated the documentation accordingly. (not applicable)

I also provide a patch for applying the same to nitropack with pnpm patch

<details>
<summary>PNPM patch to test it with nuxt example project</summary>

```jsonc
  "pnpm": {
    "patchedDependencies": {
      "nitropack@2.9.1": "patches/nitropack@2.9.1.patch"
    }
  }
```

```patch
diff --git a/dist/nitro.mjs b/dist/nitro.mjs
index 1dcaaacd17bb17e68d4f740839ce9a2f1221fd73..c2214be1bd73231be52b7ba15cdde4f729a3b0ce 100644
--- a/dist/nitro.mjs
+++ b/dist/nitro.mjs
@@ -2336,9 +2336,6 @@ async function writeTypes(nitro) {
   let autoImportExports;
   if (nitro.unimport) {
     await nitro.unimport.init();
-    autoImportExports = await nitro.unimport.toExports(typesDir).then(
-      (r) => r.replace(/#internal\/nitro/g, relative(typesDir, runtimeDir))
-    );
     const resolvedImportPathMap = /* @__PURE__ */ new Map();
     const imports = await nitro.unimport.getImports().then((r) => r.filter((i) => !i.type));
     for (const i of imports) {
@@ -2368,6 +2365,9 @@ async function writeTypes(nitro) {
       }
       resolvedImportPathMap.set(i.from, path);
     }
+    autoImportExports = await nitro.unimport.toExports(typesDir).then(
+      (r) => r.replace(/'(.*)'/g, (_, i) => `'${resolvedImportPathMap.get(i) ?? i}'`)
+    );
     autoImportedTypes = [
       (await nitro.unimport.generateTypeDeclarations({
         exportHelper: false,
```
</details>

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -131,13 +131,6 @@ export async function writeTypes(nitro: Nitro) {
 
   if (nitro.unimport) {
     await nitro.unimport.init();
-    // TODO: fully resolve utils exported from `#imports`
-    autoImportExports = await nitro.unimport
-      .toExports(typesDir)
-      .then((r) =>
-        r.replace(/#internal\/nitro/g, relative(typesDir, runtimeDir))
-      );
-
     const resolvedImportPathMap = new Map<string, string>();
     const imports = await nitro.unimport
       .getImports()
@@ -171,6 +164,12 @@ export async function writeTypes(nitro: Nitro) {
       resolvedImportPathMap.set(i.from, path);
     }
 
+    // TODO: fully resolve utils exported from `#imports`
+    autoImportExports = await nitro.unimport
+      .toExports(typesDir)
+      .then((r) =>
+        r.replace(/'(.*)'/g, (_, i) => `'${resolvedImportPathMap.get(i) ?? i}'`)
+    );
     autoImportedTypes = [
       (
         await nitro.unimport.generateTypeDeclarations({


pi0:
Thanks for the PR. And i'm certain it is valid. But please reopen when you could provide a minimal reproduction.

src/build.ts
```diff
@@ -171,6 +164,12 @@ export async function writeTypes(nitro: Nitro) {
       resolvedImportPathMap.set(i.from, path);
     }
 
+    // TODO: fully resolve utils exported from `#imports`
+    autoImportExports = await nitro.unimport
+      .toExports(typesDir)
+      .then((r) =>
+        r.replace(/'(.*)'/g, (_, i) => `'${resolvedImportPathMap.get(i) ?? i}'`)
```
pi0, line 171: This might be also something potentially unimport could directly solve. /cc @antfu

danielroe:
i believe this reproduces on a new Nuxt install without shamefully hoist.

Louis9902:
> i believe this reproduces on a new Nuxt install without shamefully hoist.

That's correct, I didn't state that very clear in my initial comment

pi0:
So it would not be possible to even reproduce it with nitro alone? 

I get the issue but having a fix in nitro that can easily cause regression in Nuxt, does seems needs better fix, possibly in unimport.

@danielroe Feel free to open a new hotfix PR if it is a regression of latest Nitro only, i trust you if you believe it needs hotfix in nitro.

danielroe:
We are already doing this (within nitro) for all the other imports. I think consistency means we should hotfix this now, even if we later move elsewhere.

I haven't yet checked the implementation above.

pi0:
Regardless i very much appreciate a clear tracker issue for this. Even if we are going with a hotfix first. 🙏

Issue #1974, opened 2023-11-30
Nitro Tasks API


pi0:
Nitro tasks allow on-off operations in runtime.

Docs: https://nitro.unjs.io/guide/tasks

- [x] Initial Implementation (#1929)
- [x] Write documentation section (PR welcome!)
- [x] Stabilize API / return values
- [x] Integrate with Cron Tasks / Background Jobs
- [ ] https://github.com/unjs/nitro/issues/2633
- [x] Vercel support
- [ ] Ability to run tasks using CLI against production build
- [ ] Task hooks (on dev start, production start and production build)
- [ ] Ability to schedule one-time run of a task for the future
- [ ] Allow optionally allow running multiple instances of task

imcm7:
But I thought it was about cron task. Mikro-orm has own commands like db:migrate  and other.

pi0:
@imcm7 check out task list it is on the roadmap :)

imcm7:
@pi0 I have a big trouble, I use mikro orm and form it with orm.em.fork(), but it have a [middleware for that](https://github.com/mikro-orm/mikro-orm/blob/cccc7561f9dc3c90f22ad293fceb7b58a67dabea/packages/core/src/utils/RequestContext.ts#L2)  but it nit work.

I do something like

```
return fromNodeMiddleware(async (
            req: IncomingMessage,
            res: ServerResponse<IncomingMessage>,
            next: (...args: any[]) => void
) =>  {
           RequestContext.create(em, next);
}
        
```

And I get result with in middleware

```
  private static storage = new AsyncLocalStorage<RequestContext>();
  console.log('>>>>',  this.storage.getStore());
  
```

But no result in mikro-rom when it try to load forked entity manager.


useEvent() say no nitro instance (may be because is called from plugin) and for that is need to change mikro-orm code.


And with that don't know what to do

```
  export const nitroAsyncContext = getContext<NitroAsyncContext>("nitro-app", {
    asyncContext: import.meta._asyncContext,
    AsyncLocalStorage: import.meta._asyncContext ? AsyncLocalStorage : undefined,
  });


  return nitroAsyncContext.use();
    
```


How to combine 2 AsyncLocalStorage

stafyniaksacha:
Do you plan to add a build time task runner? (e.g: `node .output/server/tasks.mjs run db:migrate --payload "{}"`)

Aareksio:
[...]

Issue #2196, opened 2024-02-29
Use server: {} as standard configuration namespace 


pi0:
Currently in the docs we have lots of exceptions that Nuxt and Nitro config are different. This can be easily a case for any other meta framework.

Similar how vitest supports test key in vite config, we could do the same.

Our config loader could be smart to merge different sources and detect sub key usage for backward/forward compatibility.

danielroe:
As a documentation improvement and convention (so Nuxt moves `nitro` -> `server`) I think it would make a lot of sense. 👍

(I don't think you mean this, but worth saying that I'm not a fan of Nitro resolving config from the Nuxt config file separately, though.)

pi0x:
Nitro v3 shipped with nitro.config.ts and no `server` namespace, and the thread already settled that merging downstream framework config does not belong in Nitro. Renaming the key is up to each framework's own config layer.

_Triaged with AI assistance._

PR #2219, opened 2024-03-04
Fix: Allow setting build dir when running a task from cli.

hareland:
This is useful to allow usage in Nuxt.

### 🔗 Linked issue

Related to: https://github.com/unjs/nitro/issues/1974

### ❓ Type of change

### 📚 Description

Allow to set the build directory - this allows usage within Nuxt for example.

### 📝 Checklist

I am not sure if i forgot anything for this to be appreciated, also i am not sure about the naming convention.
Feedback would be appreciated.

### DIFF
--- a/src/cli/commands/task/list.ts
+++ b/src/cli/commands/task/list.ts
@@ -20,7 +20,7 @@ export default defineCommand({
   },
   async run({ args }) {
     const cwd = resolve((args.dir || args.cwd || ".") as string);
-    const tasks = await listTasks({ cwd, buildDir: args.buildDir || '.nitro' });
+    const tasks = await listTasks({ cwd, buildDir: args.buildDir || ".nitro" });
     for (const [name, task] of Object.entries(tasks)) {
       consola.log(
         ` - \`${name}\`${task.meta?.description ? ` - ${task.meta.description}` : ""}`
--- a/src/cli/commands/task/list.ts
+++ b/src/cli/commands/task/list.ts
@@ -13,10 +13,14 @@ export default defineCommand({
       type: "string",
       description: "project root directory",
     },
+    buildDir: {
+      type: "string",
+      description: "Build directory",
+    },
   },
   async run({ args }) {
     const cwd = resolve((args.dir || args.cwd || ".") as string);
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for this PR. yes, we need to specify build-dir but I'm afraid 1) it is too early to expand experiment and 2) we might need to use programmatic API exports of Nitro for Nuxi CLI instead of depending on nitro. (I'm have it on plan ~> https://github.com/nuxt/cli/issues/353)

Nitro CLI itself can load `nitro.config` to respect the `buildDir` (PR welcome for this) but it is not going to help for Nuxt. Nitro will never involve in reading downstream configs (https://github.com/unjs/nitro/issues/2196#issuecomment-1971024371)

For clarity I'm closing this PR for now.

hareland:
@pi0, thanks for your reply!

> Thanks for this PR. yes, we need to specify build-dir but I'm afraid 1) it is too early to expand experiment and 2) we might need to use programmatic API exports of Nitro for Nuxi CLI instead of depending on nitro CLI. (I'm have it on plan ~> [nuxt/cli#353](https://github.com/nuxt/cli/issues/353))

I see, and get that.
 
> Nitro CLI itself can load `nitro.config` to respect the `buildDir` (PR welcome for this) but it won't be useful for Nuxt. Nitro will never involve in reading downstream configs ([#2196 (comment)](https://github.com/unjs/nitro/issues/2196#issuecomment-1971024371))

Could you please point me in the right direction as to how i can read the nitro config from the cli? Should i load this myself or is a there a useNitroConfig type of utility?

Also, how is nitro reading config from when you set it upstream then...? When you define a preset from nuxt.config.ts for example, it is propagated, but how does this happen, and would this not be a good place to set it by default in nuxt?

Thank sin advance!

pi0:
When nitro used in other meta frameworks (Nuxt), it won't read config itself and Nuxt provides the config after reading `nitro` key in `nuxt.config.ts` that why we cannot have this logic in Nitro but should be in Nuxt.


Options loading mechanism is here btw: https://github.com/unjs/nitro/blob/main/src/options.ts#L144

PR #2226, opened 2024-03-07
fix: respect `imports.autoImport: false`

danielroe:
### ❓ Type of change

### 📚 Description

when `imports.autoImport` is set to false we should not generate unimport declarations:

```ts
declare global {
  const __buildAssetsURL: typeof import('../../node_modules/.pnpm/nuxt@3.10.3_eslint@8.57.0_rollup@3.29.4_typescript@5.4.2_vite@5.1.5_vue-tsc@2.0.6/node_modules/nuxt/dist/core/runtime/nitro/paths')['buildAssetsURL']
  // ...
}
```

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -172,12 +172,14 @@ export async function writeTypes(nitro: Nitro) {
     }
 
     autoImportedTypes = [
-      (
-        await nitro.unimport.generateTypeDeclarations({
-          exportHelper: false,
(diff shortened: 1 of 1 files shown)


pi0 (approved):
LGTM only can you confirm it still generates #imports  type? (it is not auto)

danielroe:
It does. The only thing we are excluding is the declarations.

PR #2238, opened 2024-03-09
feat: expose `getRouteRulesForPath` from nitro runtime

danielroe:
### ❓ Type of change

### 📚 Description

This exposes `getRouteRulesForPath` from `#internal nitro`. We could use it in Nuxt ([context](https://github.com/nuxt/nuxt/pull/25841)) to avoid duplicating this logic or importing deeply from the runtime directory.

Not sure if there was a reason it was explicitly not exposed previously.

### 📝 Checklist


pi0:
We need to make (type) read-only before exposing util.

danielroe:
Should be good to go 👍

pi0 (approved):

PR #2250, opened 2024-03-12
fix: handle path negations when scanning public assets

danielroe:
### ❓ Type of change

### 📚 Description

Discovered while triaging issue with https://github.com/nuxt/nuxt/pull/26203, we basically weren't handling ignore patterns beginning with `!` if they did not also start with `!*`. (They would end up as part of a directory name.)

### 📝 Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -61,19 +61,18 @@ export async function copyPublicAssets(nitro: Nitro) {
     if (await isDirectory(srcDir)) {
       const includePatterns = [
         "**",
-        ...nitro.options.ignore
-          .map((p) => {
-            const [_, negation, pattern] = p.match(NEGATION_RE);
-            return (
-              // Convert ignore to include patterns
-              (negation ? "" : "!") +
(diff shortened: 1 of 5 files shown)


pi0:
Would you mind to add regression test to fixture?

danielroe:
Investigating it actually seems that `fast-glob` does not support negative patterns within `ignore` (upstream [issue](https://github.com/mrmlnc/fast-glob/issues/86)). That is a pretty big issue, IMHO.

What do you think about custom ignore patterns _per_ publicAssets dir? In Nuxt, we never want to ignore any build output in `dist/client/_nuxt` but want the user to be able to ignore patterns in their `~/public` directory, for example.

danielroe:
Okay, I think we're good to go here, using _include_ rather than _ignore_ patterns allows us to support negated ignore patterns without breaking current behaviour.

pi0 (approved):
Nice changes!

Issue #2215, opened 2024-03-03
Swagger trying to use hardcoded port instead .env PORT value


DreamShaded:
### Environment

NodeJS: 20.11.0
NitroPack: latest (2.9.1)

Clean Nitro, without Nuxt

### Reproduction

1. Create dotenv file at root dir with PORT and/or NITRO_PORT not equal 3000, for example, 3001
2. Set experimental.openApi as true at defineNitroConfig function
3. Write some api, for me it was greeting function returning `{ hello: 'world' }`
4. Start dev server
5. Check Nitro features working on custom port
6. Go to `/_nitro/swagger`
7. Try to execute some requests

### Describe the bug

I want to see results for executed request using my custom port.
I get the "failed to fetch" error, caused wrong request port. 

![image](https://github.com/unjs/nitro/assets/52293645/15dbaf32-e5b9-43da-a6a8-4caaf9561b13)

### Additional context

You can look at this on [StackBlitz](https://stackblitz.com/edit/github-twtf8d?file=.env), wait for init and go to `sb_url/_nitro/swagger`

PR #2251, opened 2024-03-13
fix(openapi): openapi.json ignores custom port

hanspagel:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/2215

### ❓ Type of change

### 📚 Description

Currently, we’re using a hard coded string (`http://localhost:3000`) for the base url that is added to the `openapi.json`. If a custom port is configured, it’s just ignored. This PR fixes it by using the same logic that is used in other places:

* Use `process.env.NITRO_PORT`, if it’s not available:
* Use `process.env.PORT`, if it’s not available:
* Fallback to port 3000 (the default)

While I was on it, I’ve also removed the trailing slash from the base URL. It’s not really needed, as all routes start with a `/ ` already, but it looks cleaner. ✨ 

### 📝 Checklist

### DIFF
--- a/src/runtime/routes/openapi.ts
+++ b/src/runtime/routes/openapi.ts
@@ -13,6 +13,10 @@ import { useRuntimeConfig } from "#internal/nitro";
 export default eventHandler((event) => {
   const base = useRuntimeConfig()?.app?.baseURL;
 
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR dear @hanspagel. It is fixed via https://github.com/unjs/nitro/pull/2216 (with incoming port)

PR #2229, opened 2024-03-07
fix(server-assets): mark `yaml`, `json`, `json5` and `csv` as text

manniL:
### 🔗 Linked issue

Resolves https://github.com/unjs/nitro/issues/2217

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds more extensions which are definitely not binary files.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2275, opened 2024-03-17
fix(raw): exclude `yaml` from raw plugin

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt-modules/i18n/issues/2835
* Related https://github.com/unjs/nitro/pull/2229
* Related https://github.com/unjs/nitro/pull/2239

### ❓ Type of change

### 📚 Description
Nuxt I18n is running into issues specifically with `yaml` files since nitro `2.9.2+`, I think because we're trying to parse and transform the file ourselves but it's already been handled as a raw file? 

Let me know if this is not the correct way to resolve this or if there is a way for us to exclude specific files from being handled, I'm following this PR https://github.com/unjs/nitro/pull/2239 as an example (it looks like it added checks for yaml as well anyway).

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/raw.ts
+++ b/src/rollup/plugins/raw.ts
@@ -13,8 +13,6 @@ export function raw(opts: RawOptions = {}): Plugin {
   const extensions = new Set([
     ".md",
     ".mdx",
-    ".yml",
-    ".yaml",
     ".txt",
     ".css",
     ".htm",


pi0:
Would you please make a minimal reproduction with nitro that works before 2.9 and broken with 2.9.2? 🙏🏼

BobbieGoede:
Sorry, I should have included a reproduction in the first place, I know how annoying it can be without (was a bit drained). 

Here is the reproduction: https://github.com/BobbieGoede/nuxt-i18n-yaml-nitro-raw

Interesting to note, this issue only happens when using the server side translation feature, which adds a server/nitro plugin as well as plugins to transform yaml and json5 files ([see relevant code](https://github.com/nuxt-modules/i18n/blob/main/src/nitro.ts)), there are no issues using json5 files as seen in the reproduction.

### Using nitro 2.9.1
* Install and run using `pnpm i && pnpm run dev`
* Open page and observe behaviour is as expected
* (optional) Build project using `pnpm run build`
  * Observe transformed yaml locale file at `.output/server/chunks/build/ja-[hash].mjs`

### Using nitro >2.9.2
* Remove `pnpm.overrides` from `package.json` and install and run again using `pnpm i && pnpm run dev` 
* Terminal will throw error `[worker reload] [worker init] _import_meta_url_ is not defined`, page stuck on Nuxt loading splash logging `503 status`
* (optional) Build project using `pnpm run build`
  * Observe transformed yaml locale file at `.output/server/chunks/raw/ja.mjs`

brtinney:
Just adding on that this seems to be throwing an error in my projects as well. They use the rollup yaml plugin to process the yaml files into JS objects, but after nitro 2.9.2, there's an error on dev startup saying no such file or directory. There's probably a way for me to make sure the raw plugin doesn't take over that may resolve the issue, but I assume this PR will do that too.

pi0 (approved):

PR #2284, opened 2024-03-21
 fix: make scheduledTasks types consistent

### LINKED ISSUE #2283: Scheduled tasks type on nitro options are incorrect
### Environment

Nitro 2.9

### Reproduction

— (I'll do the PR)

### Describe the bug

https://github.com/unjs/nitro/blob/8a62e7dbaeeffb6dff533c7000aba51a5c85e479/src/types/nitro.ts#L342

The types are incorrect, it should be 

```ts
 scheduledTasks?:  { [cron: string]: string[] }; 
```

I think maybe of having to handle the undefined case when we have no scheduled tasks, we could just make this property be an empty object as default (as it's assignable).

### Additional context

_No response_

### Logs

_No response_

noook:
### 🔗 Linked issue

Closes #2283, related to https://github.com/nuxt/devtools/pull/626

### ❓ Type of change

  - Not exactly breaking, but I put a deprecation warning, not sure if that's the best usage

### 📚 Description

- Make `scheduledTasks` default to `{}` instead of undefined -> Should be fixed to match the type
- Make `scheduledTasks` type consistent with the array notation -> I can revert if unwanted.

### 📝 Checklist

- [x] I have updated the documentation accordingly. (no change)

### DIFF
--- a/src/nitro.ts
+++ b/src/nitro.ts
@@ -128,6 +128,10 @@ export async function createNitro(
   nitro.options.virtual["#internal/nitro/virtual/tasks"] = () => {
     const _scheduledTasks = Object.entries(nitro.options.scheduledTasks || {})
       .map(([cron, _tasks]) => {
+        if (Array.isArray(_tasks)) {
+          nitro.logger.warn(`[DEPRECATION] \`scheduledTasks\` must be an array. This will be removed in a future release`);
+        }
+        // @todo: Remove isArray condition check when deprecation is effective
         const tasks = (Array.isArray(_tasks) ? _tasks : [_tasks]).filter(
           (name) => {
             if (!nitro.options.tasks[name]) {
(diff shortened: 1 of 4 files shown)


src/nitro.ts
```diff
@@ -128,6 +128,12 @@ export async function createNitro(
   nitro.options.virtual["#internal/nitro/virtual/tasks"] = () => {
     const _scheduledTasks = Object.entries(nitro.options.scheduledTasks || {})
       .map(([cron, _tasks]) => {
+        if (!Array.isArray(_tasks)) {
+          nitro.logger.warn(
```
pi0, line 132: I still like to support non array format.

pi0:
https://github.com/unjs/nitro/issues/2283#issuecomment-2011884154

noook:
Yes of course ! I don't mind reverting my changes at all I explained it in the PR.

However, there is still something that could be fixed, which is the default empty object for `scheduledTasks` that doesn't reflect the nitro config's typings. And it still falls under the "make type consistent". Should I reopen the PR with only this change ?

pi0:
sure

PR #2285, opened 2024-03-21
fix(options): set `scheduledTasks` to an empty object by default

### LINKED ISSUE #2283: Scheduled tasks type on nitro options are incorrect

noook:
### 🔗 Linked issue

Fixes #2283, related to https://github.com/nuxt/devtools/pull/626

### ❓ Type of change

### 📚 Description

Nitro config types indicate that `nitro.options.scheduledTasks` should always be an object, empty or not.  This PR makes the property an empty object as default as opposed to `undefined` which is not assignable to
```ts
scheduledTasks: { [cron: string]: string | string[] }
```

### 📝 Checklist

### DIFF
--- a/src/nitro.ts
+++ b/src/nitro.ts
@@ -126,7 +126,7 @@ export async function createNitro(
 
   // Virtual module for tasks (TODO: Move to rollup plugin)
   nitro.options.virtual["#internal/nitro/virtual/tasks"] = () => {
-    const _scheduledTasks = Object.entries(nitro.options.scheduledTasks)
+    const _scheduledTasks = Object.entries(nitro.options.scheduledTasks || {})
       .map(([cron, _tasks]) => {
         const tasks = (Array.isArray(_tasks) ? _tasks : [_tasks]).filter(
           (name) => {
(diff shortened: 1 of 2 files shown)


src/nitro.ts
```diff
@@ -126,7 +126,7 @@ export async function createNitro(
 
   // Virtual module for tasks (TODO: Move to rollup plugin)
   nitro.options.virtual["#internal/nitro/virtual/tasks"] = () => {
-    const _scheduledTasks = Object.entries(nitro.options.scheduledTasks || {})
+    const _scheduledTasks = Object.entries(nitro.options.scheduledTasks)
```
pi0, line 129: there is no harm to have extra runtime safety :) 

```suggestion
    const _scheduledTasks = Object.entries(nitro.options.scheduledTasks || {})
```

pi0 (approved):

PR #2287, opened 2024-03-21
fix(prerender): call `nitroApp` close hook when done prerendering

danielroe:
### ❓ Type of change

### 📚 Description

Discovered while investigating original reproduction in nuxt/cli#193.

Nitro prerenderer was never calling nitro 'close' hook which meant there was no way for runtime code to clean up after itself.

resolves nuxt/cli#193

### 📝 Checklist


pi0 (approved):
Nice idea!

PR #2288, opened 2024-03-21
fix(externals): check explicit inline rules on resolved id

pi0:
### 🔗 Linked issue

reproduction with shiki/wasm:  https://stackblitz.com/edit/unjs-nitro-starter-4aetky?file=server%2Froutes%2Findex.ts

### ❓ Type of change

### 📚 Description

Externals plugin used to only check against original id. If a resolved id is matching explicit inline patterns, it was not respected.

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/externals.ts
+++ b/src/rollup/plugins/externals.ts
@@ -150,7 +150,6 @@ export function externals(opts: NodeExternalsOptions): Plugin {
         if (!isAbsolute(originalId)) {
           const fullPath = await _resolve(originalId);
           trackedExternals.add(fullPath);
-          console.log("[externals]", id);
           return {
             id: originalId,
             external: true,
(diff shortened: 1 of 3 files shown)


src/rollup/plugins/externals.ts
```diff
@@ -138,6 +150,7 @@ export function externals(opts: NodeExternalsOptions): Plugin {
         if (!isAbsolute(originalId)) {
           const fullPath = await _resolve(originalId);
           trackedExternals.add(fullPath);
+          console.log("[externals]", id);
```
danielroe, line 153: I'm sure you're aware, but as it isn't a draft, just a quick comment to mention these three console logs...

danielroe: nice catch btw

PR #2299, opened 2024-03-22
feat: make OpenAPI and Scalar endpoint configurable

hanspagel:
### 🔗 Linked issue

–

### ❓ Type of change

### 📚 Description

1) This PR makes the OpenAPI endpoint configurable (`title`, `description`, `version`). It’s still possible to just pass `experimental.openAPI: true`, but you can also pass an object. This should make it a non-breaking addition.

2) With this PR you can also pass a configuration for Scalar (eg. to use other themes). It’s optional, but enables users to use a ton of powerful features. More on this here: https://github.com/scalar/scalar?tab=readme-ov-file#configuration

3) This PR removes the Scalar proxy. It’s really helpful to avoid CORS issues when using the API client, but you can’t use an external proxy to send request to localhost, so it’s not really helpful for local development.

Note: This PR also adds @scalar/api-reference as a dev dependency, just to import the configuration type for convenience. We could also remove it and just allow `AnyObject` (not so helpful, though) or copy the type to the package (which might be outdated quickly).

It seems not so common to add to the runtime configuration. So if there’s a better way, let me know.

I’d love to add some tests, too, but haven’t looked into that yet. That’s why I submitted it as a draft for now.

Any kind of feedback is really appreciated! We’re eager to make the OpenAPI stuff really great for everyone.

### 📝 Checklist

- [ ] I have updated the documentation accordingly. (wip)

### DIFF
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -14,6 +14,7 @@ import type { ResolvedConfig, ConfigWatcher, C12InputConfig } from "c12";
 import type { UnwasmPluginOptions } from "unwasm/plugin";
 import type { TSConfig } from "pkg-types";
 import type { ConnectorName } from "db0";
+import type { ReferenceConfiguration } from '@scalar/api-reference'
 import type { NodeExternalsOptions } from "../rollup/plugins/externals";
 import type { RollupConfig } from "../rollup/config";
 import type { Options as EsbuildOptions } from "../rollup/plugins/esbuild";
@@ -284,6 +285,7 @@ export interface NitroOptions extends PresetOptions {
       title?: string;
       description?: string;
       version?: string;
+      scalar?: ReferenceConfiguration;
     },
     /**
      * See https://github.com/microsoft/TypeScript/pull/51669
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -47,6 +47,7 @@ export interface NitroRuntimeConfig {
       [path: string]: NitroRouteConfig;
     };
   };
+  openAPI?: NitroOptions["experimental"]["openAPI"]
   [key: string]: any;
 }
 
@@ -270,6 +271,7 @@ export interface NitroOptions extends PresetOptions {
   renderer?: string;
   serveStatic: boolean | "node" | "deno" | "inline";
   noPublicDir: boolean;
+
   /**
    * @experimental Requires `experimental.wasm` to work
    *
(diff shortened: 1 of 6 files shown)


src/types/nitro.ts
```diff
@@ -278,7 +281,14 @@ export interface NitroOptions extends PresetOptions {
   wasm?: UnwasmPluginOptions;
   experimental?: {
     legacyExternals?: boolean;
-    openAPI?: boolean;
+    openAPI?:
```
pi0, line 284: I think we can start adding `openapi` top level config key preparing for 3.10 (next month release) 👍🏼 

So experimental is a boolean and top level openapi is object

src/types/nitro.ts
```diff
@@ -278,7 +281,14 @@ export interface NitroOptions extends PresetOptions {
...
+      | boolean
+      | {
+          title?: string;
+          description?: string;
+          version?: string;
+          scalar?: ReferenceConfiguration;
```
pi0, line 290: We need to namespace this. Something like `{ meta: {}, ui: { scalar: {} }` ideas welcome

hanspagel: I like meta and ui! Let’s go with that

src/types/nitro.ts
hanspagel: That makes so much sense

hanspagel: I’ve added this. Now you have to pass `experimental.openAPI: true` to activate the feature, and then just `openAPI: { … }` to configure. 👍

src/types/nitro.ts
```diff
@@ -270,12 +272,23 @@ export interface NitroOptions extends PresetOptions {
...
+  openAPI?: {
+    meta?: {
+      title?: string;
+      description?: string;
+      version?: string;
+    };
+    ui?: {
+      scalar?: ReferenceConfiguration;
+    };
+  };
```
cpreston321, line 291:
```suggestion
  openAPI?: {
    meta?: {
      title?: string;
      description?: string;
      version?: string;
    };
    ui?: {
      scalar?: ReferenceConfiguration;
    };
  } | boolean;
```

Shouldn't we add `boolean` to interface since it still supports `false` and `true` ?

pi0: i don't think it would be best. We might add `enabled` field specially because later it needs to be configurable per env (dev and prod) and not having mixed types makes relying on types easier.

pi0 (approved):

PR #2306, opened 2024-03-24
feat: generate types for runtime config

### LINKED ISSUE #1706: Automatically generate runtime config types from nitro.config the same as Nuxt
### Describe the feature

In Nuxt runtimeConfig keys/values are typed.
This improves DX considerably when accessing vis useRuntimeConfig()

Nitro has not yet implemented this, which can cause some confusion for users coming over from Nuxt

Result:
useRuntimeConfig() should be aware of the keys that exist in nitro.config.ts, and provide correct types

### Additional information

BlankParticle:
### 🔗 Linked issue

#1706

### ❓ Type of change

### 📚 Description

I have add `untyped` as a dependency and now nitro can generate types for `useRuntimeConfig()` just like Nuxt does.
It would be helpful for users to have types for `useRuntimeConfig()` which helps a lot while using nitro standalone.

For now I have utilized `nitro-config.d.ts` to keep the types in, please let me know If you would want to change the file.

Fixes #1706

### 📝 Checklist


danielroe:
It would maybe be nice to allow this to be disabled, in case there's an issue with types conflicting between auto-generated nuxt + nitro types.

BlankParticle:
> It would maybe be nice to allow this to be disabled, in case there's an issue with types conflicting between auto-generated nuxt + nitro types.

I was thinking about that, but what condition do I check against to disable it?

danielroe:
You would add a nitro option:

https://github.com/unjs/nitro/blob/d12aae10550ef8d6612da01f6065c43b02780f12/src/options.ts#L121-L128

pi0:
>  in case there's an issue with types conflicting between auto-generated nuxt + nitro types.

We can disable when framework usage is detected. What inconsistency can happen?

pi0 (approved):

danielroe (approved):

BlankParticle:
> Might be nice to add a type test to the fixture to verify that these types do in fact type the runtimeConfig. 🙏

Ok, I will add tests too

BlankParticle:
I was looking into test the types but I don't see a good way to do so, any suggestions?

PR #2319, opened 2024-03-27
fix(cloudflare-pages): use predefined wildcards in `routes.exclude`

pgraug:
### ❓ Type of change

### 📚 Description

When using Cloudflare Pages with Nitro all static files need to be present in the _routes.json file's `exclude` array.

Since only 100 entries are allowed in the file, projects can use wildcards to reduce the number entries (eg. excluding `"/_nuxt/*"` excludes all subpaths).

Nitro handles this well for `nitro.options.publicAssets` by adding the wildcard path and then not adding the individual files matching it. But if you just add a wildcard to `nitro.options.cloudflare.pages.routes.exclude` it still includes the individual files like so:

```json
{
  "version": 1,
  "include": ["/*"],
  "exclude": [
    "/cf-pages-exclude/*",
    "/cf-pages-exclude/not-in-routes-json.txt",  <-- This shouldn't be here
    ...
  ]
}
```

That's what this PR fixes. It handles both cases using `routes.exclude` (that `explicitPublicAssets` has already been appended to) for the globby ignore list, replacing the `"/*"` syntax used on Cloudflare Pages with the `"/**"` syntax that globby uses.

I added a test that fails without the fix.

### 📝 Checklist


pi0 (approved):
Thanks!

PR #2348, opened 2024-04-10
fix(prerender): only try to add `/` after `prerender:routes` hook

danielroe:
### ❓ Type of change

### 📚 Description

This is more a proposal for discussion - please feel free to close if you disagree.

As Nitro is becoming less opinionated, I think it might be good to avoid defaulting to prerendering the `/` route. It makes it difficult for frameworks to know whether this path has been _explicitly_ added by a user or module hook, or whether this is a Nitro default.

I'm looking to resolve https://github.com/nuxt/nuxt/pull/26694, and we can work around this for that PR but I think it would be simpler for a framework to add the list and not have to remove any defaults.

### 📝 Checklist


src/prerender.ts
```diff
@@ -37,12 +37,7 @@ export async function prerender(nitro: Nitro) {
     routes.add(route);
   }
 
-  // Crawl / at least if no routes are defined
-  if (nitro.options.prerender.crawlLinks && routes.size === 0) {
```
pi0, line 41: Shall we move the check after hook in meanwhile and discuss about v3 behavior later? 🙏🏼

danielroe: looks great! thank you for updating it ❤️

PR #2362, opened 2024-04-15
fix(dev): use `fs.statsSync` to determine existence of socket path

chengzhuo5:
### 🔗 Linked issue

#1115
https://github.com/nuxt/nuxt/issues/15164

### ❓ Type of change

### 📚 Description

In #1115, use `accessSync` to ensure socketPath exists. However, in some special file systems (such as the sandbox provided by security software), `accessSync` cannot determine whether the socketPath exists and will block dev server, but `statsSync` can determine correctly.

### 📝 Checklist

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -1,5 +1,5 @@
 import { Worker } from "node:worker_threads";
-import { existsSync, accessSync, promises as fsp } from "node:fs";
+import { existsSync, accessSync, statSync, promises as fsp } from "node:fs";
 import { writeFile } from "node:fs/promises";
 import { TLSSocket } from "node:tls";
 import { debounce } from "perfect-debounce";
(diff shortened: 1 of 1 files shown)


chengzhuo5:
> Thanks for PR.
> 
> > (such as the sandbox provided by security software
> 
> Can you please elaborate more which security software? I need to reproduce.

This is a sandbox software named DataCloak, which can only be used by the enterprise that purchased it, so you cannot reproduce this issue.

DataCloak on Windows will use the Windows Filtering Platform driver to take over any IO operations. When `fs.accessSync("\\?\pipe\" or "\\.\pipe\")` is called, DataCloak looks for it from the file system, but apparently it doesn't exist.

pi0:
Thanks for the explanation dear @chengzhuo5. It seems a bug/issue with this protection software and if widely unsolvable, we might propose fixes to lower-level layers (Node.js). 

I would suggest that you [contact them](https://www.datacloak.cn/apply) about it maybe they have some explanation or consider fixing it. A simpler Node.js script to reproduce might give them better context.

I have made two relevant trackers also, let's see what comes first:

- https://github.com/unjs/nitro/issues/2365 (adding support for port listening)
- https://github.com/unjs/get-port-please/issues/91 (it would still use one access but if we can reproduce sandbox issue, we can put the solution in there)

PR #2370, opened 2024-04-16
refactor: strict typechecks

pi0:
enable strict type checks across repo on 2.x branch for safer bw/fw compatibility.

There might be slight type breaking changes.


src/build.ts
```diff
@@ -499,17 +495,17 @@ async function _build(nitro: Nitro, rollupConfig: RollupConfig) {
   const rewriteRelativePaths = (input: string) => {
     return input.replace(/([\s:])\.\/(\S*)/g, `$1${rOutput}/$2`);
   };
-  if (buildInfo.commands.preview) {
+  if (buildInfo.commands!.preview) {
```
danielroe, line 498: It feels like these non null assertions might be indicative of a need to improve a type somewhere.

pi0: Well it is already set in here when constructing object https://github.com/unjs/nitro/blob/db5f87e5ccb4c7e092f48094fcef8902e781f311/src/build.ts#L471 

`NitroBuildInfo` (the public interface) is intentionally partial to make sure external consumers take this into account.

danielroe: I understand. But I’m suggesting that we should have a validator function or internal type that is more accurate than that one.

Issue #1880, opened 2023-11-03
Preset vercel: query params are ignored by cache


fmoessle:
### Environment

nitro: 2.7.0
node: v18.18.0
nuxt: 3.8.0

### Reproduction

not available

### Describe the bug

I use query params to figure out what page in a pagination the server should render, e.g. /?page=1

When configuring cache like this

```ts
routeRules: {
  '/**': {
    cache: {
      swr: true,
      maxAge: 60 * 60 * 48,
      staleMaxAge: 60 * 60 * 96
    }
  }
}
```

vercel's cache seems to ignore the query param "page". When loading "/?page=2" I receive the cached version for "/".

This doesn't seem the be an issue with Vercel, because other pages (not nuxt/nitro) don't suffer from this issue.

So my assumption is this is an issue with nitro's vercel preset.

Blemming:
This happens in node environment also, but only in production I believe

BrocksiNet:
I can confirm this issue. And also can provide a reproduction if needed.

Alex--C:
Also running into this.
Interestingly, the issue does not appear on nuxt's vercel demo: https://nuxt-vercel-isr.vercel.app/static vs https://nuxt-vercel-isr.vercel.app/static?foobar
Makes me think it might be a Vercel config issue?

aethr:
I've run into this in a Vercel preview environment when trying to migrate my API endpoints from manual configuration via `defineCachedEventHandler` to a generic config in `nuxt.config.ts`:

```ts
export default defineNuxtConfig({
  routeRules: {
    '/api/**': {
      swr: 60 * 60, // 1h
    },
  },
});
```

A bunch of API endpoints started throwing errors because they were missing a required query param. I added a log line to one of the endpoints:

```ts
  console.log('query:', JSON.stringify(getQuery(event)));
```

This shows up in the Vercel runtime logs for each request:
```
// /api/grid/10/ch4-emission?ch4Threshold=1.0276462172345253e-10&period=2024-01&page%5Blimit%5D=19
query: {}
```

A similar request against the dev server results in:
```
 query: {"ch4Threshold":"1.0276462172345253e-10","period":"2024-01","page[limit]":"195220"}
```

aethr:
[...]

PR #2375, opened 2024-04-17
feat: allow overriding allowQuery in vercel prerender-config.json

Alex--C:
### 🔗 Linked issue

#1880 

### ❓ Type of change

### 📚 Description

When using ISR on Vercel, you can define which query params should be considered for the cache using the `allowQuery` key in the `.prerender-config.json` files. This is documented in the [Build API docs](https://vercel.com/docs/build-output-api/v3/primitives#prerender-configuration-file).
This PR allows defining a custom allowQuery configuration for each routeRule.
As an example, this configuration would ensure that each query parameter is cached independently:

``` 
routeRules: {
  '/**': {
    isr: true,
    allowQuery: undefined
  },
},
```

If you consider merging this change, I will happily add some documentation for it in the vercel section. Let me know what you think.

### 📝 Checklist

### DIFF
--- a/src/presets/vercel.ts
+++ b/src/presets/vercel.ts
@@ -67,11 +67,17 @@ export const vercel = defineNitroPreset({
           funcPrefix + ".func",
           "junction"
         );
+
+        const allowQuery =
+          _hasProp(value, "allowQuery")
+            ? value.allowQuery
+            : (key.includes("/**") ? ["url"] : undefined);
+
         await writeFile(
           funcPrefix + ".prerender-config.json",
           JSON.stringify({
             expiration: value.isr === true ? false : value.isr,
-            allowQuery: key.includes("/**") ? ["url"] : undefined,
+            allowQuery,
             bypassToken: nitro.options.vercel?.config?.bypassToken,
           })
         );
--- a/src/types/nitro.ts
+++ b/src/types/nitro.ts
@@ -208,6 +208,7 @@ export interface NitroRouteRules
   extends Omit<NitroRouteConfig, "redirect" | "cors" | "swr" | "static"> {
   redirect?: { to: string; statusCode: HTTPStatusCode };
   proxy?: { to: string } & ProxyOptions;
+  allowQuery?: string[];
 }
 
 export interface NitroFrameworkInfo {


src/types/nitro.ts
```diff
@@ -208,6 +208,8 @@ export interface NitroRouteRules
   extends Omit<NitroRouteConfig, "redirect" | "cors" | "swr" | "static"> {
   redirect?: { to: string; statusCode: HTTPStatusCode };
   proxy?: { to: string } & ProxyOptions;
+  allowQuery?: string[];
+  passQuery?: boolean;
```
pi0, line 212: Since this is a vercel-specific route rule, we might need to namespace them to as `vercel?: {}`

src/presets/vercel.ts
```diff
@@ -67,11 +67,22 @@ export const vercel = defineNitroPreset({
           funcPrefix + ".func",
           "junction"
         );
+
+        let allowQuery = key.includes("/**") ? ["url"] : undefined;
```
pi0, line 71: Are there any docs that mention `url` is conventional or has a specific use case to be supported by default?

Alex--C: As far as I understand (the `url` default was there before my change and I'm pretty new to this Vercel stuff), this is necessary for catch-all routes to work at all. Otherwise each route that ends up in the catch-all would return the same cached html. In these cases `url` param gets set here: https://github.com/unjs/nitro/blob/3a69dd70bae87d12a6d9a86691fad48b2ebb044e/src/presets/vercel.ts#L238

zsotyooo:
👋 Hi,
@Alex--C @pi0 
We are facing this ISR caching issue in our Nuxt 3 application. Unfortunately, there are too many variables in the query parameters, and it's impossible to move them to route parameters. As a result of not being able to use the IRS mode, our Vercel costs skyrocketed.
I understand the difficulties of maintaining open-source projects and appreciate every amazing dev participating in them.
My intention is not to pressure you in any way, but I'd like to ask if this PR is still on your radar.

pi0:
Reworking in https://github.com/unjs/nitro/pull/2780

PR #2404, opened 2024-04-30
feat(runtime-config): export `applyEnv` and `getEnv` from nitropack

danielroe:
### ❓ Type of change

### 📚 Description

This adds `applyEnv` and `getEnv` as exports, as well as allowing them to operate on a custom `env` object instead of reading from `process.env`.

This would be a useful feature for frameworks to provide more accurate `runtimeConfig` at the build time. For example, we already inline these utils in https://github.com/nuxt/test-utils/pull/655 and would need to do the same for https://github.com/nuxt/nuxt/issues/24224 and https://github.com/nuxt/nuxt/issues/26960.

Of course, that would be fine, but it would nicer to be able to simulate the runtime behaviour of Nitro with these utils, as suggested in https://github.com/nuxt/nuxt/issues/24224#issuecomment-1805495557.

I've included them in the core `nitropack` export but I would almost prefer `nitropack/utils`, perhaps, to allow space for other similar utilities?

### 📝 Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,5 +7,7 @@ export * from "./prerender";
 export * from "./preset";
 export * from "./deps";
 
+export { applyEnv, getEnv } from "./runtime/utils.env";
+
 export { loadOptions } from "./options";
 export type { LoadConfigOptions } from "./options";
--- a/src/runtime/utils.env.ts
+++ b/src/runtime/utils.env.ts
@@ -2,15 +2,17 @@ import destr from "destr";
 import { snakeCase } from "scule";
 
 export type EnvOptions = {
+  env?: Record<string, unknown>;
   prefix?: string;
   altPrefix?: string;
   envExpansion?: boolean;
 };
 
 export function getEnv(key: string, opts: EnvOptions) {
+  const env = opts.env ?? process.env
   const envKey = snakeCase(key).toUpperCase();
   return destr(
-    process.env[opts.prefix + envKey] ?? process.env[opts.altPrefix + envKey]
+    env[opts.prefix + envKey] ?? env[opts.altPrefix + envKey]
   );
 }
 
@@ -47,7 +49,7 @@ export function applyEnv(
     }
     // Experimental env expansion
     if (opts.envExpansion && typeof obj[key] === "string") {
-      obj[key] = _expandFromEnv(obj[key]);
+      obj[key] = _expandFromEnv(obj[key], opts.env);
     }
   }
   return obj;
@@ -55,8 +57,8 @@ export function applyEnv(
 
 const envExpandRx = /{{(.*?)}}/g;
 
-function _expandFromEnv(value: string) {
+function _expandFromEnv(value: string, env: Record<string, any> = process.env) {
   return value.replace(envExpandRx, (match, key) => {
-    return process.env[key] || match;
+    return env[key] || match;
   });
 }


pi0:
Thanks for PR ❤️ 

I think this needs to be more thorough thinking before exposing from nitro runtime utils. Diverging `env` definition one more level and allowing to complete override `process.env` (while for use cases like testing makes sense) can open the door to many more mistakes and implicit issues.

Therefore putting PR on draft state until have a chance to think more. 

> I've included them in the core nitropack export but I would almost prefer nitropack/utils, perhaps, to allow space for other similar utilities?

A shared (build time) util via `nitropack/kit` seems good idea.

pi0:
As another alternative approach we can think of different methods of runtime config expansion (via a util that is not _env_) so we don't add over complexity to "environment" definition but also allow a runtime method of customizing runtime config (this would help to somehow replicate "app config" extendability feature that is going to be deprecated in Nitro v3). I love to hear your thoughts about this method.

danielroe:
It would be nice not to hard-code the use of `process.env`, even if we don't change Nitro _runtime_ behaviour - it would help to avoid duplicating these utils (which would run the risk of diverging more quickly from Nitro)... 

Are you thinking that we just rename these utils so they aren't _environment_-focused (ie. naming for clarity), so they can be used generically for expansion, either as build-time kit utils or within the runtime (first in env, and then later with new feature)? That sounds very doable - say the word and I'll refactor.

If you think this isn't so straightforward, I think we can go ahead and inline the existing nitropack behaviour in Nuxt for v3.12 like we did for nuxt/test-utils, and then take the time to get it right here, and switch downstream to using a forward-consistent implementation when it ships in `nitropack/kit`...

What do you think?

danielroe:
Thanks for the clear explanation. Yes, what was in my mind was purely about _build-time_ utilities. The change from `process.env` was not meant to affect runtime behaviour, as these utilities were not being exposed within the `runtime/` utils. (But very happy with your proposed runtime override solution!)

(The kind of usage I'm imagining was normalising `nuxt.options.runtimeConfig` - but I guess your suggestion of `useRuntimeConfig` and `updateRuntimeConfig` might be safer. Is that your take?)

pi0:
Splitting into subtopics:

- Runtime `updateRuntimeXConfig` (~> https://github.com/unjs/nitro/issues/2427)
- Build time utils (~> https://github.com/unjs/nitro/issues/2428)

I strongly belive in making a new source/definition of environment variables (as proposed with `opts.env`) in this PR can be unsafe but also understand there are cases like testing that need it and love to discuss how to approach it better.

danielroe:
Thank you! One more use-case to mention for allowing non-process environment is to 'preview' the resolved config if environment variables _were_ to be set. I'm very much not proposing using arbitrary overrides for the _actual_ resolved Nitro runtime config, just trying to make the underlying logic reusable in different contexts so it doesn't have to be duplicated.

pi0:
> to 'preview' the resolved config 

I guess at least all common runtimes allow to modify environment variables by directly modifying their env object (`Deno.env.set` and `process.env` object) if testing the exact env mapping for runtime-config is a must-have in a scenario. (*) and it is safer because:

The benefit of keeping runtime-config behavior strictly based on one standard (**) source (that is what environment variable in the runtime context means) is, that environment variables and how runtime config implicitly works with them won't diverge in a project that uses both environment variables and runtime config, behavior is consistently same. It makes tests more reliable too and also build and runtime utils consistent. 

---

(*) I understand that today, for testing modules with different sets of runtime configs, we strictly need this and also updating global env _looks_ more hacky but let's consider that 1) `updateRuntimeConfig` didn't exist and 2) updating an env in test/dev isolates gives the safety reason above.

(**) Today, we have already multiple definitions of **environment variables:** Actual system environment variables (that itself yet has no unified standard to access), `.env` augmentations, bundler **bundled** environment variables and strange runtimes like cloudflare that combine them with other values and make accessible in second stages. I belive both us and test runners have to help to at least keep on these known things rather than introducing more ways to change env variable access behavior and small places like this where we have alternatives that helps to avoid adding new sources.

pi0:
@danielroe Shall we close PR and later continue discussing about this perhaps after `updateRuntimeConfig` util landed? 🙏🏼

pi0:
Closing for triage. Followup:

- runtime: https://github.com/unjs/nitro/issues/2427
- kit util: https://github.com/unjs/nitro/issues/2428

Issue #1938, opened 2023-11-20
Netlify : undefined set-cookie


klevron:
### Environment

Node : 18.17.0
Nuxt 3.7.4 / Nitro 2.6.3
Nuxt 3.8.1 / Nitro 2.8.0

### Reproduction

Sorry I can't provide netlify link for the moment, I will try to setup one later if it's really needed (please take a look at description).

### Describe the bug

An undefined `Set-Cookie` header is sent on netlify :

![image](https://github.com/unjs/nitro/assets/4560725/fbb898f5-3676-42bc-87aa-08b724226f8f)

### Additional context

We have another site with Nuxt 3.6.5 / Nitro 2.5.2 and it works fine.

paro-paro:
I am having the exact same issue. Netlify server is setting a weird undefined cookie on every get request.  I am on latest nuxt 3.9.0 (nitro 2.8.1)

https://amazing-alfajores-2826d3.netlify.app/invest

<img width="553" alt="Screenshot 2024-01-10 at 20 12 05" src="https://github.com/unjs/nitro/assets/43136874/3c01e3d0-3ac8-4b09-8b27-fab93fcf5e40">

pi0:
A reproduction repository would be very helpful to investigate and possibly fix any issues 🙏🏼

djm-innonova:
We are experiencing this problem as well on 3.9.0 (nitro 2.8.1) - however we noticed it was already introduced when we upgraded earlier from 3.6.3 to 3.8.2 of Nuxt.

Interestingly I tried to downgrade to Nuxt 3.6.5 as originally mentioned in this post, however because of the way the dependencies are specified I anyway ended up with 2.8.1 of nitro. I then tried locking down the version of nitro to 2.5.2 (as originally specified in Nuxt 3.6.5 package.json) and still saw the `Set-Cookie: undefined` issue when deployed in Netlify.

After using the lock file we had from our previous 3.6.3 version we got it fixed now, but this indicates that the problem is not in nitro, but rather in some other dependency - I am suspecting `h3`, but did not have time to investigate if that is really the case.

klevron:
> Interestingly I tried to downgrade to Nuxt 3.6.5 as originally mentioned in this post

It works with :

```
  "nitropack": "2.5.2",
  "nuxt": "3.6.5",
```

djm-innonova:
[...]

Issue #1842, opened 2023-10-19
Netlify new functions v2 format


Rigo-m:
### Describe the feature

Netlify new v2 functions has been unveiled.
Might be worthy to implement them
https://www.netlify.com/blog/introducing-netlify-functions-2-0/

### Additional information

- [X] Would you be willing to help implement this feature?

PR #2406, opened 2024-05-01
feat(netlify): experimental support v2 functions ISR via targeted cache-control headers

serhalp:
### 🔗 Linked issue

Supersedes https://github.com/unjs/nitro/pull/1823.
Closes https://github.com/unjs/nitro/issues/1842.
Closes https://github.com/unjs/nitro/issues/1938.

### ❓ Type of change

### 📚 Description

Netlify [introduced a new Functions interface last year](https://www.netlify.com/blog/introducing-netlify-functions-2-0/). This updates Nitro to use this interface.

Previously, ISR in Nitro had been implemented on Netlify via its proprietary [On-Demand Builders](https://docs.netlify.com/configure-builds/on-demand-builders/) platform primitive, which only supports "v1" Functions. Now that there's momentum around [Targeted HTTP Cache Control](https://httpwg.org/specs/rfc9213.html) and that [Netlify supports it](https://docs.netlify.com/platform/caching/#supported-cache-control-headers), we can implement ISR in Nitro via these simple, standard HTTP headers instead.

By virtue of switching to the standard Netlify Functions interface, this also happens to:
- allow streaming responses
- add better support for "hybrid rendering", e.g. if you set `prerender: true` on a specific route, that will actually work as intended now
- remove the imposed 60s delay after a background generation request before which the new response will be served for subsequent requests (this was an ODB limitation)
- fix a bug where we were setting `set-cookie: undefined` on responses
- remove usage of an undocumented `functions-internal/server.json` Netlify config file
- ... which wasn't working when using `netlify dev` in local development anyway

This also configures a human-friendly name and some metadata on the function.

Note: I removed the `netlify_builder` preset, since there's no reason to use this anymore. I'm not even sure why it was ever exposed as a preset. I did this as a separate commit, so if you consider this a breaking change I can pull it out into a separate PR.

### 📝 Checklist


src/presets/index.ts
```diff
@@ -14,7 +14,7 @@ export { digitalOcean } from "./digital-ocean";
 export { firebase } from "./firebase";
 export { heroku } from "./heroku";
 export { edgio } from "./edgio";
-export { netlify, netlifyBuilder, netlifyEdge, netlifyStatic } from "./netlify";
+export { netlify, netlifyEdge, netlifyStatic } from "./netlify";
```
pi0, line 17: We need to keep `netlifyBuilder` for backward compatibility for users who might be using it currently

src/presets/netlify.ts
```diff
@@ -4,17 +4,32 @@ import { defineNitroPreset } from "../preset";
...
+// This is written to the functions directory. It just re-exports the compiled handler,
+// along with its config. We do this instead of compiling the entrypoint directly because
+// the Netlify platform actually statically analyzes the function file to read the config;
+// if we compiled the entrypoint directly, it would be chunked and wouldn't be analyzable.
+const generateNetlifyFunction = (handlerRelativePath: string) => {
+  return /* js */ `
+    export { default } from "${handlerRelativePath}";
+    export const config = {
+      name: "Nitro server handler",
+      generator: "${nitroPkg.name}@${nitroPkg.version}",
```
pi0, line 16: I think we can refactor this function inside hook or pass `nitro` instance to use precise generator/framework info from `nitro.options.framework`

src/runtime/entries/netlify.ts
pi0, line 1: I think we can introduce changes to new entries `netlify-v1` and `netlify-v2`

test/tests.ts
```diff
@@ -81,7 +81,7 @@ export async function setupTest(
       "vercel-edge",
       "winterjs",
     ].includes(preset),
-    isLambda: ["netlify", "aws-lambda"].includes(preset),
+    isLambda: ["aws-lambda"].includes(preset),
```
pi0, line 84: Netlify functions are not lambda anymore? or what test it fixes?

src/presets/netlify.ts
```diff
@@ -37,27 +57,6 @@ export const netlify = defineNitroPreset({
           "utf8"
         );
       }
-
-      const functionConfig = {
-        config: { nodeModuleFormat: "esm" },
-        version: 1,
-      };
```
pi0, line 44: Is v2 fully compatible with v1 for seamless migration of users?

I would prefer to keep v1 and incrementally add v2 implementation to allow opt-in for all new users unless netlify platform is not supporting v1 format anymore

pi0:
(omitted 2fc17ffcf546d33d54c4afec343c3513723c2e19 for now to reduce diff)

src/runtime/entries/netlify.ts
```diff
@@ -1,28 +1,64 @@
...
+  const r = await nitroApp.localCall({
+    url: relativeUrl,
+    headers: req.headers,
+    method: req.method,
+    body: req.body,
+  });
+
+  const headers = normalizeResponseHeaders({
+    ...r.headers,
+    ...getCacheHeaders(relativeUrl),
```
Rigo-m: Is it correct that getCacheHeaders gets spread after localCall headers?

Rigo-m:
Super @serhalp 😍

pi0 (approved):
Thanks! Merging as is mainly to allow experimenting on main branch and do pre-release refactors. 

The confusing v1/v2 scheme that I added to list commits will be removed likely before releasing next versions.

Issue #2431, opened 2024-05-10
Support Netlify-Vary header in netlify v2 preset


Rigo-m:
### Environment

See reproduction.
Relevant branches:
https://github.com/Rigo-m/netlify-v2-repro/tree/no-cdn-cache-control
https://github.com/Rigo-m/netlify-v2-repro/tree/cdn-cache-control

Check route rules.

### Reproduction

https://cache-key-variations.netlify.app/guess?value=2 you can see how this page gets cached at the edge, including the variation on the value query parameter (take a look specifically at the Netlify-Vary heade, the Cdn-Cache-Control header and the Age header for verification).

Here: https://no-cdn-cache-control--netlify-v2-nitro.netlify.app/?page=3 you can see that while the Netlify-Vary header is present, it's not actually caching the page.
Here: https://cdn-cache-control--netlify-v2-nitro.netlify.app/?page=3 you can see that by adding the Cdn-Cache-Control header everything works accordingly.

### Describe the bug

Netlify v2 preset appends a Netlify-Cdn-Cache-Control header to instruct the netlify edge platform on how to handle CDN caching.
Only setting this header makes the edge network ignore the Netlify-Vary header (somehow).
We should send the Cdn-Cache-Control header as well, so that when setting the additional header Netlify-Vary from routeRules works accordingly without handling other headers.

### Additional context

I'll drop two PRs after https://github.com/unjs/nitro/pull/2425 gets merged, one with the relevant fix, another with relevant documentation on the Netlify-Vary header for the netlify v2 preset :)

Rigo-m:
@serhalp do you have any insights on the situation?

serhalp:
Hey @Rigo-m, thanks for writing this up.

Isn't a `Vary` or `Netlify-Vary` header meaningless when no caching is enabled? I'm not sure I'm understanding why you're expecting the Netlify CDN to cache the serverless function response in this case. The function is even explicitly sending `cache-control: no-cache` on your repro site.

Rigo-m:
In the "no-cdn-cache-control" branch you can see that I've set the routeRule for caching the page [here](https://github.com/Rigo-m/netlify-v2-repro/blob/no-cdn-cache-control/nuxt.config.ts), using the new netlify v2 preset.
Now, if you visit the homepage without the query parameter it actually gets cached (check age and cache-control), whilst adding a query parameter returns no-cache.
- https://no-cdn-cache-control--netlify-v2-nitro.netlify.app/ -> cached
[...]

PR #2425, opened 2024-05-09
fix(netlify-v2): netlify not finding getGeneratorString

Rigo-m:
### 🔗 Linked issue

#2426

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Publishing to netlify, the functions fails with this error: `getGeneratorString is not defined`

@serhalp could you take a quick look?

Reproduction link in linked issue

https://github.com/Rigo-m/netlify-v2-repro/tree/patch-fix this branch contains a custom preset with the proposed fix
and it displays correctly in netlify https://patch-fix--netlify-v2-nitro.netlify.app/

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2440, opened 2024-05-13
fix(netlify): match ISR route rules when path has a query string

serhalp:
### 🔗 Linked issue

Fixes https://github.com/unjs/nitro/issues/2431.

### ❓ Type of change

### 📚 Description

See the linked issue. This was because `getRouteRulesForPath` expects a bare path with no query string, but this wasn't obvious or documented. I added a regression test and an inline doc comment.

### 📝 Checklist


Rigo-m:
That makes sense on why it wasn't working.
Thanks @serhalp!
Care to add also the quickfix present in #2425?
It makes sense to let @pi0 check only this PR and I can happily sunset the other to speed things up

pi0 (approved):
Thanks!

PR #2458, opened 2024-05-21
perf: limit open files in generateFSTree

marvin-j97:
Recreation of #2145


pi0 (approved):
Thanks for efforts on this PR. I have updated changes to reuse our existing util from prerender for handling concurrency

PR #2461, opened 2024-05-22
fix: pass custom entry filename when resolving prerenderer

### LINKED ISSUE #2444: Prerendering fails when custom serverDir and rollup.output.entryFileNames is used
### Environment

Node

### Reproduction

https://stackblitz.com/edit/github-kk3ajf?file=nitro.config.ts

1. Run the StackBlitz
2. Stop the server
3. Run npm run build
4. Note the prerender fails with a file not found error
5. Comment out either the rollup.output.entryFilenames or the output.serverDir
6. Run npm run build
7. Note the prerender finishes successfully

### Describe the bug

The prerenderer looks for an index.mjs when there is a custom entryFilenames set to a custom filename, such as server.mjs

### Additional context

_No response_

### Logs

_No response_

brandonroberts:
### 🔗 Linked issue

Closes #2444

### ❓ Type of change

### 📚 Description

This allows the `rollupConfig.options.output.entryFileNames` to be passed to the nitro prerenderer if its a string, with the fallback to `index.mjs`. This fixes an issue where prerendering fails if a custom entry filename is used.

### 📝 Checklist

### DIFF
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -79,7 +79,7 @@ export async function prerender(nitro: Nitro) {
   // Import renderer entry
   const serverFilename =
     typeof nitro.options.rollupConfig?.output?.entryFileNames === "string"
-      ? nitro.options.rollupConfig?.output?.entryFileNames
+      ? nitro.options.rollupConfig.output.entryFileNames
       : "index.mjs";
   const serverEntrypoint = resolve(
     nitroRenderer.options.output.serverDir,
--- a/src/prerender.ts
+++ b/src/prerender.ts
@@ -77,9 +77,13 @@ export async function prerender(nitro: Nitro) {
   await build(nitroRenderer);
 
   // Import renderer entry
+  const serverFilename =
+    typeof nitro.options.rollupConfig?.output?.entryFileNames === "string"
+      ? nitro.options.rollupConfig?.output?.entryFileNames
+      : "index.mjs";
   const serverEntrypoint = resolve(
     nitroRenderer.options.output.serverDir,
-    "index.mjs"
+    serverFilename
   );
   const { closePrerenderer, localFetch } = (await import(
     pathToFileURL(serverEntrypoint).href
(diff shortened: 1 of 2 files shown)


src/prerender.ts
```diff
@@ -77,9 +77,13 @@ export async function prerender(nitro: Nitro) {
   await build(nitroRenderer);
 
   // Import renderer entry
+  const serverFilename =
+    typeof nitro.options.rollupConfig?.output?.entryFileNames === "string"
+      ? nitro.options.rollupConfig?.output?.entryFileNames
```
danielroe, line 82:
```suggestion
      ? nitro.options.rollupConfig.output.entryFileNames
```

brandonroberts:
I'll check further into the failing tests

pi0:
Nice find! I think we can move the default rollup config to the [pretender preset](https://github.com/unjs/nitro/blob/main/src/presets/_nitro/nitro-prerender.ts).

brandonroberts:
> Nice find! I think we can move the default rollup config to the [pretender preset](https://github.com/unjs/nitro/blob/main/src/presets/_nitro/nitro-prerender.ts).

Cool, done!

pi0:
@brandonroberts sorry my bad just tried locally moving to preset won't work because explicit config overrides entry name still.

>  I took a different approach and just override the entry filename explicitly because the prerenderer is being built internally. Let me know if you have another recommended approach.

May i ask there was a reason to not use original approach? While since now we have only one chunk entry for pretender, if in the future we opt pretender bundle into multi chunks for less mem footprint, we might need to change solution (and since no regression test might forget). If original approach was working well, i would suggest to use that and dynamically compute `serverFilename`

brandonroberts:
That approach broke 3 test suites tests where the preset is already overriding the entryFileNames property in combination with custom serverDir and output options.

I'll take another look

pi0:
Thank you! Please let me know if could help on investing failed tests.

brandonroberts:
Should be sorted out now. I needed to use the produced `nitrorenderer` config to build the server filename dynamically.

pi0 (approved):
💚

PR #2465, opened 2024-05-24
feat: use wildcard excludes in Cloudflare Pages preset

### LINKED ISSUE #2068: Too many items generated in `_route.json` for cloudflare preset
### Environment

/

### Reproduction

Run playground from repo unsing

```sh
NITRO_PRESET=cloudflare_pages nr dev:build
```

### Describe the bug

Actually, the preset use a glob to add every files from the public directory https://github.com/unjs/nitro/blob/main/src/presets/cloudflare-pages.ts#L113. This can easily generate more than 100 paths, limit from Cloudflare, https://developers.cloudflare.com/pages/functions/routing/#limits.

Could we add a flag to generate global path like '/assets/*', '/images/*' and so on?

related to https://github.com/nuxt-modules/og-image/issues/77
related to https://github.com/unjs/website/pull/228#issuecomment-1884945571

### Additional context

_No response_

### Logs

_No response_

Brendonovich:
### 🔗 Linked issue

#2068 

### ❓ Type of change

### 📚 Description

Cloudflare Pages has a 100 entry limit across `routes.include` and `routes.exclude`. The way that `routes.exclude` is currently generated makes that limit get hit pretty quickly, since each file to be excluded is listed individually. Thankfully, CF supports using wildcards to exclude whole folders and way more than 100 files.
This PR adds the `generateExcludes` function, which attempts to combine as many of the excluded files into wildcard paths as possible, taking into account directories that contain files that have been marked specifically as non-excludable.

This setup has been working for us at [Mattrax](https://github.com/mattrax/Mattrax) for a while, but `generateExcludes` probably needs more testing and possibly some unit tests.

Resolves #2068

Also made some smaller changes:
-  The declaration of `routes` uses `satisfies` rather than a hardcoded type as it allows some non-null assertions to be removed
- `_headers` generation ignores values set for the same headers for previous sets of routes,  since I don't get the impression that headers from routeRules are intended to be additive. We rely on this as we have cache settings at `/*` that we override for certain routes (and would like nitro to override).

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare/utils.ts
+++ b/src/presets/cloudflare/utils.ts
@@ -1,5 +1,5 @@
 import { existsSync, promises as fsp } from "node:fs";
-import { resolve, join } from "pathe";
+import { resolve, join, parse } from "pathe";
 import {
   joinURL,
   withLeadingSlash,
@@ -27,11 +27,11 @@ export async function writeCFPagesStaticFiles(nitro: Nitro) {
 
 async function writeCFRoutes(nitro: Nitro) {
   const _cfPagesConfig = nitro.options.cloudflare?.pages || {};
-  const routes: CloudflarePagesRoutes = {
+  const routes = {
     version: _cfPagesConfig.routes?.version || 1,
     include: _cfPagesConfig.routes?.include || ["/*"],
     exclude: _cfPagesConfig.routes?.exclude || [],
-  };
+  } satisfies CloudflarePagesRoutes;
 
   const writeRoutes = () =>
     fsp.writeFile(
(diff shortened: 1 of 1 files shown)


GreyXor:
Hello, there's an ETA for this ? do we need to tests more ? Would be nice to have it

GreyXor:
@Brendonovich hi, gentle ping

pi0:
Thanks for PR dear @Brendonovich and sorry for late reply.

I think in order to move it forward, it would be better if we add couple of unit tests and if possible, break it down to smaller changes 🙏🏼

PR #2483, opened 2024-05-31
fix(hooks): enabled use of useAppConfig() and useRuntimeConfig() within request hooks

boenrobot:
If the nitro context isn't already initialized, it will be initialized, rather than error.

### ❓ Type of change

### 📚 Description

When calling `useAppConfig(event)` or `useRuntimeConfig(event)` from within a hook, the hook errors, because the "nitro" context option is not initialized yet at that time. It does get initialized later, making it available in middleware and route handlers. This PR extends support to include hooks as well, since although the nitro context option is not initialized, the event _is_ available, so one would expect it to work.

### 📝 Checklist


pi0:
Thanks for PR. Can you please link to a minimal reproduction please? 🙏🏼

src/types/h3.ts
```diff
@@ -20,7 +20,7 @@ declare module "h3" {
     captureError: CaptureError;
   }
   interface H3Context {
-    nitro: {
+    nitro?: {
```
HigherOrderLogic, line 23: This is a breaking change.

boenrobot: But this is already de facto the case, with or without this change.

Any Nitro plugins where builds break due to this change would be breaking the build because the plugin is now required to handle a (potential) problem when they access the "nitro" property. The behavior is not changed, just enforced on a TS level.

Any Nitro plugin that already handles the possibility of nitro being undefined when accessing it would have their builds keep working.

And any Nitro plugin that doesn't use the Nitro property isn't affected at all.

Any plugin that's currently relying on nitro already being instantiated is in fact just asserting it is, when in fact, it may not be, even today, if the code is triggered in a hook.

And Nitro itself handles it in all cases (now including also `useRuntimeConfig(event)` and `useAppConfig(event)`).

----

All of that said, I guess if there's a way to enforce Nitro's event handlers to use a context where "nitro" is not optional, while having hooks take in a context where "nitro" is optional would be best.

I'll see what I can do, and meanwhile, I amended the commit to init the errors array, as I didn't do that before.

PR #2486, opened 2024-06-01
docs: prefix `NITRO_` in `.env` file

gutenye:
From my testing, I need to prefix `NITRO_` in .env file to make it work


pi0 (commented):
Thanks!

PR #1715, opened 2023-09-11
feat: add `types:extend` hook

mukundshah:
### 🔗 Linked issue
https://github.com/unjs/nitro/issues/1710

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #1710, opened 2023-09-10
[hooks] Add `prepare:types` and `imports:extend`


mukundshah:
### Describe the feature

Having `imports:extend` and `prepare:types` will be handy to generate/alter types and imports when using custom router, like:
```ts
// server/api/hello/[...slug].ts
import { createRouter, defineEventHandler, useBase } from 'h3'

const router = createRouter()
router.get('/test', defineEventHandler(() => 'Hello World'))
export default useBase('/api/hello', router.handler)
```

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
Should be addressed by #3399

PR #2505, opened 2024-06-10
feat: add hook before prepare:types

Barbapapazes:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/core/build/types.ts
+++ b/src/core/build/types.ts
@@ -165,45 +165,27 @@ declare module "nitropack/types" {
     "export {}",
   ];
 
+  function pathToReference(path: string) {
+    return `/// <reference path="${path}" />`;
+  }
+
   const declarations = [
     // local nitropack augmentations
-    '/// <reference path="./nitro-routes.d.ts" />',
-    '/// <reference path="./nitro-config.d.ts" />',
+    "./nitro-routes.d.ts",
+    "./nitro-config.d.ts",
     // global server auto-imports
-    '/// <reference path="./nitro-imports.d.ts" />',
+    "./nitro-imports.d.ts",
   ];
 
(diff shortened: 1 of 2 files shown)


pi0:
`types:extend` hook was added via #1715 to allow extending types via a hook (#1710)

Allow extending more items like declarations is good idea, thinking to do few structure changes to allow it (we don't need a new hook, i hope)

Barbapapazes:
> `types:extend` hook was added via #1715 to allow extending types via a hook (#1710)
> 
> Allow extending more items like declarations is good idea, thinking to do few structure changes to allow it (we don't need a new hook, i hope)

_Still in my module, explained here https://github.com/unjs/nitro/pull/2506#issuecomment-2159167362_

Yes! My TypeScript knowledge led me updating what I thought was the main file `.nuxt/nuxt.d.ts` to augment Nuxt and Vue types. After, I tried to do the same to Nitro but it was not possible so I propose this change, similar to the current behavior of Nuxt.

But, after deep diving into the tsconfig file and the file created by Nuxt, it seems that Nuxt automatically add a path to the current module and because it's a directory, it's possible to add `.d.ts` file which is way more eleguant that a template file since it's not dynamic.

The Nuxt module author guide, https://nuxt.com/docs/guide/going-further/modules#adding-type-declarations, is very missleading about the usage of template and typescript and I will open an issue.

While writing these lines, I haven't check if Nitro add a similar line, a reference to the module to allow `.d.ts`, or if it's come from Nuxt but it could help to have the same (well documented) in Nitro. But there is still the question for dynamic `.d.ts` file, like the one Nitro generate for the app.config or the routes definitions.

pi0:
Thank you for initially working on this PR ❤️ It should be adressed with #3399

PR #2506, opened 2024-06-10
feat: support templates

Barbapapazes:
### 🔗 Linked issue

/

### ❓ Type of change

### 📚 Description

Hello 👋,

This PR is closely related to #2505.

With templates, modules authors could add `d.ts` file into the `buildDir` in order to augment some types. (or more usage like in Nuxt).

This is a first bare-bone implementation.

### 📝 Checklist

### DIFF
--- a/src/core/build/types.ts
+++ b/src/core/build/types.ts
@@ -248,7 +248,6 @@ declare module "nitropack/types" {
               join(nitro.options.alias["@@"] || nitro.options.rootDir, "*")
             ),
           ],
-          "#build/*": ["../*"],
           ...(nitro.options.typescript.internalPaths
             ? {
                 "nitropack/runtime": [
--- a/src/core/build/types.ts
+++ b/src/core/build/types.ts
@@ -248,6 +248,7 @@ declare module "nitropack/types" {
               join(nitro.options.alias["@@"] || nitro.options.rootDir, "*")
             ),
           ],
+          "#build/*": ["../*"],
           ...(nitro.options.typescript.internalPaths
             ? {
                 "nitropack/runtime": [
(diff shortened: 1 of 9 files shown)


pi0:
Nitro supports `vfs` option for virtual templates have you tried it? what missing features would new templates add?

Barbapapazes:
> Nitro supports `vfs` option for virtual templates have you tried it? what missing features would new templates add?

I'm trying to augment some Nitro types.

Here a Nitro plugin: 

```ts
export default defineNitroPlugin((nitro) => {
  nitro.log = console.log;
});
```

And I want to use it into a route:

```ts 
  nitroApp.log("Hello from NitroPack!");
  return 'hello';
});

```

My main issue is how to type correctly the new `log`. In a Nuxt application, adding more types to the `NuxtApp` is pretty easy by writing a new file into the `buildDir` but with Nitro, I do not know how to do it.

In the same time, I want to be able to do it from a Nitro module in order to add this new type to developer that use the module.

With a template, I can generate do:

```ts

  modules: [
    (nitro) => {
      nitro.options.templates ||= [];
      nitro.options.templates.push({
        filename: "data.mjs",
        getContents: () => {
          return `export default { hello: 'coucou' }`;
        },
      });
      nitro.options.templates.push({
        filename: "types/log.d.ts",
        getContents: () => {
          return `
declare module "nitropack/types" {
  interface NitroApp {
    log: typeof console.log;
  }
}

export {};
`;
        },
      });
    },
  ],
});
```

With #2505, I can add `log.d.ts` to the references to make everything works.

src/core/build/types.ts
```diff
@@ -248,6 +248,7 @@ declare module "nitropack/types" {
               join(nitro.options.alias["@@"] || nitro.options.rootDir, "*")
             ),
           ],
+          "#build/*": ["../*"],
```
Barbapapazes, line 251:
```suggestion
```

pi0:
Thanks for the explanation. I think we need to make the API more relevant to generated types then, not confusing with virtual source files.

pi0:
not planned for now and can be done in userland for actually writing custom files to the filesystem.

Two core functions:
- VFS to extend virtual importable files
- Typegen (to be extendable) to auto-generate types

PR #2507, opened 2024-06-10
chore(playground): does not overwrite `tsconfig.paths`

Barbapapazes:
### ❓ Type of change

### 📚 Description

Hello 👋,

This PR rework the TSConfig in the playground to not overwrite paths from the generated tsconfig file.

### 📝 Checklist


playground/nitro.config.ts
```diff
@@ -1,3 +1,17 @@
 import { defineNitroConfig } from "nitropack/config";
 
-export default defineNitroConfig({});
+export default defineNitroConfig({
+  typescript: {
+    // Because the TSConfig extends properties is not a merge but overwrite keys, we loose the paths from the NitroPack TSConfig.
```
Barbapapazes, line 5:
```suggestion
    // TSConfig `extends` property is not a merge, it overwrites keys and paths from the generated TSConfig are loosed.
```

pi0:
Thanks for noticing it. We should not need paths hack

PR #2508, opened 2024-06-11
docs: write the module guide

### LINKED ISSUE #1862: Document Nitro Modules
Implemented in #1789 (from #1788)

We shall add basic usage docs. /cc @Barbapapazes

Barbapapazes:
### 🔗 Linked issue

fix #1862

### ❓ Type of change

### 📚 Description

Hello 👋,

This PR add a new guide about modules!

I explain what is a module, how to use them, and some examples on how to add features like routes, middlewares...

### 📝 Checklist

### DIFF
--- a/docs/1.guide/96.modules.md
+++ b/docs/1.guide/96.modules.md
@@ -1,0 +1,279 @@
+---
+icon: codicon:symbol-misc
+---
+
+# Modules
+
+> Shareable and reusable Nitro code.
+
+Nitro modules are a way to **extends Nitro's runtime behavior** by accessing the Nitro context, hooking into Nitro lifecycle events, and adding new features.
+
+> [!NOTE]
+> This feature is still very new. A lot of things can be missing or broken. Please report any issues you find.
+
+## Create a local module
+
+Local modules are modules that are defined within the project. They are useful to extends the project's behavior without touching the main codebase or experimenting before creating a shareable module.
+
+To create a local module, create a folder named `modules` at the root of the project and add a file into it. Name the file as you like, for example, `my-module.ts`.
+
+Creating a module is as simple as exporting a object!
+
+```ts [modules/my-module.ts]
+import { NitroModule } from "nitropack";
+
+export default <NitroModule>{};
+```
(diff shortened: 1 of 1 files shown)


pi0:
The recipes look good, however, I strongly believe we should rethink the purpose of Nitro modules (they should be mainly for integration purposes) and first-class supporting vite might change to story so for now i do think we should keep module docs internal.

Exotelis:
Hi @pi0,

Last week, we internally evaluated Nitro as an alternative to Express and have been very impressed with Nitro's features so far. However, a core feature for are also modules. We are developing a BFF that needs to provide different API endpoints (modules) for different projects. For example, one project uses FactFinder, while another uses ElasticSearch. Would this still be within the scope of the modules, so adding endpoints via modules?
Our current solution works just fine, but we want to significantly simplify the architecture, which is where nitro comes into play.

We are delivering a POC by next week and will then make a decision. Could you provide a bit more details on the planned scope of modules in the future, maybe also a timeline?

Feel free to reach out personally if you'd like :)
Greetings from Germany

pi0:
Thanks for sharing your project status dear @Exotelis.

I believe using Nitro layers (similar concept to [Nuxt layers](https://nuxt.com/docs/getting-started/layers)) might be most fitting for your usecase. I had helped other big enterprise teams to adopt this approach for similar purpose.

Nitro can be also extended with modules but modules are not designed for this purpose.

Also since you mentioned you are migrating from a express codebase, i would suggest to first migrate to [h3](https://h3.unjs.io/) directly and then Nitro (but only a suggestion, i don't know the scope).

Nevertheless feel free to share your POC perhaps in a [discussion](https://github.com/unjs/nitro/discussions) or reach out to me (my email is in my gh profile) or via discord `@_pi0_`.

Exotelis:
Thank you very much for your incredibly quick response! Layers sound promising, will look into them 👍 
"Migrate" is such a strong word 😃 - it might even turn into a complete rewrite. That hasn't been decided yet. However, you have a valid point.

Issue #2512, opened 2024-06-12
subpath imports not working inside modules


94726:
### Environment

- Nitro: latest nightly
- Node: 21.3.0

### Reproduction

https://github.com/94726/repro-nitro-exports-undefined

1. pnpm build
2. pnpm dev

### Describe the bug

nitro-modules are imported using https://github.com/unjs/jiti 
If a module then imports from subpaths like `nitropack/kit`, it seem to search for a `node` export, which does not exist, as only `import` is defined.

### Additional context

I'll make a PR, adding `default` to each of the exports as that seems to be enough :)

### Logs

```sh
ERROR  Package subpath './kit' is not defined by "exports" in /home/maik/projects/nitro-module-exports/node_modules/nitropack/package.json

  at exportsNotFound (node:internal/modules/esm/resolve:303:10)
  at packageExportsResolve (node:internal/modules/esm/resolve:593:13)
  at resolveExports (node:internal/modules/cjs/loader:591:36)
  at Module._findPath (node:internal/modules/cjs/loader:668:31)
  at Module._resolveFilename (node:internal/modules/cjs/loader:1130:27)
  at Function.resolve (node:internal/modules/helpers:187:19)
  at _resolve (node_modules/jiti/dist/jiti.js:1:241814)
  at jiti (node_modules/jiti/dist/jiti.js:1:244531)
  at dist/index.cjs:3:13
  at evalModule (node_modules/jiti/dist/jiti.js:1:247313) 



 ERROR  Package subpath './kit' is not defined by "exports" in /home/maik/projects/nitro-module-exports/node_modules/nitropack/package.json

 ELIFECYCLE  Command failed with exit code 1.
```

94726:
resolved by https://github.com/unjs/nitro/pull/2514

PR #2513, opened 2024-06-12
fix: add default field to subpath exports

94726:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/2512

### ❓ Type of change

### 📚 Description

This adds the `default` field to each subpath export, so that these are correctly resolved by jiti, when used in modules.

Alternatively, nitro could export cjs builds, if that is preferred.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,58 +8,72 @@
   "exports": {
     "./cli": {
       "types": "./cli.d.ts",
+      "default": "./dist/cli/index.mjs",
       "import": "./dist/cli/index.mjs"
     },
     "./config": {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR but i don't think this is the right solution.

jiti has an `esmResolve` option that enables esm compatibility. We need to add it [here](https://github.com/unjs/nitro/blob/c9a7afc3ff68337c4a0e60e5e46900298e679414/src/core/module.ts#L33) feel free to make a PR for this 🙏🏼

PR #2516, opened 2024-06-12
fix(public-assets): do not shadow paths that share prefix

danielroe:
### ❓ Type of change

### 📚 Description

Encountered this as a side effect of replicating another bug.

Registering public asset dir with a base URL of, say, `/custom` will lead to paths like `/custom-thing` also being handled by the same handler (and returning a hard 404 if fallthrough is disabled). This is because we do not test if the path begins with `/custom/` but with `/custom`.

Let me know if you need me to provide a reproduction. I think the problem and fix are straightforward to see, but happy to do so!

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/public-assets.ts
+++ b/src/rollup/plugins/public-assets.ts
@@ -1,5 +1,6 @@
 import { promises as fsp } from "node:fs";
 import { relative, resolve } from "pathe";
+import { withTrailingSlash } from "ufo";
 import createEtag from "etag";
 import mime from "mime";
 import { globby } from "globby";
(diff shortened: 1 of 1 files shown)


src/rollup/plugins/public-assets.ts
```diff
@@ -112,7 +113,10 @@ export function readAsset (id) {
         const publicAssetBases = Object.fromEntries(
           nitro.options.publicAssets
             .filter((dir) => !dir.fallthrough && dir.baseURL !== "/")
-            .map((dir) => [dir.baseURL, { maxAge: dir.maxAge }])
+            .map((dir) => [
+              withTrailingSlash(dir.baseURL),
+              { maxAge: dir.maxAge },
```
pi0, line 118: I think we can fix it in config resolver instead to make sure `baseURL` always has trailing (https://github.com/unjs/nitro/blob/main/src/core/config/resolvers/assets.ts#L25)

danielroe: I assumed we had a reason for resolving without trailing slash.

https://github.com/unjs/nitro/blob/main/src/core/config/resolvers/assets.ts#L13

For example, in next lines we assume it doesn’t have one (when we generate route rules).

https://github.com/unjs/nitro/blob/main/src/core/config/resolvers/assets.ts#L48

The other issue is if the config is modified subsequently (by, for example a hook), then this issue will reoccur. It feels safer to modify within the plugin.

But of course if you prefer I can update in the config loader/resolver.

pi0 (approved):

PR #2517, opened 2024-06-12
fix(prerender): extract links from explicit html routes

danielroe:
### ❓ Type of change

### 📚 Description

Discovered while triaging https://github.com/nuxt/nuxt/issues/27490 (which is a bug in Nuxt). We are not crawling paths that are prerendered with an explicit `.html` suffix.

It feels like this might be an oversight, and we could benefit from this in Nuxt (adding prerender headers when rendering `200.html`) but if you think it might pose any issues, feel free to close.

### 📝 Checklist

### DIFF
--- a/src/core/prerender/prerender.ts
+++ b/src/core/prerender/prerender.ts
@@ -273,7 +273,7 @@ export async function prerender(nitro: Nitro) {
     }
 
     // Crawl route links
-    if (!_route.error && isImplicitHTML) {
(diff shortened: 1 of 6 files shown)


pi0 (approved):
Quick look seems good. It would be nice if you can add it to the tests mainly because i am planning to have some major refactors on prerenderer and don't want it to cause a regression with fix of this PR

PR #2519, opened 2024-06-13
feat(defineCachedEventHandler): add `event.context.cache`

Atinux:
This allow the frontend framework connected to Nitro to be aware that the page is cached, useful for handling a flag in the payload to handle a different logic on client-side (e.g: fetch the auth state after hydration)

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -349,7 +349,7 @@ export function defineCachedEventHandler<
           fetch: globalThis.$fetch as any,
         })) as $Fetch<unknown, NitroFetchRequest>;
       event.context = incomingEvent.context;
-      event.context.cached = true;
+      event.context.isCached = true;
       const body = (await handler(event)) || _resSendBody;
 
       // Collect cacheable headers
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -349,6 +349,7 @@ export function defineCachedEventHandler<
           fetch: globalThis.$fetch as any,
         })) as $Fetch<unknown, NitroFetchRequest>;
       event.context = incomingEvent.context;
+      event.context.cached = true;
       const body = (await handler(event)) || _resSendBody;
 
       // Collect cacheable headers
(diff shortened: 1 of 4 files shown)


src/runtime/internal/cache.ts
```diff
@@ -349,6 +349,7 @@ export function defineCachedEventHandler<
           fetch: globalThis.$fetch as any,
         })) as $Fetch<unknown, NitroFetchRequest>;
       event.context = incomingEvent.context;
+      event.context.cached = true;
```
manniL, line 352: Maybe `isCached`? 
```suggestion
      event.context.isCached = true;
```

Atinux: Makes more sense yes.

Also I should check if it respect the `shouldBypassCache` too

pi0:
Thanks for PR. I thibnk we should make it to `context.cache: {}` to preserve more cache state.

Atinux:
Updated to set `cache` as empty object to start, we can follow up later in other pull requests.

Also added a basic test case to improve later on too.

pi0 (approved):
Thanks!

Updates:
- Added `cache.options`
- Added h3 type augmentations (locally broken on main)

meirroth:
Hi @Atinux, Is this feature planned for Nuxt, or is it a use case you are developing? I'm curious to see how it can be used. I believe this can be very helpful for caching heavy pages while keeping them dynamic for users.

manniL:
And for "nuxt-rendered" parts (so no server api / server routes), https://github.com/nuxt/nuxt/issues/26598 is where we track it.

meirroth:
@pi0 and @manniL, thank you for your replies! 

I'm trying to understand how to apply this in practice, so an example would be really helpful if you have one.

My use case involves caching a page with SWR for all except requests with an auth header. This way, visitors will see a cached page, while logged-in users will see a personalized version of the page, which differs for each user.

I  just watched @manniL's video on Dynamic Rendering SSR/SPA. (Great video as always!) Is the approach similar? Would it involve defining server middleware that checks for an auth header and, if it exists, disables the cache with event.cache.option?

Additionally, I assume the i18n module adds complexity since you'd need to cache pages for each language.

Issue #2466, opened 2024-05-24
Possibility to render a single route directly to stdout


jrutila:
### Describe the feature

There should be a way to use nitro to generate single pages. One place to run this would be Azure Function that just outputs, using prebuilt nitro files, the desired page output when it is run and then dies out.

This should be possible through the `prerender` function and a command line tool.

Command line tool would work like this: `nitro render /path/to/page` and when you want stdout: `nitro render --stdout /path/to/page`

I have already studied the `generateRoute` code and will post a PR that will implement the needed options and the cli tool.

### Additional information

- [X] Would you be willing to help implement this feature?

jrutila:
This would also trickle down to frameworks like `nuxt` that could extend its `generate` command to support single page and stdout renderings.

jrutila:
This could be extended so that an `output` function could be provided with the `prerender` options. This function would be called with the Buffer and path so that anything could be done with the rendered output. Like, sending to email, saving to cloud storage etc.

jrutila:
You can test this feature like this:
- Add a route to `playground/routes/`, for example `abba.ts` that has this
```
export default eventHandler(async (event) => {
  return "AbbA";
});
```
- run `pnpm nitro build --dir playground` to first build the playground
- run `pnpm nitro render --dir playground --stdout abba` 
- you should now see "AbbA" in the output of the previous render command

pi0:
This is a nice idea but perhaps something to rethink for the future. (some context: https://github.com/unjs/nitro/pull/2467#issuecomment-2172866120)

jrutila:
I would like to work on this feature. You said in the PR: "Nitro has already a CLI preset intended for same thing (standalone output that can render to stdout without depending in nitro builder core) -- PR welcome to improve it". Do you mean the `static` preset for `build` command? What would be the desired way to call it from cli?

And, regarding the programmatic API, is the `src/core/prerender/prerender.ts` file's `generateRoute` the correct place to implement this (as I did on the PR)? I couldn't find the mentioned `prerenderRoute` function anywhere. Or am I in wrong branch, now?

pi0:
> Do you mean the static preset for build command? 
> What would be the desired way to call it from cli?

I meant `--preset cli` [(src)](https://github.com/unjs/nitro/blob/2ca8021b258b6519b5c972b91289c8ea388e4492/src/presets/node/preset.ts#L40).

(I guess not working on latest branch needs a simple fix to remove CJS check)

[...]

PR #2538, opened 2024-06-17
fix: do not use require.main in cli preset

jrutila:
### 🔗 Linked issue

Part of implementing #2466 

### ❓ Type of change

### 📚 Description

The generated `server/index.mjs` is no longer dependent on the `require`. I don't think there is a valid way to check if the file is the main with ES modules?

Also, it uses `Object.entries` for result headers.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
   },
   "resolutions": {
     "nitropack": "link:.",
-    "undici": "^6.18.2"
+    "undici": "^6.19.0"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


pi0:
Would you please rebase it to v2 branch? 

Also trick for ESM main check: https://github.com/unjs/unenv/blob/main/test/node-coverage.ts#L75

jrutila:
Uhh... sorry, need to fix more...

PR #2550, opened 2024-06-20
fix(types): infer types correctly when method is omitted

danielroe:
### ❓ Type of change

### 📚 Description

Not sure when it was introduced, but at some point there was a regression in the types for `$fetch` when a method is omitted. Previously, the `get` method was selected, but now a union of all possible endpoints is provided.

This is the remaining issue blocking types in https://github.com/nuxt/nuxt/pull/27702.

~~This PR adds a type test - I plan to update when the fix is found.~~

cc: @pi0 

### 📝 Checklist

### DIFF
--- a/src/types/fetch/fetch.ts
+++ b/src/types/fetch/fetch.ts
@@ -78,7 +78,13 @@ export interface $Fetch<
   >(
     request: R,
     opts?: O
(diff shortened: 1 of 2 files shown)


pi0:
https://github.com/unjs/nitro/commit/9a71720396ca1c7af804a1ec588732aa78ae1a21

PR #2563, opened 2024-06-26
fix: backward compatibility types for v2

pi0:
- Add missing type exports from `nitropack` to avoid breaking changes
- Portal augmented types for RuntimeConfig and AppConfig if imported from `nitropack` to extend from ones from `nitropack/types`


src/types/config.ts
```diff
@@ -341,10 +341,12 @@ export type DatabaseConnectionConfigs = Record<
 >;
 
 // Runtime config
+
```
pi0, line 344:
```suggestion
```

src/types/config.ts
```diff
@@ -341,10 +341,12 @@ export type DatabaseConnectionConfigs = Record<
...
 export interface NitroRuntimeConfigApp {
   baseURL: string;
   [key: string]: any;
 }
+
```
pi0, line 349:
```suggestion
```

Issue #2569, opened 2024-06-27
Compute runtimeConfig at startup


cjpearson:
### Describe the feature

Our Nuxt application has a fairly large runtime configuration and we've noticed that performance of the application decreases as the size of this config grows. The call to `applyEnv` takes up a large portion of the execution time. This problem is exacerbated by the fact that the runtime config is calculated for every request. 

We've prototyped a simple plugin to compute the runtime config at startup and re-use for each request. 

```
export default defineNitroPlugin((nitroApp: NitroApp) => {
  const runtimeConfig = useRuntimeConfig() as RuntimeConfig
  nitroApp.hooks.hook('request', (event) => {
    event.context.nitro = event.context.nitro ?? {}
    event.context.nitro.runtimeConfig = runtimeConfig
  })
})
``` 

So far the results seem to be promising, and I think it might also be beneficial for those with smaller runtime configs. Do you think it makes sense to include something like this in the nitro core or is there a need for each request to have a fresh runtime config?

### Additional information

- [X] Would you be willing to help implement this feature?

pqnet:
Is is not enough to use [lazy event handlers](https://h3.dev/guide/basics/handler#lazy-handlers) to compute it once for each handler?

Togetic:
A production datapoint for this, plus a reframing that I think makes it a much easier "yes" for v2 — because the expensive part turns out not to be the part this issue proposes to change.

**Cost, measured on a real build.** Large-ish Nuxt app (crypto exchange, SSR on Node): the inlined runtime config resolves to **2726 keys / 43.9 KB**, so `applyEnv` performs 2726 `getEnv` calls per request — each a `snakeCase(key).toUpperCase()` plus two `process.env` reads and a `destr()`, i.e. ~5.4k environment reads per request. Split:

| | per request |
|---|---|
| `klona(_inlineRuntimeConfig)` | **0.04 ms** |
| `applyEnv` walk | **1.27 ms** |
| total | **1.31 ms** |

End-to-end, with @cjpearson's plugin from the issue body versus without it — two full production builds, 30k requests at concurrency 16, comparing the server process's own CPU time:

| | CPU/req | rps |
|---|---|---|
| without | 1.66 ms | 638 |
| with | 0.26 ms | 4385 |

[...]

PR #2575, opened 2024-07-01
fix(app): reuse computed runtime config across requests

cjpearson:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/2569

### ❓ Type of change

### 📚 Description

In our Nuxt application, we noticed that as the size of the runtime config the performance decreases significantly. Loading the runtime config occurs in every request and was making up a significant portion of the request time. Since there is not a need to recompute the config every request, we created a plugin to load the config at startup instead of for each request. 

```
export default defineNitroPlugin((nitroApp: NitroApp) => {
  const runtimeConfig = useRuntimeConfig() as RuntimeConfig
  nitroApp.hooks.hook('request', (event) => {
    event.context.nitro = event.context.nitro ?? {}
    event.context.nitro.runtimeConfig = runtimeConfig
  })
})
```

But I think this performance improvement would also be beneficial for other Nitro users and would make sense in core. (Unless my assumptions here are wrong). 

I've done some very rough tests with autocannon against a cached page to show the impact we're seeing. It's nearly twice as fast in the average case.

**Before Change**

```
┌─────────┬───────┬───────┬───────┬───────┬─────────┬──────────┬─────────┐
│ Stat    │ 2.5%  │ 50%   │ 97.5% │ 99%   │ Avg     │ Stdev    │ Max     │
├─────────┼───────┼───────┼───────┼───────┼─────────┼──────────┼─────────┤
│ Latency │ 23 ms │ 28 ms │ 36 ms │ 39 ms │ 31.3 ms │ 53.74 ms │ 1370 ms │
└─────────┴───────┴───────┴───────┴───────┴─────────┴──────────┴─────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 45      │ 45      │ 344     │ 355     │ 314.4   │ 90.13   │ 45      │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 12.3 MB │ 12.3 MB │ 94.3 MB │ 97.3 MB │ 86.2 MB │ 24.7 MB │ 12.3 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘
┌──────┬───────┐
│ Code │ Count │
├──────┼───────┤
│ 200  │ 3144  │
└──────┴───────┘
```

**After Change**

```
┌─────────┬───────┬───────┬───────┬───────┬──────────┬──────────┬─────────┐
│ Stat    │ 2.5%  │ 50%   │ 97.5% │ 99%   │ Avg      │ Stdev    │ Max     │
├─────────┼───────┼───────┼───────┼───────┼──────────┼──────────┼─────────┤
│ Latency │ 13 ms │ 15 ms │ 21 ms │ 25 ms │ 17.75 ms │ 48.12 ms │ 1142 ms │
└─────────┴───────┴───────┴───────┴───────┴──────────┴──────────┴─────────┘
┌───────────┬─────┬──────┬────────┬────────┬────────┬─────────┬────────┐
│ Stat      │ 1%  │ 2.5% │ 50%    │ 97.5%  │ Avg    │ Stdev   │ Min    │
├───────────┼─────┼──────┼────────┼────────┼────────┼─────────┼────────┤
│ Req/Sec   │ 0   │ 0    │ 616    │ 642    │ 547.6  │ 188.61  │ 472    │
├───────────┼─────┼──────┼────────┼────────┼────────┼─────────┼────────┤
│ Bytes/Sec │ 0 B │ 0 B  │ 169 MB │ 176 MB │ 150 MB │ 51.7 MB │ 129 MB │
└───────────┴─────┴──────┴────────┴────────┴────────┴─────────┴────────┘
┌──────┬───────┐
│ Code │ Count │
├──────┼───────┤
│ 200  │ 5476  │
└──────┴───────┘
```

### 📝 Checklist


jamaluddinrumi:
thanks @cjpearson 

i use this approach to make sure runtime config is filled by environment variables.

```typescript
// plugins/runtime-config.ts

export default defineNitroPlugin((nitroApp: NitroApp) => {
  nitroApp.hooks.hook('request', (event) => {
    event.context.nitro = event.context.nitro ?? {}
    event.context.nitro.runtimeConfig = {
      cloudflarePagesWebhookUrl: process.env.CLOUDFLARE_PAGES_WEBHOOK_URL,
      turnstileSecretKey: process.env.NITRO_TURNSTILE_SECRET_KEY,
      turnstileEndpoint: process.env.NITRO_TURNSTILE_ENDPOINT,
    }
  })
})
```

pi0:
Sorry, this was left unattended for a long time. 

It would be a breaking change to do because top-level runtime config is non-shared in Nitro v2 (long story, it was initially like this only because Cloudflare Workers only made env variables available only after the first request, so we had to invent this cloning mechanism  so that there is a cloned version bound to request that has env variables available)

A solution like @jamaluddinrumi suggested to override `event.context.nitro.runtimeConfig` should work.

---

But good news is that, in Nitro v3, not only runtime config is evaluated once and shared, it won't be even evaluated or be in bundle unless explicitly using `useRuntimeConfig()` from `nitro/runtime-config`.

cjpearson:
Thanks for the explanation! Totally understand the concerns around breaking v2 and happy to hear that something similar will be v3

PR #2577, opened 2024-07-01
docs: move `await` to explicit line for clarity

iiio2:
### ❓ Type of change

### 📚 Description

Async function always returns a promise. There is no need to explicitly write await unless it is within try/catch. I think here await is kind of redundant.  

https://eslint.org/docs/latest/rules/no-return-await
https://dev.to/danielbellmas/return-await-in-a-trycatch-4oh

### 📝 Checklist

### DIFF
--- a/docs/1.guide/8.assets.md
+++ b/docs/1.guide/8.assets.md
@@ -116,6 +116,6 @@ Then you can use the `assets:templates` base to retrieve your assets.
 
 ```ts [handlers/success.ts]
 export default defineEventHandler(async (event) => {
-  return await useStorage('assets:templates').getItem(`success.html`)
+  return useStorage('assets:templates').getItem(`success.html`)
 })
 ```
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR but probably await can stay there it was mostly to show users it returns a promise.

docs/1.guide/8.assets.md
```diff
@@ -116,6 +116,6 @@ Then you can use the `assets:templates` base to retrieve your assets.
 
 ```ts [handlers/success.ts]
-  return await useStorage('assets:templates').getItem(`success.html`)
+  return useStorage('assets:templates').getItem(`success.html`)
```
pi0, line 119:
```suggestion
  const html = await useStorage('assets:templates').getItem(`success.html`)
  return html
```

iiio2:
Thanks @pi0.

PR #2591, opened 2024-07-07
fix: hide unhandled error messages in prod

pi0:
We should hide error messages (to client) in case of unhandled errors because they can include internal server state.

Thanks for @galactichypernova for discovering this! (issue was if a chunk is missing, we were showing full path but it can be extended to other causes)

Note: This might be a minor behavior change but since we only hide unhandled errors, (`new Error` not `createError`) i guess should be safe for v2 as patch.

### DIFF
--- a/src/runtime/internal/utils.ts
+++ b/src/runtime/internal/utils.ts
@@ -55,33 +55,41 @@ export function isJsonRequest(event: H3Event) {
   );
 }
 
-export function normalizeError(error: any) {
+export function normalizeError(error: any, isDev?: boolean) {
   // temp fix for https://github.com/unjs/nitro/issues/759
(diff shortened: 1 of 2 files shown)


src/runtime/internal/utils.ts
```diff
@@ -55,33 +55,41 @@ export function isJsonRequest(event: H3Event) {
...
-          (line.includes("node_modules") && !line.includes(".cache")) ||
-          line.includes("internal") ||
-          line.includes("new Promise"),
-      };
-    });
+
+  // Hide details of unhandled/fatal errors in production
+  const hideDetails = !isDev && (error.unhandled || error.fatal);
+
+  const stack = hideDetails
```
pi0, line 66: (we were hiding stack previously. this is a an additional guard added as part of refactor)

Issue #1992, opened 2023-12-08
SWR cached routes won't ever update content if response is empty or response is 404


mdarrik:
### Environment

##Reproduction env: 
- Stackblitz
- Nitro v2.8.1
- Node v18.18.0

## Env where encountered

- Operating System: Linux
- Node Version:     v18.17.1
- Nuxt Version:     3.8.1
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  yarn@1.22.19

### Reproduction

https://stackblitz.com/edit/github-wlae5p?file=routes%2Findex.ts

1. Start the site either with `npm run dev` or `npm build && node .output/server/index.mjs`
2. Navigate to the `add-value-to-storage` endpoint
3. Navigate to the `swr` endpoint (will see `nitro: is-awesome`)
4. Navigate to the `not-swr` endpoint (will see `nitro: is-awesome`)
5. Navigate to the `remove-value-from-storage` endpoint
6. Navigate to the `not-swr` route and refresh until the cache expires
7. Will eventually get a blank response
8. Navigate to the `swr` endpoint
9. It'll continue to return the `nitro: is-awesome` response indefinitely.

### Describe the bug

Routes cached by SWR cannot be updated to return empty/404 values without fully clearing the cache. 

## Description
We currently have a high-traffic Nuxt 3 site that's backed by a CMS. We have the pages and a Nuxt/Nitro API set to cache with SWR for 60 minutes.

When content editors delete a page, our CMS API will return `null`, which is what we'd expect our Nitro API to return. However, if the nitro API returns `null` or manually returns a 404 response, Nitro never seems to invalidate the SWR cache. This means we can't delete pages without manually clearing entries from the cache. This feels unexpected since there might be valid reasons for an API to return a 404 or 204 response.

infabo:
Why is this issue stale? This is a real-world issue.

pi0:
This is now fixed in [unjs/ocache](https://github.com/unjs/ocache) (used by Nitro v3):

- When SWR background revalidation throws (e.g. 404 error), the stale cache entry is evicted from storage
- When SWR background revalidation returns a value that fails validation (e.g. empty/undefined), the stale cache entry is evicted from storage

This prevents the "stale forever" behavior where deleted content keeps being served.

See https://github.com/unjs/ocache/pull/9

PR #2611, opened 2024-07-12
fix(cache): remove cached value when revalidation results in an error…

cjpearson:
… or invalid value

### 🔗 Linked issue

#1992 

### ❓ Type of change

### 📚 Description

When an SWR revalidation occurs the existing value is not initially deleted from the cache storage. Instead it only overwritten after the revalidation is complete. However, if the revalidation results in an error or invalid value, the original value remains in the cache storage. This can result in a stale value persisting in the cache throughout many revalidations, long after it should have been replaced. 

Another approach could be to preemptively delete the stored value. This may simplify the implementation, but it means that other requests which arrive during the revalidation will not be served the cached value.

### 📝 Checklist


src/runtime/internal/cache.ts
```diff
@@ -137,6 +137,18 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
       _resolvePromise.catch((error) => {
         console.error(`[nitro] [cache] SWR handler error.`, error);
         useNitroApp().captureError(error, { event, tags: ["cache"] });
+
+        // SWR revalidation failed, remove existing entry so we do
+        // not continue to return the cached value indefinitely.
+        const promise = useStorage()
+          .removeItem(cacheKey)
```
pi0, line 144: we do not store invalid (error, empty) values in cache -- they can be expired though. if SWR handler fails, old value should be served to requests until it can successfully be revalidated.

I'm reverting this part to unblock PR, we can discuss later if you think we still should.

dl-ias:
Hi! Is it still under construction?

momi-aura:
Is this still active? 👀

infabo:
Why got this PR no attention?

pi0x:

Nitro v3 now delegates caching to the standalone [`ocache`](https://github.com/unjs/ocache) package, and the exact behavior proposed in this PR is already implemented there. In `src/cache.ts`, the storage entry is evicted from all matched tiers when the resolver throws during revalidation, and is likewise evicted when a freshly-resolved value fails the `validate()` check — preventing errors or invalid values from being cached or silently served afterward. This matches the intent of this PR precisely, just implemented at the ocache level rather than in nitro directly. A human maintainer should confirm and decide whether to close this in favor of the ocache implementation.

PR #2619, opened 2024-07-16
fix: only hide message of unhandled errors in prod

pi0:
Fix regression of https://github.com/unjs/nitro/pull/2591 that hides messages of `fatal: true` errors. 

Context: As pointed out in https://github.com/nuxt/nuxt/pull/28156#discussion_r1679720615, the current assumption was that `fatal` can be also used by user-thrown errors. 

In the future, we might introduce a new `internal` flag to h3 errors to clearly separate internal and user errors from each other and keep (internal) fatal also private.

### DIFF
--- a/src/runtime/internal/utils.ts
+++ b/src/runtime/internal/utils.ts
@@ -60,36 +60,35 @@ export function normalizeError(error: any, isDev?: boolean) {
   // TODO: investigate vercel-edge not using unenv pollyfill
   const cwd = typeof process.cwd === "function" ? process.cwd() : "/";
 
-  // Hide details of unhandled/fatal errors in production
-  const hideDetails = !isDev && (error.unhandled || error.fatal);
-
-  const stack = hideDetails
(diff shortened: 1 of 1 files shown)


src/runtime/internal/utils.ts
```diff
@@ -60,36 +60,35 @@ export function normalizeError(error: any, isDev?: boolean) {
...
-          return {
-            text,
-            internal:
-              (line.includes("node_modules") && !line.includes(".cache")) ||
-              line.includes("internal") ||
-              line.includes("new Promise"),
-          };
-        });
+  const stack =
+    !isDev && (error.unhandled || error.fatal)
```
danielroe, line 64: do we want to process the stack on handled/non-fatal errors in production, even though it will not be used at all in production? it feels like it will be more performant not to process the stack at all.

```suggestion
    !isDev
```

danielroe (approved):

Issue #2634, opened 2024-07-26
Task cli forces .nitro as buildDir


TMBL-DEV:
### Environment

nitro: 2.9.6
node: v20

### Reproduction

i think the description together with the link to the code that causes this is more then clear for anyone to understand.

### Describe the bug

# .nitro forced as buildDir for tasks cli

## details
Like the title says currently when you want to run a task through the cli like [the docs show](https://nitro.unjs.io/guide/tasks#using-cli) you will run into the issue that it forces .nitro to be your build directory.

This is caused by [this line](https://github.com/unjs/nitro/blob/62a74df2cac30fb6e73d4a83959e24daa3bcfa73/src/cli/commands/task/run.ts#L47)

This causes the issue that if you have a custom build directory or if you use nuxt that the task cli won't work.

## proposal
utilize the `nitro.options.buildDir` as the default buildDir for the task cli.

PR #2635, opened 2024-07-26
fix(cli): respect custom `buildDir` option for `tasks` sub command

TMBL-DEV:
updated list & run to get a nitro instance and grab the buildDir from the nitro options and at the close the nitro instance.

### 🔗 Linked issue
[#2634 Task cli forces .nitro as buildDir](https://github.com/unjs/nitro/issues/2634)

### ❓ Type of change

#### Changes
In run.ts & list.ts from the task command i used `createNitro()` to have a nitro instance so i could access the options.
i use the `nitro.options.builDir` to replace the hardcoded `".nitro"` buildDir that is passed to the runTask function.
at the end of the command run function i also close the the nitro instance that i created.

#### Why
Now if your `buildDir` in the `nitro.config.ts `is not the default .nitro directory instead of getting an error it uses the correct buildDir and your command will run.
This will also make sure that the task cli works with nuxt, because nuxt uses the the buildDir .nuxt.

### 📝 Checklist


src/cli/commands/task/list.ts
```diff
@@ -16,13 +16,17 @@ export default defineCommand({
   },
   async run({ args }) {
     const cwd = resolve((args.dir || args.cwd || ".") as string);
-    const tasks = await listTasks({ cwd, buildDir: ".nitro" });
+    const nitro = await createNitro({ rootDir: cwd });
```
pi0, line 19: I think you can use lighter `loadConfig` to load nitro config only

TMBL-DEV: I'll take a look at it today

TMBL-DEV: i used the loadOptions() because i couldn't find an example on the loadConfig function.
I hope that's okay.

TMBL-DEV:
@pi0 is this good?

pi0 (approved):

Issue #2657, opened 2024-08-13
500 error on invalid paths for cached event handlers


cjpearson:
### Environment

Node v20.9.0, nitropack@2.9.7

### Reproduction

https://stackblitz.com/edit/github-qj829h?file=server%2Froutes%2Findex.ts

### Describe the bug

For a cached event handler, sending a request an invalidly encoded path will cause a 500 error. This is similar to #717. I would expect either a 404 error or for nitro to ignore the invalid path.

### Logs

```sh
[nitro] [request error] [unhandled] URI malformed
  at decodeURI (<anonymous>)
  at getKey (./.nitro/dev/index.mjs:305:35)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async ./.nitro/dev/index.mjs:273:17
  at async Object.handler (./.nitro/dev/index.mjs:461:22)
  at async ./node_modules/h3/dist/index.mjs:1975:19
  at async Object.callAsync (./node_modules/unctx/dist/index.mjs:72:16)
  at async Server.toNodeHandle (./node_modules/h3/dist/index.mjs:2266:7)
```

cjpearson:
Fixed in 2.10.

PR #2658, opened 2024-08-13
fix(cache): try decode path

cjpearson:
### 🔗 Linked issue

#2657

### ❓ Type of change

### 📚 Description

This follows the same approach as https://github.com/unjs/nitro/pull/1459, replacing the `decodeURI` with `decodePath` from `ufo`. This means that an invalid path will no longer throw an error but instead use the undecoded text.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -220,8 +220,14 @@ export function defineCachedEventHandler<
       // Auto-generated key
       const _path =
         event.node.req.originalUrl || event.node.req.url || event.path;
-      const _pathname =
-        escapeKey(decodeURI(parseURL(_path).pathname)).slice(0, 16) || "index";
+      let _pathname: string;
+      try {
+        _pathname =
+          escapeKey(decodeURI(parseURL(_path).pathname)).slice(0, 16) ||
+          "index";
+      } catch {
+        _pathname = "-";
+      }
       const _hashedPath = `${_pathname}.${hash(_path)}`;
       const _headers = variableHeaderNames
         .map((header) => [header, event.node.req.headers[header]])


src/runtime/internal/cache.ts
```diff
@@ -221,7 +221,7 @@ export function defineCachedEventHandler<
       const _path =
         event.node.req.originalUrl || event.node.req.url || event.path;
       const _pathname =
-        escapeKey(decodeURI(parseURL(_path).pathname)).slice(0, 16) || "index";
+        escapeKey(decodePath(parseURL(_path).pathname)).slice(0, 16) || "index";
```
pi0, line 224: This is arguably _less_ safe because `decodePath` falls back to user input (which can be dangerous with unseen paths).

pi0: We might have a local try catch system that _removes_ (fallback to empty string or "??" or something like tat) in case of decode failure.

cjpearson: So my original approach looked like:

```
let _pathname: string;
  try {
    _pathname = escapeKey(decodeURI(parseURL(_path).pathname)).slice(0, 16) || "index"
  } catch {
    throw createError({
      message: 'Invalid URL',
      statusCode: 404,
    });
  }
```

But then I changed to match the behavior of the static handler. Would something like that work better?

> We might have a local try catch system that removes (fallback to empty string or "??" or something like tat) in case of decode failure.

I don't think we can use a common fallback because then the cache key will be identical for different paths. If the handler that is wrapped by the cache handler doesn't mind the invalid encoding it may return different responses for different paths and expect them to be cached separately.

But maybe there's another way to mitigate the risk? Could you explain where the risk is since I'm not sure I understand? We are already using the user input to generate the cache key when URI decoding does not fail.

pi0: > the cache key will be identical for different paths

Full path is always added as hash (L225). `_pathname` is only a human friendly part to assist identify keys.

cjpearson: Oh, I see now. So you're thinking something like this?

let _pathname: string;
try {
  _pathname = escapeKey(decodeURI(parseURL(_path).pathname)).slice(0, 16) || "index"
} catch {
  _pathname = "??"
}

pi0: Yes.. Althoigh i guess `?` is illegal in windows. Perhaps `-`?

cjpearson: Tested it out locally and this approach works as well. 👍 Thanks for the feedback!

pi0 (approved):
Thanks!

PR #2681, opened 2024-08-22
fix(scanner)!: limit patterns scanned from `modules/` dir

### LINKED ISSUE #1947: Error: [worker reload] Cannot read properties of null (reading 'address')
Related: https://github.com/nuxt/nuxt/issues/24423

This seems a recent regression

### LINKED ISSUE #2680: Nitro `scanModules` returns every files
### Environment

nitropack-nightly@3.0.0-beta-28665895.e727afda-nitro-nightly-mirror

### Reproduction

1. Create the dir `modules`
2. Add the file `log/index.ts` and `log/runtime/routes/log.ts`
3. Inspect the return of https://github.com/unjs/nitro/blob/d20ffcbd16fc4003b774445e1a01e698c2bb078a/src/core/scan.ts#L44

### Describe the bug

You'll see that the return will have both the `index.ts` file which define the module and the `log.ts` file which define an event handler. This makes the local modules system unusable.

Like Nuxt, this could be nice to only scan two patters (https://nuxt.com/docs/guide/directory-structure/modules#:~:text=The%20auto%2Dregistered,modules/*.ts)

> The auto-registered files patterns are:
> - modules/*/index.ts
> - modules/*.ts

### Additional context

_No response_

### Logs

_No response_

Barbapapazes:
Hello 👋,

fix #2680 in order to limit files scanned as modules.


pi0:
Can you please rebase against v2? 🙏🏼

Barbapapazes:
> Can you please rebase against v2? 🙏🏼 (also no new fixture is required)

Yes!

pi0 (approved):
Thanks! rebased (back) to the main (v3) again since it is technically behavior breaking change.

PR #2692, opened 2024-08-26
feat: add liara provider

A-Najmabadi:
### ❓ Type of change

### 📚 Description

I've added Liara Provider docs to nitro, thanks for reading.

### 📝 Checklist

### DIFF
--- a/docs/2.deploy/20.providers/liara.md
+++ b/docs/2.deploy/20.providers/liara.md
@@ -1,0 +1,41 @@
+# Liara
+
+> Deploy Nitro apps to Liara.
+
+**Preset:** `liara`
+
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the PR.

1. It looks like an amazing service 🚀 
2. We need to add a dedicated provider for deployment providers in addition to docs
3. Is this service internationally available? It seems the signup process requires an Iranian SSN. In this case, I'm afraid that it might probably not be best to list our current deployment providers. If the service became internationally available we might reconsider. I am sorry.

A-Najmabadi:
Unfortunately it is not internationally available, I'd make a PR when we could achieve that! 
Thanks for reading this.

PR #2700, opened 2024-09-01
refactor: improve type definitions for CacheOptions

Rewwoken:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

Hi!

This PR adds argument type inheritance from `defineCachedFunction` and `defineCachedEventHandler` to the arguments of functions in the `CacheOptions` interface.

I think this will improve DX, because:
1.  No need to explicitly type arguments for every `CacheOptions` object.
2. TypeScript will show an error in case of type incompatability

#### Example
We have some function we want to cache:
```ts
// some service.ts
import type { FooObj } from '...';

export function foo(a: number, obj: FooObj) {
  // ...
}
```
Before:
```ts
// some cache.ts
import { foo } from '...';
import type { FooObj } from '...';

export const cachedFoo = cachedFunction(
  foo,
  {
    // Arguments are type of `any`, so need to import `FooObj` and hardcode first argument as `number`
    getKey: (a: number, obj: FooObj) => generateFooCacheKey(obj),
    maxAge: 1000,
  },
);
```
After:
```ts
// some cache.ts
import { foo } from '...';

export const cachedFoo = cachedFunction(
  foo,
  {
    // Arguments type is inherited from foo, no need to import `FooObj` or hardcode
    getKey: (a, obj) => generateFooCacheKey(obj),
    maxAge: 1000,
  },
);
```
I don't think docs are needed to updated, since explicitly typing is still a valid case.

**TypeScript _WILL_ complain if arguments are improperly typed explicitly. <ins>I’m curious if this counts as a breaking change.</ins>**
### 📝 Checklist


src/types/runtime/cache.ts
```diff
@@ -32,10 +32,10 @@ export interface ResponseCacheEntry<T = any> {
 }
 
 export interface CachedEventHandlerOptions<T = any>
-  extends Omit<CacheOptions<ResponseCacheEntry<T>>, "transform" | "validate"> {
-  shouldInvalidateCache?: (event: H3Event) => boolean | Promise<boolean>;
-  shouldBypassCache?: (event: H3Event) => boolean | Promise<boolean>;
-  getKey?: (event: H3Event) => string | Promise<string>;
```
pi0, line 38: in case of not explicitly typing `event` arg, this will cause to event be `any`

pi0 (approved):
Looks nice thanks ❤️

PR #2626, opened 2024-07-19
fix: fix types for `cloudflare:queue` args

dalbodeule:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

The queue event(MessageBatch) is wrong types.
What was originally supposed to be MessageBatch<unknown>, I looked it up today and found that it was written as CloudflareEmailContext.
I just fixed it.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

https://github.com/unjs/nitro/issues/2587#issuecomment-2240766921

PR #2487, opened 2024-06-02
feat(cloudflare-module): support `cloudflare:email` and `cloudflare:queue` hooks

dalbodeule:
### 🔗 Linked issue

[https://github.com/unjs/nitro/issues/2480](https://github.com/unjs/nitro/issues/2480)

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

The existing Cloudflare-module preset did not have a function that could handle Email.

To solve this, we worked on email handlers.
Similar work was performed with the existing task handler. I'm not sure if it's right to do this.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #2480, opened 2024-05-30
Request to add cloudflare worker email handler


dalbodeule:
### Describe the feature

The cloudflare worker has a feature that allows email handling by declaring the `email` function [added as beta](https://developers.cloudflare.com/email-routing/email-workers/).
I would like to use this feature. However, nitro does not officially support this feature.

Therefore, I hope you will consider using this feature.
In addition, I hope that the queue consumer is also well created.

### Additional information

- [ ] Would you be willing to help implement this feature?

dalbodeule:
I implemented the Email function in [my fork's emailTask ​​branck](https://github.com/dalbodeule/nitro/tree/emailtask). Is it okay to implement it this way?

pi0:
I think we can simply expose a hook from cloudflare one and allow integrating it since this feature is mostly vendor specific (cloudflare preset calls `cloudflare:email(event, context)` hook)

dalbodeule:
`cloudflare:email` and `cloudflare:queue` handler is mareged. close this issue!

dalbodeule:
<img width="783" alt="image" src="https://github.com/unjs/nitro/assets/11470513/4410da78-ccaa-43d7-a275-1c98e14432aa">

after [the chore](https://github.com/unjs/nitro/blob/63526f24b03e5e5de65934c697c92dd00f1e1ab1/src/presets/cloudflare/types.ts#L88), "cloudflare:email" and "cloudflare:queue" hook is not detected properly. Could you possibly know a solution to this?

dalbodeule:
<img width="1236" alt="image" src="https://github.com/unjs/nitro/assets/11470513/51662925-1d7a-4843-a3e8-f125ac1ace40">

I placed the plugin in the exact location just in case, but it's still the same. It seems that TypeScript does not recognize the existence of the "cloudflare:email" and "cloudflare:queue" hooks.

dalbodeule:
issue moved [here](https://github.com/unjs/nitro/issues/2587)

zou-yu:
> `cloudflare:email` and `cloudflare:queue` handler is mareged. close this issue!

@dalbodeule Good job! Thank you for your contribution!! I really need this feature (email handler)

PR #2711, opened 2024-09-05
fix(cloudflare-module): use correct types for email and queue events/hooks

zsilbi:
### 🔗 Linked issue

Original issue: https://github.com/unjs/nitro/issues/2480

### ❓ Type of change

### 📚 Description

https://github.com/unjs/nitro/pull/2487 introduced custom types for these events and hooks that contain typos. (for example: `adk()` instead of `ack()`)

There is an open PR: https://github.com/unjs/nitro/pull/2626 but it's still not correct.

There are native types provided by `@cloudflare/workers-types` that we can use for these. 
I updated the exported handler and the hooks accordingly.

### 📝 Checklist


pi0 (approved):
Thanks ❤️

PR #2713, opened 2024-09-05
feat(dev-server): redirect requsts without base url

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/cli/issues/469

### ❓ Type of change

### 📚 Description

This is a quality-of-life improvement rather than a bugfix per se.

If a `baseURL` is provided, this means that we can redirect user to it if they access a different URL.

### 📝 Checklist

### DIFF
--- a/src/core/dev-server/server.ts
+++ b/src/core/dev-server/server.ts
@@ -233,8 +233,12 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
   app.use(
     eventHandler(async (event) => {
       await reloadPromise;
-      if (nitro.options.baseURL && nitro.options.baseURL !== '/' && !event.path.startsWith(nitro.options.baseURL)) {
-        return sendRedirect(event, nitro.options.baseURL, 307)
+      if (
+        nitro.options.baseURL &&
+        nitro.options.baseURL !== "/" &&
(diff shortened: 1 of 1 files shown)


src/core/dev-server/server.ts
```diff
@@ -232,6 +233,13 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
   app.use(
     eventHandler(async (event) => {
       await reloadPromise;
+      if (
+        nitro.options.baseURL &&
+        nitro.options.baseURL !== "/" &&
+        !event.path.startsWith(nitro.options.baseURL)
+      ) {
+        return sendRedirect(event, nitro.options.baseURL, 307);
```
pi0, line 241: What if we do redirect with baseURL+path so `/foo` will become `/bar/foo`

src/core/dev-server/server.ts
```diff
@@ -232,6 +233,13 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
   app.use(
     eventHandler(async (event) => {
       await reloadPromise;
+      if (
+        nitro.options.baseURL &&
+        nitro.options.baseURL !== "/" &&
```
pi0, line 238: `?.length > 1` can merge both conditions

PR #2733, opened 2024-09-12
docs(routing): add error handling section

horvbalint:
### ❓ Type of change

### 📚 Description

While migrating some servers to Nitro (in Nuxt), I came across an odd behaviour that took some time to unveil. For some of the routes errors were handled correctly, while for others the client could not determine the error type. With some trial and error I figured out, that there is a difference between how an error is sent back to the client depening on which route directory the route is placed.

The `routes/` dir uses a Content-Type of `text/html`, while the `api/` dir `application/json` by default. In this PR I tried to clarify this in the docs, wich previously stated that the *only* difference is the `/api/` prefix.

As a side note I would be very interested, in the reasoning behind this behaviour. I feel like it is quite footgun-y, while I don't see the benefits.

### 📝 Checklist

### DIFF
--- a/docs/1.guide/0.index.md
+++ b/docs/1.guide/0.index.md
@@ -72,7 +72,7 @@ The `routes/` directory contains your application handlers. You can create subdi
 
 ### `api/`
 
-The `api/` directory is similar to `routes/` with the main difference that routes inside it will be prefixed with `/api/` for convenience.
+The `api/` directory is similar to `routes/` with the only difference that routes inside it will be prefixed with `/api/` for convenience.
 
 :read-more{to="/guide/routing"}
 
--- a/docs/1.guide/0.index.md
+++ b/docs/1.guide/0.index.md
@@ -72,7 +72,7 @@ The `routes/` directory contains your application handlers. You can create subdi
 
 ### `api/`
 
-The `api/` directory is similar to `routes/` with only difference that routes inside it will be prefixed with `/api/` for convenience.
+The `api/` directory is similar to `routes/` with the main difference that routes inside it will be prefixed with `/api/` for convenience.
 
 :read-more{to="/guide/routing"}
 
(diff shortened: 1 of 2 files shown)


docs/1.guide/2.routing.md
```diff
@@ -50,6 +50,11 @@ Some providers like Vercel use a top-level `api/` directory as a feature, theref
 You will have to use `routes/api/`.
 ::
 
+#### Differences between the two directories
+The `api/` directory is mainly for convenience, routes inside it will be prefixed with `/api/`.
+
+A notable difference between the two directories is how errors are handled by default: in `routes/` they are sent back with with a `Content-Type` of `text/html`, while in `api/` with `application/json`. This behaviour is overridden by some request properties (eg.: `Accept` or `User-Agent` headers).
```
pi0, line 56: It is not difference between directories. any route that starts with `/api` uses JSON content type in errors (both `api/foo.ts` and `server/api/foo.ts` are same behavior)

docs/1.guide/0.index.md
```diff
@@ -72,7 +72,7 @@ The `routes/` directory contains your application handlers. You can create subdi
 
 ### `api/`
 
-The `api/` directory is similar to `routes/` with only difference that routes inside it will be prefixed with `/api/` for convenience.
+The `api/` directory is similar to `routes/` with the main difference that routes inside it will be prefixed with `/api/` for convenience.
```
pi0, line 75: `api/*` is a shortcut to `routes/api/*` (there is no other difference)

horvbalint:
@pi0 Thanks for the quick response! I went to sleep sorry 😃

I tried to fix the docs according your comments. I found no appropriate place for this note in the current docs, so I created an 'Error handling' section in the routing guide (similar how it is in the H3 docs). Hopefully this is ok :)

Can you tell me the reasoning behind this route path based content type behaviour?

pi0 (approved):

pi0:
Thanks again for the PR and sorry for my late reply.

The reason is that when using tools like `curl` or HTTP clients like `fetch` (instead of directly) making request, it is easier to inspect a json object instead.

For `/api` routes, and with higher order frameworks like Nuxt, it might make sense that we use HTML in browser at least. i belive this something doable in Nuxt (since it has it's own copy version of this)

PR #2742, opened 2024-09-20
feat: upgrade to jiti v2

pi0:
Dependencies:

- [x] c12 (beta)
- [ ] listhen
- [ ] untyped


pi0:
(merging to test via nightly. if there are issues with hosting of 2 other deps, this way we can also see them)

Issue #2762, opened 2024-09-29
preset and docs for coolify


pi0:
https://github.com/coollabsio/coolify

/cc @justserdar if you want to take on this :)

sandros94:
Here are some of my notes on the topic:
Coolify, by default, uses [Nixpacks](https://nixpacks.com/docs/getting-started) under the hood to deploy a project.

As you can see in the example, Nixpacks uses the `build` and `start` scripts inside `package.json`. One drawback is that the final Docker image, by default built at with each push, will include all the source code and `node_modules`,  effectively making it way larger than required.

Nuxt's new `experimental.buildCache` might be easier to handle via Nixpacks compared to Dockerfiles. As the first natively supports `node_modules` caching between each build stage, but does require a custom `nixpacks.toml` (although I haven't succeeded, yet).

I started building my own Dockerfiles (available in [this gist](https://gist.github.com/sandros94/03675514546f17af1fd6db3863c043b4)) when deploying to production-like envs. On average this results in my projects going from >600MB to 150-190MB per image, per build, (with 46MB being the base image, thus shared).

To conclude:
- With the current Coolify version, we are more talking about a "Nixpacks Preset" rather than a Coolify one.
- If someone would like to focus on image size, a custom approach is currently a better choice.

P.S.: The shared Gist has multiple Dockerfiles, one solely based on `bun`, one on `pnpm+bun`, while the last one is based on `pnpm+node`. Based on the project I usually switch between one or the other based on build/runtime stability or req/s performance.

pi0:
Thanks for notes dear @sandros94 . (i suggest to read my new [writeup](https://github.com/nitrojsdev/community/discussions/7)). 

Docker support (and Nixpack/Buildpack) are also long time in track and indeed it is the main pain point that they depend on packing root instead of output dir.

Also are should **always** have dedicate presets for each deployment method.

We should only bundle `.output` dir in output in docker images. By all these, i think our first goal should be see how we can have a coolify setup that achieves this.

justserdar:
Happy to pick this up @pi0 and thank you @sandros94 for sharing your notes.

Incorporating the Coolify preset into Nitro will help ensure our efforts align with the existing standards. Currently, deploying a Nitro or Nuxt app would fail due to the absence of a start command (which is empty by default through Nixpacks and not present in default Nuxt/Nitro CLI output). 

[...]

PR #2767, opened 2024-10-01
feat: add coolify preset

justserdar:
A Nitro Coolify preset to standardise and improve security when deploying to Coolify.
Attached issue [2762](https://github.com/unjs/nitro/issues/2762).

### ❓ Type of change

Docs and Preset for Coolify.

- [ x ] ✨ New feature (a non-breaking change that adds functionality)
- [ x ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)

### 📚 Description

As discussed earlier over multiple channels with pi0 this is the right step moving forward.
Requesting feedback, the preset now correctly adds the start script to the output `package.json`.

### 📝 Checklist


sandros94:
Currently I wouldn't call this a `Coolify` preset but rather a `Nixpacks` preset, since it has more to do with that compared to Coolify. 

In particular, because if someone during a Coolify deployment picks a different `Build Pack` than this preset will be just useless or might become counterproductive in future versions.
image

To add onto this, since there will be a dedicated Nixpacks preset, we could create a dedicated [`nixpacks.toml' or `nixpacks.json` configuration file](https://nixpacks.com/docs/configuration/file) solely for proper Nitro/Nuxt deployments. Coping into the final image only the required folders and files (eg: `.output`).

While this is still not as optimized Dockerfile+Alpine images, it is still an improvement. With a downside that I still have no idea how to handle additional folders like `content` from `@nuxt/content`.

pi0:
Thanks for your pull-request dear @justserdar and feedback @sandros94.

Few notes:

0. Please read [nitro core principles](https://github.com/nitrojsdev/community/discussions/7) in detail i have tried to explain.
1. #2762 is requires **nitro** + **"coolify"** recommended deployment method regardless of details. nixpacks/decker could be another preset we extend from but it is a detail.
2. We still have to find a way to deploy **build output** directory. Not project root.

docs/2.deploy/20.providers/coolify.md
```diff
@@ -0,0 +1,23 @@
...
+
+:read-more{title="coolify.io" to="https://coolify.io"}
+
+::note
+This is the recommended preset for Coolify deployments.
+::
+
+::note
+Integration with this provider is possible with [zero configuration](/deploy#zero-config-providers).
+::
```
pi0, line 15: [zero config](https://nitro.unjs.io/deploy#zero-config-providers) is explained in docs. It is for providers that nitro can **auto-detect their CI/CD** (with an environment variable), coolify is not this way.

playground/nitro.config.ts
```diff
@@ -2,4 +2,5 @@ import { defineNitroConfig } from "nitropack/config";
 
 export default defineNitroConfig({
   compatibilityDate: "2024-09-29",
+  preset: 'coolify',
```
pi0, line 5: Please do not modify playground. You can build with `nitro build --preset coolify` to test.

docs/2.deploy/20.providers/coolify.md
```diff
@@ -0,0 +1,23 @@
...
+
+Coolify supports deploying both static and server-side rendered apps with zero configuration.
+
+## Starter templates
+
+If you want to quckly get started with coolify and nitro you can use the public nitro example in the [coolify examples repository](https://github.com/coollabsio/coolify-examples) to get started with from the Coolify dashboard.
```
pi0, line 21: I have explained this in the [Deployment via presets only](https://github.com/nitrojsdev/community/discussions/7) section. Boilerplates are not acceptable because they cause vendor-lockin. We should instead explain all steps here in the docs

playground/nitro.config.ts
justserdar: Removed the modification to the playground, thank you for the advice!

docs/2.deploy/20.providers/coolify.md
justserdar: I removed the zero config provider lines accordingly.

docs/2.deploy/20.providers/coolify.md
justserdar: I replaced this with a `Getting started` section that takes you through the process step by step now.

justserdar:
As we take this step by step, I feel the PR is currently at the point of:
**Preset:**
- [x] Append 'start' script.
- [ ] Solution for handling the **build output** directory.
- [ ] Making Coolify a provider by adding the neccesary deploy code.

**Docs:**
- [x] Steps to get started in Coolify using Nixpacks (server)
- [ ] Steps to get started in Coolify using Nixpacks (static)
- [ ] Steps to get started in Coolify using Docker (server/static)

docs/2.deploy/20.providers/coolify.md
```diff
@@ -0,0 +1,38 @@
...
+
+::note
+This is the recommended preset for Coolify deployments.
+::
+
+Coolify supports deploying both static and server-side rendered apps with zero configuration.
+
+## Set up your web app
+
+In your project, set Nitro preset to `coolify`.
```
pi0, line 17: Instead, we should instruct the build commands to use `SERVER_PRESET=coolify` environment variable.

docs/2.deploy/20.providers/coolify.md
```diff
@@ -0,0 +1,38 @@
...
+
+```js
+export default {
+  nitro: {
```
pi0, line 21: This config works in nuxt only :)

PR #2769, opened 2024-10-02
fix(azure-functions): follow symlinks in zip bundle

### LINKED ISSUE #2114: Azure functions not working
See https://github.com/unjs/nitro-deploys/issues/48 and https://github.com/unjs/nitro/pull/2092 for some context.

Unfrounatly i don't have any more information what exactly is the issue nor had a chance to investigate. So here is a placeholder.🤷‍♂️

---

(ref) all magic we have for azure support:

azure swa:
- https://github.com/unjs/nitro/blob/main/src/runtime/entries/azure.ts
- https://github.com/unjs/nitro/blob/main/src/presets/azure.ts

azure fns:
-  https://github.com/unjs/nitro/blob/main/src/runtime/entries/azure-functions.ts
- https://github.com/unjs/nitro/blob/main/src/presets/azure-functions.ts

itpropro:
The archiver instance was not configured to follow symlinks.
Fixes #2114

### DIFF
--- a/src/presets/azure/utils.ts
+++ b/src/presets/azure/utils.ts
@@ -210,7 +210,7 @@ function _zipDirectory(dir: string, outfile: string): Promise<undefined> {
 
   return new Promise((resolve, reject) => {
     archive
-      .directory(dir, false)
+      .glob("**/*", { cwd: dir, nodir: true, dot: true, follow: true })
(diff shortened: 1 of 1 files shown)


pi0:
Thanks. rebased to v2 / active branch.

pi0 (approved):

PR #2780, opened 2024-10-07
feat(vercel): allow fine-grained isr config via route rules

pi0:
- rework of #2375
- related to #1880

This PR adds the ability to define `isr` route rule as an object:

```ts
{
  expiration?: number | false;
  group?: number;
  fallback?: string;
  allowQuery?: string[];
  passQuery?: boolean;
}
```

**Example:**

```ts
  routeRules: {
    "/products/**": { isr: { allowQuery: ["q"] } },
  },
});
```

These route rules will be used to generate [vercel prerender config](https://vercel.com/docs/build-output-api/v3/primitives#prerender-configuration-file)

```json
{
  "expiration": false,
  "allowQuery": [
    "q",
    "url"
  ]
}
```


pi0:
Merging to try on nightly.

PR #2812, opened 2024-10-24
fix: `vercel_edge` enable code splitting by default

oscartbeaumont:
### ❓ Type of change

### 📚 Description

Previously when building using the `vercel_edge` preset the entire function code would be exported as a single JS file at `.vercel/functions/__nitro.func/index.mjs`. This means all dyanmic `import`'s are flattened to be inline.

This change copies the rollup configuration from the Cloudflare preset to the Vercel edge preset so this is no longer the case and dynamic imports actually lazily import code.

This changes allows me to lazy load less important code instead of requiring it all to be loaded and parsed at server startup.

This change aligns the `vercel_edge` preset with Nitro's existing Cloudflare presets which seems logical because Vercel's Edge functions run on Cloudflare Workers.

### 📝 Checklist

### DIFF
--- a/src/presets/vercel/preset.ts
+++ b/src/presets/vercel/preset.ts
@@ -56,7 +56,9 @@ const vercelEdge = defineNitroPreset(
     },
     rollupConfig: {
       output: {
-        format: "module",
+        entryFileNames: "index.js",
+        format: "esm",
+        inlineDynamicImports: false,
       },
     },
     unenv: {


pi0:
First thanks for PR.

Vercel edge, is not based on cloudflare CDN.

I have tested your PR against our e2e deployments (commit: https://github.com/unjs/nitro-deploys/commit/fb1f640e1f04edbeca131f9f81a339f4e098c971)

It seems they don't support dynamic imports (at least via this change alone)

image

I will try to reach out to them for more info (feel free to open a tracker issue) but in the meantime it is not possible at is seems.

pi0:
Almost sure they don't host it on top of cloudflare workers with all defferences. Not sure if they built it on top of opensource [workerd](https://github.com/cloudflare/workerd) or not ([docs](https://vercel.com/docs/functions/runtimes/edge-runtime) and [docs](https://vercel.com/docs/edge-network/overview) don't mention it at all and they only provide [edge-runtime](https://edge-runtime.vercel.app/) -- a Node.js based emulation layer for development time, if it was workerd based there was no reason to built such thing)

---

On cloudflare/workerd side, indeed. they are working on a new "module registry" system to allow lazy chunk imports and also unblock Nitro to avoid second build step.

PR #2814, opened 2024-10-25
fix: apply headers from `routeRules` in dev server

### LINKED ISSUE #2749: `routeRules` does not work in DEV mode for static files
### Environment

Node Version: v18.20.3
Nitro Version: 2.9.7

### Reproduction

In [this example](https://stackblitz.com/edit/nitro-route-rules-dev?file=nitro.config.ts) `/hello.txt` route should return `Content-type: application/json` header, and it does work in the PROD mode, but doesn't work in the DEV mode, where `Content-type: text/plain; charset=UTF-8` is returned.

Try to access `/hello.txt` route and check the headers of the response:
1. In `DEV` mode.
2. In `PROD` mode.

#### public/hello.txt

```
World
```

#### nitro.config.ts

```ts
// https://nitro.unjs.io/config
  srcDir: 'server',
  routeRules: {
    '/hello.txt': {
      headers: {
        'Content-Type': 'application/json',
      },
    },
  },
});
```

### Describe the bug

When `nitro.config.ts` contains a `routeRules` section that sets headers for a static file (from `public` dir), then these headers are not set when the server is running in DEV mode (`npm run dev`). The headers are set, though, when the server is running in PROD mode (`npm run build && npm run preview`).

### Additional context

The bug was originally created as nuxt/nuxt#26740.

### Logs

_No response_

peterthenelson:
### 🔗 Linked issue

#2749

### ❓ Type of change

- [x ] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

As detailed in the linked bug, `routeRules` fail to apply to public assets when using the dev server. The reason this happens is, IIUC, that the handler for `routeRules` runs in the reloadable worker (like the majority of the server logic), while the public asset serving happens locally in the app created in `dev-server/server.ts`. Requests for public assets get served without ever proxying anything to the worker.

I solve this by adding a handler to the dev server itself that partially duplicates some of the logic in `route-rules.ts`. Some notes:
- This only applies the headers from `routeRules`. The actual handler in the worker also allows for redirects and proxying. It's not clear to me whether it makes any sense to apply those to public assets, and in the case of proxying, it requires the "localFetch" object.
- This redundantly sets the headers in cases where the dev server was already setting them (i.e., the non public assets). I considered a couple alternatives, but they would either require duplicating or inverting the URL-path-to-file-path logic from the "send" package (which is what "serve-static" uses).

Anyway, it's totally possible I'm going about this the wrong way, but that's my reasoning for why I did it this way.

Resolves #2749

### 📝 Checklist

- [x ] I have linked an issue or discussion.

### DIFF
--- a/src/core/dev-server/server.ts
+++ b/src/core/dev-server/server.ts
@@ -8,6 +8,7 @@ import { type FSWatcher, watch } from "chokidar";
 import {
   type H3Error,
   type H3Event,
+  appendResponseHeaders,
   createApp,
   createError,
   eventHandler,
@@ -22,15 +23,18 @@ import type {
   Nitro,
   NitroBuildInfo,
   NitroDevServer,
+  NitroRouteRules,
   NitroWorker,
 } from "nitropack/types";
 import { resolve } from "pathe";
 import { debounce } from "perfect-debounce";
 import { servePlaceholder } from "serve-placeholder";
 import serveStatic from "serve-static";
-import { joinURL } from "ufo";
+import { joinURL, withoutBase } from "ufo";
 import defaultErrorHandler from "./error";
 import { createVFSHandler } from "./vfs";
+import { createRouter as createRadixRouter, toRouteMatcher } from "radix3";
+import defu from "defu";
 
 function initWorker(filename: string): Promise<NitroWorker> | undefined {
   if (!existsSync(filename)) {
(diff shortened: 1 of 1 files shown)


peterthenelson:
@pi0 let me know if there's anything else I should do before you review this. Also, thanks for all your work on the many unjs projects; I've been finding them really useful :)

src/core/dev-server/server.ts
```diff
@@ -166,6 +170,26 @@ export function createDevServer(nitro: Nitro): NitroDevServer {
   // Debugging endpoint to view vfs
   app.use("/_vfs", createVFSHandler(nitro));
 
+  // Apply headers based on routeRules. In cases where the dev server is being
+  // used, these would otherwise fail to be applied to public assets.
+  const routeRulesMatcher = toRouteMatcher(
+    createRadixRouter({ routes: nitro.options.routeRules })
+  );
+  app.use(
+    "/",
```
pi0, line 179: I think we could add middleware only to static asset paths (and ideally only when matching them)

peterthenelson:
I had a followup question (see thread) before I make the requested changes.

src/core/dev-server/server.ts
peterthenelson: I revisited this, and I was indeed missing an easier approach. The serveStatic setHeaders callback indeed gets passed the file path, but the response object can get at the original url path, so no need to do the rube-goldberg nonsense I was worrying about.

pi0:
Added tests, two issues:

- `path` passed inside middleware omits base `/build/test.txt` gets `/test.txt` therefore rules wasn't matching.
- Currently, nitro has an implicit behavior that ignores `maxAge` of public assets (which makes a route rule) in development mode which means this patch can cause behavior change (long term asset caches in dev). I am thinking for Nitro v2, we should avoid setting `cache-control` from route rules to avoid it)

prettypet2137:
@pi0 
I'm using Nuxt 3.12.4 with Nitro 2.9.7 and I have also same issue as above.
Can you please check this PR and merge?
Thank you.

dataexcess:
is there an ETA on when this fix will be shipped? I would prefer not have to pin my nitro version to this branch. Plus Im using nuxt, so is that even possible? (I'm a noob)

pi0:
Thanks for PR dear @peterthenelson and sorry left unattended. Moving to #3926

PR #2818, opened 2024-10-29
chore(types): annotate `h3Exports` types

iiio2:
### ❓ Type of change

### 📚 Description

`resolveModuleExportNames` function returns `Promise<string[]>`. Here we have called it but `h3Exports` is returning `any`. I have explicitly annotated `string[]` for `h3Exports`. Also below we are filtering `h3Exports`. I think `filter` can be used only with arrays. That is why I think we may explicitly annotate `h3Exports` here.  

### 📝 Checklist

### DIFF
--- a/src/core/config/resolvers/imports.ts
+++ b/src/core/config/resolvers/imports.ts
@@ -15,7 +15,7 @@ export async function resolveImportsOptions(options: NitroOptions) {
   options.imports.presets.push(...getNitroImportsPreset());
 
   // Add h3 auto imports preset
-  const h3Exports = await resolveModuleExportNames("h3", {
+  const h3Exports: string[] = await resolveModuleExportNames("h3", {
     url: import.meta.url,
   });
   options.imports.presets ??= [];


src/core/config/resolvers/imports.ts
```diff
@@ -15,7 +15,7 @@ export async function resolveImportsOptions(options: NitroOptions) {
   options.imports.presets.push(...getNitroImportsPreset());
 
   // Add h3 auto imports preset
-  const h3Exports = await resolveModuleExportNames("h3", {
+  const h3Exports: string[] = await resolveModuleExportNames("h3", {
```
pi0, line 18: Type is already inferred: 

image

pi0:
Thanks for PR, but in this case, simply relying on inference should be good enough internally.

src/core/config/resolvers/imports.ts
iiio2: oh. i see. thanks.

iiio2:
Ok @pi0 . Thank you so much.

PR #2464, opened 2024-05-23
fix(vercel): support custom `baseURL`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26886
resolves https://github.com/nuxt/nuxt/issues/21780
resolves https://github.com/unjs/nitro/issues/1348

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

It's possible to deploy sites to Vercel with a custom `baseURL` and the renderer will correctly handle 404 on `/` and correct responses from `/<baseURL>/` but the public assets aren't configured to respect the `baseURL`.

Published minimal reproduction & patch to https://github.com/danielroe/nuxt-3-vercel-baseurl-test - you can see deployed at https://nuxt-3-vercel-baseurl-test-tan.vercel.app/subdir.

You can confirm that correct headers are added to a request like:

```sh
http --headers https://nuxt-3-vercel-baseurl-test-tan.vercel.app/subdir/_nuxt/C7CLbNSw.js
> HTTP/1.1 200 OK
> Access-Control-Allow-Origin: *
> Age: 88
> Cache-Control: public,max-age=31536000,immutable
> Connection: keep-alive
> Content-Disposition: inline; filename="C7CLbNSw.js"
> Content-Encoding: gzip
> Content-Type: application/javascript; charset=utf-8
> Date: Thu, 23 May 2024 16:37:55 GMT
> Etag: W/"c44aff3c70a5f49fad126ee0c615c297"
```

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2821, opened 2024-10-29
fix(cloudflare): support custom `baseURL`

atinux:
### 🔗 Linked issue

Similar to #2464 

Resolves https://github.com/nuxt-hub/core/issues/356

### ❓ Type of change

### 📚 Description

Add support of `baseURL` to the Cloudflare preset.

- Without the fix (current version): https://with-base-url.nuxt.dev/base/ (see the error in the browser console)
- With current fix: https://e162454b.with-base-url.pages.dev/base/

I also tested the result without the baseURL and it works like before.


src/presets/cloudflare/utils.ts
```diff
@@ -118,7 +118,7 @@ async function writeCFPagesHeaders(nitro: Nitro) {
     ([_, routeRules]) => routeRules.headers
   )) {
     const headers = [
-      path.replace("/**", "/*"),
+      path.replace("/**", "/*").replace(/^\//, nitro.options.baseURL || "/"),
```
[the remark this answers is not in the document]
atinux: Fixed, I directly used `withBase` as it can also support if the base is already there (useful for redirects).

Should be ready now

src/presets/cloudflare/utils.ts
```diff
@@ -118,7 +119,7 @@ async function writeCFPagesHeaders(nitro: Nitro) {
     ([_, routeRules]) => routeRules.headers
   )) {
     const headers = [
-      path.replace("/**", "/*"),
+      withBase(path.replace("/**", "/*"), nitro.options.baseURL),
```
pi0, line 122: `withBase` might cause one implicit issue:

Imagine if baseURL is `/test` and we have a route roule also for `/test/**` (So full URL would be `/test/test/**`) but in this case, we apply it to `/test/**` -- effectively `/**` without base)

atinux: Yes but we can have another issue if the user set the baseURL to `/base/` and in the routeRules:

```typescript
baseURL: '/base/',
routeRules: {
  '/base/foo': { redirect: '/base/bar' }
}
```

But I guess we should advice to not include the base URL in the route rules so the user should do:

```typescript
baseURL: '/base/',
routeRules: {
  '/foo': { redirect: '/bar' }
}
```

atinux, line 122:
```suggestion
      joinURL(nitro.options.baseURL, path.replace("/**", "/*")),
```

src/presets/cloudflare/utils.ts
```diff
@@ -161,7 +162,7 @@ async function writeCFPagesRedirects(nitro: Nitro) {
   )) {
     const code = routeRules.redirect!.statusCode;
     contents.unshift(
-      `${key.replace("/**", "/*")}\t${routeRules.redirect!.to}\t${code}`
+      `${withBase(key.replace("/**", "/*"), nitro.options.baseURL)}\t${withBase(routeRules.redirect!.to, nitro.options.baseURL)}\t${code}`
```
atinux, line 165: Here this is more complex as redirect might be an absolute URL so with base works best in this scenario compared to joinURL...

src/presets/cloudflare/utils.ts
pi0: Yes indeed. baseURL is a dynamic config prefixing everything, including rules.

src/presets/cloudflare/utils.ts
pi0: I think it is same to advice: rules (including redirects) should be unware of base and we add base regardless (otherwise we cannot allow first segment of any URL to be same as dynamic base URL)

pi0: cff5807527ee24b6759be9a64cbd49eeb5b14cb5

PR #2822, opened 2024-10-30
perf(netlify, netlify-edge): exclude static paths from server handler

serhalp:
### 🔗 Linked issue

N/A - discussed on Discord

### ❓ Type of change

### 📚 Description

The netlify preset was already using `preferStatic: true` which lets existing static assets take precedence over the function, but this now also avoids a function invocation for static paths that don't exist, i.e. this avoids an unnecessary dynamic 404 that could be served directly from the CDN.

The `netlify-edge` preset wasn't excluding anything, so this addresses both the 404 case and the existing asset case.

The 404 case is important because browsers frequently attempt to request hashed assets from previous deploys that have been invalidated. There's no reason for this to go through functions, since we know that the whole parent path is static.

This PR also adds tests for the `netlify-edge` preset, which were missing. They're mostly copied from the `netlify` preset tests.

### 📝 Checklist


src/presets/netlify/utils.ts
```diff
@@ -92,6 +92,13 @@ export async function writeHeaders(nitro: Nitro) {
   await fsp.writeFile(headersPath, contents);
 }
 
+export function getStaticPaths(nitro: Nitro): string[] {
+  const publicAssets = nitro.options.publicAssets.filter(
+    (dir) => dir.fallthrough !== true && dir.baseURL && dir.baseURL !== "/"
```
serhalp, line 97: @pi0 is this `dir.baseUrl !== "/"` case right? I was seeing this be included when running the tests and I saw similar exclusions in some other presets. I'm not sure I understand what this is doing though.

src/presets/netlify/utils.ts
```diff
@@ -92,6 +92,13 @@ export async function writeHeaders(nitro: Nitro) {
...
+  );
+  return ["/.netlify/*", ...publicAssets.map((dir) => `${dir.baseURL}/*`)];
```
serhalp, line 99: @pi0 are these `baseURL`s guaranteed to be prefixed with a leading slash? I see some presets handling that but I'm not sure if it's necessary. (This Netlify config does require it.)

test/presets/netlify.test.ts
```diff
@@ -86,6 +118,24 @@ describe("nitro:preset:netlify", async () => {
...
+          "utf8"
+        );
+        expect(serverFunctionFile).toEqual(
+          `
+export { default } from "./main.mjs";
+export const config = {
+  name: "server handler",
+  generator: "mock-framework@1.2.3",
+  path: "/*",
+  excludedPath: ["/.netlify/*","/icons/*","/nested/fonts/*","/build/*"],
```
serhalp, line 133: @pi0 where is `/build/*` coming from here? I couldn't quite trace it. I'd like to configure it explicitly in the test setup if possible.

test/presets/netlify.test.ts
```diff
@@ -137,3 +187,139 @@ describe("nitro:preset:netlify", async () => {
...
+        const res = await handler(req, {} as EdgeFunctionContext);
+        if (!(res instanceof Response))
+          // The Netlify Edge Function handler API allows returning `undefined` but this
+          // test helper only supports a Response or this shape.
+          return {
+            data: undefined,
+            status: 404,
+            headers: {},
+          };
+        return res;
```
serhalp, line 258: @pi0 I've got one test failure from the `testNitro()` cases that I don't understand:

```
 FAIL  test/presets/netlify.test.ts > nitro:preset:netlify-edge > API Works
AssertionError: expected undefined to match object { message: 'Hello API' }

- Expected:
Object {
  "message": "Hello API",
}

+ Received:
undefined

 ❯ test/tests.ts:229:26
    227|   it("API Works", async () => {
    228|     const { data: helloData } = await callHandler({ url: "/api/hello" });
    229|     expect(helloData).to.toMatchObject({ message: "Hello API" });
       |                          ^
    230|
    231|     if (ctx.nitro?.options.serveStatic) {
```

It's hitting this case where the edge function handler is returning `undefined`. Any tips? I'm not sure what's being tested here and why it would be behaving differently for netlify-edge.

src/presets/netlify/utils.ts
pi0: It is correct. The check is to make sure we don't include root `/` patterns (that imply fallthrough behavior) to exclude. Nitro usually infers `fallthrough` properly but it might not be set correctly (ie, after creating nitro instance via wrapper frameworks)

test/presets/netlify.test.ts
pi0: `/api/hello` is expected to be prerendered (an static asset in output public dir)

test/presets/netlify.test.ts
```diff
@@ -1,14 +1,44 @@
...
+import type { Context as EdgeFunctionContext } from "@netlify/edge-functions";
 import { resolve } from "pathe";
 import { describe, expect, it } from "vitest";
 import { getPresetTmpDir, setupTest, testNitro } from "../tests";
 
 describe("nitro:preset:netlify", async () => {
+  const publicDir = resolve(getPresetTmpDir("netlify"), "dist");
   const ctx = await setupTest("netlify", {
     config: {
+      framework: {
```
pi0, line 12: I think we can directly add `framework` to the main fixture nitro.config file

test/presets/netlify.test.ts
pi0: test fixture i suppose

test/presets/netlify.test.ts
```diff
@@ -137,3 +187,139 @@ describe("nitro:preset:netlify", async () => {
...
+
+describe("nitro:preset:netlify-edge", async () => {
+  const publicDir = resolve(getPresetTmpDir("netlify-edge"), "dist");
+  const ctx = await setupTest("netlify-edge", {
+    config: {
+      framework: {
+        name: "mock-framework",
+        version: "1.2.3",
+      },
+      publicAssets: [
```
pi0, line 199: This seems like duplicate logic we can at least share between two netlify/edge tests.

Ideally, I would love to extend the main fixture with more cases of public assets so that we increase overall coverage instead of per-preset tests.

pi0:
Thanks for the PR dear @serhalp 

It seems main blockers on PR are addition of tests (which are so nice) -- although i don't want the main improvement to be missed for 2.10 release, perhaps we can extract work on tests to an additional PR? 

Main addition seems could be tested easily both manually or with a simple unit test for `getStaticPaths` utility.

test/tests.ts
```diff
@@ -223,16 +224,22 @@ export function testNitro(
     _handler = await getHandler();
   }, 25_000);
 
-  it("API Works", async () => {
-    const { data: helloData } = await callHandler({ url: "/api/hello" });
-    expect(helloData).to.toMatchObject({ message: "Hello API" });
+  // netlify-edge intentionally ignores prerendered routes
+  it.skipIf(ctx.preset === "netlify-edge")(
```
serhalp, line 228: https://github.com/serhalp/nitro/blob/ce72ea3fd1ae59cf722a9868d8e9cc99eb6b0aa1/src/presets/netlify/runtime/netlify-edge.ts#L12-L14

happy to handle this a different way in a follow-up PR if you'd like

src/presets/netlify/utils.ts
pi0: it should be normalized but we can use `joinURL` from `ufo` to make sure of it.

src/presets/netlify/utils.ts
```diff
@@ -92,6 +92,13 @@ export async function writeHeaders(nitro: Nitro) {
   await fsp.writeFile(headersPath, contents);
 }
 
+export function getStaticPaths(nitro: Nitro): string[] {
```
pi0, line 95: This utility can accept `publicAssets` directly. This way we can write simple unit tests for it without all new complexity being added (both fixture and tests and be simplified)

test/tests.ts
pi0: I would suggest to use unit test for new functionality and avoid any changes here for now.

serhalp: so you'd prefer to split out the addition of the missing tests for netlify-edge into a separate PR?

pi0: Sure

serhalp:
@pi0 I made all the requested changes.

Here's a manual QA against a nuxt site:
- site homepage is rendered via SSR: https://672396fce8c6141da780bc41--nuxt-test-static-assets-fix.netlify.app
- a nonexistent path under `/_nuxt` returns a 404 directly from the CDN without invoking a function: https://672396fce8c6141da780bc41--nuxt-test-static-assets-fix.netlify.app/_nuxt/fake-old-file-123.css
- a real path under `/_nuxt` returns a 200 directly from the CDN without invoking a function: https://672396fce8c6141da780bc41--nuxt-test-static-assets-fix.netlify.app/_nuxt/entry.CQDx9R-V.css

serhalp:
And for good measure here's one against `netlify-edge`:
- site homepage is rendered via SSR: https://67239a3eeb45272890169516--nuxt-test-static-assets-fix.netlify.app
- a nonexistent path under `/_nuxt` returns a 404 directly from the CDN without invoking a function: https://67239a3eeb45272890169516--nuxt-test-static-assets-fix.netlify.app/_nuxt/fake-old-file-123.css
- a real path under `/_nuxt` returns a 200 directly from the CDN without invoking a function: https://67239a3eeb45272890169516--nuxt-test-static-assets-fix.netlify.app/_nuxt/entry.CQDx9R-V.css

test/tests.ts
```diff
@@ -224,22 +224,16 @@ export function testNitro(
     _handler = await getHandler();
   }, 25_000);
 
-  // netlify-edge intentionally ignores prerendered routes
```
serhalp: oops, forgot to undo this one, my bad! thanks for cleaning up

pi0 (approved):

pi0:
Thanks ❤️

Issue #2819, opened 2024-10-29
Clamp to valid Vercel Node versions


cjpearson:
### Describe the feature

The Vercel preset currently sets its runtime node version based on the version of node used to build the application. https://github.com/unjs/nitro/blob/v2/src/presets/vercel/utils.ts#L19 This caused an issue when updating from Node 20 to 22 (the LTS recently changed) because Vercel [does not support](https://vercel.com/docs/functions/runtimes/node-js/node-js-versions) Node 22. 

What do you think about clamping the value to valid versions? So if you build with v22 or v23, v20 will be used as the default for the runtime?

Alternatively, maybe follow the [same rules](https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#version-overrides-in-package.json) as Vercel does with the `engine` field? I was originally a bit confused with the invalid runtime error because I assumed those rules would apply (and we have `>= 20` in our engines), but Nitro setting the runtime takes precedence. 

### Additional information

- [X] Would you be willing to help implement this feature?

pi0x:
Fixed in v3 — the Vercel runtime is now clamped to the versions Vercel actually supports and can be overridden with `vercel.functions.runtime`, see https://github.com/nitrojs/nitro/blob/main/src/presets/vercel/utils.ts.

_Triaged with AI assistance._

PR #2824, opened 2024-10-31
fix(vercel): clamp runtime config to valid node versions 

cjpearson:
### 🔗 Linked issue

#2819 

### ❓ Type of change

### 📚 Description

This change ensures that only supported node versions will be used as the default `runtime` value for the vercel preset. If a non-LTS version of node is used, it will use the previous LTS. If the version is outside the range supported by vercel, it will be clamped to the min/max of the supported range. 

The `runtime` value can still be manually overridden as needed.

### 📝 Checklist


src/presets/vercel/utils.ts
```diff
@@ -10,13 +10,31 @@ import type {
...
 export async function generateFunctionFiles(nitro: Nitro) {
   const buildConfigPath = resolve(nitro.options.output.dir, "config.json");
   const buildConfig = generateBuildConfig(nitro);
   await writeFile(buildConfigPath, JSON.stringify(buildConfig, null, 2));
 
-  const systemNodeVersion = process.versions.node.split(".")[0];
-  const runtimeVersion = `nodejs${systemNodeVersion}.x`;
+  const systemNodeVersion = getSystemNodeVersion();
+  const closestLtsVersion =
+    systemNodeVersion % 2 === 0 ? systemNodeVersion : systemNodeVersion - 1;
```
pi0, line 31: I think instead we can use an array of supported Node.js versions in vercel and clamp (up) to closest

cjpearson: Are you thinking something like this? 

```ts
  let usedNodeVersion = 0;
  for (const version of SUPPORTED_NODE_VERSIONS) {
    if (Math.abs(systemNodeVersion - version) <= Math.abs(systemNodeVersion - usedNodeVersion)) {
      usedNodeVersion = version
    }
  }
```

pi0: Something like this could be nice 👍🏼 I think if `SUPPORTED_NODE_VERSIONS` is sorted array, we can find first upper abound with simple find as well.

src/presets/vercel/utils.ts
```diff
@@ -13,15 +13,30 @@ import { isTest } from "std-env";
 
 // https://vercel.com/docs/build-output-api/configuration
 
+// https://vercel.com/docs/functions/runtimes/node-js/node-js-versions
+const SUPPORTED_NODE_VERSIONS = [16, 18, 20, 22];
```
pi0: 16 is not supported by nitro

```suggestion
const SUPPORTED_NODE_VERSIONS = [18, 20, 22];
```

src/presets/vercel/utils.ts
```diff
@@ -13,15 +13,30 @@ import { isTest } from "std-env";
...
 
+// https://vercel.com/docs/functions/runtimes/node-js/node-js-versions
+const SUPPORTED_NODE_VERSIONS = [18, 20, 22];
+
+function getSystemNodeVersion() {
+  const systemNodeVersion = Number.parseInt(
+    process.versions.node.split(".")[0]
+  );
+
+  return Number.isNaN(systemNodeVersion) ? 20 : systemNodeVersion;
```
pi0:
```suggestion
  return Number.isNaN(systemNodeVersion) ? 22 : systemNodeVersion;
```

PR #2831, opened 2024-11-02
docs(cloudflare): update `cloudflare-module`

gxres042:
### ❓ Type of change

### 📚 Description

With Cloudflare Workers Static Assets, we only need to bind `ASSETS` and don't need to configure `bucket` anymore.Also, a link at Cloudflare Service Workers part has an unnecessary `/`, making the server return 404 to visitor.

This PR resolves these problems.

### 📝 Checklist

### DIFF
--- a/docs/2.deploy/20.providers/cloudflare.md
+++ b/docs/2.deploy/20.providers/cloudflare.md
@@ -9,7 +9,7 @@
 :read-more{title="Cloudflare Pages" to="https://pages.cloudflare.com/"}
 
 ::note
-This is the recommended preset for Cloudflare deployments, please consider using the alternative ones only if you have special requirements or needs.
+This is the recommended preset for Cloudflare deployments, please consider using the alternative ones if you need special features.
 ::
 
 ::note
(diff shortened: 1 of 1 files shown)


docs/2.deploy/20.providers/cloudflare.md
```diff
@@ -85,16 +85,14 @@ The following shows a typical `wrangler.toml` file for a Nitro application:
 name = "playground"
 main = "./.output/server/index.mjs"
 workers_dev = true
-compatibility_date = "2023-12-01"
+compatibility_date = "2024-11-01"
```
pi0, line 88: I think we also need to document `compatibilityDate` has to be set in nitro/nuxt config to enable new preset (we could also use two tabs perhaps for now for legacy and new one)

gxres042:
I add a sample `nitro.config.ts` file into it. 🤔 

For `compatibilityDate` option, we should document it at https://nitro.build/config too.

pi0 (approved):
Thanks! ❤️

adamdehaven:
@pi0 took me a bit of digging when upgrading `nuxt` to `3.14.x` to realize why my app that utilizes the `cloudflare-module` preset was broken 😬 

Turns out, the change to static assets requires this change in my `wrangler.toml`:

```diff
- [site]
- bucket = ".output/public"
+ assets = { directory = "./.output/public/", binding = "ASSETS" }
```

This PR addresses the change in the docs; however, it may be worth a specific callout in the `2.10` release notes

pi0:
Yes, this is required with latest compatibility date being set (otherwise nitro uses legacy `cloudflare_worker` preset that uses old assets handling)

PR #2855, opened 2024-11-06
docs(alwaysdata): fix casting

iiio2:
`alwaysdata` starts with lowercase `a`. It is there naming convention I think 

https://www.alwaysdata.com/en/ 
https://www.linkedin.com/company/alwaysdata/


pi0:
In docs, for consistency, first case is always upper

image

PR #2877, opened 2024-11-13
Don't use options object for listening on unix socket

Tofandel:
This change is so nitro works on a passenger server with autoinstall turned on

https://www.phusionpassenger.com/library/indepth/nodejs/reverse_port_binding.html

The gist is that if we have multiple server.listen on passenger, we need to disable autoinstall and instead call `server.listen('passenger', ...)`

It is very important that the parameter is a string, it cannot be passed as an object like `{path: 'passenger'}` or `{port: 'passenger'}`, otherwise passenger doesn't pick up on it and instead the server will be registered on a unix socket `passenger` in the cwd

### DIFF
--- a/src/presets/node/runtime/node-server.ts
+++ b/src/presets/node/runtime/node-server.ts
@@ -29,7 +29,7 @@ const host = process.env.NITRO_HOST || process.env.HOST;
 const path = process.env.NITRO_UNIX_SOCKET;
 
 // @ts-ignore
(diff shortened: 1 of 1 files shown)


Tofandel:
Mm, passenger is still node so it wouldn't make much sense, it's just an extension if you will that manages the server

pi0:
I would highly appreciate if you can help on making reproduction steps for a problematic deployment first thing.

Issue #2892, opened 2024-11-19
Running node-cluster preset build throwing EADDRINUSE errors


aisteneciunaitevalantic:
### Environment

Node Version: v18.20.3
Nitro Version: 2.10.4

### Reproduction

https://stackblitz.com/edit/github-mrhh35?file=nitro.config.ts

```
> npm run build
> npm run preview
```

### Describe the bug

Running application with preset 'node-cluster' is throwing EADDRINUSE errors

### Additional context

No such issue in v2.9.7

### Logs

```sh
[nitro] [uncaughtException] Error: bind EADDRINUSE null:3000
    at __node_internal_captureLargerStackTrace2 (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:101:5335)
    at __node_internal_exceptionWithHostPort2 (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:101:7017)
    at listenOnPrimaryHandle (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:285:12098)
    at rr (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:55:2291)
    at Worker.<anonymous> (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:55:2716)
    at _0x54dcec.onInternalMessage (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:59:458)
    at _0x54dcec.emit (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:30:11015)
    at emit (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:52:6956)
    at processTicksAndRejections (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/builtins.ddb8d84d.js:192:1093)
    at https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/blitz.d3acb9e1.js:40:198427
    at _0x40724c (https://qgnirvpgrqgithub-iuqv.w-credentialless-staticblitz.com/blitz.d3acb9e1.js:40:198508)
[...]

PR #2893, opened 2024-11-19
fix(node-cluster): fix node cluster preset

OskarLebuda:
### 🔗 Linked issue
Fix https://github.com/nitrojs/nitro/issues/2892

### ❓ Type of change

### 📚 Description
Resolves [#2892](https://github.com/nitrojs/nitro/issues/2892)

This change addresses the issue where, after the `nitro build` (unbuild), the code from `node-server` ends up in the same scope as `node-cluster` (see screen bellow). In the current implementation, the primary cluster executes the server code, and subsequently, when workers attempt to execute the same code, they encounter an error because the port is already in use by the primary. By wrapping the server code into a separate, exported function, we avoid these scope conflicts and port clashes, ensuring that each worker operates independently without interference.

builded `index.mjs`:
image

### 📝 Checklist

### DIFF
--- a/src/presets/node/runtime/node-server.ts
+++ b/src/presets/node/runtime/node-server.ts
@@ -12,60 +12,65 @@ import {
   trapUnhandledNodeErrors,
 } from "nitropack/runtime/internal";
 
-const cert = process.env.NITRO_SSL_CERT;
-const key = process.env.NITRO_SSL_KEY;
+export const serverWorker = async () => {
 
-const nitroApp = useNitroApp();
-
-const server =
-  cert && key
-    ? new HttpsServer({ key, cert }, toNodeListener(nitroApp.h3App))
-    : new HttpServer(toNodeListener(nitroApp.h3App));
-
-const port = (destr(process.env.NITRO_PORT || process.env.PORT) ||
-  3000) as number;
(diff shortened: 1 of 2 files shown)


src/presets/node/runtime/node-server.ts
pi0, line 1: This entry is also used for default nitro server-preset that requires the entrypoint itself to immediately start listening and this PR breaks it!

image

I think we should find better fix or at least duplicate this entry to be specific solution for `node_cluster`.

pi0:
Thanks for PR and note about unbuild output it was important to help finding issue ❤️ 

I think I found a better alternative specific to this issue: ~> https://github.com/nitrojs/nitro/pull/2894

src/presets/node/runtime/node-server.ts
OskarLebuda: Eh.. I didn't know about that :/ Sorry, my mistake 😭

OskarLebuda:
> Thanks for PR and note about unbuild output it was important to help finding issue ❤️
> 
> I think I found a better alternative specific to this issue: ~> #2894

Thanks for the reply 😄 Yea, your solution is much better then mine

PR #2895, opened 2024-11-19
fix(firebase-app-hosting): change preset to use relative server path in runCommand

Yuangwang:
### 🔗 Linked issue

Discussed on discord, needed to support frameworks that change the location of serverDir

### ❓ Type of change

### 📚 Description
This addresses and issue where the hard coded runCommand doesn't work if the server directory location is changed

### 📝 Checklist

### DIFF
--- a/src/presets/firebase/preset.ts
+++ b/src/presets/firebase/preset.ts
@@ -88,7 +88,7 @@ const firebaseAppHosting = defineNitroPreset(
           stringifyYAML({
             version: "v1",
             runConfig: {
-              runCommand: "node .output/server/index.mjs",
+              runCommand: `node ${join(nitro.options.output.serverDir, "index.mjs")}`,
               ...(nitro.options.firebase as AppHostingOptions)?.appHosting,
             },
             metadata: {


src/presets/firebase/preset.ts
```diff
@@ -88,7 +88,7 @@ const firebaseAppHosting = defineNitroPreset(
           stringifyYAML({
             version: "v1",
             runConfig: {
-              runCommand: "node .output/server/index.mjs",
+              runCommand: `node ${join(nitro.options.output.serverDir, "index.mjs")}`,
```
pi0, line 91: Let's generate it relative to `options.rootDir` (or `options.workspaceDir` depending on current app hosting behavior) instead of absolute path.

pi0, line 91: We should generate it relative to `options.rootDir` or `options.workspaceDir`, currently it leads to an absolute path like this:

```
 runCommand: node /Users/pooya/Code/nitro/playground/.output/server/index.mjs
```

pi0: 19c674a09485cf9d8e241fd5fd78b45c4b148b7e

pi0:
Thanks.

PR #2902, opened 2024-11-25
fix(config): avoid serialization warning when primitives have prototype changed

SoaresMG:
### 🔗 Linked issue

No issue, do I need to open it?

### ❓ Type of change

### 📚 Description

Unfortunately there are a few libs that inject utils into prototypes, with the current implementation of `checkSerializableRuntimeConfig` if `String` or any other primitive has an util injected into the prototype a warning is thrown.

I don't see any direct consequence of just ignoring it during the check, if that prototype is indeed needed it will be there once `JSON.parse` is used.

An example of this is [`forEachEmbeddedItem`](https://github.com/contentstack/contentstack-utils-javascript/blob/master/src/extensions/index.ts#L7) where the ContentStack team added an util into `String.prototype`.

I don't think libs should be polluting prototypes needlessly when it could very well be an exported function, but nitro shouldn't also be concerned about it since there is no consequence at all (right?).

[Reproduction](https://stackblitz.com/edit/github-xyyb7z?file=nitro.config.ts)

### 📝 Checklist


pi0 (approved):
❤️

Issue #2072, opened 2024-01-11
Support AWS Lambda behind Load Balancer (ALB)


zoellner:
### Describe the feature

The current aws-lambda present assumes that the lambda is behind an API Gateway (tries to distinguish between v1 and v2). But there are other infrastructure configuration options and it is possible to have the lambda behind an ALB (see https://aws.amazon.com/blogs/networking-and-content-delivery/lambda-functions-as-targets-for-application-load-balancers/)

The event format is slightly different in that case. ALB can also have a multi value header option turned on or off (https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#enable-multi-value-headers) which changes the event format as well. 

Related to #505 (split presets into different one for these use cases) and #79 (lambda-edge support)

This repo does a good job at distinguishing between the different event formats and normalizing it: https://github.com/dougmoscrop/serverless-http

### Additional information

- [ ] Would you be willing to help implement this feature?

pi0:
(BTW latest Nitro v3 / srvx should support ALB requests)

PR #2914, opened 2024-12-02
feat: add `aws-lambda-alb` support

marcogrcr:
### 🔗 Linked issue

#2072

### ❓ Type of change

### 📚 Description

WHAT?

Add support for ALB lambda function targets.

WHY?

`ALBEvent` is similar to `APIGatewayProxyEvent`. However, it has key differences such as:

- Query string parameters are not URL-decoded automatically.
- The events either have `headers`/`queryStringParameters` or `multiValueHeaders`/`multiValueQueryStringParameters`, but
  not both.
- Headers without values are removed from the event's input.

Without this change, using `nitro` with Lambda + ALB yields unexpected behaviors like double URL encoding.

RESOLVES?

#2072

### 📝 Checklist

### DIFF
--- a/test/presets/aws-lambda.test.ts
+++ b/test/presets/aws-lambda.test.ts
@@ -1,65 +1,418 @@
 import type { APIGatewayProxyEvent, APIGatewayProxyEventV2 } from "aws-lambda";
 import destr from "destr";
 import { resolve } from "pathe";
-import { describe } from "vitest";
-import { setupTest, testNitro } from "../tests";
+import { describe, expect, it } from "vitest";
+import {
+  setupTest,
+  testNitro,
+  type Context,
+  type TestHandlerResult,
+} from "../tests";
+import type { DeepPartial } from "../../src/types/_utils";
+
+/**
+ * Enumerates the types of HTTP Lambdas:
+ *
+ * - `alb-single`: Application Load Balancer without multi-value query/headers enabled.
+ * - `alb-multi`: Application Load Balancer with multi-value query/headers enabled.
+ * - `rest`: API Gateway REST API.
+ * - `http-v2`: API Gateway HTTP API with v2.0 paylaod.
+ *
+ * Note: API Gateway HTTP API with v1.0 payload is omitted as it's the same as `rest` for practical purposes.
+ */
+type HttpLambdaType = "alb-single" | "alb-multi" | "rest" | "http-v2";
+
+/**
+ * Creates query string parameters with the format of an HTTP Lambda type.
+ * @param params The URL query string parameters.
+ * @param type The HTTP Lambda type.
+ */
+function createQueryStringParameters(
+  params: URLSearchParams,
+  type: HttpLambdaType
+) {
+  const queryStringParameters: Record<string, string | undefined> | undefined =
+    type === "alb-multi" ? undefined : {};
+  const multiValueQueryStringParameters:
+    | Record<string, string[] | undefined>
+    | undefined = type === "alb-single" ? undefined : {};
+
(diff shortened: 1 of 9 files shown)


test/presets/aws-lambda.test.ts
marcogrcr, line 1: **Note:** The tests were based on the behaviors observed in [here](https://gist.github.com/marcogrcr/1cce0db789ead5ae6a2d812936ba4b2e).

pi0:
Thanks for PR.

Considering major differences, i think it would be nicer if we develop a standalone subpreset + entry such as `aws-lambda-alb` and avoid changes to current `aws-lambda` (that as linked issue states is purpose built for gateway proxy)

test/presets/aws-lambda.test.ts
pi0: When work finished, would be nice if we can also setup for [nitro-deploys](https://github.com/nitrojs/nitro-deploys) to make sure preset works also in actual environment.

marcogrcr:
Thanks for the reply @pi0!

> Considering several differences, i think it would be nicer if we develop a standalone subpreset + entry such as `aws-lambda-alb`

Just to confirm, do you mean something like the following?

Update:

https://github.com/nitrojs/nitro/blob/f982f99fbab57d3bfc51d155f5fff73c7ab2a4ec/src/presets/aws-lambda/types.ts#L1-L3

to:

```ts
export type AwsLambdaOptions =
  | {
    // backwards compatibility, the presence of `type` overrides this
    streaming?: boolean;
  }
  | {
    /**
     * The Lambda function type:
     *
     * - `alb`: An ALB Lambda target group.
     * - `api-gateway`: An API Gateway REST API Lambda proxy integration, HTTP API Lambda integration, or Lambda function URL.
     * - `function-url`: A synonym of `api-gateway`.
     * - `function-url-stream`: A Lambda function URL with response streaming.
     *
     * @default "api-gateway"
     */
    type?: "api-gateway" | "alb" | "function-url" | "function-url-stream";
  };
```

And update:

https://github.com/nitrojs/nitro/blob/f982f99fbab57d3bfc51d155f5fff73c7ab2a4ec/src/presets/aws-lambda/preset.ts#L10-L16

to:

```ts
hooks: {
  "rollup:before": (nitro, rollupConfig) => {
    if ("type" in nitro.options.awsLambda) {
      if (nitro.options.awsLambda.type === "alb") {
        (rollupConfig.input as string) += "-alb";
      } else if (nitro.options.awsLambda.type === "function-url-stream") {
        (rollupConfig.input as string) += "-streaming";
      }
    } else if (nitro.options.awsLambda.streaming) {
      (rollupConfig.input as string) += "-streaming";
    }
  },
},
```

test/presets/aws-lambda.test.ts
marcogrcr: > When work finished, would be nice if we can also setup for [nitro-deploys](https://github.com/nitrojs/nitro-deploys) to make sure preset works also in actual environment.

I'm not sure how to action this since there doesn't seem to be any workflow in place that deploys to AWS (AWS account, credentials and infra would would be a pre-req). Also, it seems all deployments except [alwaysdata.yml] are currently [disabled].

[alwaysdata.yml]: https://github.com/nitrojs/nitro-deploys/blob/53976f9fa3337ee25832f6c22e948799ba423956/.github/workflows/alwaysdata.yml
[disabled]: https://github.com/nitrojs/nitro-deploys/tree/53976f9fa3337ee25832f6c22e948799ba423956/.github/disabled

pi0:
I mean making new variant / export in presets (similar to cloudflare/)

Issue #2932, opened 2024-12-12
Cache stream responses


kerwanp:
## Current behavior

When a stream is used in a `cachedEventHandler` the cached body is the serialized stream and not the result of this stream.
Resulting in an undesired behavior as the response sent when taken from the cache is the serialized stream.

## Desired behavior

When a cachedEventHandler returns a stream, the stream should be return and the response of the stream should be cached as the body.

### Additional information

- [X] Would you be willing to help implement this feature?

pi0x:
Works in v3 — a cached handler returning a `ReadableStream` now caches the consumed body, so repeat hits serve the stream content rather than a serialized stream.

_Triaged with AI assistance._

PR #2933, opened 2024-12-12
feat(cache): cache event handler stream response

kerwanp:
### 🔗 Linked issue

- #2932 

### ❓ Type of change

#### Problem

When a stream is used in a cachedEventHandler the cached body is the serialized stream and not the result of this stream.
Resulting in an undesired behavior as the response sent when taken from the cache is the serialized stream.

#### Solution

This pull request add the ability to use `defineCachedEventHandler` with stream responses.
When the `eventHandler` returns a stream, the response of this stream is cached rather than the stream itself.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -5,6 +5,7 @@ import {
   fetchWithEvent,
   handleCacheHeaders,
   isEvent,
+  isStream,
   splitCookiesString,
 } from "h3";
 import type { EventHandlerRequest, EventHandlerResponse, H3Event } from "h3";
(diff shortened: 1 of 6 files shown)


kerwanp:
As I am not aware of all the implications this implementation would have on all the different runtimes, I would be happy to discuss further on how this feature can get merged! @pi0

pi0x:

This looks resolved in nitro v3, which now uses the standalone [`ocache`](https://github.com/unjs/ocache) package for its caching layer instead of the old inline implementation. The bug this PR targeted — caching the raw, unresolved stream object instead of its actual content — can't happen in ocache's current `defineCachedHandler` (`src/http.ts`): it always fully awaits the response via `res.text()` before building the cache entry, so it stores resolved text content rather than a stream reference.

One nuance: this only covers *text-safe* streams. True binary/streaming-to-client caching (e.g. arbitrary binary payloads, chunked passthrough) is still an open gap in ocache today (fixing soon)

PR #2949, opened 2024-12-20
feat(cloudflare): generated wrangler configuration

pi0:
Nitro has introduced an experimental (opt-in) way to generate/extend `wrangler.toml` for cloudflare deployments (#2353, #2355) (However, overwriting the user config is not ideal or recommended, so it remain experimental)

The end goal is to leverage Cloudflare platform features such as Node compat v2 (hybrid) and Async Context with zero-config and ensure we can reliably enable these features.

Thanks to the huge huge efforts and collab with the Cloudflare team and @petebacondarwin, a new solution (~> cloudflare/workers-sdk#7442) is coming to allow frameworks like Nitro to generate a merged `wrangler.toml` in dist and config that redirects/points wrangler CLI and pages CI to it.

Docs: https://developers.cloudflare.com/workers/wrangler/configuration/#generated-wrangler-configuration

## Remarks

- `cloudflare.noWranglerDeployConfig: true` can be used to opt-out from generating deploy-config in `.wrangler/deploy/config.json` 
- User config `wrangler.toml|json` is read and always merged with the highest precedence
- Modules and frameworks can extend config by extending `cloudflare.wrangler` object
- Nitro preset adds defaults
   - `compatibility_date` defaults to `compatibilityDate` from framework config
   - `compatibility_flags` defaults to `["nodejs_compat", "no_nodejs_compat_v2"]`

## Pages

`pnpm nitro build playground --preset cloudflare_pages`

`dist/_worker.js/wrangler.json`:

```json
{
  "compatibility_date": "2024-09-19",
  "compatibility_flags": [
    "nodejs_compat",
    "no_nodejs_compat_v2"
  ],
  "pages_build_output_dir": ".."
}
```

## Modules

`pnpm nitro build playground --preset cloudflare_module`

`.output/server/wrangler.json`:

```json
{
  "compatibility_date": "2024-09-19",
  "compatibility_flags": [
    "nodejs_compat",
    "no_nodejs_compat_v2"
  ],
  "main": "index.mjs",
  "assets": {
    "binding": "ASSETS",
    "directory": "../public"
  }
}
```


src/presets/cloudflare/utils.ts
```diff
@@ -182,37 +186,106 @@ async function writeCFPagesRedirects(nitro: Nitro) {
...
+ * Merge user config with extra config
+ *
+ * - Objects/Arrays are merged
+ * - User config takes precedence over extra config
+ */
+function mergeWranglerConfig(
+  userConfig: WranglerConfig = {},
+  extraConfig: WranglerConfig = {}
+): WranglerConfig {
+  // TODO: Improve logic with explicit merging
```
atinux, line 289: We need to dedupe the compatibility flags here to be sure

petebacondarwin:
Given this is using the cloudflare_pages plugin, this should not have an `assets` section in the user's `wrangler.toml`.
This is likely to break the Pages CI.
Not a problem in real life because the user wouldn't have that. But in terms of testing, I will remove that from the wrangler.toml

petebacondarwin:
For Pages projects that have a `wrangler.toml` it is required to have a `pages_build_output_dir` field in order that the config is considered for deployment. Otherwise it is completely ignored.

Depending on what you want the DX, you could either require the user to add that field themselves, or you could add it as part of the generation of `wrangler.json`. I think the latter is better, since I believe you want to generate a `wrangler.json` even if there is no user `wrangler.toml`.

src/presets/cloudflare/utils.ts
```diff
@@ -182,37 +186,116 @@ async function writeCFPagesRedirects(nitro: Nitro) {
...
+
+  // Write config
+  // https://github.com/cloudflare/workers-sdk/pull/7442
+  const configRedirect = !!process.env.EXPERIMENTAL_WRANGLER_CONFIG;
+  if (configRedirect) {
+    if (mergedConfig.pages_build_output_dir) {
+      throw new Error(
+        "`pages_build_output_dir` wrangler config should not be set."
+      );
+    }
```
petebacondarwin, line 216: Oh I see that you actively do not want the user to set this.
OK. But in that case you need to provide it automatically.

petebacondarwin: Perhaps add after the check?

```
    mergedConfig.pages_build_output_dir = nitro.options.output.publicDir;
```

src/presets/cloudflare/preset.ts
```diff
@@ -173,8 +183,8 @@ const cloudflareModule = defineNitroPreset(
     entry: "./runtime/cloudflare-module",
     exportConditions: ["workerd"],
     commands: {
-      preview: "npx wrangler dev ./server/index.mjs --assets ./public/",
-      deploy: "npx wrangler deploy",
+      preview: "npx wrangler dev -c ./server/wrangler.json",
```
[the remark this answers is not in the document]
pi0: ~> https://github.com/cloudflare/workers-sdk/pull/7994

PR #2968, opened 2025-01-05
fix `no-misused-promises` eslint, when creating plugin with async body

CorrM:
### ❓ Type of change

### 📚 Description

When creating plugin with async body eslint will detect that the body should be function that returns `void`, while its okay to make the body of the plugin async.
_IMO this is the correct return type `void | Promise<void>`._

Here is the eslint message:
```
Promise returned in function argument where a void return was expected. `eslint@typescript-eslint/no-misused-promises`
```

### 📝 Checklist

### DIFF
--- a/src/types/runtime/nitro.ts
+++ b/src/types/runtime/nitro.ts
@@ -16,7 +16,7 @@ export interface NitroApp {
 }
 
 export interface NitroAppPlugin {
-  (nitro: NitroApp): void;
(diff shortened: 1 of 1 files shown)


pi0:
I appreciate the PR but actually current nitro plugins are **expected** to not be async, this is the correct type and warning. 

You should use async hooks within plugin body instead.

pi0:
where do you see them?

PR #2983, opened 2025-01-08
fix(loader): clone preset object

huang-julien:
### 🔗 Linked issue

https://github.com/nuxt/test-utils/issues/537
https://github.com/nuxt/test-utils/issues/1043

### ❓ Type of change

Hi :wave:

This PR fix an issue with presets and will allow multiple Nuxt to be ran within the same process (like for vitest workspaces or its vscode extension)

Linked issues above are trying to run multiple nuxt at the same time. It means we need also to be able to start multiple nitro.

Here's the race condition. 

Presets are objects and https://github.com/nitrojs/nitro/blob/e8d609908e9758094d3639b83fe2956ddf872fa8/src/core/config/loader.ts#L141
returns directly the reference of the preset.

If we load 2 nitro at the same time, C12 can delete _layers on the preset object
https://github.com/unjs/c12/blob/67cf916ca727e584390196ce3d3bfb76aa102ec1/src/loader.ts#L267
and the other other may need _layer at that time (calling .push() for example) which will fail and throw an error.

(will this pr be merge in nitro 2 or does it need another pr ? )

### 📝 Checklist

### DIFF
--- a/src/core/config/loader.ts
+++ b/src/core/config/loader.ts
@@ -7,8 +7,8 @@ import type {
   NitroConfig,
   NitroOptions,
   NitroPresetMeta,
-} from "nitropack/types";
-
+} from "nitro/types";
+import { defu } from "defu"
 import { NitroDefaults } from "./defaults";
 
 // Resolvers
@@ -138,7 +138,7 @@ async function _loadUserConfig(
       });
       if (preset) {
         return {
-          config: preset,
+          config: defu(preset),
         };
       }
     },
--- a/src/core/config/loader.ts
+++ b/src/core/config/loader.ts
@@ -1,6 +1,5 @@
 import { loadConfig, watchConfig } from "c12";
 import { type CompatibilityDateSpec, resolveCompatibilityDates } from "compatx";
-import { defu } from "defu";
 import { klona } from "klona/full";
 import type { PresetName } from "nitropack/presets";
 import type {
--- a/src/core/config/loader.ts
+++ b/src/core/config/loader.ts
@@ -1,5 +1,6 @@
 import { loadConfig, watchConfig } from "c12";
 import { type CompatibilityDateSpec, resolveCompatibilityDates } from "compatx";
+import { defu } from "defu";
 import { klona } from "klona/full";
 import type { PresetName } from "nitropack/presets";
 import type {
--- a/src/core/config/loader.ts
+++ b/src/core/config/loader.ts
@@ -8,7 +8,6 @@ import type {
   NitroOptions,
   NitroPresetMeta,
 } from "nitro/types";
-import { defu } from "defu"
 import { NitroDefaults } from "./defaults";
 
 // Resolvers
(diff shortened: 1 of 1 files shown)


pi0:
We should rebase it to v2. But first, can you please try helping to scope-down issue with a nitro-only reproduction? (or if we might narrow-down to c12 to avoid overriding input objects for `_layers`. shallow defu clone does not look like best option here)

src/core/config/loader.ts
```diff
@@ -125,7 +125,7 @@ async function _loadUserConfig(
       });
       if (preset) {
         return {
-          config: preset,
+          config: defu(preset),
```
pi0, line 128: `defu` is not safe as a clone utility. we might need to use krona.

huang-julien: haha i think @danielroe would be interested to see that :D . Why isn't it safe ?

I'll change move it to klona thanks ! :heart:

pi0: becaue defu only creates new object on top level (same effect here is { …preset } + when merging two namespaces. It is not for safe clone.

huang-julien: Oh that makes sense 👍👍

pi0:
@huang-julien can you please rebase your PR against v2 branch?

src/core/config/loader.ts
```diff
@@ -7,8 +7,7 @@ import type {
   NitroConfig,
   NitroOptions,
   NitroPresetMeta,
-} from "nitropack/types";
-
+} from "nitro/types";
```
pi0, line 10:
```suggestion
} from "nitropack/types";
```

src/core/config/loader.ts
```diff
@@ -8,7 +8,6 @@ import type {
   NitroOptions,
   NitroPresetMeta,
 } from "nitropack/types";
```
pi0, line 10:
```suggestion
} from "nitropack/types";

```

pi0 (approved):

pi0:
@huang-julien Thanks for PR. as this change adds runtime cost (and there is no easy way to reproduce/confirm with nitro only) can you please try on nightly channel if it actually solved linked issues?

huang-julien:
Thanks for merging this so fast :heart:

https://github.com/nuxt/nuxt/tree/repro/nitro-preset-shared-object here's the branch for reproduction in Nuxt. You need to run `npx jiti ./loadNuxt.ts` in `test/fixtures/multiple-nuxt`

It doesn't solve the issue in linked issues but allows to create the PR that will solve them in Nuxt.

pi0:
That's a Nuxt reproduction. it was generally good safety fix thanks ❤️

huang-julien:
ahhh so sorry about that 

https://stackblitz.com/edit/unjs-nitro-starter-qxgcfv55?file=createNitro.ts&title=Nitro%20Starter
https://stackblitz.com/edit/unjs-nitro-starter-y8dt79gv?file=createNitro.ts,patches%2Fnitropack.patch,package.json,nitro.config.ts&title=Nitro%20Starter

you can run `npx jiti createNitro.ts` in both stackblitz

PR #2984, opened 2025-01-08
feat: provide config idle timeout for bun runtime.

jamaluddinrumi:
```
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
```

we need to be able to configure idle timeout for bun runtime, because if there are any [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch) inside event handler sometimes need more than 10 seconds.

we can leverage built in `NITRO_SHUTDOWN_TIMEOUT ` environment variable to configure this.

references:
- https://bun.sh/docs/api/http#idletimeout
- https://nitro.build/deploy/runtimes/node#environment-variables

### ❓ Type of change

### 📝 Checklist


src/presets/bun/runtime/bun.ts
```diff
@@ -12,6 +12,7 @@ const ws = import.meta._websocket
 
 // @ts-expect-error
 const server = Bun.serve({
+  idleTimeout: process.env.NITRO_SHUTDOWN_TIMEOUT || 10,
```
pi0, line 15: These are not exactly same options, we need bun specific environment variable for this

jamaluddinrumi: `NITRO_BUN_IDLE_TIMEOUT` is good for this, isn't it?

btw, luckily, i can use `bun patch` to overcome this for now. so, no need to rush to fix this as soon as possible in my side.

jamaluddinrumi:
continued in https://github.com/nitrojs/nitro/pull/3293

PR #3046, opened 2025-01-29
fix(netlify): avoid second bundling

pi0:
Context: Found while investigating https://github.com/nuxt/nuxt/issues/30789#issuecomment-2622552133

Netlify functions (both user and framework internals) are bundled by default via a last esbuild step. ([netlify docs](https://docs.netlify.com/frameworks-api/#bundling-2)) this is an unnecessary step for Nitro as nitro already makes an optimized bundle output.

In the legacy netify v1 preset, we were using `nodeModuleFormat: "esm"` to preserve ESM compatibility however it is broken currently with v2 format (https://github.com/nitrojs/nitro/pull/2406) in 2.10.0

With this PR, ESM compat of output remains as-is

### DIFF
--- a/src/presets/netlify/utils.ts
+++ b/src/presets/netlify/utils.ts
@@ -116,6 +116,8 @@ export const config = {
   name: "server handler",
   generator: "${getGeneratorString(nitro)}",
   path: "/*",
+  nodeBundler: "none",
+  includedFiles: ["**"],
   excludedPath: ${JSON.stringify(getStaticPaths(nitro.options.publicAssets, nitro.options.baseURL))},
   preferStatic: true,
 };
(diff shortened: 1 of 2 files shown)


src/presets/netlify/utils.ts
```diff
@@ -116,6 +116,8 @@ export const config = {
   name: "server handler",
   generator: "${getGeneratorString(nitro)}",
   path: "/*",
+  nodeBundler: "none",
+  includedFiles: ["**"],
```
pi0, line 120: @serhalp this pattern is scoped to function dir right?

src/presets/netlify/utils.ts
```diff
@@ -116,6 +116,8 @@ export const config = {
   name: "server handler",
```
pi0, line 116: @serhalp (unrelated) from [docs](https://docs.netlify.com/frameworks-api/#configuration-options-2) i noticed it should be `displayName`? 🤔 (are there some types we could reuse perhaps for this object?)

pi0:
Merging to try on nightly

PR #3047, opened 2025-01-29
feat(cloudflare): `cloudflare:fetch` and `cloudflare:durable:fetch` hooks

pi0:
This PR enables two new hooks:

- `cloudflare:fetch(event)`
- `cloudflare:durable:fetch(event)`

These hooks (while adding to complexity and runtime overhead) allow custom extensions to normal nitro lifecycle.

The main use case of this is for durable objects to allow **redirect** some requests from main worker to the single instance durable worker.

```ts
export default defineNitroPlugin((nitroApp: NitroApp) => {
  nitroApp.hooks.hook("cloudflare:fetch", async (event) => {
    // Force redirects all fetch requests to the durable object
    return event.durableFech();
  });

  nitroApp.hooks.hook("cloudflare:durable:fetch", async (event) => {
    console.log("Handling response in durable object:", event.url);
  });
});
```

With the above plugin, `event` in event handlers has ` event.context.cloudflare.context` with the value of DurableObjectState type

---

(this is mainly a draft idea; An alternative is to simply expose a method of direct `fetch` to durable)

### DIFF
--- a/playground/wrangler.toml
+++ b/playground/wrangler.toml
@@ -10,4 +10,4 @@ class_name = "$DurableObject"
 
 [[migrations]]
 tag = "v1"
-new_classes = ["$DurableObject"]
+new_sqlite_classes = ["$DurableObject"]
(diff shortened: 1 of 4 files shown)


pi0:
Closing in favor of #3048, too much complexity and makes more edge cases if used wrongly, especially since we completely loose control over lifecycle

PR #2801, opened 2024-10-18
feat: experimental `cloudflare-durable` preset

pi0:
This PR adds a new experimental `cloudflare-durable` (based on `cloudflare-module`) preset which allows:

- out of the box websocket support with pub/sub
- `cloudflare:durable:init` and `cloudflare:durable:alarm` for custom hooks via plugins

Not documenting yet to experiment

PR #3048, opened 2025-01-30
feat(cloudflare-durable): allow redirect fetch to durable object

pi0:
This PR updates the experimental cloudflare_durable preset (#2801) with new context util `durableFetch()` allowing redirecting fetches to durable instance for handling.

This is useful for running code in a durable instance handler, such as access to the storage or connected WebSocket peers.

```ts
  const { durable, durableFetch } = event.context.cloudflare;

  if (durableFetch) {
    return durableFetch(); // Redirect to durable object
  }

  // This runs inside durable object
  console.log("Durable object:", durable);
  console.log("context:", event.context.cloudflare.context); // DurableObjectState

  return "check console logs";
});

```

Notes:
- `new_classes` in `wrangler` config should be changed to `new_sqlite_classes` to allow storage access
- Durable instance name changed from `crossws` (websocket-only) to `server` (generic)
- This API is part of the experimental preset and might change
- In nitro v2 / h3 v1, response handling (result of `durableFetch`) is limited. this PR aims for stable support in nitro v3 / h3 v2 with native `Response` handler

(alternative to #3047)

### DIFF
--- a/playground/wrangler.toml
+++ b/playground/wrangler.toml
@@ -10,4 +10,4 @@ class_name = "$DurableObject"
 
 [[migrations]]
 tag = "v1"
-new_classes = ["$DurableObject"]
+new_sqlite_classes = ["$DurableObject"]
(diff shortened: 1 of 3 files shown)


EnjoyKojima:
Hi !
Could you please let me know when the next release is scheduled?

PR #3050, opened 2025-01-30
feat(deno-server): deno v2 compatibility

marvinhagemeister:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/issues/3049

### ❓ Type of change

### 📚 Description

This PR updates the `deno-server` preset for Deno 2.x (released on October 9, 2024). It removes a couple of polyfills that were injected for features natively supported in Deno. This makes the generated server run faster.

The following changes have been made:

- Remove `process` polyfill -> natively supported in Deno (from 2.x onwards)
- ~~Remove `fetch()` polyfill -> natively supported in Deno~~ 
- Remove `web-streams-polyfill` -> natively supported in Deno
- Use Rollup's `external` option instead of a plugin for `https://` ~~or builtin Node module imports~~
- ~~Remove unnecessary `--config=./deno.json` argument from preview command~~
- Add `--unstable-node-globals` to `start` command to ensure that Node globals like `Buffer` and others are available globally
- Update test case for deno preset as it just needed the `cwd` arg passed when spawning the fixture server instead of manually resolving the config path

Resolves https://github.com/nitrojs/nitro/issues/3049

### 📝 Checklist


src/presets/deno/preset.ts
```diff
@@ -1,12 +1,7 @@
...
-import { findStaticImports } from "mlly";
 import { defineNitroPreset } from "nitropack/kit";
 import { writeFile } from "nitropack/kit";
-import { isAbsolute, resolve } from "pathe";
-
-// nitro/src/rollup/plugin/import-meta.ts
-const ImportMetaRe = /import\.meta|globalThis._importMeta_/;
+import { resolve } from "pathe";
 
 const denoDeploy = defineNitroPreset(
```
pi0, line 6: We should probably create new preset and enable via compatibilityDate (I will push)

src/presets/deno/preset.ts
```diff
@@ -43,109 +38,28 @@ const denoServer = defineNitroPreset(
...
-                map: s.generateMap({ includeContent: true }),
-              };
-            },
-          },
-        },
-      ],
+    },
+    alias: {
+      // Deno supports `fetch()` natively everywhere
+      "node-fetch-native": "unenv/runtime/mock/empty",
```
pi0, line 54: Note that `node-fetch-native` main subpath is **not** a polyfill. It uses [export conditions](https://github.com/unjs/node-fetch-native/blob/main/package.json#L12) and in case of deno, it is using native fetch.

This line should be avoided, because unenv redirects legacy packages in dependency tree (like `node-fetch` usages) to node-fetch-native (as stub)

marvinhagemeister: Without this line it always includes the `fetch` polyfill in the generated server code. Maybe something else is not working correctly.

src/presets/deno/preset.ts
pi0: 90a6484577da7f6ca75204d8ce8de1a19244e241

marvinhagemeister: Ah looks like you already pushed that 👍

src/presets/deno/preset.ts
marvinhagemeister: Nice, didn't know about the playground. That makes it a lot easier to work on presets. I stand corrected, looks like I was looking at an old build. Can confirm that it's not present. Thanks for taking a look 👍

pi0: Removed alias in d0e9c73686d6adaa7c753ae971065e29703e263a (screenshot above is from the state without using alias)

pi0:
Thanks for nice PR and description dear @marvinhagemeister ❤️ 

I have pushed few changes:
- removed extra alias
- updated build command to be as before. notice the CLI output will rewrite `./` as correct path like `✔ You can preview this build using deno task --config playground/.output/deno.json start`
- only have `https://` external condition. others, are handled by rollup config as we extend from node preset

src/presets/deno/preset.ts
```diff
@@ -45,107 +42,19 @@ const denoServer = defineNitroPreset(
...
     },
     hooks: {
       async compiled(nitro) {
-        // https://deno.com/manual@v1.34.3/getting_started/configuration_file
+        // https://docs.deno.com/runtime/fundamentals/configuration/
         const denoJSON = {
           tasks: {
             start:
-              "deno run --allow-net --allow-read --allow-write --allow-env --unstable-byonm ./server/index.mjs",
+              "deno run --allow-net --allow-read --allow-write --allow-env --unstable-byonm --unstable-node-globals ./server/index.mjs",
```
lucacasonato, line 57: Unstable byonm should not be needed

PR #3058, opened 2025-02-02
feat: add `bun-cluster` preset

### LINKED ISSUE #3057: Add bun-cluster preset
### Describe the feature

Hello,
We could utilize bun's [`reusePort`](https://bun.sh/guides/http/cluster) option in `Bun.serve` to allow multiple child processes to share the listening port. this allows a similar clustering approach used in `node-cluster` preset to make a new preset named `bun-cluster`.

### Additional information

Saeid-Za:
### 🔗 Linked issue

resolves #3057

### ❓ Type of change

### 📚 Description

This PR tries to create a new preset to support clustering mode similar to `node-cluster` in bun.

### 📝 Checklist


pi0:
Thanks for the PR. Since bun already supports `node:http` and `node:cluster`, can you please try a benchmark to see if output of `node-cluster` on bun is faster or slower than this new `bun-cluster`? 

(since it adds complexity to the codebase, I prefer to recommend `node-cluster` to bun users. it makes a difference in nitro v3 with native fetch support which we can consider)

pi0:
Yes indeed (and we do use native fetch for long time). 

I meant `Bun.serve({ fetch() => {} })` (as server interface). While it is more convenient, `node:http` implementation in Bun is sometimes faster.

Saeid-Za:
I used [`oha`](https://github.com/hatoo/oha) to benchmark a simple GET endpoint built with both `node-cluster` and `bun-cluster`.
Here is the `index.ts` in routes: 
```ts
let number = 1

export default eventHandler(async () => {
  return "Hello World! (request number: " + number++ + ")";
});
```

There were some unexpected/interesting results.

## 1. Using bun to run `node-cluster` : 31K Req/s

bun running node-cluster

## 2. Using Bun to run `bun-cluster` : 61K Req/s

bun running bun-cluster

## 3. Using node to run `node-cluster`: 72K Req/s

node running node-cluster


---

This means than if we set Bun running with `bun-cluster` as baseline:
1. bun with `bun-cluster` is slower than node with `node-cluster` (14.55% slower)
2. bun with `bun-cluster` is faster than running `node-cluster` preset (93.91% faster)

Saeid-Za:
Based on the benchmark data, it appears that using the `node-cluster` preset with the Node.js runtime is more performant than Bun, even when utilizing Bun's own HTTP server and clustering mechanism. This contradicts claims that Bun's approach is faster.

Given this information, it seems that this pull request primarily adds code complexity without providing significant benefits. We could revisit this PR at a later stage after additional releases from Bun to see if future benchmark results change.

I will close this for now.

Thank you for your input!

PR #3068, opened 2025-02-05
feat(prerender): use ultrahtml to extract links

danielroe:
resolves https://github.com/nitrojs/nitro/issues/2113

### ❓ Type of change

### 📚 Description

it's very tricky to extract valid links with a regexp without getting false positives (JSON payload, text content of the page, etc.)

This uses https://github.com/natemoo-re/ultrahtml to parse the HTML and extract links from all tags with a `href` - I would suggest we consider narrowing this down somewhat in future?

### 📝 Checklist


src/core/prerender/utils.ts
```diff
@@ -34,12 +33,17 @@ export function extractLinks(
...
   // Extract from any <TAG href=""> to crawl
   if (crawlLinks) {
-    _links.push(
-      ...[...html.matchAll(LINK_REGEX)]
-        .map((m) => escapeHtml(m[1]))
-        .filter((m) => !decodeURIComponent(m).startsWith("#"))
-        .filter((link) => allowedExtensions.has(getExtension(link)))
-    );
+    walkSync(parseHTML(html), (node) => {
+      if (node.attributes.href) {
```
pi0, line 37: Can you please refactor this to do with early return?

Also we can have a new crawler option that gets node and returns false if node can be ignored wdyt?

src/core/prerender/utils.ts
```diff
@@ -34,12 +33,17 @@ export function extractLinks(
 
   // Extract from any <TAG href=""> to crawl
   if (crawlLinks) {
-    _links.push(
-      ...[...html.matchAll(LINK_REGEX)]
-        .map((m) => escapeHtml(m[1]))
-        .filter((m) => !decodeURIComponent(m).startsWith("#"))
-        .filter((link) => allowedExtensions.has(getExtension(link)))
-    );
+    walkSync(parseHTML(html), (node) => {
```
pi0, line 36: Not sure about perf tradeoffs but as saw first example in ultrahtml readme is using async iteratable, maybe worth we also try async version?

src/core/prerender/utils.ts
danielroe: oh, I like the idea, but it would tie us to this particular AST. maybe better seeing if we need it? Or maybe adding a hook that gets called with extracted links instead?

pi0: my thinking was to give full control/awareness to filter callback (node type, other atts combo, etc). personally, I'm fine with betting on ultrahtml AST. But also fine with first adding that generic hook, feel free to drop another PR 👍🏼

danielroe:
seems behaviour is slightly different between walk + walkSync - attrs can be undefined in former case. updated.

PR #3085, opened 2025-02-12
feat: allow stacking error handlers

pi0:
This PR allows stacking multiple `errorHandler`s with fallback to Nitro default error handler.

With this, frameworks (and users) can add overlay error handlers and fallback to nitro default handlers by not handling response (note) conditionally which allows the leverage new built-in error handler DX introduced in #3002

(note): With h3 v2 / nitro v3, it should be by explicitly not returning a value, generic change in error handlers.

### DIFF
--- a/src/core/config/resolvers/error.ts
+++ b/src/core/config/resolvers/error.ts
@@ -4,9 +4,13 @@ import { join } from "pathe";
 
 export async function resolveErrorOptions(options: NitroOptions) {
   if (!options.errorHandler) {
-    options.errorHandler = join(
-      runtimeDir,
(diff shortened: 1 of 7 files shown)


danielroe (commented):
is there a way to access the content of the nitro default error handler, and modify it? (ie by wrapping it or adding  content?

I'm imagining displaying it as an overlay on top of the user's own error page (if they have one).

pi0:
Not with this one, but we can also expose a hook from dev error handler to allow this kind of overlay.

For API routes (and non HTML requests) I suggest we always bypass to nitro handler as SSR handler is not much helpful.

PR #3093, opened 2025-02-13
fix(prerender): ignore prefixed public assets

danielroe:
### ❓ Type of change

### 📚 Description

Currently there's no checking to see if a file already exists in public assets, and this has to be added manually to the ignore list. I think it's generally safe to ignore public assets paths as otherwise we have multiple possible files resolving to the same path.

Context: https://github.com/nuxt/nuxt/issues/25378

### 📝 Checklist


pi0:
This makes sense as general fix (since in general we always check static before hitting routes, prerender should not at least override files).

But with current approach, i'm afraid it will at least add overhead (of two times scanning nested public dir) and also implicit behavior change of stopping crawler if there is an overlapping static asset file. 

Maybe we can add a guard to [canWriteToDisk](https://github.com/nitrojs/nitro/blob/v2/src/core/prerender/prerender.ts#L131) to skip if file entry exists already?

Also perhaps Nuxt can explicitly ignore important files such as `manifest.json` from prerender?

danielroe:
We do ignore the Nuxt manifest:

https://github.com/nuxt/nuxt/blob/f56e29fb1caed4b502ec82baf6c3844767667767/packages/nuxt/src/core/nitro.ts#L266-L268

The linked issue is for a user `manifest.json` file, and the canWriteToDisk check would pass as `subfolderIndex` is on so it was actually generating `manifest.json/index.html`. It's a needless call to the SSR renderer (as static assets aren't reachable in prerenderer by default).

We could also add all public assets to ignore list but seemed better to fix in Nitro.

Perhaps we could cache scanned public asset files on the nitro instance so we don't scan it twice?

pi0:
Even if we say it is needless operation, it is behavior change. we do render during prerendering to both write staric pages but also crawl them all (even urls we don’t finally write like the ones with query params and this hidden case)

when subfolderIndex is on, we can also check for existence of parrent without index.HTML no? 

Im more lean towards dynamic write ignore since gives more flexibility (for example if we support emitting static assets from prerender routes)

danielroe:
no worries, i can implement this in nuxt directly if you don’t think it’s worth a broader nitro change

i regard the crawling of the path as a bug even if we don’t write it to disk, so personally i wouldn’t worry about a behaviour change

pi0:
not suggesting to diverge behavior (and main blast of behavior change is for nuxt users anyway)

Maybe this:
- fs override safeguard by default 
- opt-in option for current PR effect  nuxt can use

Also thinking we could indeed cache result of static writes (it happens before prerender, we just can keep it in nitro internal)

danielroe:
As it happens, `copyPublicAssets` is called by Nuxt _after_ the prerender step (because Nuxt writes some files to public assets _after_ prerendering based on the result) so caching the result may be tricky. We can work around in Nuxt by removing the cache but it seems ... not optimal. Do you have suggestions?

This also means FS protection step won't work (as it's order-dependent).

src/core/prerender/prerender.ts
```diff
@@ -120,6 +128,11 @@ export async function prerender(nitro: Nitro) {
       }
     }
 
+    // do not prerender any files in the public assets
+    if (publicAssets.has(route)) {
```
pi0, line 132: One thing we can optimize (for almost free -- no cache && no glob), are `nitro.options.publicAssets` with dir (no fallthrugh) such as client assets and build manifest dir.

We can also do it by default since this dirs (prefixes) were already explicitly configured as static (not mixed prefixes)

Another benefit is that regardless of prerender/copy order it works same.

src/core/build/assets.ts
```diff
@@ -8,10 +8,17 @@ import { compressPublicAssets } from "../utils/compress";
...
 const PARENT_DIR_GLOB_RE = /!?\.\.\//;
 
-export async function copyPublicAssets(nitro: Nitro) {
+export async function getPublicAssets(nitro: Nitro) {
   if (nitro.options.noPublicDir) {
     return;
   }
+
+  if (nitro._cachedPublicAssets) {
+    return nitro._cachedPublicAssets;
```
pi0, line 17: I think this cache will be unreliable if nuxt emits assets after prerendering.

We can keep `copyPublicAssets` intact and introduce a util that scans only unprefixed public assets (usable for opt-in feature of pretender)

src/core/prerender/prerender.ts
```diff
@@ -161,6 +174,14 @@ export async function prerender(nitro: Nitro) {
       }
     }
 
+    // Check if file already exists, for example copied across from public assets
+    const fileAlreadyExists =
+      existsSync(join(nitro.options.output.publicDir, route.fileName!)) ||
+      existsSync(join(nitro.options.output.publicDir, route.route));
```
pi0, line 180: We can optimize to reduce one stat if we check `fileName` ending with `/index.html`, then skip on parent existence (`manifest.json/index.html`)

pi0 (approved):
Looks good, only if we can find better name for `scanPublicAssets` (internal) and `prerender.ignorePublicAssets` (exposed for v2 -- we might drop for v3) that is clear it is about non fallthrough would be nice.

Issue #2974, opened 2025-01-07
Allow dynamic code within `defineRouteMeta`


pi0:
Related:
 - #2641
 - #2912

Experimental route meta ([#2102](https://github.com/nitrojs/nitro/pull/2102)) ([docs](https://nitro.build/guide/routing#route-meta)) only allows static definition like this:

```js
defineRouteMeta({
  openAPI: {
    tags: ["test"],
    description: "Test route description",
    parameters: [{ in: "query", name: "test", required: true }],
  },
});
```

However, users might want to generate it with other tools like `zod`.

In order to support this, we need to update [handlers-meta](https://github.com/nitrojs/nitro/blob/v2/src/rollup/plugins/handlers-meta.ts) rollup plugin to detect `defineRouteMeta` and actually **evaluate** (run) the code necessary in place since by-design this information should be statically extracted into the bundle.

LouisHaftmann:
Got auto imports to work! Now it's only missing `#nitro-internal-virtual/` packages:

```typescript
import type { Nitro, NitroEventHandler } from 'nitropack'
import type { Plugin } from 'rollup'
import { createJiti } from 'jiti'

const virtualPrefix = '\0nitro-handler-meta:'

export function handlersMeta(nitro: Nitro) {
  const jiti = createJiti(nitro.options.rootDir, {
    alias: nitro.options.alias,
    moduleCache: false,
    fsCache: false,
  })
  return {
    name: 'nitro:handlers-meta',
    async resolveId(id, importer, resolveOpts) {
      if (id.startsWith('\0')) {
        return
      }
      if (id.endsWith(`?meta`)) {
        const resolved = await this.resolve(id.replace(`?meta`, ``), importer, resolveOpts)
        if (!resolved) {
          return
        }
        return virtualPrefix + resolved.id
      }
    },
    load(id) {
      if (id.startsWith(virtualPrefix)) {
        const fullPath = id.slice(virtualPrefix.length)
        return this.load({
          id: fullPath,
        }).then((r) => r.code)
      }
    },
    async transform(code, id) {
      if (!id.startsWith(virtualPrefix)) {
        return
      }
      let meta: NitroEventHandler['meta'] | null = null

      try {
        const handlerFilePath = id.slice(virtualPrefix.length)
        const evaluated = jiti.evalModule(code, {
          id: handlerFilePath,
        })

        meta =
          evaluated && typeof evaluated === 'object' && 'meta' in evaluated
[...]

PR #3096, opened 2025-02-15
feat: dynamic data in defineRouteMeta

horvbalint:
### 🔗 Linked issue

#2974

### ❓ Type of change

### 📚 Description

In this PR I tried to add a very basic implementation of dynamic data in defineRouteMeta. It does not evaluate the meta in build time so if an approach like this is not desired, feel free to close this PR :)

The basic idea is that since we are already creating a virtual file that exports the route meta object for every route, it would make sense to have the computation in these files as well. I am searching through the route file code until the macro invocation is found. Along the way relative imports are updated with their absolute path. If the macro is found, I discard the lines after it and replace it with a default export of its argument.

As an example this route:

```ts
import utilityFunction from './utilityPath'
const openAPI = { description: 'The routes description' }

defineRouteMeta({ openAPI })

export default defineEventHandler(() => utilityFunction())
``` 

is transformed into this virtual file:
```ts
import utilityFunction from 'absolute/path/utilityPath'
const openAPI = { description: 'The routes description' }

export default { openAPI }
``` 

Though this is a very simple idea, it is working surprisingly well, allowing us to do very nice things with it like:

```ts
import { z } from 'zod'
import { extendZodWithOpenApi, createSchema } from 'zod-openapi'

extendZodWithOpenApi(z)

const bodyValidator = z.object({
    foo: z.enum(["bar", "baz"]).openapi({ description: "Property description" }),
})
const { schema, components } = createSchema(bodyValidator);

defineRouteMeta({
  description: routeDescription,
  requestBody: {
    content: {
      "application/json": {
        schema,
      },
    },
  },
  $global: {
    components,
  },
});

  const body = await readValidatedBody(event, validator.parse);
});

```

The meta or parts of it can also come from imports, so we can create utilities, like:

```ts
// server/utils/openapi.ts

import { z } from "zod";
import { extendZodWithOpenApi, createSchema } from "zod-openapi";

extendZodWithOpenApi(z);

export function createValidator(routeDescription: string, callback: (zod: typeof z) => z.AnyZodObject) {
  const validator = callback(z);
  const { schema, components } = createSchema(validator);

  return {
    validator,
    openAPI: {
      description: routeDescription,
      requestBody: {
        content: {
          "application/json": {
            schema,
          },
        },
      },
      $global: {
        components,
      },
    },
  };
}
```

```ts
// server/routes/index.post.ts

import { createValidator } from "~/utils/openapi";

const { validator, openAPI } = createValidator("Route description", (z) =>
  z.object({
    foo: z.enum(["bar", "baz"]).openapi({ description: "Property description" }),
  })
);

defineRouteMeta({ openAPI });

  const body = await readValidatedBody(event, validator.parse);
});
```

Of course there is no shortage of drawbacks/todos either:
- [x] ~~Relative imports can not be used in the macro~~
- [ ] Every code above the macro runs twice
  - this sounds bad, but users can be encouraged to place the macro as far up in the file as possible. In nuxt they can only appear at the very top
- [ ] If the code above the macro has side-effects, those will happen multiple times
  - I don't think this can be prevented in any way
- [ ] Variables placed under the macro can not be used in it, even if they are usually hoisted (functions, vars)

### 📝 Checklist


Navidtm:
I'm very excited about this feature :heart_eyes:.
I hope this feature will be added as soon as possible.

MahmoodKhalil57:
Hello guys, any updates on this feature? Would love to get my hands on this asap!

hacxy:
Hi, Has there been any progress on this feature recently?

Davvii1:
Waiting for this feature to be added, this look really cool. Any progress or news on this?

horvbalint:
The PR is waiting on a review from Pooya, but it might not be merged at all, since it does not implement the feature in a way that was previously requested by him. I am also rooting for this feature, since I would use it a lot, but please don't leave comments with the intention to speed the process up, cause that will have an opposite effect. Pooya has many many stuff to work on (eg.: upcoming v3 release) and I am sure he will get to this PR once he has the time. If you want to show support, please simply leave a reaction :)

pi0:
Thanks for the PR and patience @horvbalint.

We really need to keep supporting build-time evaluation for `defineRouteMeta` to be able to use it for route rules and any other feature that needs deterministic behavior based on route meta. Also this is important for runtime performance to not be forced to load all route chunks only to see meta.

We might use either jiti or esbuild (or rolldown in the future), which are already deps to evaluate extracted chunks.

Closing PR for clarity (tracker: #2974) if you could make a POC with above in implementations against `v3` it is more than welcome ❤️

horvbalint:
@pi0 Thanks for the review. I would love to make this happen at build-time, but I have very little knowledge on how to do that. I suspected that jiti will be the solution and will try to poke around with it, but likely I will need some help on how to structure something like this. Is there a discrod or something where I can ask for some help if needed?

On a sidenote:  I know this PR is not exactly what you were looking for, but when setting `openAPI.production = 'prerender'` this ended up being staticly added to the bundle. Is this not a viable solution? Are middlewares before files in the public dir impossible?

PR #3097, opened 2025-02-16
docs(cache): add return type constraint for cached functions

Buzut:
### ❓ Type of change

- [ x ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)

### 📚 Description

Cached functions are serialised as JSON regardless of the caching store. This excludes limits the return values of cached functions.

I added a warning to specify this (same way it is specified in the Nuxt docs regarding useState). It's IMO really important to mention it as I spent a good hour trying to understand why my function wasn't behaving as expected, trying to cache a JS Map.

### 📝 Checklist

- [ x ] I have updated the documentation accordingly.

### DIFF
--- a/docs/1.guide/6.cache.md
+++ b/docs/1.guide/6.cache.md
@@ -65,6 +65,8 @@ The stars will be cached in development inside ```.nitro/cache/functions/ghStars
 {"expires":1677851092249,"value":43991,"mtime":1677847492540,"integrity":"ZUHcsxCWEH"}
 ```
 
+::important Because the cached data is serialized to JSON, it is important that the cached function does not return anything that cannot be serialized, such as Symbols, Maps, Sets… ::
+
 ::note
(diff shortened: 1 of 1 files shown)


docs/1.guide/6.cache.md
```diff
@@ -65,6 +65,8 @@ The stars will be cached in development inside ```.nitro/cache/functions/ghStars
 {"expires":1677851092249,"value":43991,"mtime":1677847492540,"integrity":"ZUHcsxCWEH"}
 ```
 
+::important Because the cached data is serialized to JSON, it is important that the cached function does not return anything that cannot be serialized, such as Symbols, Maps, Sets… ::
```
pi0, line 68: This is valid BTW for **any** response. Exception / current limitation with cache vs normal routes is that it cannot handle _binary_ and _streams_

pi0 (approved):

PR #3120, opened 2025-02-22
fix(types): deduplicate aliases

danielroe:
### 🔗 Linked issue

context: https://github.com/nuxt/nuxt/issues/27161

### ❓ Type of change

### 📚 Description

Nuxt is setting paths to match its aliases. We can likely remove this code as it's now implemented in Nitro. But Nitro probably can safely deduplicate the paths too.

### 📝 Checklist

### DIFF
--- a/src/core/build/types.ts
+++ b/src/core/build/types.ts
@@ -280,23 +280,25 @@ declare module "nitropack/types" {
     });
 
     for (const alias in tsConfig.compilerOptions!.paths) {
-      const paths = tsConfig.compilerOptions!.paths[alias];
-      tsConfig.compilerOptions!.paths[alias] = await Promise.all(
-        paths.map(async (path: string) => {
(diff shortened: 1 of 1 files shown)


src/core/build/types.ts
```diff
@@ -280,23 +280,25 @@ declare module "nitropack/types" {
...
+            return relativeWithDot(
+              tsconfigDir,
+              stats?.isFile()
+                ? path.replace(/(?<=\w)\.\w+$/g, "") /* remove extension */
+                : path
+            );
+          })
+        )
       );
+      tsConfig.compilerOptions!.paths[alias] = [...paths];
```
pi0, line 301: Code style, we could deduplicate paths directly here (this isn't immediately visible to me since we spit deduplicate (new Set + to array) in two places

pi0 (approved):
Thnx!

PR #3149, opened 2025-03-04
docs(vercel): update docs

leerob:
More on Fluid here: https://vercel.com/blog/introducing-fluid-compute

More on KV -> Marketplace here: 

CleanShot 2025-03-04 at 07 33 47@2x


pi0 (approved):
Thnx!

PR #3155, opened 2025-03-05
fix(app): initialize `event.context.nitro` before calling `request` hook

zsilbi:
### 🔗 Linked issue

Fixes https://github.com/nitrojs/nitro/issues/3150

### ❓ Type of change

### 📚 Description

`event.context.nitro` is currently initialized in the generic event handler of `h3App`.
The `request` hook is called in `h3App.onRequest()` which runs before the generic event handler.
By moving the nitro context initialization there, we can provide the context for the listeners of the `request` hook.

### 📝 Checklist


pi0 (approved):
Thanks, `onRequest` was added later we could move other init logic early here as well!

PR #3156, opened 2025-03-05
fix(cli): set `NODE_ENV` when running build/dev commands

zsilbi:
### 🔗 Linked issue

Fixes https://github.com/nitrojs/nitro/issues/3052

### ❓ Type of change

### 📚 Description

When built from Nuxt the `$production` and `$development` config for Nitro is loaded correctly:
```ts
  $production: {
    nitro: {
      ignore: ['routes/dev/**']
    }
  },
  $development: {
    nitro: {
      ignore: ['routes/dev/**']
    }
  },
});
```

When built using the Nitro CLI, the same config does not work:

```ts
  $production: {
    ignore: ['routes/dev/**']
  },
  $development: {
    ignore: ['routes/dev/**']
  },
});
```

Setting the correct `NODE_ENV` before running the commands (like Nuxt CLI does) resolve this issue.

### 📝 Checklist

### DIFF
--- a/src/cli/commands/build.ts
+++ b/src/cli/commands/build.ts
@@ -35,7 +35,7 @@ export default defineCommand({
     },
   },
   setup() {
-    overrideEnv('production')
+    overrideEnv("production");
   },
   async run({ args }) {
     const rootDir = resolve((args.dir || args._dir || ".") as string);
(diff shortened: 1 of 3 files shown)


pi0:
thanks for PR. while it was an old issue with some (nuxt) packages, it is not applicable generally. and for c12 I think we should find a way to apply production/development without overriding `NODE_ENV`, we should apply conditions for c12 somehow (this also makes sure other meta frameworks work without being force to set this env)

PR #3157, opened 2025-03-05
fix(cache): fixed `defineCachedFunction` to use the function name

### LINKED ISSUE #3019: defineCachedFunction ignores function name
### Environment

nitropack 2.10.4

### Reproduction

https://stackblitz.com/edit/github-9uizvfes?file=server%2Froutes%2Findex.ts

Misbehaving code from the repro:

```ts
async function _foo() {
  return 'FOO';
}

const foo = defineCachedFunction(_foo);

async function _bar() {
  await new Promise((resolve) => setTimeout(resolve, 1));
  return 'BAR';
}

const bar = defineCachedFunction(_bar);

async function main() {
  // Expected: FOO BAR
  // Actual result: FOO FOO
  console.log(await foo());
  await new Promise((resolve) => setTimeout(resolve, 100));
  console.log(await bar());
}

main();
```

### Describe the bug

`defineCachedFunction` is supposed to guess cache key from the function name, and only use `_` as a fall back:

> Guessed from function name if not provided, and falls back to '_' otherwise.

However, name is never taken from the function, `_` is always used. This is due to bug where the default is taken before looking up function name:

https://github.com/nitrojs/nitro/blob/005a74f5173ba81d6f49772dfc6e35570a56beb0/src/runtime/internal/cache.ts#L40-L46

As you can see `opts.name` is taken first, which comes from `defaultCacheOptions()`, which always returns `{ name: "_" }`.

https://github.com/nitrojs/nitro/blob/005a74f5173ba81d6f49772dfc6e35570a56beb0/src/runtime/internal/cache.ts#L25-L32

### Additional context

_No response_

### Logs

```sh

```

zsilbi:
### 🔗 Linked issue

Fixes https://github.com/nitrojs/nitro/issues/3019

### ❓ Type of change

### 📚 Description

The provided function name was overwritten by the defaults.
I fixed that by removing it from the defaults when it's merged with the options.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -37,13 +37,14 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
   fn: (...args: ArgsT) => T | Promise<T>,
   opts: CacheOptions<T, ArgsT> = {}
 ): (...args: ArgsT) => Promise<T> {
-  opts = { ...defaultCacheOptions(), ...opts };
+  const defaults = defaultCacheOptions();
+  opts = { ...defaults, name: undefined, ...opts };
 
   const pending: { [key: string]: Promise<T> } = {};
 
   // Normalize cache params
   const group = opts.group || "nitro/functions";
-  const name = opts.name || fn.name || "_";
+  const name = opts.name || fn.name || defaults.name;
   const integrity = opts.integrity || hash([fn, opts]);
   const validate = opts.validate || ((entry) => entry.value !== undefined);
 


pi0:
This is valid fix thanks but it can cause cache busting (for any current usage) can you checm this?

In case of changing hashes we should do it for nitro v3 only (and all cache utils being externalized)

zsilbi:
> This is valid fix thanks but it can cause cache busting (for any current usage) can you checm this?
> 
> In case of changing hashes we should do it for nitro v3 only (and all cache utils being externalized)

Only those are affected by this who use `defineCachedFunction()` without providing the name option.

I think they might have already noticed that this is function not working correctly because the `cacheKey` for these is always `/cache:nitro/functions:_:.json` unless they provided something explcitly.

Cached event handlers provide the default name `_` in `opt`, custom `getKey()` and custom integrity hash.
An example of a default `cachedEventHandler` before/after (no change):

```ts
// Before  console.log([group, name, integrity], opts);
[ 'nitro/handlers', '_', 'lU0KWljkQ4' ] {
  name: '_',
  base: '/cache',
  swr: true,
  maxAge: 1,
  getKey: [AsyncFunction: getKey],
  validate: [Function: validate],
  group: 'nitro/handlers',
  integrity: 'lU0KWljkQ4'
}

// Cache key: /cache:nitro/handlers:_:index.il7asoJjJE.json
```

```ts
// After console.log([group, name, integrity], opts);
['nitro/handlers', '_', 'lU0KWljkQ4' ] {
  name: '_',
  base: '/cache',
  swr: true,
  maxAge: 1,
  getKey: [AsyncFunction: getKey],
  validate: [Function: validate],
  group: 'nitro/handlers',
  integrity: 'lU0KWljkQ4'
}

// Cache key: /cache:nitro/handlers:_:index.il7asoJjJE.json
```

HiroyukiYANO:
Any progress on this? I'm having the same issue. 
As a workaround, I'm using the name option.

pi0x:

This fix landed in [unjs/ocache#63](https://github.com/unjs/ocache/pull/63) (breaking change): `name` now resolves as `opts.name || fn.name || anon_<hash(fn)>`, restoring `fn.name` in the cache key, plus a hashed fallback so distinct anonymous functions no longer share/thrash a single key.

Merged but not yet released as part of a tagged version; will be available once nitro v3 upgrades to latest [ocache](https://github.com/unjs/ocache)

PR #3208, opened 2025-03-11
feat!: migrate to h3 v2

pi0:
[heavy WIP]

This PR migrates Nitro to use h3 v2 (see [migration guide](https://github.com/unjs/h3/blob/main/MIGRATION.md))

H3 is a full rewrite with backward-compatible utils. But the biggest difference is that we are finally moving to adopt Web (req => res) first class for servers.

(TODO)


pi0:
We have couple of todo items (including websocket) but i think ready to iterate from main with smaller updates.

PR #3211, opened 2025-03-11
feat: experimental rolldown support

pi0:
This PR adds experimental support for [rolldown](https://rolldown.rs/) bundler (for development and production nitro Builds) as a faster alternative to [rollup](https://rollupjs.org/).

In the next steps, we will add [Vite](https://vite.dev/) support as a bundler as well (which allows using either vite as a standalone bundler or using Nitro as an environment build target for Vite).

Supporting both rolldown and vite is easy in the long term as both 3 share a common rollup plugin API. 

This feature will be available initially on `nitro-nightly` channel and can be opted in using `builder: "rolldown"` config or `NITRO_BUILDER=rolldown` environment variable + installing `rolldown` peer dependency.

It is likely that all 3 `rolldown`, `vite`, and `rollup` become optional peer dependencies of Nitro so users can choose.

/cc @yyx990803 @hyf0 @IWANABETHATGUY @underfin @Boshen @shulaoda @sapphi-red

----

Current failing tests: (4/39)

```
pnpm test:rolldown test/presets/node.test.ts
```

- JSX support (there is no neutral option to avoid react)
- Static build flags (`import.meta.dev`, etc) plugin fails
- Raw loader (rolldown wrongly processes `\0path\to\asset.json` and fails)
- Externals bundled
- Async context


pi0:
(merging to iterate faster on `v3` branchs. feedbacks and contributions are more than welcome 🙏🏼 )

TheAlexLichter:
>     * JSX support (there is no neutral option to avoid react)

`preserve` + handle it via plugin(s) is no option?

>     * Raw loader (rolldown wrongly processes `\0path\to\asset.json` and fails)

I think this might be due to [module types](https://rolldown.rs/guide/in-depth/module-types#how-module-types-affect-users).

TheAlexLichter:
> - Static build flags (`import.meta.dev`, etc) plugin fails

Got it working with defines for most of it. You probably still want `replace` for e.g. runtime config - but the rolldown replace plugin [doesn't support functions](https://github.com/rolldown/rolldown/issues/4257) (yet?).

PR #3213, opened 2025-03-11
refactor!: avoid `import.meta` polyfill

pi0:
We have introduced a polyfill for `import.meta` in nitro v2 for platform compatibility however with advancements of runtimes, there will be more leverages of platform native `import.meta`.

Downside of depending on platform `import.meta` is that the value differs between source and bundle.

For internal use of serveStatic (to resolve public dir relative to server entry) a lightweight `globalThis.__nitro_meta__` injection is used instead. (hopefully one day we can use `import.meta.main` instead: https://github.com/nodejs/node/issues/49440)


TheAlexLichter:
https://github.com/nodejs/node/pull/57804 is supported in LTS now (since Node 22.18.0) 👏🏻

PR #3215, opened 2025-03-11
 refactor!: remove firebase functions in favor of firebase app hosting

Yuangwang:
### ❓ Type of change

### 📚 Description
Uncomment the previously commented out firebase app hosting documentation

### 📝 Checklist


docs/2.deploy/20.providers/firebase.md
```diff
@@ -27,198 +23,7 @@ You can integrate with this provider using [zero configuration](/deploy/#zero-co
     - Choose a region.
     - Import a GitHub repository (you’ll need to link your GitHub account).
     - Configure deployment settings (project root directory and branch), and enable automatic rollouts.
-    - Choose a unique ID for deployment.
-4. Wait for the first release to complete.
-
--->
-
-## Firebase hosting
```
pi0, line 35: Firebase hosting is still a supported preset, unless we want to deprecate it for (upcoming) Nitro v3.

Yuangwang: I discussed this with the Product Manager on our side and we would like to push users over to App Hosting + Nitro as opposed to Hosting classic + Nitro so they mentioned that they would prefer if the previous sections were removed. https://github.com/nitrojs/nitro/pull/2864/files#diff-264aa1b3e0b6fffa263d3f31dbf83811b5aeb513a38826853e2c00eb54b3e8e9

I know you mentioned that we could do that for the v3 docs but could we also do the same for the v2 docs? 

Perhaps we could also compromise and add a "deprecated" note in front of the Hosting section or a "recommended" note in front of the App Hosting section. 

Also lets continue the discussion here https://github.com/nitrojs/nitro/pull/3217!

pi0:
Note: We are removing legacy preset as it is on outdated versions and not actively maintained.

If in the future there is user request to add it back, we can definitely do this.

PR #3216, opened 2025-03-12
fix(lib/meta): apply compatible export solution for package version

markthree:
### ❓ Type of change

### 📚 Description

wechat_2025-03-12_152500_694

The current `​Import Assertions` syntax is excellent, but it has poor support in practice when running on Node.js, even in the v22 LTS version.
Perhaps we should switch to file reading instead. Though slightly slower, it would offer the best compatibility.

### 📝 Checklist

### DIFF
--- a/lib/meta.mjs
+++ b/lib/meta.mjs
@@ -1,1 +1,10 @@
-export { version } from "../package.json" with { type: "json" };
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { readFileSync } from "node:fs";
+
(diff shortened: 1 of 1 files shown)


lib/meta.mjs
```diff
@@ -1 +1,10 @@
-export { version } from "../package.json" with { type: "json" };
```
pi0, line 1: i think we can simply change it to `import pkg; export const version of pkg.version`

pi0 (approved):
Thanks ❤️

PR #3220, opened 2025-03-14
fix(core): scan handlers when initialising nitro

danielroe:
### ❓ Type of change

### 📚 Description

`writeTypes` relies on handlers having previously been scanned, so we need a way to do this. (See failing type tests in https://github.com/nuxt/nuxt/pull/31356 for example.)

This PR:

1. ~~exports `scanHandlers` for manual control~~
2. calls `scanHandlers` when initialising nitro (see https://github.com/nitrojs/nitro/pull/2546 for an identical previous PR)

### 📝 Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,5 +17,4 @@ export { runTask, listTasks } from "./task";
 export { build } from "./build/build";
 export { copyPublicAssets } from "./build/assets";
 export { prepare } from "./build/prepare";
-export { scanHandlers } from "./scan";
 export { writeTypes } from "./build/types";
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -17,4 +17,5 @@ export { runTask, listTasks } from "./task";
 export { build } from "./build/build";
 export { copyPublicAssets } from "./build/assets";
 export { prepare } from "./build/prepare";
+export { scanHandlers } from "./scan";
```
danielroe, line 20:
```suggestion
```

pi0:
Closing for now as main fix https://github.com/nitrojs/nitro/pull/2546 is added back to v3.

In the long term, I love to see if we can introduce `nitro.reload()` or `nitro.scan(filters?)` API for more control if needed 👍🏼

PR #3237, opened 2025-03-20
feat(tasks): add one-time tasks

schplitt:
### ❓ Type of change

### 📝 Checklist

Implements scheduled one time tasks for #1974

Croner only ever runs on full seconds, so the actual wait time may be slightly after the specified time.  
I added a test that runs a scheduled task. Sadly, I didn't find a way to use Vitest's mock timers to artificially advance the Cron job.

### DIFF
--- a/docs/1.guide/10.tasks.md
+++ b/docs/1.guide/10.tasks.md
@@ -108,6 +108,24 @@ export default eventHandler(async (event) => {
 });
 ```
 
+## Scheduled one time tasks
+
+To run a scheduled task once in the future, you can use `runTask(name, { context: { waitFor: 10 }})` to run the task in 10 seconds.
+
+**Example:**
+
+```ts [server/api/runOnce.ts]
+export default eventHandler(async (event) => {
+  runTask("future", { payload, context: { waitFor: 10 } });
+  
+  return { result: "Scheduled to run in 10 seconds" };
+});
+```
+
+> [!IMPORTANT]
+> Scheduled one time tasks are only ever run on full seconds.
+> If the actual time would be `YYYY-MM-DD HH:MM:10.250`, the task will be run on `YYYY-MM-DD HH:MM:11`.
+
 ## Run tasks with dev server
 
 Nitro's built-in dev server exposes tasks to be easily executed without programmatic usage.
--- a/src/runtime/internal/task.ts
+++ b/src/runtime/internal/task.ts
@@ -30,7 +30,9 @@ export async function runTask<RT = unknown>(
     context = {},
   }: { payload?: TaskPayload; context?: TaskContext } = {}
 ): Promise<TaskResult<RT>> {
-  if (__runningTasks__[name]) {
+  const runImmediately = !context?.waitFor;
+
+  if (!runImmediately && __runningTasks__[name]) {
     return __runningTasks__[name];
   }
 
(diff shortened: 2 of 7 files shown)


docs/1.guide/10.tasks.md
```diff
@@ -108,6 +108,24 @@ export default eventHandler(async (event) => {
...
+  runTask("future", { payload, context: { waitFor: 10 } });
+  
+  return { result: "Scheduled to run in 10 seconds" };
+});
+```
+
+> [!IMPORTANT]
+> Scheduled one time tasks are only ever run on full seconds.
+> If the actual time would be `YYYY-MM-DD HH:MM:10.250`, the task will be run on `YYYY-MM-DD HH:MM:11`.
+
```
iBobik, line 128: Maybe I am dump, maybe this needs clarification :-)

Does it depend on some scheduled task configured in the Cloudflare? E.g. there needs to be configured scheduled task to run every minute and nitro will execute all one-time tasks scheduled to run since last execution to now.

Or of this depends on some external API it is good to link it.

schplitt: You are right, this isn't very clear.

I think there isn't really a way to have the current implementation work on any edge worker.

The implementation creates a cron job that will we executed after the given time. This works fine in dev, node-server, bun and deno-server as the instance is always running.

This however won't work in edge workers.

In edge workers it can't be assured that there is enough wall time to wait for a crown job to run.

I am not sure how to even tackle it with edge workers.

I will update the docs to specify that this only works in long running servers.

src/runtime/internal/task.ts
```diff
@@ -50,14 +52,41 @@ export async function runTask<RT = unknown>(
...
+    } finally {
+      delete __runningTasks__[name];
+    }
   }
+
+  const nextExecution = new Date();
+  nextExecution.setSeconds(nextExecution.getSeconds() + context.waitFor!);
+
+  const p = new Promise<TaskResult<RT>>((resolve, reject) => {
+    const job = new Cron(
```
pi0, line 70: What would be even benefit of using corner if we are setting a timer in same running process?

schplitt: I thought having the same way of scheduling would be a good idea.
What would be better in your opinion?

pi0: Ideally, we need to persist future tasks and use runtime timer to load them.

We also need this to be mapped to other providers.

docs/1.guide/10.tasks.md
schplitt: Maybe i am misunderstanding this time, so please confirm: 

You mean we would deploy a workflow to invoke the task.
As an example: the main application being deployed as e.g. a worker and a workflow to handle (scheduled one-time) tasks?

src/runtime/internal/task.ts
schplitt: What are you referring to with runtime timer?

As to other providers, I would think for edge workers there would need to be a solution to let the provider schedule some kind of invokation to the task.

If you want I will remodel scheduling to have it persist and maybe have an internal handler for this to also cancel.
Any specifics regarding the implementation that would be necessary from your point of view?

hairyf:
Why not adopt a simpler pattern ?, such as specifying to run the task immediately when defining it

```ts
export default defineTask({
  immediate: true,

  meta: {
    name: "db:migrate",
    description: "Run database migrations",
  },
  run({ payload, context }) {
    console.log("Running DB migration task...");
    return { result: "Success" };
  },
});
```

pi0x:
Rebased this on latest main locally. A lot has moved since March: the dev task routes are now in `src/runtime/internal/routes/dev-tasks.ts` (and already parse the body safely), fixtures live under `test/fixture/server/`, and the docs page is `docs/1.docs/50.tasks.md`.

Two real bugs:

- The dedupe guard is inverted. `if (!runImmediately && __runningTasks__[name])` means normal `runTask` calls are no longer deduplicated at all.
- `TaskContext` is empty on purpose. Adding a named field turns on excess-property checks, so every preset that passes `context: { waitUntil }` stops compiling.

On design, @pi0 asked for persisted tasks that map to each provider. The in-process croner timer is lost on restart and cannot work on edge, so this still needs rework.

PR #3243, opened 2025-03-23
refactor: update NitroOptions to allow DotenvOptions for dev property

productdevbook:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -1,5 +1,5 @@
 import type { RollupCommonJSOptions } from "@rollup/plugin-commonjs";
-import type { C12InputConfig, ConfigWatcher, ResolvedConfig } from "c12";
+import type { C12InputConfig, ConfigWatcher, DotenvOptions, ResolvedConfig } from "c12";
 import type { WatchConfigOptions } from "c12";
 import type { ChokidarOptions } from "chokidar";
 import type { CompatibilityDateSpec, CompatibilityDates } from "compatx";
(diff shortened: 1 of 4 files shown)


src/types/global.ts
```diff
@@ -1,10 +1,12 @@
...
+import type { DotenvOptions } from "c12";
+import type { NitroConfig, NitroOptions } from "./config";
+import type { NitroModule } from "./module";
 
 export interface NitroStaticBuildFlags {
   _asyncContext?: boolean;
   _websocket?: boolean;
   _tasks?: boolean;
-  dev?: boolean;
+  dev?: boolean | DotenvOptions;
```
pi0, line 9: This is unrelated, should be always boolean

src/types/config.ts
```diff
@@ -172,8 +153,8 @@ export interface NitroOptions extends PresetOptions {
   ignore: string[];
 
   // Dev
-  dev: boolean;
-  devServer: NitroDevServerOptions;
+  dev: boolean | DotenvOptions;
```
pi0, line 156: We need a new `dotEnv` config if want to introduce it

pi0:
I think we should simply update `nitro dev` (CLI command) to load dot env by default.

No need to change other places.

pi0x:
Checked this against latest main. It looks obsolete now.

Main already has a separate `dotenv` option on `LoadConfigOptions` (added in #3274), which is what @pi0 asked for here. `nitro dev` also loads `.env` and `.env.local` by default. `dev` stayed a plain `boolean`, so the type change in this PR is no longer needed.

I did not rebase it, since there is nothing left to keep. Suggest closing.

Thanks for the original push on this, it did lead to the change landing.

PR #3246, opened 2025-03-24
refactor: add wildcard support for internal runtime module paths

productdevbook:
### ❓ Type of change

### 📚 Description

if we continue to use them in this way, there will be road and ts errors, and we need to fix them to prevent this.

```
function getNitroImportsPreset(): Preset[] {
  return [
    {
      from: "nitro/runtime/internal/app",
      imports: ["useNitroApp"],
    },
    {
      from: "nitro/runtime/internal/config",
      imports: ["useRuntimeConfig"],
    },
    {
      from: "nitro/runtime/internal/plugin",
      imports: ["defineNitroPlugin", "nitroPlugin"],
    },
```

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "./presets": "./dist/presets/index.mjs",
     "./runtime": "./dist/runtime/index.mjs",
     "./runtime/internal": "./dist/runtime/internal/index.mjs",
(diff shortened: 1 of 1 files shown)


pi0:
I think we can update the preset instead to directly import all utils from `nitro/runtime` (feel free to open a PR) internals are not to be exposed.

PR #3247, opened 2025-03-24
feat: initialize global nitro object and expose useRuntimeConfig

productdevbook:
### 🔗 Linked issue
V2 brach

[V3 pr](https://github.com/nitrojs/nitro/pull/3242)

close: https://github.com/nitrojs/nitro/issues/1472
cc: https://github.com/nuxt/nuxt/issues/14919

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/config.ts
+++ b/src/runtime/internal/config.ts
@@ -38,6 +38,12 @@ export function useRuntimeConfig<
   return runtimeConfig;
 }
 
+// Initialize global nitro object and expose useRuntimeConfig
+if (!globalThis.__nitro__) {
+  globalThis.__nitro__ = {};
+}
+globalThis.__nitro__.useRuntimeConfig = useRuntimeConfig;
+
 // App config
 const _sharedAppConfig = _deepFreeze(klona(_inlineAppConfig));
 export function useAppConfig(event?: H3Event) {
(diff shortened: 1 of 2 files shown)


cjpearson:
Would it also be possible to add a reference to the nitro app to the `__nitro__` object?

productdevbook:
> Would it also be possible to add a reference to the nitro app to the `__nitro__` object?

Hi, @cjpearson  Can you explain in a little more detail? I don't think I understand. And why you need this.

cjpearson:
I mean this object https://github.com/nitrojs/nitro/blob/v3/src/runtime/internal/app.ts#L213 

I'm working on tracing a nuxt/nitro app and it would be helpful to have access to the app instance outside a plugin. It's possible to work around, but if there's going to be this global `__nitro__` object it might be a nice addition.

pi0:
I appreciate the pull request. I think we need to think this through more and try a new system for Nitro v3 integrations first.

I have concerns about global side effects and multi-app support limitations

The goal is not to expose internals for external usage, but only better full-stack (nuxt) integration as described in https://github.com/nuxt/nuxt/issues/14919.

I think for the purpose of tracing, using async context (`useEvent()`) would probably be the best way to access Nitro event context, and for any other purpose, depending on the standard and universally accessible `process.env` allows external libraries to use the same source of trust.

PR #3268, opened 2025-04-01
fix(task): use `http.request` to support `socketPath` in dev fetch

### LINKED ISSUE #3264: Task CLI commands fail when dev server isn't using default port 3000
### Environment

Node: v22.14.0
Nitro: 2.11.6

### Reproduction

1. Start Nitro dev server with a custom port (not 3000)
2. Run `nitro task list` or another CLI command
3. See error message: ` "http://localhost:3000/_nitro/tasks": <no response> fetch failed`

### Describe the bug

When running the Nitro development server on a port other than the default 3000, CLI commands that communicate with the dev server (such as nitro task list) fail despite the server running properly. The errors show port 3000 is always used with the fetch.

Issue stems from below. 

```Typescript
// nitro/src/task.ts

  const devFetch = ofetch.create({
    baseURL: `http://${buildInfo.dev.workerAddress.host || "localhost"}:${
      buildInfo.dev.workerAddress.port || "3000"
    }`,
    // @ts-expect-error
    socketPath: buildInfo.dev.workerAddress.socketPath,
  });
``` 

Below is a potential fix that does work in my case, to properly allow dev server communication.

```Typescript
import { Agent } from 'undici';

 const devFetch = ofetch.create({
    baseURL: `http://${buildInfo.dev.workerAddress.host || "localhost"}:${buildInfo.dev.workerAddress.port || "3000"
      }`,
    dispatcher: new Agent({
      connect: {
        socketPath: buildInfo.dev.workerAddress.socketPath
      }
    })
  });
```

This fixes the issue, though it results in bundling warnings:
- Implicitly bundling "undici"
- Potential implicit dependencies found: node:sqlite
ERROR  Exiting with code (1). You can change this behavior by setting failOnWarn: false.

### Additional context

_No response_

### Logs

```sh

```

kricsleo:
### 🔗 Linked issue

resolves #3264

### ❓ Type of change

### 📚 Description

Since `ofetch` doesn't support `socketPath`, this PR directly uses `http.request`.
The implementation is pretty simple and only handles `JSON` responses, but that should be enough since the task's response is required to be `JSON`.

Otherwise, we'd need to use `undici`'s `Agent` for `socketPath` support, which would require adding `undici` as a dependency or bundling it—both options seem to add extra overhead in my opinion 🤔

Looking forward to your feedback! ❤️

### 📝 Checklist

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -6,8 +6,11 @@ import type {
   TaskEvent,
   TaskRunnerOptions,
 } from "nitro/types";
-import { ofetch } from "ofetch";
 import { normalize, resolve } from "pathe";
+import http from "node:http";
+import type { RequestOptions } from "node:http";
+import { withBase, withQuery } from "ufo";
+import type { QueryObject } from "ufo";
 
 /** @experimental */
 export async function runTask(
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks ❤️  

I have not tested PR and [fetching socket](https://github.com/nitrojs/nitro/blob/main/src/runner/proxy.ts#L56-L121) is acually possible in Node, Deno and Bun but it can get buggy so we need to move away from it. Your code in PR would be a good start!

PR #3284, opened 2025-04-04
refactor(cloudflare): enable node compat and deploy config progressiv…

pi0:
`cloudflare.nodeCompat` and `cloudflare.deployConfig` flags introduced in recent versions allow Nitro deploys to leverage native node compatibility and deploy config features for optimized zero config deploys.

However, we kept this flag disabled by default as there is a risk of overriding the user dashboard configuration with the next deployment.

This PR adds a new progressive method to opt in:
- Enable `deployConfig` if not configured in new cloudflare workers CI deployments
- Enable `nodeCompat` if not configured if `deployConfig` is enabled


petebacondarwin (commented):
LGTM

PR #3293, opened 2025-04-08
feat: provide config idle timeout for bun runtime

jamaluddinrumi:
```
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
```

we need to be able to configure idle timeout for bun runtime, because if there are any [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch) inside event handler sometimes need more than 10 seconds.

we can use, let's say, `BUN_IDLE_TIMEOUT ` environment variable to configure this.

references:
- https://bun.sh/docs/api/http#idletimeout
- https://nitro.build/deploy/runtimes/node#environment-variables

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/presets/bun/runtime/bun.ts
+++ b/src/presets/bun/runtime/bun.ts
@@ -12,7 +12,7 @@ const ws = import.meta._websocket
 
 // @ts-expect-error
 const server = Bun.serve({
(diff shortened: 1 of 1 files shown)


jamaluddinrumi:
continued in https://github.com/nitrojs/nitro/pull/3294

PR #3319, opened 2025-04-16
docs(database): update default db location

### LINKED ISSUE #3318: Default database location moved without notice
### Environment

Node v22.14.0
Nuxt v3.16.2

### Reproduction

Enable the experimental database mode with better-sqlite3 as the connector

### Describe the bug

The documentation states that the database file defaults to `.data/db.sqlite3`. This was also the case in previous versions. Somewhere along the lines this default however must have changed as starting an updated Nuxt server and accesing a route which uses `useDatabase()` results in the creation of `.data/db.sqlite3`

### Additional context

_No response_

### Logs

```sh
A lot of lines about missing tables as the newly created database is empty

ℹ Error: no such table: entities
```

septatrix:
Fixes #3318

### DIFF
--- a/docs/1.guide/5.database.md
+++ b/docs/1.guide/5.database.md
@@ -6,7 +6,7 @@ icon: ri:database-2-line
 
 > Nitro provides a built-in and lightweight SQL database layer.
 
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks ❤️ (this PR is landing for v3 WIP docs)

Issue #3256, opened 2025-03-25
Support edge function caching with netlify-edge preset


peralta:
### Describe the feature

The `netlify-edge` preset does not honour `isr` caching settings from Nuxt config. This seems to be intended behaviour and properly documented (https://nuxt.com/docs/guide/concepts/rendering#route-rules), although it would be great to support it for edge functions.

Besides the preset change to add the relevant headers, the generated `manifest.json` file should set caching to manual.

Refs: 
- https://docs.netlify.com/edge-functions/optional-configuration/#response-caching
- https://docs.netlify.com/edge-functions/create-integration/

### Additional information

- [ ] Would you be willing to help implement this feature?

PR #3321, opened 2025-04-17
fix(netlify-edge): enable function caching

peralta:
Currently, netlify is ignoring any cache directives because caching is not enabled via the integration manifest.
A config like the following would have no effect:
```
  routeRules: {
    '/hello-cache': {
      cache: {
        maxAge: 38400,
      },
    },
  }
```

See: https://docs.netlify.com/edge-functions/optional-configuration/#response-caching

### 🔗 Linked issue

#3256 

### ❓ Type of change

### 📚 Description

This PR contributes to getting the netlify preset to properly support caching (see #3256). It should be backwards compatible in the sense that the preset change has no effect for existing deployments.

Follow up work would still be needed as, after the patch, the only way to control edge function caching is through additional header config in `routeRules`:
```
    '/hello-cache-headers': {
      headers: {
        'Cache-Control': 'public, max-age=3600, must-revalidate',
        'Netlify-CDN-Cache-Control': 'public, max-age=38400, stale-while-revalidate=60, durable',
      },
    }
```
 
### 📝 Checklist

### DIFF
--- a/src/presets/netlify/preset.ts
+++ b/src/presets/netlify/preset.ts
@@ -95,6 +95,7 @@ const netlifyEdge = defineNitroPreset(
               name: "edge server handler",
               function: "server",
               generator: getGeneratorString(nitro),
+              cache: "manual",
             },
           ],
         };


serhalp (approved):
You're absolutely right — thanks for the fix!

PR #3324, opened 2025-04-18
fix(open-api): load virtual files from rollup rather than fs

danielroe:
### ❓ Type of change

### 📚 Description

this addresses an issue found by @atinux in https://github.com/nuxt-ui-pro/chat/pull/10.

minimal reproduction: https://stackblitz.com/edit/github-eddeyrez?file=nitro.config.ts

the issue is that the handlersMeta plugin always tries to load files from disk, but in some cases this is impossible. We bail out if a null byte is detected, which will cover nitro virtual files as well as any virtual files present from other plugins.

### 📝 Checklist


pi0 (approved):
Thnx!

PR #3326, opened 2025-04-22
fix: explicitly set moduleType to `js` for raw loader

TheAlexLichter:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/pull/3211#issuecomment-2820298178

### ❓ Type of change

### 📚 Description

Resolves the issue that `.json` files are not "handled correctly" by setting the [module type](https://rolldown.rs/guide/in-depth/module-types#how-module-types-affect-users) of all non-binary raw files to `js` while keeping the transform code.

Another option would be setting the `moduleType` to `json` (which is the default for `.json` files) and keep the file as is, removing the need for transforms. In my test, that clashed with other types of files though.

### 📝 Checklist

### DIFF
--- a/src/build/plugins/raw.ts
+++ b/src/build/plugins/raw.ts
@@ -71,9 +71,9 @@ export function raw(opts: RawOptions = {}): Plugin {
         };
       }
       return {
-        code,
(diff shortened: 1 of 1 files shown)


pi0 (approved):
thnx!

PR #3330, opened 2025-04-22
fix(rolldown): use rollup replace plugin for now

TheAlexLichter:
Related Issue/PR: https://github.com/nitrojs/nitro/pull/3211

This PR resorts back to the rollup replace plugin for now. After https://github.com/rolldown/rolldown/issues/4257 is resolved, switching to the rolldown one should be drop-in.

I'd still consider using `define` for most scenarios as it is less brittle (AST replacement instead of direct one). It does *not* support functions though.

### DIFF
--- a/src/build/rolldown/config.ts
+++ b/src/build/rolldown/config.ts
@@ -32,7 +32,10 @@ export const getRolldownConfig = (nitro: Nitro): RolldownOptions => {
     ],
     plugins: [
       ...(baseBuildPlugins(nitro, base) as RolldownPlugin[]),
-      replace({ preventAssignment: true, values: base.replacements }) as RolldownPlugin,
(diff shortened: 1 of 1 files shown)


src/build/rolldown/config.ts
```diff
@@ -31,6 +32,10 @@ export const getRolldownConfig = (nitro: Nitro): RolldownOptions => {
     ],
     plugins: [
       ...(baseBuildPlugins(nitro, base) as RolldownPlugin[]),
+      replace({
```
pi0, line 35:
```suggestion
      // https://github.com/rolldown/rolldown/issues/4257
      replace({
```

pi0 (approved):
thnx!

Issue #3337, opened 2025-04-28
CLI always overrides rootDir


unknown:
### Environment

nitropack@2.11.9

### Reproduction

https://stackblitz.com/edit/github-6kzuj6fp

```ts
//https://nitro.unjs.io/config
export default defineNitroConfig({
  rootDir: 'src',
  srcDir: 'server',
});
```

### Describe the bug

Configuring `root_dir` is effectively useless, because it is always overridden by the CLI.

https://github.com/nitrojs/nitro/blob/b5e29ed5ab4f8ff11afd72c8b9e31a7a7bdbc95a/src/cli/commands/dev.ts#L39
https://github.com/nitrojs/nitro/blob/b5e29ed5ab4f8ff11afd72c8b9e31a7a7bdbc95a/src/cli/commands/prepare.ts#L16
https://github.com/nitrojs/nitro/blob/b5e29ed5ab4f8ff11afd72c8b9e31a7a7bdbc95a/src/cli/commands/build.ts#L46

### Additional context

This seems like a bug to me, but looking at the git history, it seems like this behavior has been a part of Nitro for quite some time.

### Logs

```sh

```

pi0:
`rootDir` config is mainly for programmatic usage (with other wrapper frameworks) to specify where is root.

Any reason you simply don't use `srcDir`?

```js
export default defineNitroConfig({
  srcDir: 'src/server',
});
```

pi0:
> I imagine this may be used in a monorepo to configure the root dir at be the root of the monorepo

There is another explicit option for this `workspaceDir`.

unknown:
> There is another explicit option for this `workspaceDir`.

Ah, that works, but as far as I can tell it's not mentioned in the docs. Should I open a PR to add a note for `rootDir` and add `workspaceDir` to the docs?

pi0:
Yes please 🙏🏼

pi0:
(seperate PR for docs is always best :)

PR #3340, opened 2025-05-01
fix(cli): respect `rootDir` config in CLI

kricsleo:
### 🔗 Linked issue

resolves https://github.com/nitrojs/nitro/issues/3337

### ❓ Type of change

### 📚 Description

The CLI always overrides `rootDir` with `process.cwd()` when no `dir` argument is specified, ignoring the `rootDir` in the user's config file. But it should respect the user's config when no `dir` is provided and only override the user's config when `dir` is otherwise specified.

(If this fix is okay, I'll port it to v3)

### 📝 Checklist

### DIFF
--- a/src/core/config/loader.ts
+++ b/src/core/config/loader.ts
@@ -96,7 +96,7 @@ async function _loadUserConfig(
       : loadConfig<NitroConfig & { _meta?: NitroPresetMeta }>
   )({
     name: "nitro",
-    cwd: configOverrides.rootDir,
+    cwd: opts.cwd,
     dotenv: opts.dotenv ?? configOverrides.dev,
     extend: { extendKey: ["extends", "preset"] },
     overrides: {
--- a/src/core/config/loader.ts
+++ b/src/core/config/loader.ts
@@ -96,7 +96,7 @@ async function _loadUserConfig(
       : loadConfig<NitroConfig & { _meta?: NitroPresetMeta }>
   )({
     name: "nitro",
-    cwd: opts.cwd,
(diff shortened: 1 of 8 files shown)


src/core/config/loader.ts
```diff
@@ -96,7 +96,7 @@ async function _loadUserConfig(
       : loadConfig<NitroConfig & { _meta?: NitroPresetMeta }>
   )({
     name: "nitro",
-    cwd: configOverrides.rootDir,
+    cwd: opts.cwd || configOverrides.rootDir,
```
pi0, line 99: Generally, `configOverrides` should have must priority also for cwd.

pi0:
Thanks for the PR however, it is tricky. `rootDir` config is mainly for programmatic usage to specify project root (which also contains config) but with CLI arg, we actually specify rootdir of which project should be loaded.

Only possible use I can think about is that config being in different dir but it at least breaks main assumption that config is in root dir!

pi0:
Closing PR for now until we can discuss more about the validity of the case in #3337.

We can reopen it or rework.

kricsleo:
> main assumption that config is in root dir!

I agree with this assumption; it's conventional for real-world usage. 👍
However, since the current config rule allows users to break this convention, and since we mainly don't want to support this, do you think we should warn users about this abnormal config (config file not in the root directory)?

pi0:
Warning makes sense 👍🏼 (feel free to make PRs to `v3` directly, I backport some with cherry-pick to old release)

PR #3344, opened 2025-05-02
refactor: warn for config file outside of `rootDir`

kricsleo:
### 🔗 Linked issue

#3337

Context: https://github.com/nitrojs/nitro/pull/3340#issuecomment-2845098364

### ❓ Type of change

### 📚 Description

This PR adds a warning for nitro config file outside of `rootDir`.

### 📝 Checklist

### DIFF
--- a/src/config/loader.ts
+++ b/src/config/loader.ts
@@ -29,6 +29,8 @@ import { resolveURLOptions } from "./resolvers/url";
 import { resolveErrorOptions } from "./resolvers/error";
 import { resolveUnenv } from "./resolvers/unenv";
 import { resolveBuilder } from "./resolvers/builder";
+import { resolve } from "pathe";
+import consola from "consola";
 
 const configResolvers = [
   resolveCompatibilityOptions,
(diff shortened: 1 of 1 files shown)


src/config/loader.ts
```diff
@@ -110,6 +112,19 @@ async function _loadUserConfig(
           configs.rc?.[key] ??
           configs.packageJson?.[key]) as NitroConfig[K];
 
+      const configRootDir =
+        configs.main?.rootDir ??
+        configs.rc?.rootDir ??
+        configs.packageJson?.rootDir;
+      if (
+        configRootDir &&
+        resolve(configRootDir) !== resolve(configOverrides.rootDir || ".")
```
pi0, line 121: We could have a better sanity check [here](https://github.com/kricsleo/nitro/blob/refactor/warn-root-dir/src/config/loader.ts#L78) after resolving everything. `loadedConfig` includes `layers` (and top level layer, if exists should not have `rootDir` with different value

src/config/loader.ts
```diff
@@ -110,6 +112,19 @@ async function _loadUserConfig(
...
+      ) {
+        consola.warn(
+          `Nitro config file should be in the \`rootDir: '${configRootDir}'\`.`
```
pi0, line 124: The warning is a little misleading. Simply, we are warning that the `rootDir` specified in the config has been overridden by loader (Nitro CLI or other framework)

But honestly, I'm not even sure if it is best idea or if makes more false positive warnings.. We could also make `defineNitroConfig` type better to remove `rootDir` (as breaking change)

kricsleo: Removing `rootDir` from `defineNitroConfig` would be better if it's not meant to be used in the config file, IMO. Less work and also be less confusing.

PR #3348, opened 2025-05-06
refactor(prerender): show details for 5xx handled errors

### LINKED ISSUE #2992: Show prerender error details
### Environment

Nitro: 2.10.4
Node: 18.20.3 (stackblitz) and 20.16.0 (local test)

### Reproduction

https://stackblitz.com/edit/github-3wg9izmg?file=nitro.config.ts&view=editor

### Describe the bug

When returning an error like this:
```ts
export default eventHandler((event) => {
  return Error("Returning error");
});
```

The error reason is not shown in the console when prerendering:

```
Errors prerendering:
  ├─ /returnError (2ms)
  │ └── [500]
```

This case is the same if an error is thrown, but then at least a stack trace will appear:
```ts
export default eventHandler((event) => {
  throw Error("Throwing error");
});
```
```
ℹ Prerendering 3 routes
[nitro] [request error] [unhandled] Throwing error
  at Object.eval (./.nitro/prerender/chunks/routes/throwError.mjs:19:9)  
  at eval (./node_modules/h3/dist/index.mjs:2016:43)  
  at async Object.eval (./node_modules/h3/dist/index.mjs:2088:19)  
  at async toNodeHandle (./node_modules/h3/dist/index.mjs:2380:7)  
  at async ufetch (./node_modules/unenv/runtime/fetch/index.mjs:28:17)  
  at async generateRoute (./node_modules/nitropack/dist/core/index.mjs:1984:17)
  ├─ /throwError (4ms)
  │ └── [500]
```

### Additional context

This issue stems from [this Nuxt issue](https://github.com/nuxt/nuxt/issues/30527), which stems from trying to debug [this Pinia issue](https://github.com/vuejs/pinia/issues/2826)

In the Nuxt issue, I suggested a solution in the now removed [`normalizeError`](https://github.com/nitrojs/nitro/blob/728180981a1209cb1e27053ad167c230c408ad3a/src/runtime/internal/error/utils.ts#L33-L74) function, but that solution doesn't apply here since the function was removed.

~The only way I have found that an error will show nicely, is when setting the response status to 500, and providing text.~
Personally I would also like to see some sort of stack trace for further debugging, but I will leave that up to be debated about.

As [mentioned by Daniel Roe](https://github.com/nuxt/nuxt/issues/30527#issuecomment-2586631737), using `createError` is indeed the preferred way of returning an error. But that still does not properly show the error if you don't explicitly set `statusMessage`

### Logs

_No response_

kricsleo:
### 🔗 Linked issue

resolves #2992

### ❓ Type of change

### 📚 Description

This PR shows the prerender error details for better DX. 
(Not sure if it's the right place to handle errors, looking forward to your feedback ❤️)

|Before|After|
|--|--|
|image|image|

### 📝 Checklist


src/presets/_nitro/runtime/nitro-prerenderer.ts
```diff
@@ -1,9 +1,26 @@
...
+import consola from "consola";
+import { getRequestHeader, getRequestURL, isEvent } from "h3";
 import { useNitroApp } from "nitro/runtime";
 import { trapUnhandledNodeErrors } from "nitro/runtime/internal";
 
 const nitroApp = useNitroApp();
 
+nitroApp.hooks.hook("error", (error, context) => {
+  if (
+    isEvent(context.event) &&
```
pi0, line 11: Why we need this check?

pi0:
This is a really nice idea, but only one small issue that when tying locally, errors can be displayed twice now. We need to investigate.

image

src/presets/_nitro/runtime/nitro-prerenderer.ts
```diff
@@ -1,9 +1,26 @@
...
+    getRequestHeader(context.event, "x-nitro-prerender")
+  ) {
+    const url = getRequestURL(context.event).href;
+    consola.error(
+      `[prerender errror]`,
```
pi0, line 16:
```suggestion
      `[prerender error]`,
```

kricsleo:
Yeah, I noticed that; the second error log comes from the [`defaultHandler`](https://github.com/nitrojs/nitro/blob/v3/src/runtime/internal/error/prod.ts#L49) which is the default behavior. I'm not sure which one to skip.

Suppose skipping in this added hook:

```diff
+   !error.unhandled &&
    isEvent(context.event) &&
    getRequestHeader(context.event, "x-nitro-prerender")
```

Suppose skipping in the [`defaultHandler`](https://github.com/nitrojs/nitro/blob/v3/src/runtime/internal/error/prod.ts#L49):

```diff
-  if (isSensitive && !opts?.silent) {
+  if (isSensitive && !opts?.silent && !getRequestHeader(event, "x-nitro-prerender")) {
    // prettier-ignore
    const tags = [error.unhandled && "[unhandled]", error.fatal && "[fatal]"].filter(Boolean).join(" ")
    console.error(`[request error] ${tags} [${event.method}] ${url}\n`, error);
  }
```

Which do you prefer?

pi0:
Only catching non unhandled might make better sense.

We might also make it-opt in with an option/flag for showing all error details in addition to 404 exclude.

src/presets/_nitro/runtime/nitro-prerenderer.ts
```diff
@@ -1,9 +1,27 @@
...
+import { getRequestHeader, getRequestURL, H3Error, isEvent } from "h3";
 import { useNitroApp } from "nitro/runtime";
 import { trapUnhandledNodeErrors } from "nitro/runtime/internal";
 
 const nitroApp = useNitroApp();
 
+nitroApp.hooks.hook("error", (error, context) => {
+  if (
+    isEvent(context.event) &&
+    !(error as H3Error).unhandled &&
```
pi0, line 12: We should also ignore 404/40x (or maybe all non 500x?) errors. When crawling enabled, we hit routes that don't exist or need authentication or are (30x) redirect. We need better cases for this whether verbosity is needed or not.

codeflorist:
Is there any way to get this into nitro v2?

I assume, that this change could save thousands of wasted dev-hours per day.

pi0:
No reason it can't be. Feel free to open a PR @codeflorist. A small test in fixture is recommanded to verify it.

megrooms:
I'm glad this got fixed in v3 but could a similar fix be applied to v2 please?

PR #3361, opened 2025-05-13
fix: automatically trace package aliases from `package.json`

### LINKED ISSUE #3034: Nitro does not transpile libraries from external registry (JSR)
### Environment

- Operating System: Linux
- Node Version: v22.12.0
- Nitro Version: 2.10.4
- Package Manager: pnpm@9.15.4

### Reproduction

https://github.com/arunanshub/nuxt-jsr-bug-repro/tree/nitro

### Describe the bug

For some reason Nitro does not transpile libraries from external registry such as JSR even though its build output is supposed to be standalone. I am trying to use [@std/ulid](https://jsr.io/@std/ulid@1.0.0) from JSR but in runtime Nitro cannot find it.

Also see https://github.com/nuxt/nuxt/issues/30737

### Additional context

_No response_

### Logs

```sh
[nitro] [request error] [unhandled] Cannot find package '@std/ulid' imported from /app/server/chunks/routes/api/ulid.get.mjs
```

kricsleo:
### 🔗 Linked issue

resolves #3034

### ❓ Type of change

### 📚 Description

This PR automatically adds package aliases from `package.json` for tracing, so the final bundle can correctly link these dependencies.

For example:

```json
"dependencies": {
  "@std/ulid": "npm:@jsr/std__ulid@1.0.0-rc.4",
  "lodash4": "npm:lodash@4.17.21"
}
```

### 📝 Checklist

### DIFF
--- a/src/build/plugins/externals.ts
+++ b/src/build/plugins/externals.ts
@@ -301,6 +301,33 @@ export function externals(opts: NodeExternalsOptions): Plugin {
         }
       }
 
+      const userPkg = await readPackageJSON(
+        opts.rootDir || process.cwd()
+      ).catch(() => ({}) as PackageJson);
+
+      const userPkgAliases = Object.fromEntries(
+        [
+          ...Object.entries(userPkg.dependencies || {}),
+          ...Object.entries(userPkg.devDependencies || {}),
+        ]
+          .map(([name, version]) => {
+            // TODO: optimize this package name parsing
+            // { "lodash4": "npm:lodash@4.17.21" }              => [ "lodash" ~> "lodash4" ]
+            // { "@std/ulid": "npm:@jsr/std__ulid@1.0.0-rc.4" } => [ "@jsr/std__ulid" ~> "@std/ulid" ]
+
+            if (!version.includes(":")) {
+              return null;
+            }
+
+            const parts = version.split(":")[1]!.split("@");
+            return parts.length >= 3
+              ? ["@" + parts[1], name]
+              : [parts[0], name];
+          })
+          .filter(Boolean) as [string, string][]
+      );
+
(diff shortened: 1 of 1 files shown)


src/build/plugins/externals.ts
```diff
@@ -301,6 +301,33 @@ export function externals(opts: NodeExternalsOptions): Plugin {
...
+          .map(([name, version]) => {
+            // TODO: optimize this package name parsing
+            // { "lodash4": "npm:lodash@4.17.21" }              => [ "lodash" ~> "lodash4" ]
+            // { "@std/ulid": "npm:@jsr/std__ulid@1.0.0-rc.4" } => [ "@jsr/std__ulid" ~> "@std/ulid" ]
+
+            if (!version.includes(":")) {
+              return null;
+            }
+
+            const parts = version.split(":")[1]!.split("@");
```
pi0, line 322: This is not a reliable way.

- Aliases can be done using `resolutions`, `overrides`, or aliases like this
- Rules can also be from workspace settings
- Package aliases can happen **at any nested level** and are respected by package managers.

kricsleo: I see, then what about adding package aliases here:
https://github.com/nitrojs/nitro/blob/32e1850760fa19466ab69d31636ead05a856dc6c/src/build/plugins/externals.ts#L137-L139

If the imported package name doesn't match the resolved package name (also considering "Subpath export and full paths"), then there might be a package alias. We could automatically add this alias to `traceAlias`. How does that sound? 🤔

pi0:
Sorry it got stalled. Not sure how much relavant this will be in v3 but i have opened a followup issue in nf3 to potentially implement it https://github.com/unjs/nf3/issues/16

PR #3369, opened 2025-05-21
feat(cache): add serialize option to CacheOptions for custom serialization

### LINKED ISSUE #3367: Caching with `routeRules.cache.maxAge` is broken when returning `ReadableStream` from event handler
### Environment

nitropack v2.11.12
node v23.9.0

### Reproduction

```
git clone https://github.com/pajecawav/nitro-maxage-stream-issue
pnpm install
pnpm build && pnpm preview
```

Then navigate to `http://localhost:3000/`, wait for response to stream in and refresh the page

Note that `server/routes/index.ts` is taken from [h3 example](https://h3.dev/examples/stream-response#send-a-stream)

### Describe the bug

Setting `routeRules.cache.maxAge` causes `nitro` to cache an event handler response (in memory?), but caching breaks when event handler returns a `ReadableStream`. First (uncached) request is served correctly, but for all subsequent requests `nitro` serves an empty JSON object from the cache 

### Additional context

_No response_

### Logs

```sh

```

Narixius:
### 🔗 Linked issue

#3367, #2932, #2933

### ❓ Type of change

### 📚 Description

In some cases, we return a `ReadableStream` as the result of a request (such as when streaming React DOM), and the response needs to be cached. However, `ReadableStream` instances are not directly serializable, and attempting to cache them as-is results in an empty `{}` being stored.

To handle such cases, I’ve introduced a new method called `serialize` to the `CacheOptions` interface. This method allows defining custom serialization logic before storing the cache entry in the store. For example, you can convert a stream into a full HTML string or apply any other custom transformation as needed.

The existing `transform` method can then be used to deserialize the stored value when reading from the cache.

This change enables proper support for caching stream-based or otherwise non-serializable data structures.

Event handler example:
```tsx
// routes/index.tsx
export default defineCachedEventHandler(
  (e) => {
    return renderToReadableStream(<App />);
  },
  {
    swr: true,
    maxAge: 60,
    async serialize(entry, e) {
      if (entry.value.body instanceof ReadableStream) {
        const reader = entry.value.body
          .pipeThrough(new TextDecoderStream())
          .getReader();
        let html = "";
        while (true) {
          const { value, done } = await reader.read();
          if (done) break;
          html += value;
        }
        entry.value.body = html;
      }
      return entry;
    },
  }
);
```

Cached function example:
```tsx
const renderPage = defineCachedFunction(
  async (e: H3Event) => {
    return renderRoute(e); // which renders the route and returns a ReadableStream
  },
  {
    swr: true,
    maxAge: 60,
    getKey(e) {
      return `path-${btoa(e.path)}`;
    },
    async serialize(entry, e) {
      if (entry.value instanceof ReadableStream) {
        const reader = entry.value
          .pipeThrough(new TextDecoderStream())
          .getReader();
        let html = "";
        while (true) {
          const { value, done } = await reader.read();
          if (done) break;
          html += value;
        }
        entry.value = html;
      }
      return entry;
    },
  }
);

export default defineEventHandler(renderPage)
```

Resolves #3367, #2932, #2933

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -46,13 +46,14 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
   const name = opts.name || fn.name || "_";
   const integrity = opts.integrity || hash([fn, opts]);
   const validate = opts.validate || ((entry) => entry.value !== undefined);
-  const serializeValue = opts.serializeValue || ((value) => value);
+  const serialize = opts.serialize || (async (entry) => entry);
 
   async function get(
     key: string,
     resolver: () => T | Promise<T>,
     shouldInvalidateCache?: boolean,
-    event?: H3Event
+    event?: H3Event,
+    args: ArgsT = [] as any
   ): Promise<ResolvedCacheEntry<T>> {
     // Use extension for key to avoid conflicting with parent namespace (foo/bar and foo/bar/baz)
     const cacheKey = [opts.base, group, name, key + ".json"]
@@ -105,7 +106,7 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
       }
 
       try {
-        entry.value = await serializeValue(await pending[key]);
+        entry.value = await pending[key];
       } catch (error) {
         // Make sure entries that reject get removed.
         if (!isPending) {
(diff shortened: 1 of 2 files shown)


pi0x:

Implemented (independently) in [unjs/ocache#59](https://github.com/unjs/ocache/pull/59): a write-time `serialize(entry, ...args)` hook on `CacheOptions`, the write-side counterpart of the existing `transform`. It runs exactly once per resolution — folded into the shared in-flight promise so concurrent/deduplicated callers all observe the same serialized value — for resolver outputs a storage backend can't persist as-is (raw `ReadableStream`, class instances).

Merged but not yet released as part of a tagged version; will be available once nitro v3 upgrades to the latest [ocache](https://github.com/unjs/ocache) version.

PR #3370, opened 2025-05-22
fix(bun) remove `worker` condition from Bun

### LINKED ISSUE #3043: Bun subpaths imports are not resolved properly
### Environment

- nitro: 2.10.4
- bun: 1.2
- @clerk/backend: 1.23.7

### Reproduction

https://github.com/p9f/nitro-clerk-bun-repro

### Describe the bug

It seems that one can't install @clerk/backend on a bun + nitro server.
It fails with the error: 

> [nitro] [request error] [unhandled] Cannot find package '#crypto' from '/Users/p9f/repos/p9f/nitro-clerk-bun-repro/.output/server/node_modules/@clerk/backend/dist/chunk-G4VEKB6A.mjs'

My understanding is:
- clerk use a [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) for their crypto code ([package.json](https://github.com/clerk/javascript/blob/main/packages/backend/package.json#L16-L26))
- at build time, nitro is using the `worker` key to get the file path -> `./dist/runtime/browser/crypto.mjs`
  - this is what i figured playing with the different value, but it seems consistent with the bun preset config ([here](https://github.com/nitrojs/nitro/blob/v2/src/presets/bun/preset.ts#L8)): since `bun` is not defined, `worker` is next in line.
- at run time, bun is using the `node.imports` key to get the file path -> `./dist/runtime/node/crypto.mjs`, which causes the issue.
  - it seems consistent with [this section](https://bun.sh/docs/runtime/modules#importing-packages) of their documentation.  

We could remove the `worker` item of the bun preset `exportConditions` (which would fix the issue) - but i am unsure about the implications.
This happened with `@clerk/backend`, but i am assuming it would be the case with any other dependency using a subpath imports.

### Additional context

_No response_

### Logs

```sh

```

RiskyMH:
### 🔗 Linked issue

fixes #3043

### ❓ Type of change

  - technically it could be classified as breaking, but it shouldn't break much

### 📚 Description

Bun doesn't use the `worker` condition normally. The only time it should be used is when actually using worker on backend and that's a very low chance compared to people using [`@clerk/backend`](https://github.com/clerk/javascript/blob/main/packages/backend/package.json#L18). I decided to fix it here instead of clerk because high chance other packages could be broken.

Now as I mentioned it can hurt people trying to use bun workers *with package that has different code using condition* but I think its an important change because lower chance of usage. At least putting it below node/import could help as more packages define worker than bun (and also assume worker is browser). Also it's bold assumption to make that you would only want to use the worker and then only copy that code, instead a proper fix could be bundling conditions with more context on how its being imported.

### 📝 Checklist

### DIFF
--- a/src/presets/bun/preset.ts
+++ b/src/presets/bun/preset.ts
@@ -5,7 +5,7 @@ const bun = defineNitroPreset(
     entry: "./runtime/bun",
     serveStatic: true,
     // https://bun.sh/docs/runtime/modules#resolution
-    exportConditions: ["bun", "worker", "node", "import", "default"],
+    exportConditions: ["bun", "node", "import", "default"],
     commands: {
       preview: "bun run ./server/index.mjs",
     },


Jarred-Sumner:
@claude fix the compile errors. `--json` needs to be added in clap. fix the unused variable. 
25-05-21 12:38:24 PDT | mq zig build-obj bun ReleaseSafe x86_64-windows.win10_rs5-msvc (+3 more reused dependencies)
-- | --
  | 2025-05-21 12:38:24 PDT | src/cli/package_manager_command.zig:399:23: error: unused local constant
  | 2025-05-21 12:38:24 PDT | const path = bun.getcwd(&cwd_buf) catch {
  | 2025-05-21 12:38:24 PDT | ^~~~
  | 2025-05-21 12:38:24 PDT | src/deps/zig-clap/clap/comptime.zig:187:17: error: --json is not a parameter.
  | 2025-05-21 12:38:24 PDT | @compileError(name ++ " is not a parameter.");
  | 2025-05-21 12:38:24 PDT | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  | 2025-05-21 12:38:24 PDT | src/deps/zig-clap/clap/comptime.zig:123:45: note: called from here
  | 2025-05-21 12:38:24 PDT | const param = comptime findParam(name);
  | 2025-05-21 12:38:24 PDT | ~~~~~~~~~^~~~~~

Jarred-Sumner:
wrong pr

pi0 (approved):

mags-sully:
@RiskyMH  @pi0 are you able to merge? 🙏

PR #3393, opened 2025-06-05
refactor(netlify): disable `autoSubfolderIndex` for prerendered pages

alan-agius4:
### ❓ Type of change

### 📚 Description
This prevents an unnecessary redirect from `/page/` to `/page` when accessing prerendered content.

### 📝 Checklist


serhalp:
Adding some links for posterity (I referenced these for my own understanding here!):
- `prerender.autoSubfolderIndex` nitro option: https://nitro.build/config#prerender
- why this avoids a 301 redirect in some cases: https://answers.netlify.com/t/support-guide-how-can-i-alter-trailing-slash-behaviour-in-my-urls-will-enabling-pretty-urls-help/31191#p-86000-im-ssg-lessmy-ssg-uses-static-routing-4

src/presets/netlify/preset.ts
```diff
@@ -22,6 +22,9 @@ const netlify = defineNitroPreset(
       dir: "{{ rootDir }}/.netlify/functions-internal",
       publicDir: "{{ rootDir }}/dist/{{ baseURL }}",
     },
+    prerender: {
+      autoSubfolderIndex: false,
+    },
```
serhalp: Could you please add a comment here explaining why we set this? You can use the links I just shared in a comment.

serhalp: We should set this in all three presets: `netlify` (this one), `netlifyEdge`, and `netlifyStatic`. 🙏🏼

serhalp (commented):
Thank you! Good catch!

@pi0 this is marked as a breaking change, but I think with preset compat date mechanism we should instead bump the presets' compat dates? is that right? or maybe this is a safe enough that we just consider it a nonbreaking fix.

pi0:
Thanks for sharing context @serhalp I am totally fine with later backporting this also to v2, perhaps via compat date if you confirm this is certainly a better default and low chance of breaking behavior 👍🏼

src/presets/netlify/preset.ts
pi0: @serhalp Do you think perhaps later we can add a failing test to [nitro-deploys](https://github.com/nitrojs/nitro-deploys) about trailing redirects behavior with prerendered routes?

alan-agius4: Comments added.

alan-agius4: Added

pi0 (approved):

PR #3394, opened 2025-06-07
fix(prerender): pin `serverDir` to subdir of `buildDir`

kricsleo:
### 🔗 Linked issue

resolves https://github.com/nitrojs/nitro/issues/3392

### ❓ Type of change

### 📚 Description

This PR ensures that the prerender's `entryFileNames` respects the current preset's `entryFileNames` config. Without this change, there could be multiple entry files in the final server bundle. I'd appreciate your feedback! ❤️

### 📝 Checklist


src/core/prerender/prerender.ts
```diff
@@ -68,9 +69,14 @@ export async function prerender(nitro: Nitro) {
     ...nitro.options._config,
     static: false,
     rootDir: nitro.options.rootDir,
+    rollupConfig: {
+      output: {
+        entryFileNames: nitro.options.rollupConfig?.output?.entryFileNames,
```
pi0: I think in contrary prerender preset should _override_ this to fixed path in temporary `buildDir`. prerender is only and only meant to run during build step not affected by deployment settings.

kricsleo: Thanks for pointing that out! I was wondering about this before 😁. Now updated.

pi0:
reverted in #3478 we should revisit issue later

PR #3406, opened 2025-06-13
docs(nightly): include `bun.lock` for lockfile

iiio2:
bun basically generates `bun.lock` by default now from v1.2. I think we may add `bun.lock` also with `bun.lockb`. 
https://bun.sh/docs/install/lockfile


pi0 (approved):

iiio2:
Thanks @pi0 .

PR #3411, opened 2025-06-19
fix: remove ignoreInitial on the reloadWatcher

### LINKED ISSUE #3410: Plugins directory is not watched
### Environment

nitro v2
node 22

### Reproduction

https://stackblitz.com/edit/unjs-nitro-starter-e2qqum9j?file=.gitignore&title=Nitro%20Starter

- add the dir plugins in the server dir 

### Describe the bug

nitro doesn't reload in dev mode when a folder is being added into one of the scanDirs

### Additional context

original issue https://github.com/nuxt/nuxt/issues/32425

### Logs

```sh

```

huang-julien:
### 🔗 Linked issue
fix #3410
fix https://github.com/nuxt/nuxt/issues/32425

### ❓ Type of change

### 📚 Description
 
After looking at chokidar's doc. `ignoreInitial` is so confusing as a name. setting it to `false` (default value) makes chokidar emit `addDir` for undiscovered dir when creating the watcher

### 📝 Checklist

### DIFF
--- a/src/core/build/dev.ts
+++ b/src/core/build/dev.ts
@@ -31,7 +31,7 @@ export async function watchDev(nitro: Nitro, rollupConfig: RollupConfig) {
   ]);
 
   const watchReloadEvents = new Set(["add", "addDir", "unlink", "unlinkDir"]);
-  const reloadWatcher = watch(watchPatterns, { ignoreInitial: true }).on(
+  const reloadWatcher = watch(watchPatterns).on(
     "all",
     (event) => {
       if (watchReloadEvents.has(event)) {


huang-julien:
CI failing doesn't seems related to this PR 🤔

pi0:
Removing `ignoreInitial`, causes an event to be triggered immediately after `watch` is being called (dev server init) for all matching watch patterns. you can verify it by adding a debug line:

image

pi0:
We should perhaps initiate a new watcher instance for watching the top level scannable dirs.

Additionally, calling `reload` in the dev server watcher does not rescans plugins. `scanAndSyncOptions` is needed which is currently only called once on startup [see here](https://github.com/nitrojs/nitro/blob/046a8c291a84f6f75e294b2ec8ee30aaa42d0014/src/core/nitro.ts#L40-L42).

Lets keep track from #3410, feel free to followup with new PR (i will add comment there)

PR #3412, opened 2025-06-19
fix(dev): remove prepare output dir from dev command

huang-julien:
### 🔗 Linked issue

closes https://github.com/nuxt/nuxt/issues/32432

### ❓ Type of change

### 📚 Description

`prepare()` doesn't seems to be necessary for dev mode

### 📝 Checklist

### DIFF
--- a/src/cli/commands/dev.ts
+++ b/src/cli/commands/dev.ts
@@ -67,7 +67,6 @@ export default defineCommand({
       const server = createDevServer(nitro);
       const listhenOptions = parseArgs(args);
       await server.listen(listhenOptions.port || 3000, listhenOptions);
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR @huang-julien. I think this at least won't be affecting nuxt (nuxt does not uses nitro CLI for `dev` command).

huang-julien:
oops yes, shouldn't have added the `close` before the issue

huang-julien:
Should `prepare()` be used only for build ? If so we could add a JSDoc comment so frameworks don't call it when starting their dev server

pi0:
~> https://github.com/nitrojs/nitro/pull/3413

Issue #3409, opened 2025-06-18
Default error handler always uses `Server Error` as default


septatrix:
### Environment

Node 22
Nitro from git main (see reference code below for exact commit)

### Reproduction

See reference code below, otherwise a simple endpoint creating a 403 error without an explicit statusText suffices

### Describe the bug

When an error without an explicit statusText is produced somewhere in an application, Nitro set's a default one of "Server Error".
This suppresses the logic of user agents to display a reasonable string based on the status code (e.g. Unauthorized for 403).

The relevant code can be seen below:
- https://github.com/nitrojs/nitro/blob/4971b434db0461371c54425da68a4ef2f7d4256d/src/runtime/internal/error/prod.ts#L23
- https://github.com/nitrojs/nitro/blob/4971b434db0461371c54425da68a4ef2f7d4256d/src/runtime/internal/error/dev.ts#L31

### Additional context

Comment by @pi0 confirming this should not be done and is undesired:
https://github.com/h3js/h3/pull/1090#discussion_r2154490407

### Logs

```sh

```

pi0:
Thanks for issue. Feel free to open a PR to change it from v3 branch 👍🏼

PR #3416, opened 2025-06-24
refactor!: do not set default `statusText` (#3409)

septatrix:
Setting a generic message when the error
does not have an explicit message associated
suppresses correct inferrence by user-agents.

### 🔗 Linked issue

#3409

### ❓ Type of change

### 📚 Description

Not setting a generic fallback message will instead leave it to the client to show a reasonable message. These are generally inferred from the status code (e.g. 404 -> "Not found"). There might be clients which do not have an internal lookup table and instead don't show any message but we don't really lose anything there over the generic "Server Error". Browsers and modern tools are generally capable of deducing a message because HTTP/2 does not even send these message any longer.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/error/dev.ts
+++ b/src/runtime/internal/error/dev.ts
@@ -28,7 +28,6 @@ export async function defaultHandler(
 ): Promise<InternalHandlerResponse> {
   const isSensitive = error.unhandled;
   const status = error.status || 500;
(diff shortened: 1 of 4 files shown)


pi0:
@septatrix can you check failing CI errors?

pi0:
Thnx!

Issue #3418, opened 2025-06-25
Building Nitro apps fail when using Bazel because root `node_modules` is read-only


clementguillot:
### Environment

Nitro v2.11.12
Node v22.15.1

### Reproduction

Using this repo: https://github.com/clementguillot/testbazel

Command `bazel build //apps/frontend:build` to build frontend app (Tanstack Start based) fails with logs:
```
Another command (pid=16437) is running. Waiting for it to complete on the server (server_pid=8691)...
INFO: Analyzed target //apps/frontend:build (736 packages loaded, 8068 targets configured).
ERROR: /home/clement.guillot/sources/testbazel/apps/frontend/BUILD:91:14: JsRunBinary apps/frontend/.output failed: (Exit 1): vite failed: error executing JsRunBinary command (from target //apps/frontend:build) bazel-out/k8-opt-exec-ST-d57f47055a04/bin/apps/frontend/vite_/vite build
[plugin vite:css-post] Sourcemap is likely to be incorrect: a plugin (vite:css-post) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help

[nitro]  ERROR  Error: EACCES: permission denied, open '/home/clement.guillot/.cache/bazel/_bazel_clement.guillot/a1bdd2ced324bc220bb8cd8598cd04ce/execroot/_main/bazel-out/k8-fastbuild/bin/apps/frontend/.output/server/node_modules/tiny-warning/package.json'


undefined

error during build:
Error: EACCES: permission denied, open '/home/clement.guillot/.cache/bazel/_bazel_clement.guillot/a1bdd2ced324bc220bb8cd8598cd04ce/execroot/_main/bazel-out/k8-fastbuild/bin/apps/frontend/.output/server/node_modules/tiny-warning/package.json'
    at async open (node:internal/fs/promises:633:25)
    at async Object.writeFile (node:internal/fs/promises:1207:14)
    at async writePackage (file:///home/clement.guillot/.cache/bazel/_bazel_clement.guillot/a1bdd2ced324bc220bb8cd8598cd04ce/execroot/_main/bazel-out/k8-fastbuild/bin/node_modules/.aspect_rules_js/nitropack@2.11.12/node_modules/nitropack/dist/rollup/index.mjs:545:9)
    at async Promise.all (index 7)
[...]

PR #3419, opened 2025-06-25
fix: add an option to change output `node_modules` permission

clementguillot:
### 🔗 Linked issue

#3418

### ❓ Type of change

### 📚 Description

When Nitro uses a read-only `node_modules`, the Rollup `externals` plugins copies files and preserves the read-only permissions.
This results in a failure because the plugin needs to modify those files later in the build.

This PR applies `chmod 644` to all copied external files in the `.output/server/node_modules/*` folder.

### 📝 Checklist

### DIFF
--- a/src/rollup/plugins/externals.ts
+++ b/src/rollup/plugins/externals.ts
@@ -321,6 +321,9 @@ export function externals(opts: NodeExternalsOptions): Plugin {
           const dst = join(opts.outDir, "node_modules", pkgPath, subpath);
           await fsp.mkdir(dirname(dst), { recursive: true });
           await fsp.copyFile(src, dst);
+          if (opts.chmod) {
+            await fsp.chmod(dst, opts.chmod === true ? 0o644 : opts.chmod);
+          }
         }
 
         // Copy package.json
--- a/src/types/rollup.ts
+++ b/src/types/rollup.ts
@@ -49,6 +49,7 @@ export interface NodeExternalsOptions {
   exportConditions?: string[];
   traceInclude?: string[];
   traceAlias?: Record<string, string>;
+  chmod?: boolean | number;
 }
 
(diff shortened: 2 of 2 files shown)


src/rollup/plugins/externals.ts
```diff
@@ -321,6 +321,7 @@ export function externals(opts: NodeExternalsOptions): Plugin {
           const dst = join(opts.outDir, "node_modules", pkgPath, subpath);
           await fsp.mkdir(dirname(dst), { recursive: true });
           await fsp.copyFile(src, dst);
+          await fsp.chmod(dst, 0o644);
```
pi0: It looks like a very specific fix and might also have unwanted effects for current users.

Do you mind adding it behind a flag like `externals.chmod?: boolean | number` (`false` is default, `true` means same `0o644` as proposed)

clementguillot: Can you please confirm that the `chmod` flag should be added to `NodeExternalsOptions` interface?

pi0: Yes i think that's the one (once you pushed i can double check)

clementguillot: Done, option `chmod` type is `number | string | undefined` because its default value is `undefined` and we could simply pass `0o644` or `+x`. What do you think?

pi0: Nice. i think `true` might still make sense to chmod into a _normliezed_ permission (644/rw-r--r-- seems a reasonable default) but also happy to keep it explicit to users.

pi0 (approved):
Lets go with explicit version!

clementguillot:
@pi0 just pushed your suggested change

pi0 (approved):

PR #3436, opened 2025-07-02
fix(cloudflare): update wrangler types

luc122c:
### ❓ Type of change

### 📚 Description

I noticed that the types in [environment.ts](https://github.com/nitrojs/nitro/edit/v3/src/presets/cloudflare/wrangler/environment.ts) were out of date. This PR updates them with the [latest changes](https://github.com/cloudflare/workers-sdk/blob/29e911a/packages/wrangler/src/config/environment.ts) as of [this commit](https://github.com/cloudflare/workers-sdk/commit/29e911abbbd12385aec201cb9589cccd832fb400).

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare/wrangler/environment.ts
+++ b/src/presets/cloudflare/wrangler/environment.ts
@@ -6,6 +6,8 @@
  * Source: https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/environment.ts
  */
 
+import type { Json } from "miniflare";
+
 /**
  * The `Environment` interface declares all the configuration fields that
  * can be specified for an environment.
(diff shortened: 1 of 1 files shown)


luc122c:
Could this also be backported to current Nitro?

src/presets/cloudflare/wrangler/environment.ts
```diff
@@ -6,6 +6,8 @@
  * Source: https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/environment.ts
  */
 
+import type { Json } from "miniflare";
```
pi0: We should avoid direct dependency. Can you please inline it?

luc122c: No problem 👍

pi0:
(please check failed CI type issue. it was reason we had simplified JSON type — i think we can keep that as is)

luc122c:
No problem, I've reverted it to `Record<string, unknown>` and removed the troublesome type.

luc122c:
@pi0 Is it possible for this to be backported to v2?

pi0:
Sure 👍🏼 e6624f7b47e1bf09ca42077eca25b0485db78998

PR #3443, opened 2025-07-03
fix(content-negotiation) respect encoding file presence for Vary header

### LINKED ISSUE #3077: Nitro Content Negotiation fails
### Environment

nitro.config.js is 

```
//https://nitro.unjs.io/config
  srcDir: "server",
  compressPublicAssets: true,
  compatibilityDate: "2025-02-07"
});
```

Nitropack is 2.10.4

node --version
v20.18.0

### Reproduction

https://github.com/paulcrussell/minimal-nitro
To reproduce you will need to have a chrome extension which strips the Accept-Encoding header.
Or you could request one of the public assets with Postman and omit the Accept-Encoding header.

### Describe the bug

Nitro performs content negotiation however if the Accept-Encoding request header is stripped then the server returns the plain text version yet sets no Vary header in the response
This makes it possible for a client to poison a shared cache for all other clients with the plain text version

paulcrussell:
### 🔗 Linked issue

#3077

### ❓ Type of change

### 📚 Description

The build process produces compressed public assets based on nitro option 'compressPublicAssets' if those assets are present then it means the response may 'Vary' based on Accept-Encoding request header.

Currently the 'Vary' based on Accept-Encoding request header is included only if the Accept-Encoding request header exists. This introduces a problem that the Vary header can get omitted if the client sends a request without Accept-Encoding header. This would cause the server to respond with a uncompressed file that would then be cached in shared caches and served to everyone.

Shifting the logic so that the header is always sent if the server is optionally configured to return uncompressed, gz or br public assets removes the possibility for a client to poison the cache for all.

Resolves #3077

### 📝 Checklist


paulcrussell:
@pi0  I've updated the commit to fix the failing tests.

src/runtime/internal/static.ts
```diff
@@ -49,9 +56,6 @@ export default eventHandler((event) => {
       .sort(),
     "",
   ];
-  if (encodings.length > 1) {
-    appendResponseHeader(event, "Vary", "Accept-Encoding");
-  }
 
   for (const encoding of encodings) {
```
pi0, line 53: We are already iterating on encoding variations, on match, we can alternatively also append the header instead of additional loop

paulcrussell: Hi @pi0 the fix here is to change the dependency of the Vary response header.
current: dependent on the request accept-encoding request header
proposed: dependent on whether the application has the capability to Vary the response

So it needs to iterate all possible encodings (i.e. gz or .br) and validate if at least one encoding is enabled. In this case the server is capable of Varying the response so the Vary header can be sent.

pi0: Please let me double check this, we can land fix in next releases.

PR #3459, opened 2025-07-11
docs(cache): add note about normalizing keys

MrKaashee:
# Added documentation for Nitro v2 cache key normalization

This update adds detailed documentation on how Nitro v2 automatically normalizes cache keys to improve compatibility and security.

## Key points included:

- Explanation of the automatic normalization process that removes non-alphanumeric characters such as `/` and `-` from cache keys.
- Examples illustrating how raw keys like `/api/products/sale-items` are transformed into normalized keys such as `api/productssaleitems.json`.
- Introduction of the new `escapeKey(key: string): string` utility function exported by Nitro.
- Guidance on how to use `escapeKey()` to manually reproduce Nitro’s normalized cache key format, especially useful when manually invalidating cache entries.
- Code samples demonstrating both the normalization and manual invalidation processes.

This helps developers better understand cache key handling and ensures consistency when interacting with Nitro's cache storage system.

### DIFF
--- a/docs/1.guide/6.cache.md
+++ b/docs/1.guide/6.cache.md
@@ -303,7 +303,9 @@ await useStorage('cache').removeItem('nitro:functions:getAccessToken:default.jso
 
 ### Normalizing Cache Keys
 
+::important
 **Cache keys are automatically normalized** using an internal utility that removes non-alphanumeric characters such as `/` and `-`. This behavior helps ensure compatibility across various storage backends (e.g., `file systems`, `key-value` stores) that might have restrictions on characters in `keys`, and also prevents potential path traversal vulnerabilities.
+::
 
 For example:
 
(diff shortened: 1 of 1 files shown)


docs/1.guide/6.cache.md
```diff
@@ -300,3 +300,40 @@ You can invalidate the cached function entry with:
...
+
+Would generate a key like:
+
+```ts
+api/productssaleitems.json
+```
+
+This behavior may result in keys that look different from the original route or identifier.
+
+🔑 `escapeKey(key: string): string`
```
pi0: We can show a simple snippet instead + note that behavior might change so only using as advanced usage.

mrkaashee: Thanks for the feedback! Actually, I’m using the cache in my e-commerce app, and on the detail pages, I encountered issues due to `UUIDs` and the path formatting. That’s why I decided to add this explanation and the `escapeKey` snippet — to help handle these cases consistently.

pi0: I see. My suggestion is as alternative to #3458, directly document this:

```
 function escapeKey(key: string | string[]) {
  return String(key).replace(/\W/g, "");
}
```

PR #3463, opened 2025-07-11
fix(vite): set default consumer to "client" in client environment

markthree:
### ❓ Type of change

### 📚 Description

In [nitrojs/vite-examples](https://github.com/nitrojs/vite-examples), it appears impossible to detect the client environment consumer, resulting in failure to generate the correct `globalThis.__VITE_MANIFEST__`. I'm uncertain whether this is an issue with Vite itself; this is a simple patch solution.

#### Error Snapshot

image

### 📝 Checklist

### DIFF
--- a/src/build/vite/plugin.ts
+++ b/src/build/vite/plugin.ts
@@ -84,6 +84,9 @@ export async function nitro(
 
         // Add Nitro as a Vite environment
         environments: {
+          client: {
+            consumer: userConfig.environments?.client?.consumer || "client",
+          },
           ...createServiceEnvironments(ctx),
           nitro: createNitroEnvironment(ctx),
         },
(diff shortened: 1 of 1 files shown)


src/build/vite/plugin.ts
```diff
@@ -111,7 +111,8 @@ export async function nitro(
 
     // Modify environment configs before it's resolved.
     configEnvironment(name, config) {
-      if (config.consumer === "client") {
+      const mayBeClient = config.consumer === undefined && name === "client";
```
pi0: We could instead assign `consumer: "client"` in config hook as default value for `client` environment.

markthree: Thanks, I'll look into this tomorrow

pi0: No worries if you are happy with direction i can also push on your PR

markthree: done

pi0 (approved):

PR #3471, opened 2025-07-13
feat: netlify dev preset

pi0:
This PR adds a dev preset for netlify local emulation based on new dev presets (#3467) similar to cloudflare (#3470), and based on https://github.com/netlify/primitives/pull/344 as base implementation.

(experimental) We probably have to improve all implementations as a runtime middleware/plugin for a closer to prod experience, and also forward compatibility with Nitro v3 and its Vite env presets.


pi0:
Moving forward to e2e test via nightly channel. This certainly needs rework in next patch releases to be a runtime augmentation.

PR #3494, opened 2025-07-18
fix(vite): use buildApp hook and construct manifest from bundle

schiller-manuel:
use `buildApp` hook from vite 7 instead of configuring a builder

this ensures compatibility with frameworks that configure a builder themselves

### DIFF
--- a/src/build/vite/plugin.ts
+++ b/src/build/vite/plugin.ts
@@ -140,11 +140,17 @@ export async function nitro(
     generateBundle: {
       handler(_options, bundle) {
         const { root } = this.environment.config;
+        const services = ctx.pluginConfig.services || {};
+        const serviceNames = Object.keys(services);
+        const isRegisteredService = serviceNames.includes(
+          this.environment.name
+        );
+
         // find entry point of this service
         let entryFile: string | undefined;
         for (const [_name, file] of Object.entries(bundle)) {
           if (file.type === "chunk") {
-            if (file.isEntry) {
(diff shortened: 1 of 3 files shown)


src/build/vite/plugin.ts
```diff
@@ -14,6 +13,7 @@ import * as rou3 from "rou3";
 import * as rou3Compiler from "rou3/compiler";
 import { resolveModulePath } from "exsolve";
 import { prettyPath } from "../../utils/fs";
+import path from "node:path";
```
pi0: We could use pathe (already a dep i guess vite also uses same) to make sure paths are always normalized

src/build/vite/plugin.ts
```diff
@@ -124,13 +126,50 @@ export async function nitro(
...
+      handler(_options, bundle) {
+        const { root } = this.environment.config;
+        // find entry point of this service
+        let entryFile: string | undefined;
+        for (const [_name, file] of Object.entries(bundle)) {
+          if (file.type === "chunk") {
+            if (file.isEntry) {
+              if (entryFile !== undefined) {
+                this.error(
+                  `Multiple entry points found for service "${this.environment.name}". Only one entry point is allowed.`
```
pi0, line 155: Maybe we only show warning if this.environment is server consumer (or more specifically a defined servicr/ssr)

pi0 (approved):
Thanks ❤️

PR #3495, opened 2025-07-18
fix(vite): don't write service outputs to disk, instead use VFS

schiller-manuel:


pi0 (approved):
Awesome, Sorry for late merge. Small strange change i noticed chunk names in output are different and we are not testing sourcemap generation will invesgtigate both latet 👍🏼

PR #3497, opened 2025-07-21
fix: plugins can break prerenderer

sotasan:
### 🔗 Linked issue

There's no issue, because it's a small bug, but still annoying.

### ❓ Type of change

### 📚 Description

This is a small annoying bug, when you use the prerender feature.

Normally when you don't specify a route and just tell it to crawl all the links, it works perfectly fine.

But I stumbled upon this, when I installed `nuxt-llms`. This plugin adds a prerender hook, which adds the `llms.txt` file, which upon itself is fine. But in this context, it breaks the application's already working prerenderer.

This PR fixes this issue and adds the index route, after the application has no defined routes and `crawlLinks` is turned on, but the hooked routes, which come from plugins most of the time, are only applied afterwards.

### 📝 Checklist

### DIFF
--- a/src/prerender/prerender.ts
+++ b/src/prerender/prerender.ts
@@ -46,17 +46,17 @@ export async function prerender(nitro: Nitro) {
     routes.add(route);
   }
 
+  // Crawl / at least if no routes are specified
+  if (routes.size === 0 && nitro.options.prerender.crawlLinks) {
+    routes.add("/");
+  }
+
   // Allow extending prerender routes
   await nitro.hooks.callHook("prerender:routes", routes);
 
   // Skip if no prerender routes specified
   if (routes.size === 0) {
-    // Crawl / at least if no routes are specified
-    if (nitro.options.prerender.crawlLinks) {
-      routes.add("/");
-    } else {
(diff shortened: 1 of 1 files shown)


pi0:
Would you able to please provide a minimal nitro reproduction?

sotasan:
Here is a reproducible example with the exact same problem I encountered and a patch, which is this PR:
https://github.com/sotasan/damaging-glass

Output with the plugin disabled:
image

Output with the plugin enabled:
image

Output with my patch applied:
image

pi0:
Thanks, but you are reporting a reproduction with Nuxt and `nuxt-llms`, which is using Nitro v2 and Nuxt-specific setup.

The proposed change in PR is a behavior change targeting Nitro v3 (unreleased).

We deliberately changed the order in https://github.com/nitrojs/nitro/pull/2348 to avoid adding `/` as default to solve https://github.com/nuxt/nuxt/pull/26694 (/cc @danielroe)

I would suggest opening an issue in the `nuxt` repo or `nuxt-llms` module repo for a fix 🙏🏼

PR #3500, opened 2025-07-22
fix(vercel): valid regex for o11y routes

pi0:
Reported in https://github.com/nuxt/nuxt/issues/32718

We have introduced o11y route hints in 2.12 (https://github.com/nitrojs/nitro/pull/3474), it should make sure to generate PCRE-compliant named capture groups (404 in case of reported issue is invalid since it shouldn't start with a digit)

This PR also uses a less conflicting anonymous group for `/**` routes.

### DIFF
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -377,11 +377,11 @@ function normalizeRouteSrc(route: string): string {
 }
 
 // Valid PCRE capture group name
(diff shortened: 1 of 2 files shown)


JenuelDev:
when will this be released?

pi0:
shortly you can use nightly channel or https://github.com/nuxt/nuxt/issues/32718#issuecomment-3101436117 as workaround i suggested.

PR #3511, opened 2025-07-30
refactor!: remove default `~` and `@` aliases

schiller-manuel:
> (@pi0) 
> We are removing default `~/` and `@/` aliases to make the default Nitro behavior less opinionated and less chance to conflict with user custom aliases.
> We might add them back later, either via a flag or as the lowest priority in resolution. 
> This PR in meanwhile, unblocks vite plugin adoption


pi0:
Thanks for PR. I think we need to update test fixtures i will check on it asap.

PR #3512, opened 2025-07-30
fix(vite): autoconfigure ssr service with predefined input

schiller-manuel:


pi0:
Can you please only include fix change? Diff is hard to follow. 

Splitting into multiple plugins makes sense but for start i was thinking to keep plugins simple to read.

PR #3516, opened 2025-07-30
feat(dev): support `.env.local`

### LINKED ISSUE #3486: Support multiple dotenv files (`.env.local`)
### Describe the feature

`nitro dev` currently only loads env variables from the `.env` file. This was following [dotenv strict recommendation](https://github.com/motdotla/dotenv/tree/v16.4.0?tab=readme-ov-file#should-i-have-multiple-env-files) to prefer `.env` (I noticed docs [updated](https://github.com/motdotla/dotenv/commit/a44cb3dd00abdd77f3268be111cf9dd68e9a8e91) recently.)

Other tools, such as [`vite dev`](https://vite.dev/guide/env-and-mode.html#env-files) which we are [defaulting to](https://github.com/nitrojs/nitro/discussions/3460) in v3 also support [`.env.local`](https://vite.dev/guide/env-and-mode.html#env-files).

I think it is reasonable that we now allow `.env.local` also for `nitro dev`.

### Additional information

gioboa:
> [!NOTE]
We add `.env` and `.env.local` as default in case of fallback and only in dev.

Close #3486
 

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/config/loader.ts
+++ b/src/config/loader.ts
@@ -87,7 +87,7 @@ async function _loadUserConfig(
 
   const _dotenv =
     opts.dotenv ??
-    (configOverrides.dev && { cwd: "???", fileName: [".env", ".env.local"] });
+    (configOverrides.dev && { fileName: [".env", ".env.local"] });
   const loadedConfig = await (
     opts.watch
       ? watchConfig<NitroConfig & { _meta?: NitroPresetMeta }>
@@ -95,6 +95,7 @@ async function _loadUserConfig(
   )({
     name: "nitro",
     cwd: configOverrides.rootDir,
+    // @ts-expect-error
     dotenv: _dotenv,
     extend: { extendKey: ["extends", "preset"] },
     defaults: NitroDefaults,
--- a/src/config/loader.ts
+++ b/src/config/loader.ts
@@ -85,6 +85,7 @@ async function _loadUserConfig(
   // prettier-ignore
   let preset: string | undefined = (configOverrides.preset as string) || process.env.NITRO_PRESET || process.env.SERVER_PRESET
 
+  const _dotenv = opts.dotenv ?? configOverrides.dev;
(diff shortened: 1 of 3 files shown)


src/config/loader.ts
```diff
@@ -85,14 +85,18 @@ async function _loadUserConfig(
   // prettier-ignore
   let preset: string | undefined = (configOverrides.preset as string) || process.env.NITRO_PRESET || process.env.SERVER_PRESET
 
+  const _dotenv = opts.dotenv ?? configOverrides.dev;
```
pi0: Let's keep it little simpler. We only need to default `.env, .env.local` in case of fallback and only in dev (after ??)

gioboa: ok, to create a DotenvOptions object I need a value for cwd though.
I'm not sure that process.cwd() is ok for all the scenarios.

src/config/loader.ts
```diff
@@ -85,14 +85,17 @@ async function _loadUserConfig(
   // prettier-ignore
   let preset: string | undefined = (configOverrides.preset as string) || process.env.NITRO_PRESET || process.env.SERVER_PRESET
 
+  const _dotenv =
+    opts.dotenv ??
+    (configOverrides.dev && { cwd: "???", fileName: [".env", ".env.local"] });
```
gioboa:
```
cwd: "???"
```
This is my concern, what value should we use?

pi0: Oh that's a type issue 🤦🏼‍♂️ please just add a `@ts-expect-error` above it and omit cwd

gioboa: Done

pi0 (approved):
Thanks ❤️

gioboa (commented):
It's always a pleasure to collaborate with you @pi0 👏

PR #3517, opened 2025-07-31
fix(cloudflare-module): output _headers file to correct directory

adamdehaven:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/issues/3441

### ❓ Type of change

### 📚 Description

In https://github.com/nitrojs/nitro/pull/3442 we output the `_headers` file for applying `routeRules` when utilizing the `cloudflare-module` build preset.

According to [Cloudflare's documentation](https://developers.cloudflare.com/workers/static-assets/headers/#custom-headers), we should output to the static assets directory (essentially `.output/public/`) rather than the current implementation that writes the file to the root of the `.output/` directory.

**The current output path does not actually apply the headers when deployed.**

When I put up https://github.com/nitrojs/nitro/pull/3442 I had to do a manual upload of the `_headers` file for verification but mistakenly dropped it into the _correct_ directory.

I can verify with the change in the PR outputting to the `.output/public/_headers` path and uploading a worker version to Cloudflare does properly apply the route rule in this example:

```ts

  compatibilityDate: "latest",
  srcDir: "server",
  preset: 'cloudflare-module',
  routeRules: {
    '/_assets/**': { headers: { 'Cache-Control': 'public, max-age=2592000, immutable' } }, /* 30 days */
  }
});
```

cache-control verification

@pi0 this will need to be ported forward to `v3` as well.

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare/preset.ts
+++ b/src/presets/cloudflare/preset.ts
@@ -60,7 +60,7 @@ const cloudflarePages = defineNitroPreset(
       async compiled(nitro: Nitro) {
         await writeWranglerConfig(nitro, "pages");
         await writeCFRoutes(nitro);
-        await writeCFHeaders(nitro, false);
+        await writeCFHeaders(nitro, "output");
         await writeCFPagesRedirects(nitro);
       },
     },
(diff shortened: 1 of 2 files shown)


adamdehaven:
@pi0 is there anything else blocking here?
The output looks the same with `deployConfig: true` and `deployConfig: false`; according to the docs, the path should be valid 👍🏼

pi0 (approved):
Thanks!

PR #3519, opened 2025-07-31
fix(vite): don't force client manifest

schiller-manuel:
it's not necessary for the nitro vite plugin to function


pi0:
Can you share more context? we could allow explicit false value for example to opt out.

schiller-manuel:
we don't need the manifest to be written for nitro since we analyze the bundle in the generate bundle hook 
do you need the manifest for anything else?

pi0:
Yes for building simple SSR rendered apps it is required.

Can we use ??= true perhaps to avoid force overriding?

Update: remembered we are now collecting `ctx._manifest` still

pi0 (approved):

Issue #3501, opened 2025-07-22
"Vary: Accept-Encoding" header is added on every response despite nitro not doing any runtime compression


paro-paro:
### Environment

node: v20.19.1
nitro: 2.12.4

### Reproduction

[reproduction](https://stackblitz.com/edit/github-ba3yqpkf)

Build the app, preview it, curl the index route and verify that no compression is done by nitro but the `Vary: Accept-Encoding` response header is added.

### Describe the bug

I have a nitro server behind traefik (reverse proxy) that handles gzip compression for me since, as far as I know, nitro does not support runtime response compression (only for static assets) #1007 

Everything works fine but I have noticed that the `Vary: Accept-Encoding` is duplicated on every route response from my nitro server.

Maybe I am wrong, but I believe this header is used by servers to tell caches and proxies that the response changes based on the client's Accept-Encoding header, but nitro will always send the same response (same Content-Length) regardless the client's request. 

Why is it adding the header then? Since traefik gzip compression middleware is actually doing the work and adding the header, I get the duplicate. 

Is this expected behaviour? Is this a traefik issue?

Thanks in advanced

### Logs

```sh

```

pi0:
Thanks for repro. You are right, we should avoid using [EncodingMap](https://github.com/nitrojs/nitro/blob/v2/src/runtime/internal/static.ts#L28) unless `compressPublicAssets` config is enabled. 

Would you be able to work on a PR to fix it? (we need to pass a static flag in rollup config to conditionally enable it)

paro-paro:
Yes, I would love to!

PR #3521, opened 2025-07-31
fix(static): only add vary response header when serving public assets if compression is enabled

paro-paro:
### 🔗 Linked issue

#3501

### ❓ Type of change

### 📚 Description

This PR aims to fix the current implementation of the `vary` response header from the nitro server.

Right now, every request that includes the `accept-encoding` header will get back a response with `vary: accept-encoding` despite nitro not serving different content. This should only happen when nitro serves different compressed (on build time) versions of public assets (when the `compressPublicAssets` config option is enabled).

This PR fixes that and also handles this other related issue #3077, where the vary header is wrongly omitted if the encoding request header is not sent (or if set to a different value than the ones supported by EncodingMap) for compressed assets.

### 📝 Checklist

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -222,6 +222,15 @@ export const getRollupConfig = (nitro: Nitro): RollupConfig => {
     DEBUG: nitro.options.dev,
   };
 
+  const getCompressPublicAssetsOption = () => {
+    const option = nitro.options.compressPublicAssets;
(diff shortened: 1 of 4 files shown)


mapirellidontouch:
i'm waiting for the merge of this PR too. 
It appears that Firebase's automatic HTML compression is failing because of the headers configured by Nitro in the response.

PR #3527, opened 2025-08-08
fix(dev): allow access `/_vfs` endpoint via unix socket

huang-julien:
### 🔗 Linked issue

fix https://github.com/nuxt/nuxt/issues/32894

### ❓ Type of change

### 📚 Description

Hey :wave: 

This PR allows proxied request to reach the VFS endpoint. The reason is that nuxt/cli is using a proxy in dev mode 

https://github.com/nuxt/cli/blob/6b80c59a310bde67256d946e67ffad158924ac41/packages/nuxi/src/dev/index.ts#L84C1-L86C76

### 📝 Checklist

### DIFF
--- a/src/core/dev-server/vfs.ts
+++ b/src/core/dev-server/vfs.ts
@@ -3,7 +3,7 @@ import type { Nitro } from "nitropack/types";
 
 export function createVFSHandler(nitro: Nitro) {
   return eventHandler(async (event) => {
-    const ip = getRequestIP(event, { xForwardedFor: false });
+    const ip = getRequestIP(event, { xForwardedFor: true });
     const isLocalRequest = ip && /^::1$|^127\.\d+\.\d+\.\d+$/.test(ip);
     if (!isLocalRequest) {
       throw createError({


src/core/dev-server/vfs.ts
```diff
@@ -3,7 +3,7 @@ import type { Nitro } from "nitropack/types";
 
 export function createVFSHandler(nitro: Nitro) {
   return eventHandler(async (event) => {
-    const ip = getRequestIP(event, { xForwardedFor: false });
+    const ip = getRequestIP(event, { xForwardedFor: true });
```
pi0, line 6: A bad actor can easily bypass restrictions over the network by passing `x-forwarded-for` header in their requests to access the filesystem. This is why this option is opt-in to enable only when requests are trusted (behind a CDN or Reverse proxy) -- unix sockets could be considered truested too.

I suggest 2 alternatives to move forward:
- Nuxt CLI can set `event.context.clientAddress` ([getRequestIP](https://github.com/h3js/h3/blob/ec77d6bc148e4ff7629ba56577697055cc0fcf2e/src/utils/request.ts#L386)) in socket mode
- We somehow detect event request is in socket mode to fix here

pi0 (commented):
we need alternative method

src/core/dev-server/vfs.ts
huang-julien: :+1: I think we can try to fix that in nuxt/cli first

huang-julien:
thanks ! :heart:

PR #3531, opened 2025-08-09
fix(vite): pass on request props in dispatcher

schiller-manuel:


src/runtime/internal/vite/dispatcher.mjs
```diff
@@ -8,5 +8,16 @@ export default defineHandler(async (event) => {
   }
   return fetch(new URL(`/${m.params?._ || ""}${event.url.search}`, event.url), {
     viteEnv: m.data.service,
+    method: event.req.method,
+    headers: event.req.headers,
```
pi0, line 13: Note for later: We might need to clone headers properly (HOST in particular should be overriden and we need to add x-forwarded headers)

pi0:
```suggestion
    // TODO: Clone headers or full RequestInit
    headers: event.req.headers,
```

pi0 (approved):
Thnx! (added a note for future refactor but it will likely be included in my bigger refactor on dispatcher/service routing)

Issue #3535, opened 2025-08-13
OpenAPI auto generated path params not merged correctly


amandesai01:
### Environment

Nitro `2.12.4`
Node `v22.13.1`

### Reproduction

Step 1:
use `definePageMeta` in any route.

Step 2:
make sure that `openAPI` object contains `parameters` array, even if empty.
`/api/{projectId}/index.get.ts`
```js
openAPI: {
    description: 'Get a specific project by ID',
    tags: ['projects'],
    summary: 'Get project by ID',
    operationId: 'getProjectById',
    parameters: [],
  }
```

Step 3:
Open swagger docs and notice that path variable is **_not_** auto-populated.

### Describe the bug

Nitro has this feature of auto detecting API path variables and populating them as path parameters in Open API spec. However, this feature is not working as intended.

### Additional context

This can be easily fixed. Just use DEFU!!!

### Logs

```sh

```

amandesai01:
I will raise a PR for this.

amandesai01:
Here is what I am planning to do:

1. Fix this feature (of course)
2. Do tests exist for this? If yes, planning to cover this case as well.
3. Write better docs for everything OpenAPI.

PR #3536, opened 2025-08-13
fix(openapi): recursively merge openapi paramteres

amandesai01:
### 🔗 Linked issue
https://github.com/nitrojs/nitro/issues/3535

### ❓ Type of change

📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
🐞 Bug fix (a non-breaking change that fixes an issue)
⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/routes/openapi.ts
+++ b/src/runtime/internal/routes/openapi.ts
@@ -68,10 +68,11 @@ function getHandlersMeta(): {
   let globals: OpenAPIGlobals = {};
 
   for (const h of handlersMeta) {
-    const { route, parameters } = normalizeRoute(h.route || "");
+    const { $global, ...openAPI } = h.meta?.openAPI || {};
+    const { route, parameters: _parameters } = normalizeRoute(h.route || "", openAPI);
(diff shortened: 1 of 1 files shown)


amandesai01:
Hi team, can we get this in? looks harmless to me. Anything I can do to make this quick?

pi0:
Can you please add it to be covered in test/fixture?

okydk:
Guess this will only be fixed in v3?

pi0x:
Rebased this on current `main` (applies cleanly) and tested it against a route with a path param.

It does not actually fix #3535:

- `defu(array, array)` returns `{}`, not a merged array — defu only merges plain objects at the top level.
- `...openAPI` is still spread *after* `parameters`, so a user-defined `parameters` array overwrites the merged value anyway.
- Routes with no user `parameters` now emit `"parameters": {}`, which is invalid OpenAPI.

Fix: drop `defu` here, merge the two arrays by hand (user entry wins when `name` + `in` match) and set `parameters` *after* the spread.

Also note #4560 rewrites the same file, so these two will conflict.

PR #3539, opened 2025-08-19
fix(vercel): workaround for isr route rule with `passQuery: true`

pi0:
Normally, with ISR matched routes, there is an `x-now-route-matcher` header that contains the full URL. And Nitro vercel entry uses this header value to restore the original URL, but when `passQuery` is set, this header won't exist anymore, and the path is invalid (ISR function name) (example deploy). We cannot always assume `url` query param is for ISR functions (it can be an actual query in normal routes!)

This PR is a proposed fix to only use `url` param as ISR url if route starts with `/__fallback` (default function name), but still it can overlap with a real `__fallback` route and break websites if for any reason they are using it.

### DIFF
--- a/src/presets/vercel/runtime/vercel.ts
+++ b/src/presets/vercel/runtime/vercel.ts
@@ -2,18 +2,26 @@ import "#nitro-internal-pollyfills";
 import { useNitroApp } from "nitropack/runtime";
 
 import { type NodeListener, toNodeListener } from "h3";
-import { parseQuery } from "ufo";
+import { parseQuery, withQuery } from "ufo";
 
 const nitroApp = useNitroApp();
 
 const handler = toNodeListener(nitroApp.h3App);
 
 const listener: NodeListener = function (req, res) {
-  const query = req.headers["x-now-route-matches"] as string;
-  if (query) {
-    const { url } = parseQuery(query);
+  const isrRoute = req.headers["x-now-route-matches"] as string;
+  if (isrRoute) {
+    const { url } = parseQuery(isrRoute);
     if (url) {
       req.url = url as string;
+    }
+  } else if (req.url?.startsWith("/__fallback")) {
+    // Workaround for ISR functions with passQuery: true
+    // /__fallback--api-weather?url=%2Fapi%2Fweather%2Famsterdam&units=123"
+    const urlQueryIndex = req.url.indexOf("?url=");
+    if (urlQueryIndex !== -1) {
+      const { url, ...params } = parseQuery(req.url.slice(urlQueryIndex));
+      req.url = withQuery((url as string) || "/", params);
     }
   }
   return handler(req, res);


mkucmus:
hey @pi0, do you plan to release the fix in upcoming version maybe?

fmoessle:
Would love to see this merged. Using
```
  "resolutions": {
    "nitropack": "https://pkg.pr.new/nitrojs/nitro/nitropack@3539"
  }
```
works but is not ideal.

fmoessle:
The problem I face with https://pkg.pr.new/nitrojs/nitro/nitropack@3539: Building and starting the application with node-presets causes an issue with the public folder resolution in .output

Reproduction: https://github.com/fmoessle/nitro-vercel-isr/tree/test-3539-fix

```
pnpm build
pnpm preview
```

results in errors like this:
```
request error] [unhandled] [GET] http://[::]:3000/favicon.ico
 H3Error: ENOENT: no such file or directory, open 'XXX/florian/projects/test/nitro-vercel-isr/.output/server/chunks/public/favicon.ico'
    at async open (node:internal/fs/promises:634:25)
    ... 2 lines matching cause stack trace ...
    at async Server.toNodeHandle (file:///XXX/.output/server/chunks/nitro/nitro.mjs:1917:7) {
  cause: Error: ENOENT: no such file or directory, open 'XXX/.output/server/chunks/public/favicon.ico'
      at async open (node:internal/fs/promises:634:25)
      at async Object.readFile (node:internal/fs/promises:1238:14)
      at async Object.handler (file:///XXX/.output/server/chunks/nitro/nitro.mjs:1646:19)
      at async Server.toNodeHandle (file:///XXX/.output/server/chunks/nitro/nitro.mjs:1917:7) {
    errno: -2,
    code: 'ENOENT',
    syscall: 'open',
    path: 'XXX/.output/server/chunks/public/favicon.ico'
  },
  statusCode: 500,
  fatal: false,
  unhandled: true,
  statusMessage: undefined,
  data: undefined
}
```

There is no **public** folder in .output/server/chunks

pi0:
@coderabbitai review

pi0:
@fmoessle I couldn't reproduce error on your branch... This PR is also only affecting vercel preset not sure how can it break node-server. If you continue to having them on nightly or next release please feel free to raise an issue with more steps for reproduction.

ps-20x:
> Would love to see this merged. Using
> 
> ```
>   "resolutions": {
>     "nitropack": "https://pkg.pr.new/nitrojs/nitro/nitropack@3539"
>   }
> ```
> 
> works but is not ideal.

As this branch was now merged and deleted and as far as I can see there is no new release in nitro v2. What's the proposed way now? Should it now target the commit?

pi0:
You can use nightly channel: https://nitro.build/guide/nightly

PR #3544, opened 2025-08-24
docs(websocket): provide async await

iiio2:
I think here we need to provide async await in the defineEventHandler function. 
Also in the doc, for the chat demo Nitro Websocket API this example is given (https://nuxt-chat.pi0.io/), which does not work now. 
Also the fetch example link (https://raw.githubusercontent.com/unjs/crossws/main/examples/h3/public/index.html) shows `404: Not Found`.


pi0:
Thanks for PR but it doesn't makes a different. PS this code is synced from `examples/websocket/routes/index.ts` should not directly modify markdown for automd sections.

iiio2:
Ok @pi0 . Thanks.

PR #3547, opened 2025-08-25
fix(bun): update server host configuration and log URL

### LINKED ISSUE #3546: NITRO_HOST not working when using preset: "bun" in Nuxt
### Environment

bun: 1.2.20
nitro: 2.12.4
nuxt: 4.0.3
nodejs: v24.4.0

### Reproduction

https://github.com/nitrojs/nitro/blob/v2/src/presets/bun/runtime/bun.ts#L14-L41

### Describe the bug

When using Nuxt with Nitro and setting:

```js
preset: "bun"
```

I noticed that `NITRO_HOST` has no effect — only `NITRO_PORT` works.

In the built `index.mjs`, the Bun server is started like this:

```js
me = Bun.serve({
    port: t.env.NITRO_PORT || t.env.PORT || 3e3,
    websocket: de.websocket,
    async fetch(t, e) {
        if ("websocket" === t.headers.get("upgrade"))
            return de.handleUpgrade(t, e);
        const s = new URL(t.url);
        let c;
        if (t.body) c = await t.arrayBuffer();
        return ue.localFetch(s.pathname + s.search, {
            host: s.hostname,
            protocol: s.protocol,
            headers: t.headers,
            method: t.method,
            redirect: t.redirect,
            body: c
        });
    }
});
```

As you can see, only the `port` option is read from the environment.
`NITRO_HOST` is never used, so there’s no way to set a custom host — it always binds to the default.

How can I enable or set a custom host (like 0.0.0.0 or 127.0.0.1) when using the Bun preset? Is there a supported way or workaround?

### Additional context

https://github.com/nitrojs/nitro/blob/v2/src/presets/bun/runtime/bun.ts#L14-L41

### Logs

```sh

```

wling-art:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/issues/3546

### ❓ Type of change

### 📚 Description

Resolves #3546 

### 📝 Checklist

### DIFF
--- a/src/presets/bun/runtime/bun.ts
+++ b/src/presets/bun/runtime/bun.ts
@@ -13,6 +13,7 @@ const ws = import.meta._websocket
 // @ts-expect-error
 const server = Bun.serve({
   port: process.env.NITRO_PORT || process.env.PORT || 3000,
+  host: process.env.NITRO_HOST || process.env.HOST || "localhost",
   websocket: import.meta._websocket ? ws!.websocket : (undefined as any),
   async fetch(req: Request, server: any) {
     // https://crossws.unjs.io/adapters/bun
@@ -38,7 +39,7 @@ const server = Bun.serve({
   },
 });
 
-console.log(`Listening on http://localhost:${server.port}...`);
+console.log(`Listening on ${server.url}...`);
 
 // Scheduled tasks
 if (import.meta._tasks) {


src/presets/bun/runtime/bun.ts
```diff
@@ -13,6 +13,7 @@ const ws = import.meta._websocket
 // @ts-expect-error
 const server = Bun.serve({
   port: process.env.NITRO_PORT || process.env.PORT || 3000,
+  host: process.env.NITRO_HOST || process.env.HOST || "localhost",
```
pi0: This is a breaking change. Bun's default behavior is listening on all interfaces (while it _says_ it is only listening on localhost with `server.url`!)

qgustavor:
Shouldn't it be `hostname` instead of `host` as [per the documentation](https://bun.sh/docs/api/http)?

pi0:
standard env variable is `HOST` (but config name is `hostname`)

qgustavor:
Just to be sure, the `Bun.serve` config is `hostname` but it was set to `host`, thus it's not respecting nor `HOST` nor `NITRO_HOST`, thus, it is a bug, correct? At the moment I'm patching the server code to workaround that.

pi0:
Ah sorry you are right! This is currently wrong. Please feel free to drop a PR to change.

PR #3554, opened 2025-08-27
feat: add sevalla preset

olawanlejoel:
### ❓ Type of change

### 📚 Description

[Sevalla](https://sevalla.com/) is a modern cloud platform for hosting applications, databases, object storage and static sites, offering a simple and scalable deployment experience.

This PR introduces a new preset named `sevalla`, allowing users to easily deploy their Nitro apps to Sevalla with Node server support. The preset ensures compatibility and provides a straightforward deployment workflow.

Additionally, the documentation has been updated with a dedicated Sevalla deployment guide, making it easy for users to get started and configure their environment.

### 📝 Checklist


docs/2.deploy/20.providers/sevalla.md
```diff
@@ -0,0 +1,44 @@
...
+**Preset:** `sevalla`
+
+:read-more{title="sevalla.com" to="https://sevalla.com"}
+
+## Setup
+
+1. [Create a new app](https://docs.sevalla.com/applications/get-started/add-an-application) on Sevalla and connect it to your Git repository.
+2. Set the **application name**, **region**, and **pod size** (you can start with 0.5 CPU / 1GB RAM).  
+3. Click **Create**, but skip the deploy step for now.  
+4. In your Sevalla application dashboard, click **Environment variables** on the sidebar and set `NITRO_PRESET` to `sevalla`.
```
pi0, line 14: Could svella CI somehow set this env? (or if it has a known env variable, we can add it to [std-env](https://github.com/unjs/std-env/blob/main/src/providers.ts) for zero config

docs/2.deploy/20.providers/sevalla.md
```diff
@@ -0,0 +1,44 @@
...
+    :::code-group
+    ```json [package.json]
+    {
+      "scripts": {
+        "build": "nitro build",  // or `nuxt build` if using Nuxt
+        "start": "node .output/server/index.mjs"
+      }
+    }
+    ```
+    :::
```
pi0: This step seems unnecessary, we have these commands in the default starter templates

docs/2.deploy/20.providers/sevalla.md
```diff
@@ -0,0 +1,44 @@
...
+2. Set the **application name**, **region**, and **pod size** (you can start with 0.5 CPU / 1GB RAM).  
+3. Click **Create**, but skip the deploy step for now.  
+4. In your Sevalla application dashboard, click **Environment variables** on the sidebar and set `NITRO_PRESET` to `sevalla`.
+5. Ensure your project specifies a supported Node version in `package.json`:
+
+    :::code-group
+    ```json [package.json]
+    {
+      "engines": {
+        "node": "22.x"
```
pi0, line 21: Is this a hard requirement to deploy on svella or a recommendation? (If recommendation only, I suggest using a `> [!TIP]` note below and reducing initial steps)

src/presets/sevalla/preset.ts
```diff
@@ -0,0 +1,14 @@
+import { defineNitroPreset } from "../_utils/preset";
+
+const sevalla = defineNitroPreset(
+  {
+    extends: "node-server",
+    serveStatic: true,
```
pi0: Can't svella CDN serve static assets itself?

pi0:
@olawanlejoel feel free to join [Nitro discord](https://discord.nitro.build) if you like to followup over chat.

src/presets/sevalla/preset.ts
olawanlejoel: Thanks for the feedback! You're absolutely right. Sevalla offers a free Cloudflare-powered CDN that can serve static assets efficiently.

I've updated the preset to remove `serveStatic: true` and added a "Static Assets" section in the documentation explaining how users can enable Sevalla's CDN for optimal performance. I think this approach is cleaner and leverages Sevalla's infrastructure rather than having the Node server handle static files.

docs/2.deploy/20.providers/sevalla.md
olawanlejoel: True. I've removed it

docs/2.deploy/20.providers/sevalla.md
olawanlejoel: It's actually a hard requirement. Nixpacks is our default builder, and it uses Node 18, while modern Nitro apps require Node 20+.

Without specifying the Node version in `package.json`, deployments fail with an oxc-parser error related to optional dependencies. Setting "node": "22.x" in the engines field is necessary to ensure Nixpacks uses the correct Node version for the build.

I've updated the documentation to clarify this requirement and explain why it's needed.

docs/2.deploy/20.providers/sevalla.md
olawanlejoel: I am trying to know how we can go about this internally. But at the moment this works this way. Can we maybe merge this way and then later the team can come back to this @pi0 .

olawanlejoel:
Hi @pi0, I attended to the comments you raised

pi0x:
I rebased this onto latest `main` locally and regenerated `_all.gen.ts` / `_types.gen.ts` with `pnpm gen-presets`. I could not push, so the branch here is unchanged.

Two small things needed fixing after the rebase:

- The import needs the `.ts` extension (`../_utils/preset.ts`), which is now required.
- `url: import.meta.url` is no longer part of the preset meta type and fails typecheck. I dropped it.

With those, typecheck and lint pass. The preset itself looks right: `extends: "node-server"` already brings `serveStatic: true`, so the CDN has something to cache from.

Still open from @pi0's earlier review: no zero-config detection. If Sevalla sets a known build env variable, adding it to `std-env` would let Nitro pick the preset automatically.

Note: #4025, #3970 and #4554 also add presets and touch the same generated files, so whoever lands first forces the rest to rebase.

PR #3556, opened 2025-08-28
build: remove `@netlify/functions` dependency

pi0:
Currently we have a direct dependency on [`@netlify/functions`](https://npmjs.com/@netlify/functions) which is only used for internal types and `netlify-builder` preset (which is [deprecated](https://nitro.build/deploy/providers/netlify#on-demand-builders)

To avoid chances of breaking changes, small part of builder logic is inlined. /cc @nitrojs/netlify 

This change will reduce [~76MB](https://packagephobia.com/result?p=@netlify/functions) of install size.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
   },
   "dependencies": {
     "@cloudflare/kv-asset-handler": "^0.4.0",
-    "@netlify/functions": "^3.1.10",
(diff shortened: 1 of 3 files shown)


leefreemanxyz:
Thanks Pooya, this is so dope getting rid of all these dependencies.

PR #3559, opened 2025-08-29
chore: use `node:lts` image in devcontainer

iiio2:
i think in devcontainer we can use `node:lts` image, since node:lts has some support like typescript.


iiio2:
Thanks @pi0 . :blush:

PR #3564, opened 2025-09-07
docs(routing): update error handling description

azrikahar:
### ❓ Type of change

### 📚 Description

> If the route starts with `/api/` (either because it is placed in `api/` or `routes/api/`) the default will change to `application/json` and a JSON object will be sent.

I believe the above excerpt is no longer applicable due to the changes from #3002. 

It was true when the response still uses a util function called `isJsonRequest` and it has a specific check for `event.path.startsWith("/api/")` to make sure routes in that path will return JSON:

https://github.com/nitrojs/nitro/blob/e8d609908e9758094d3639b83fe2956ddf872fa8/src/runtime/internal/error/utils.ts#L11-L24

but as it was removed and #3002 specifically split the error handler into dev and prod handlers, I've opted to update the docs based on the new error handling changes.

That being said, should the following existing sentence: 

> This behaviour can be overridden by some request properties (e.g.: `Accept` or `User-Agent` headers).

be removed since it might be redundant now? or at least the `User-Agent` part might not really work, assuming if I didn't misunderstand the new error handling logic of course.

### 📝 Checklist

### DIFF
--- a/docs/1.guide/2.routing.md
+++ b/docs/1.guide/2.routing.md
@@ -271,7 +271,7 @@ export default defineEventHandler((event) => {
 
 You can use the [utilities available in H3](https://h3.dev/guide/basics/error) to handle errors in both routes and middlewares.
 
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks ❤️

PR #3566, opened 2025-09-09
chore(playground): add `"type": "module"` to `package.json`

iiio2:
though it is in playground, i had read `"type": "module"` has some performance related issue if we do not explicitly mention it. that is why I've thought we can add it here. (https://nodejs.org/en/blog/release/v20.19.0#module-syntax-detection-is-now-enabled-by-default)


iiio2:
Thanks @pi0 .

PR #3568, opened 2025-09-10
chore: mention `strictRequires` to `true`

iiio2:
Actually in the comment it is said to make it as `true` in v3. Since we are now on v3 beta, that's why I thought may be we can change it now.


pi0:
Appreciate PR but i have to keep TODO until properly investigating change effects (it also needs to be in sync with rollup)

iiio2:
I see. Thanks @pi0 .

PR #3572, opened 2025-09-15
refactor(aws-amplify, azure, vercel): update node to 20, 22

iiio2:
Actually there are some runtime versions which are changed. For example, in aws-amplify now only supported node.js runtime is 20 and 22. (https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported). 
In azure, 18, 20, 22 (https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=typescript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#supported-versions). We can go below 18 but it requires specific file structure. 
In vercel, 20, 22. (https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#default-and-available-versions)
That is why I thought may be we can update something in the codebase as well.

### DIFF
--- a/src/presets/azure/utils.ts
+++ b/src/presets/azure/utils.ts
@@ -9,7 +9,7 @@ export async function writeSWARoutes(nitro: Nitro) {
   };
 
   // https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=typescript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#supported-versions
-  const supportedNodeVersions = new Set(["18", "20", "22"]);
+  const supportedNodeVersions = new Set(["20", "22"]);
   let nodeVersion = "18";
   try {
     const currentNodeVersion = JSON.parse(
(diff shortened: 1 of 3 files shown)


src/presets/azure/utils.ts
```diff
@@ -10,7 +9,7 @@ export async function writeSWARoutes(nitro: Nitro) {
   };
 
   // https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=typescript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#supported-versions
-  const supportedNodeVersions = new Set(["16", "18", "20"]);
+  const supportedNodeVersions = new Set(["18", "20", "22"]);
```
pi0:
```suggestion
  const supportedNodeVersions = new Set(["20", "22"]);
```

pi0:
We should sync back to v2 as well (preserving older versions)

iiio2:
Thanks @pi. :blush:

snoozbuster:
@iiio2 do you know if this was ever backported to v2? I assume this is just a types thing but it would be nice to have this on the older version.

PR #3580, opened 2025-09-19
feat: add support for streaming cache responses

### LINKED ISSUE #1894: Caching fails when returning a Buffer (PNG image)

lukekarrys:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/issues/1894

### ❓ Type of change

### 📚 Description

This is one possible way to support binary responses in `cachedEventHandler`. When `stream: true` is set on the cache options, the response body will not be read at all and will instead be teed with one stream sent to `setItemRaw()` and the other returned as the response.

This can be useful for text responses as well when proxying (such as large JSON data) because the response does not need to be read into memory. It can be streamed directly to the client while writing to the cache does not block the response.

When using the `stream: true` option it will require special handling of the stream in the `setItemRaw` method of any storage adapter. So I think this is an advanced use-case and doesn't automatically fix #1894.

I experimented with another fix that would use `res.arrayBuffer()` and convert it to an `Uint8Array` so the response could still be serialized as JSON. But for my use-case (tar files that were potentially 100s of MB) it was too slow. I'd be happy to amend this PR to support that as well.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -122,12 +122,32 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
           if (opts.maxAge && !opts.swr /* TODO: respect staleMaxAge */) {
             setOpts = { ttl: opts.maxAge };
           }
-          const promise = useStorage()
-            .setItem(cacheKey, entry, setOpts)
-            .catch((error) => {
+          let cacheEntry = entry;
+          let setItem = useStorage().setItem;
+          if (
+            opts.stream &&
+            entry.value &&
+            typeof entry.value === "object" &&
+            "body" in entry.value &&
+            entry.value.body instanceof ReadableStream
+          ) {
+            const [cacheStream, resStream] = entry.value.body.tee();
+            cacheEntry = {
+              ...entry,
+              value: {
+                ...entry.value,
+                body: cacheStream,
+              },
+            };
+            entry.value.body = resStream;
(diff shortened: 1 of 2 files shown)


pi0:
I appreciate the PR dear @lukekarrys it had been now landed in in [ocache](https://ocache.unjs.io/docs/handler#streaming-the-first-response) available in Nitro v3 nightlies.

Please drop me a DM if had any issues if had a chance to try!

PR #3589, opened 2025-09-24
fix(azure): preserve query string in adapter handler

### LINKED ISSUE #3588: Azure Functions adapter does not forward query string to localCall
### Environment

- Nitro version: v2
- Nuxt version: 4.1.1
- Target: azure_functions (via NITRO_PRESET=azure_functions nuxi build)
- Hosting: Azure Functions (Linux consumption plan)

### Reproduction

Building with `NITRO_PRESET=azure_functions` and deploying to Azure Functions.  
Any Nuxt server route (`defineEventHandler`) that calls `getQuery(event)` will always return `{}` when using query parameters in the request.

Example request:

```
GET /api/v1/org/123/site/456/events?scenario_id_applied=abc&event_type=all
```

Server code:

```ts
defineEventHandler((event) => {
  return getQuery(event);
});
```

Response is `{}` instead of the expected object.

### Describe the bug

The generated Azure Functions entry (`.output/server/chunks/nitro/nitro.mjs`) builds the request URL like this:

```js
const url = "/" + (req.params.url || "");
```

This only forwards the path from `{*url}` and discards the query string.  
However, `req.url` and `req.originalUrl` in Azure already contain the proper query parameters.  

Because the query is never appended to the `localCall()`, Nitro cannot parse it and `getQuery(event)` always returns empty.

### Additional context

- This is an adapter bug specific to the `azure_functions` preset.
- Queries are visible in both `req.url` and `req.originalUrl`; only the preset drops them.
- A fix is trivial append of the query portion

### Logs

```sh
Excerpt from `context.log` in Azure Functions:


req.params.url api/v1/org/.../events
req.url https://.../events?scenario_id_applied=6ab9...&event_type=all
req.originalUrl https://.../events?scenario_id_applied=6ab9...&event_type=all
```

xMorthi:
### 🔗 Linked issue
- Resolves #3588 

### ❓ Type of change

### 📚 Description

The `azure_functions` preset discards query parameters when constructing the
URL passed to `nitroApp.localCall`. Currently it only uses:

```js
const url = "/" + (req.params.url || "");
```

As a result, any query string is lost (`getQuery(event)` always returned `{}`).

This PR updates the handler to preserve the query portion from `req.url`
(if present):

```js
let path = "/" + (req.params.url || "");
if (req.url && req.url.includes("?")) {
  path += req.url.slice(req.url.indexOf("?"));
}
```

### 📝 Checklist

### DIFF
--- a/src/presets/azure/runtime/azure-functions.ts
+++ b/src/presets/azure/runtime/azure-functions.ts
@@ -10,7 +10,11 @@ import type { HttpRequest, HttpResponse } from "@azure/functions";
 const nitroApp = useNitroApp();
 
 export async function handle(context: { res: HttpResponse }, req: HttpRequest) {
-  const url = "/" + (req.params.url || "");
+  let url = "/" + (req.params.url || "");
+
+  if (req.url && req.url.includes("?")) {
+    url += req.url.slice(req.url.indexOf("?"));
+  }
 
   const { body, status, statusText, headers } = await nitroApp.localCall({
     url,


src/presets/azure/runtime/azure-functions.ts
```diff
@@ -10,7 +10,11 @@ import type { HttpRequest, HttpResponse } from "@azure/functions";
 const nitroApp = useNitroApp();
 
 export async function handle(context: { res: HttpResponse }, req: HttpRequest) {
-  const url = "/" + (req.params.url || "");
+  let url = "/" + (req.params.url || "");
+
+  if (req.url && req.url.includes("?")) {
+    url += req.url.slice(req.url.indexOf("?"));
```
pi0, line 16: Wondering why we are not using `req.url` in the first place instead of `req.params.url`? (/cc @danielroe if you remember any of the context)

xMorthi: To my understanding, its because `req.url` is full urlls `https://domain...` while `req.params.url` actually starts from `api/`

pi0: That's actually a good thing! We can use `new URL(req.url)` one step + also provide valid `host` header not sure if it is right or not currently.

xMorthi: I'm already running a few apps on Azure and haven't encountered any issues with accessing headers on the Nitro side. We pass the entire `req.headers` directly.

pi0: I meant it can also use proper hostname without dependency on host header.

PR #3595, opened 2025-09-26
fix(vercel): always add `url` to src regex for `isr` routes

pi0:
Resolves https://github.com/nuxt/nuxt/issues/33316

In nitro [2.12.6](https://github.com/nitrojs/nitro/releases/tag/v2.12.6), we have introduced fixes for overlapping route rule functions + better names.

This revealed an undicovered issue where in past ISR route for `/path` would generate a function called `/path.func` and even without ISR path rewrites (`$url` param) it was working as name of function and path were same.

This PR fixes issue by 1) always normalizing rou3 patterns to PCRE and 2) Always wrapping regex with `$url` matcher.

### DIFF
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -207,7 +207,7 @@ function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
     ...rules
       .filter(([key, value]) => value.isr !== undefined && key !== "/")
       .map(([key, value]) => {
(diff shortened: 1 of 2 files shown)


pi0:
Fix is released in [nitropack@2.12.7](https://github.com/nitrojs/nitro/releases/tag/v2.12.7)

ps-20x:
I updated nitropack to version 2.12.7 (double checked in my package-lock.json) but for me the issue still persists.

pi0:
@ps-20x would you please open a new issue with reproduction so we can track and investigate 🙏🏼

PR #3606, opened 2025-10-03
ci: bump actions/setup-node version

gioboa:
- [X] 🧹 Chore (updates to the build process or auxiliary tools and libraries)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: npm i -g --force corepack && corepack enable
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with: { node-version: lts/*, cache: pnpm }
       - uses: oven-sh/setup-bun@v2
         if: ${{ matrix.os != 'windows-latest' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
```
pi0: Can you please also check others i think checkout is also now on v5. 

(we have to also fix renovate config, it should do it automatically for `main` + `v2` + `v3`)

gioboa: yep, you are right `actions/checkout@v5`
let me change it

gioboa: Changed

gioboa: is it `unjs/renovate-config` ?

pi0: Yes. We can extend it in https://github.com/nitrojs/nitro/blob/v3/renovate.json (+ https://github.com/nitrojs/nitro/blob/v2/renovate.json)

Example: https://github.com/nuxt/nuxt/blob/main/renovate.json#L8

gioboa: #3608 :+1:

PR #3609, opened 2025-10-03
fix(vite): enforce hooks as pre

yuusheng:
### 🔗 Linked issue
https://github.com/nitrojs/nitro/issues/3551

### ❓ Type of change

### 📚 Description
As Vite doesn't support importing a path that starts with "#" from a virtual file when the package.json imports field is set. We should pre the #nitro-internal-virtual handler to avoid Vite's inner plugin (vite:import-analysis) handling that first.

reproduction: https://stackblitz.com/edit/vitejs-vite-ckpgd5rk?file=package.json

### 📝 Checklist

### DIFF
--- a/src/build/vite/plugin.ts
+++ b/src/build/vite/plugin.ts
@@ -304,7 +304,7 @@ function nitroServicePlugin(ctx: NitroPluginContext): VitePlugin {
         }
 
         // Run rollup resolve hooks in dev (VFS support)
-        if (ctx.nitro?.options.dev && (id.startsWith("#") || id.startsWith("\0"))) {
+        if (id.startsWith("#") || id.startsWith("\0")) {
           for (const plugin of ctx.rollupConfig!.config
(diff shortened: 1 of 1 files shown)


src/build/vite/plugin.ts
```diff
@@ -390,8 +392,8 @@ function nitroServicePlugin(ctx: NitroPluginContext): VitePlugin {
           return `export const findService = ${rou3Compiler.compileRouterToString(router)};`;
         }
 
-        // Run rollup load hooks in dev (VFS support)
-        if (ctx.nitro?.options.dev) {
```
yuusheng, line 394: I think this shouldn't run only in dev mode.

pi0 (approved):
Thanks

src/build/vite/plugin.ts
pi0: This should normally work for production as we also provide these as rollup plugins (only in dev vite does not calls plugins)

PR #3628, opened 2025-10-07
refactor!: remove `rollup-plugin-visualizer` (for now)

pi0:
rollup-plugin-visualizer has a really big [dependency graph](https://npmgraph.js.org/?q=rollup-plugin-visualizer) and is incompatible with both vite and traced node modules. /cc @43081j

This PR for now removes `analyze` feature from nitro v3 but it can be fairly easily added back by users. 

---

Ideally i want to adopot [vite-bundle-analyzer](https://github.com/nonzzz/vite-bundle-analyzer) which  as rollup/rolldown/vite support and zero dep 🚀 but couldn't work it with rollup/rolldown (does nothing) and with vite, env api seems not supported (ssr and nitro envs are not displayed) -- once fixed we might add this as official integration. /cc @nonzzz @kricsleo

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,6 @@
     "ofetch": "^1.4.1",
     "ohash": "^2.0.11",
     "rollup": "^4.52.4",
-    "rollup-plugin-visualizer": "^6.0.4",
     "rou3": "^0.7.7",
     "scule": "^1.3.0",
     "source-map": "^0.7.6",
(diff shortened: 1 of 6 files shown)


kricsleo:
Would be happy to work on this recently 😆

pi0:
@kricsleo surely would be awesome if have time!

PR #3634, opened 2025-10-08
refactor!: lighter shared runtime config

pi0:
We previously had to make two versions of runtime config (one frozen global and one per request) since Cloudflare Workers didn't support global env access.

This PR simplifies `useRuntimeConfig()` to be fully tree-shakable and lazy inits its value on first call (globally) 

Also, env expansion simplified, we avoid `destr` on values which was causing issues like wrong number parses.

### DIFF
--- a/src/config/resolvers/imports.ts
+++ b/src/config/resolvers/imports.ts
@@ -62,7 +62,7 @@ function getNitroImportsPreset(): Preset[] {
       imports: ["useNitroApp"],
     },
     {
-      from: "nitro/runtime/internal/config",
+      from: "nitro/runtime/internal/runtime-config",
       imports: ["useRuntimeConfig"],
     },
     {


danielroe:
nice! 👌

test/unit/utils.env.test.ts
```diff
@@ -1,23 +1,18 @@
...
+import { applyEnv } from "../../src/runtime/internal/runtime-config";
 
 describe("env utils", () => {
   describe("applyEnv", () => {
     const tests = [
       {
-        config: { a: 1, b: 2 },
+        config: { a: "1", b: "2" },
         env: { NITRO_A: "123" },
-        expected: { a: 123, b: 2 },
```
productdevbook, line 10: Could this cause a problem? Why can't we get number types?

pi0: because environment variables are string-only. We were trying to smartly parse them before however it could make its own edge cases

PR #3635, opened 2025-10-08
refactor!: provide `$fetch` as auto import or via `nitro/deps/ofetch`

pi0:
With direct fetch being globally supported, global injection of `$fetch` is not necessary and adds to the minimal bundle size.

`$fetch` instead can be imported by:

- auto imports
- `nitro/deps/ofetch`
- `ofetch` (added to `dependencies`)

### DIFF
--- a/src/runtime/internal/client.ts
+++ b/src/runtime/internal/client.ts
@@ -1,7 +1,0 @@
-import type { $Fetch, NitroFetchRequest } from "nitro/types";
-// Client polyfill
-import { $fetch } from "ofetch";
-
-if (!globalThis.$fetch) {
-  globalThis.$fetch = $fetch as $Fetch<unknown, NitroFetchRequest>;
-}
--- a/src/types/fetch/fetch.ts
+++ b/src/types/fetch/fetch.ts
@@ -112,14 +112,4 @@ export interface $Fetch<
   ): $Fetch<T, R>;
 }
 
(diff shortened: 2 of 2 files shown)


MickL:
Hey @pi0 ! Could you explain "With direct fetch being globally supported" a bit more? I was just wondering that `$fetch` doesnt exist anymore and the docs have no word about fetching. 

I guess because every runtime now implements [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) we can just use `fetch()`, but what about internal fetches? Are they no more possible with v3?

In any case, I guess a word about `$fetch` being removed should be added to the [migration page](https://v3.nitro.build/docs/migration) :)

P.S. Great talk today and amazing work on v3!

pi0:
> what about internal fetches? Are they no more possible with v3?

Thats what we added. internal fetches work with just global fetch()


Previously we only had it for $fetch via a custom instance now $fetch from original ofetch support relative fetch too in nitro context.


Feel free to add PR mention it in migration guide (using auto imports no action is necessary it works as before)

MickL:
Oh so it works "magically". Thats cool! This info should be added to the docs. 

For migration I am unsure because you said it should work as before, but I used $fetch which is not available anymore. So I needed add ofetch to my project (for typings) and replace all $fetch.

PR #3645, opened 2025-10-11
feat: export PresetName, PresetNameInput, and PresetOptions types from nitro/presets

productdevbook:
## Description

This PR adds support for importing preset types directly from `nitro/presets` subpath.

## Changes

- Add `src/presets/index.ts` to build entries to generate presets index files
- Update package.json exports to point to `dist/presets/index.mjs`
- Add `@cloudflare/workers-types` to externals to fix build error with rollup-plugin-dts

## Usage

After this change, users can import preset types directly:

```typescript
import { PresetName, PresetOptions, PresetNameInput } from 'nitro/presets'
```

## Motivation

Currently, these types are generated in `src/presets/_types.gen.ts` but are not exported from any public subpath, making them unavailable for TypeScript users who need to reference preset names or options in their code.

## Testing

- Built the project successfully with `pnpm build`
- Verified types are properly exported in `dist/presets/index.d.mts`
- Confirmed package.json exports point to correct files

### DIFF
--- a/src/presets/index.d.cts
+++ b/src/presets/index.d.cts
@@ -1,0 +1,1 @@
+export * from "/home/runner/work/nitro/nitro/src/presets/index.js";
--- a/src/presets/index.d.mts
+++ b/src/presets/index.d.mts
@@ -1,0 +1,1 @@
+export * from "/home/runner/work/nitro/nitro/src/presets/index.js";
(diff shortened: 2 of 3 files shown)


src/presets/index.d.cts
productdevbook, line 1: I don't understand why mts and cts files are coming.

src/presets/index.d.mts
productdevbook, line 1: I don't understand why mts and cts files are coming.

pi0:
We can simply export `PresetName` from `nitro/types` (src/types.ts)

`PresetOptions` is merged and available from `NitroOptions['preset']` not sure if it might be very helpful same as `PresetNameInput`

src/presets/index.d.cts
productdevbook: These files came with auto fix.

PR #3650, opened 2025-10-13
feat(cli): add TAB completions

AmirSa12:
### ❓ Type of change

### 📚 Description
This PR introduces tab completion functionality for the nitro cli, improving developer experience by allowing shell(zsh, powershell, fish, bash) autocompletion for commands, options, values, and flags. With this PR, users can navigate available commands, options and values faster!

A small video of how this looks:

https://github.com/user-attachments/assets/56c6b671-6573-4385-8a62-cfd2aa91735c

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "@apphosting/common": "^0.0.9",
     "@azure/functions": "^3.6.0",
     "@azure/static-web-apps-cli": "^2.0.10",
-    "@bomb.sh/tab": "^0.0.5",
+    "@bomb.sh/tab": "^0.0.7",
     "@cloudflare/workers-types": "^5.20260904.1",
     "@cloudflare/workers-utils": "^0.36.0",
     "@deno/types": "^0.0.1",
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "@apphosting/common": "^0.0.9",
     "@azure/functions": "^3.6.0",
     "@azure/static-web-apps-cli": "^2.0.10",
-    "@bomb.sh/tab": "^0.0.7",
+    "@bomb.sh/tab": "^0.0.9",
     "@cloudflare/workers-types": "^5.20260904.1",
     "@cloudflare/workers-utils": "^0.36.0",
     "@deno/types": "^0.0.1",
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "typecheck": "tsc --noEmit --skipLibCheck"
   },
   "dependencies": {
-    "@bomb.sh/tab": "^0.0.5",
     "consola": "^3.4.2",
     "crossws": "^0.4.12",
     "db0": "^0.4.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
     "undici": "^7.11.0"
   },
   "dependencies": {
+    "@bomb.sh/tab": "^0.0.5",
```
pi0: (similar to vite) We are bundling as much as possible can be moved to devDependencies

src/cli/completions.ts
```diff
@@ -0,0 +1,148 @@
...
+                complete("iis-node", "IIS Node");
+                complete("winterjs", "WinterJS");
+                complete("standard", "Standard runtime");
+            };
+        }
+
+        const minifyOption = buildCommand.options.get("minify");
+        if (minifyOption) {
+            minifyOption.handler = (complete) => {
+                complete("true", "Enable minification");
```
pi0: Could we somehow integrate it with citty core to avoid custom logic for each arg? (we have an object of args with their meta)

src/cli/completions.ts
```diff
@@ -0,0 +1,148 @@
...
+            };
+        }
+    }
+
+    const buildCommand = completion.commands.get("build");
+    if (buildCommand) {
+        const presetOption = buildCommand.options.get("preset");
+        if (presetOption) {
+            presetOption.handler = (complete) => {
+                complete("node-server", "Node.js server");
```
pi0: (TODO for myself): List can come from auto gen file

src/cli/completions.ts
AmirSa12: Yes, that's what I'm trying to do.
There is an open PR that integrates tab with citty core:
https://github.com/unjs/citty/pull/211

with this, we can avoid this part.

AmirSa12:
Salam @pi0! I heard there were concerns about tab depending on the `mri` package. I just wanted to let you know that tab no longer depends on mri (tab now has zero dependencies), and I’ll be publishing a new version soon!
Thank you!

pi0:
Amazin @AmirSa12! I will also try to find some time moving citty PR forward to make it happen!

pi0x:
Rebased onto `main` and pushed. Only `pnpm-lock.yaml` and the `devDependencies` block conflicted; I took main's versions and relocked. `nitro complete zsh` and `--help` work on the rebased branch.

Two things I fixed on top:
- `import ... from "./completions"` needs an explicit `.ts` extension here — `pnpm typecheck` failed without it.
- Reformatted `completions.ts` to the repo's 2-space style via `pnpm fmt`.

One thing worth a look before this lands: `@bomb.sh/tab/citty` awaits every entry in `subCommands` to build the completion tree, so the lazy `() => import(...)` loaders in `src/cli/index.ts` all run on every invocation. On a stub build `nitro --version` went from ~58ms to ~193ms. Probably worth only initialising completions when the `complete` command is actually being run.

Leaving the citty-core integration and the hardcoded preset list alone, since you and @pi0 already have those covered.

PR #3663, opened 2025-10-17
fix(cloudflare): add `cloudflare:workers` alias in dev

RihanArfan:
### ❓ Type of change

### 📚 Description

Adds support for `import { env } from 'cloudflare:workers'` during development Cloudflare emulation. `env` is mapped to `getPlatformProxy()`'s environment/bindings.

As Nitro's native `nitro-cloudflare-dev` isn't documented yet, no docs were updated. We're also missing types for `cloudflare:workers`.

image

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare/dev.ts
+++ b/src/presets/cloudflare/dev.ts
@@ -77,7 +77,7 @@ export async function cloudflareDev(nitro: Nitro) {
 
   // Add plugin to inject bindings to dev server
   nitro.options.plugins = nitro.options.plugins || [];
-  nitro.options.plugins.push(
+  nitro.options.plugins.unshift(
     fileURLToPath(new URL("runtime/plugin.dev", import.meta.url))
   );
 }


src/presets/cloudflare/plugin-env-alias.ts
```diff
@@ -0,0 +1,42 @@
...
+    resolveId(id) {
+      if (id === "cloudflare:workers") {
+        return "\0cloudflare-workers-virtual";
+      }
+      return null;
+    },
+    load(id) {
+      if (id === "\0cloudflare-workers-virtual") {
+        return `
+export const env = new Proxy({}, {
```
pi0: It doesn't have any dynamic logic in it. We can directly use a `cloudflare/runtime/imports/workers.ts` file  + add it to preset's `unenv: { alias }`

pi0:
Thanks for PR. Other than moving to runtime file, can you please take a quick look about other possible exports and add them as comment in same file (so later if needed we can add more than `env`

src/presets/cloudflare/preset.ts
```diff
@@ -113,6 +113,12 @@ export const cloudflareDev = defineNitroPreset(
       async (nitro) =>
         await import("./dev").then((m) => m.cloudflareDev(nitro)),
     ],
+    unenv: {
+      alias: {
+        "cloudflare:workers":
+          "nitropack/presets/cloudflare/runtime/imports/workers",
```
pi0: We can resolve it using `presetsDir` from `nitropack/runtime/meta`

src/presets/cloudflare/runtime/imports/workers.ts
```diff
@@ -0,0 +1,24 @@
+export const env = new Proxy(
+  {},
+  {
+    get(_target, prop) {
+      return globalThis.__env__?.[prop as keyof typeof globalThis.__env__];
```
pi0: Do we really need a proxy wrapper? (considering `__env__` should be bound to CF proxy itself already and i guess itself is a proxy no?

RihanArfan: With `export const env = globalThis.__env__`, things doesn't resolve so I needed to use a proxy.

Proxy:
image

Direct:
image

src/presets/cloudflare/preset.ts
RihanArfan: Does this exist in Nitro v2? I can see it in main but presetsDir doesn't seem to be a thing in the v2 codebase 
image

PR #3664, opened 2025-10-17
chore(readme): add deepwiki badge

franklin-tina:
[DeepWiki](https://deepwiki.com) provides in-depth documentation for public (and, in some cases, private) repositories. Adding this badge ensures that the content is automatically refreshed and always stays up to date.
See [Nitro docs in DeepWiki](https://deepwiki.com/nitrojs/nitro)

P.S. [Rolldown](https://github.com/rolldown/rolldown/blob/main/README.md) and [Hono](https://github.com/honojs/hono/blob/main/README.md) are among the popular repositories adopting this as well.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/nitrojs/nitro)
+
 # Nitro
 
 > [!NOTE]


pi0:
Looks sick! I will review generated content once had little more time. We should add it for [h3](https://h3.dev) too!

Small question: Is there a way to refine content if there was a mistake?

franklin-tina:
> Small question: Is there a way to refine content if there was a mistake?

I'm not entirely sure, since all the documentation is generated from the available codebase. Definitely something to keep an eye out for.

But, the more up to date it is when the repo last indexed, the more accurate the generated docs will be.

medz:
> Looks sick! I will review generated content once had little more time. We should add it for [h3](https://h3.dev) too!
> 
> Small question: Is there a way to refine content if there was a mistake?

There's only one way to fix it: force yourself to write detailed code comments for every prop or function.

Adding a badge automatically updates you weekly, but you'll need to pay for more frequent updates.

I'm using DeekWiki to write documentation instead of myself. But I've noticed that the documentation isn't updated regularly, forcing me to write a lot of meaningless code comments.

medz:
I'm already looking for alternatives or reverting to writing my own documentation. DeepWiki's original intentions and goals are great, but without a subscription, there are many uncontrollable issues, such as over-interpretation and over-promotion of the project.

For example, I have a state tool for Flutter, which is a technical experiment. Then DeepWiki boasts that it "completely changes the state of Flutter state management," which is very embarrassing and I have no way to correct it. 😂

franklin-tina:
> I'm already looking for alternatives or reverting to writing my own documentation. DeepWiki's original intentions and goals are great, but without a subscription, there are many uncontrollable issues, such as over-interpretation and over-promotion of the project.
> 
> For example, I have a state tool for Flutter, which is a technical experiment. Then DeepWiki boasts that it "completely changes the state of Flutter state management," which is very embarrassing and I have no way to correct it. 😂

I don't think the aim of adding this badge by any of the aforementioned repos above and the thousands of other repos, or by me, via this PR, was to depend on generated AI docs!
Sure, you'll still have to write documentation where need be, especially for complex implementations.
The aim here is not to replace Nitro or any other repo's documentation!

Keeping these AI models up to date with recent changes or updates is good for the ecosystem, especially when debugging or writing code where some internal APIs were recently deprecated, updated etc.

Their MCP server in VS Code is really helpful.

As for the frequency of re-indexed docs, one a week is better than nothing don't you think?
At least when seach engines or other automated tools access these docs, don't you think getting something that is one week old is better than months old?

_One of the main reasons I looked into adding this badge, apart from the fact that I saw these other repos do it was the fact that I was reading some docs about a feature and noticed the internals had been updated and it behaved a bit differently. 
Then I noticed the docs were last indexed in August._

franklin-tina:
> **I'm using DeekWiki to write documentation instead of myself**. But I've noticed that the documentation isn't updated regularly, forcing me to write a lot of meaningless code comments...

Bro, this is akin to vibe coding and depending fully on the generated code 😅

franklin-tina:
By the way, thank you @medz 
I appreciate these conversations, especially last week when we were debugging the Cloudflare issue with Nitro v3. 💚
(of course, Pooya too)

medz:
> As for the frequency of re-indexed docs, one a week is better than nothing don't you think?

Absolutely agree. It is very useful to be able to track the latest changes at least once a week, especially when developers don't have much time to update the documentation website.

> I appreciate these conversations, especially last week when we were debugging the Cloudflare issue with Nitro v3. 💚

@franklin-tina  This is what I'm supposed to do. I'm constantly monitoring the development of H3 and Nitro. (Ehmmmm~) Because I'm using Nitro v3 directly with Cloudflare in a large application.

This forces me to keep a close eye on things, as they haven't reached a production release yet. I have to ensure that every deployment of my application doesn't affect existing functionality, even digging deep to find workarounds.

PR #3671, opened 2025-10-23
fix(build): avoid replacing unexpected words

9romise:
### 🔗 Linked issue

https://github.com/vueuse/vueuse/issues/5114

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

The default [`delimiters`](https://github.com/rollup/plugins/tree/master/packages/replace#delimiters) in `@rollup/plugin-replace` incorrectly matches variable names containing `$`, as shown in the [linked issue ](https://github.com/vueuse/vueuse/issues/5114)where `typeof window$1` was transformed into `"undefined"$1`, causing the subsequent error.
This PR narrows down the matching scope to prevent accidental matches with unexpected variables.

Before:
`\btypeof window\b(?!\.)` - https://regex101.com/r/1gkfNj/1

After:
`\btypeof window(?![\w.$])` - https://regex101.com/r/cyyShK/1

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #3672, opened 2025-10-23
fix(build): avoid replacing `window$1`

pi0:
Backporting another fix of #3671 by @9romise to v2 to address https://github.com/vueuse/vueuse/issues/5114 in current versions (used by Nuxt).

(thanks to @43081j helping ❤️)

### DIFF
--- a/test/fixture/routes/replace.ts
+++ b/test/fixture/routes/replace.ts
@@ -1,5 +1,6 @@
 export default defineEventHandler((event) => {
   // #3672
+  globalThis.$window1 = false;
   const { window: window$1 = globalThis } = {};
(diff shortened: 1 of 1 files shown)


pi0:
After a few investigations, changing `delimiters` is not easily backportable to Nitro v2, as we need raw code replacements like `import.meta.*` (this is not the case for upcoming Nitro v3 anymore, but the issue is today...)

One alternative hotfix was be to specially handle the `window$1` case and preserve it as is (without treeshaking like `window`)

/cc @9romise @antfu @OrbisK

pi0:
Merging limited replacement as a temporary hotfix for v2.

PR #3675, opened 2025-10-23
refactor(vite)!: early init nitro

pi0:
## Context

Nitro Vite plugin needs to inject other rollup plugin hooks in dev and to resolve those plugins it needs to instantiate based on env (`dev|build` + `root` + user config are required to init) 

Currently, we do it with a very hacky [workaround to manually call](https://github.com/nitrojs/nitro/blob/2d19936975fa7357bcd40bab2604de9af32893ee/src/build/vite/plugin.ts#L397-L439) rollup plugin hooks which is broken in many cases as we don't chain hooks nor respect order.

## New workaround

We could infer `userConfig` somehow from global (`NODE_ENV` and `process.argv`) and root dir from last arg + cwd.

We cannot anymore also access to user config from top level...

## Alternatives 

Ideally it would be great if vite could allow exending plugins with `config` hook ([here](https://github.com/vitejs/vite/blob/29cdb390374689e4dec9017b21fefe88b6ce4203/packages/vite/src/node/config.ts#L2135)) -- giving the responsibility to plugin authors to only inject valid/resolved plugins in there with intended order.

One other alternative would be that vite supporting callable functions as plugins input and call/instnatiate them itself using params similar to [filter](https://github.com/vitejs/vite/blob/29cdb390374689e4dec9017b21fefe88b6ce4203/packages/vite/src/node/config.ts#L1104)

A 3rd alternative would be keeping current solution but an improved `dynamicRollupPlugins()` util that adds hooks with respect of order and chaining (for dev).

### DIFF
--- a/src/build/vite/plugin.ts
+++ b/src/build/vite/plugin.ts
@@ -47,21 +47,21 @@ export async function nitro(
 
   ctx.rollupConfig = await getViteRollupConfig(ctx);
 
-  let additionalPlugins: VitePlugin[] = [];
+  let rollupDevPlugins: VitePlugin[] = [];
   if (ctx.nitro.options.dev) {
     await ctx.nitro.hooks.callHook(
       "rollup:before",
       ctx.nitro,
       ctx.rollupConfig.config
     );
-    additionalPlugins = ctx.rollupConfig.config.plugins as VitePlugin[];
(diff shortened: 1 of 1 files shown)


pi0:
#3676

PR #3682, opened 2025-10-26
test: add minimal fixture for bundle size testing

pi0:
Adding a minimal fixture with `server.ts` and `standard` preset to continuensly measure bundle sizes + test fetch functionality with rollup, rolldown and vite.

Current results: (<33kB/20kB gzip)

```
┌─────────┬───────────────────────┬─────────────┬─────────┐
│ (index) │ builder               │ size        │ time    │
├─────────┼───────────────────────┼─────────────┼─────────┤
│ 0       │ 'rollup'              │ '34.13 kB'  │ '303ms' │
│ 1       │ 'rollup (minified)'   │ '20.20 kB'  │ '264ms' │
│ 2       │ 'rolldown'            │ '188.91 kB' │ '28ms'  │
│ 3       │ 'rolldown (minified)' │ '188.91 kB' │ '21ms'  │
│ 4       │ 'vite'                │ '34.12 kB'  │ '156ms' │
│ 5       │ 'vite (minified)'     │ '16.25 kB'  │ '109ms' │
└─────────┴───────────────────────┴─────────────┴─────────┘
```

(rolldown tree-shaking seems somehow not working properly and we forgot about minify support have to investigate it, /cc @TheAlexLichter @sapphi-red if had time to take a look)

### DIFF
--- a/test/minimal/minimal.test.ts
+++ b/test/minimal/minimal.test.ts
@@ -8,6 +8,12 @@ import { glob } from "tinyglobby";
 const fixtureDir = fileURLToPath(new URL("./", import.meta.url));
 const tmpDir = fileURLToPath(new URL(".tmp", import.meta.url));
 
+const sizeThresholds: Record<string, [number, number]> = {
+  rollup: [35, 21],
+  rolldown: [200, 200],
+  vite: [34, 16],
+};
+
 describe("minimal fixture", () => {
   const builders = ["rollup", "rolldown", "vite"] as const;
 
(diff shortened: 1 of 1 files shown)


sapphi-red:
@pi0 I think it's because you are including the sourcemap size. Rolldown has the values in `sourcesContent`, while Rollup has `null`s. Also the output of `rolldown (minified)` is not minified. I think you need to add `minify: nitro.options.minify,` at https://github.com/nitrojs/nitro/blob/13f66ac2b1655f63dc681d35c50be1a09ada44a3/src/build/rolldown/config.ts#L85-L131

pi0:
Thnx for checking. I think we should also make rolldown source contents to null to be similar (error handler can read from fs and normalize stack)

sapphi-red:
I guess it's because `sourcemapExcludeSources: true` is not set for Rolldown & Rolldown doesn't support the option yet (https://github.com/rolldown/rolldown/issues/5554).
https://github.com/nitrojs/nitro/blob/13f66ac2b1655f63dc681d35c50be1a09ada44a3/src/build/rollup/config.ts#L151

pi0:
Is there a workaround? or perhaps maybe we disable sourcemaps for now in rolldown?

sapphi-red:
You can parse the sourcemap and replace the `sourcesContent` in `generateBundle` hook as a workaround.

pi0:
@sapphi-red Do you know a hook to modify sourcemap json before write? (im worried to not cause additional build time for read/write)

sapphi-red:
`generateBundle` hook runs before the file is written out.

PR #3689, opened 2025-10-27
fix: set explicit delimiters to allow $ in replacements

43081j:
The default delimiter inside the replace plugin is `\b(?!\.)`.

By changing this to `\b(?![\.$])`, we are basically saying _don't_
replace matches which are followed by a `$`.

Makes sense in all replacements, it seems since we'd never want to do
`import.meta.foo$` for example.

### ❓ Type of change

### 📝 Checklist


pi0:
Can you please update fixture too? 🙏🏼

PR #3711, opened 2025-10-31
chore(node, bun, deno): validate port range

iiio2:
Actually the port number range is between 0 to 65535. That is why I've tried to give a validation range for port. I dont know how it would be though.

### DIFF
--- a/src/presets/bun/runtime/bun.ts
+++ b/src/presets/bun/runtime/bun.ts
@@ -9,7 +9,7 @@ const host = process.env.NITRO_HOST || process.env.HOST;
 const cert = process.env.NITRO_SSL_CERT;
 const key = process.env.NITRO_SSL_KEY;
 
-if (port < 0 || port > 65535) {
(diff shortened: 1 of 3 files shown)


pi0:
Thank you for the PR! It is a nice idea but might be better we add to shared utils in srvx. (PS: check to do it only if runtimes themselves don’t validate and fail already)

iiio2:
@pi0 ,

Bun and Deno both validate port range. 
In Bun, when port range is out of range, it is ignored. (https://bun.com/reference/node/url/URL/port) 
In Deno, there is PR where it throws error for out of port range. (https://github.com/denoland/deno/pull/24399/files) 
For Node.js runtime, I've not found such thing maybe where node.js itself validate port range. That is why I've opened a PR in srvx on it. 
https://github.com/h3js/srvx/pull/137

pi0:
🙏 Thnx! Lets follow it there

PR #3713, opened 2025-11-01
refactor: tiny refactoring

iiio2:
actually i think when key and value are same in object, we can just write this way.


iiio2:
Thanks @pi0 . :blush:

PR #3717, opened 2025-11-03
feat: support `format: 'node'`

pi0:
This PR adds a new `format: 'node'` option for server handlers.

This allows one line config (or ideally later zero-config) express support!

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -455,12 +455,6 @@ importers:
 
   test/fixture:
     devDependencies:
-      '@types/express':
-        specifier: ^5.0.5
-        version: 5.0.5
-      express:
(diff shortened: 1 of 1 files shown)


src/types/handler.ts
```diff
@@ -56,6 +53,13 @@ export interface NitroEventHandler extends NitroHandlerCommon {
    */
   handler: string;
 
+  /**
+   * Event handler type.
+   *
+   * Default is `"web"`. If set to `"node"`, the handler will be converted into a web compatible structure.
```
pi0:
```suggestion
   * Default is `"web"`. If set to `"node"`, the handler will be converted into a web compatible handler.
```

PR #3726, opened 2025-11-04
feat: export `defineConfig` and runtime utils from `"nitro"`

pi0:
Nitro v3 moved build utils to `nitro/builder`, now we can use main `"nitro"` subpath for generic/runtime utils.

Some runtime utils (still) depend on virtual import but type helpers and `defienConfig` could be exported from top level.

```js
import {
  defineConfig,
  definePlugin,
  defineRouteMeta,
} from "nitro";
```

### DIFF
--- a/src/runtime/nitro.ts
+++ b/src/runtime/nitro.ts
@@ -1,15 +1,12 @@
 // Config
 import type { NitroConfig } from "nitro/types";
 
-export function defineNitroConfig(
+export function defineConfig(
   config: Omit<NitroConfig, "rootDir">
 ): Omit<NitroConfig, "rootDir"> {
   return config;
 }
 
-export { defineNitroConfig as defineConfig };
-
 // Type (only) helpers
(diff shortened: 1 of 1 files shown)


test/fixture/nitro.config.ts
```diff
@@ -1,8 +1,9 @@
+import { defineConfig } from "nitro";
+
 import { fileURLToPath } from "node:url";
-import { defineNitroConfig } from "nitro/config";
 import { dirname, resolve } from "node:path";
 
-export default defineNitroConfig({
```
productdevbook, line 5: Actually, this name was good for distinguishing.

pi0: You can do `{ defineConfig as defineNitroConfig }`. Shorter is simpler and similar to vite's `defineConfig`

productdevbook: Actually, it's easy to avoid this complexity when reading function names with the CLI.

PR #3745, opened 2025-11-06
refactor: use `node_modules/.nitro` for `buildDir` and rely less on it

pi0:
This PR moves default buildDir from `.nitro` to `node_modules/.nitro` to generate less user code.

Previously we were relyong on build dir to generate chunk names which was opinionated this had been removed.

Temporary (dev) `nito.json` will be standalone generated to `node_modules/.nitro/nitro.dev`

Typescript files will be generated to ``node_modules/.nitro/types` with a new standalone `typescript.generatedTypesDir` (/cc @danielroe)

### DIFF
--- a/src/build/vite/env.ts
+++ b/src/build/vite/env.ts
@@ -68,7 +68,7 @@ export function createServiceEnvironment(
     build: {
       rollupOptions: { input: serviceConfig.entry },
       minify: ctx.nitro!.options.minify,
-      outDir: join(ctx.nitro!.options.buildDir, "vite", "services", name),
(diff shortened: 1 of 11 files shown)


src/build/vite/env.ts
```diff
@@ -68,7 +68,7 @@ export function createServiceEnvironment(
     build: {
       rollupOptions: { input: serviceConfig.entry },
       minify: ctx.nitro!.options.minify,
-      outDir: join(ctx.nitro!.options.buildDir, "vite", "services", name),
+      outDir: join(ctx.nitro!.options.buildDir, "vite/services", name),
```
productdevbook, line 71: I wondered what the purpose was here. :D

PR #3749, opened 2025-11-06
feat(vercel): switch to `node` compat if one handler format is `node`

pi0:
Vercel preset supports both `node` and `web` handler compatibility format.

When at least one handler needs node compatibility (like Express), it is safer at runtime to switch to node handler format and use srvx compatibility layer as this way we can directly call express middleware with node `IncominMessage`/`ServerResponse` instead of doing web=>node conversion.

### DIFF
--- a/examples/express/nitro.config.ts
+++ b/examples/express/nitro.config.ts
@@ -4,4 +4,9 @@ export default defineConfig({
   routes: {
     "/**": { handler: "./server.ts", format: "node" },
   },
+  vercel: {
+    functions: {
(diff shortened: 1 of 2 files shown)


moshetanzer:
Excuse my ignorance should we also be supporting bum on vercel like they just announced can be done?

pi0:
@moshetanzer already landed in latest nightly! https://github.com/nitrojs/nitro/pull/3678

Bun also supports both web and node handler formats + in case of Express + Bun + Vercel, node handler format is better.

PR #3757, opened 2025-11-06
feat(core): load project `tsconfig.json`

pi0:
As many tools already infering options from `tsconfig.json`, nitro core can use it too.

Using lighweight [tsconfck](https://npmjs.com/tsconfck) + cache.

TODO:
- [x] Load
- [x] Aliases 
- [x] JSX options

### DIFF
--- a/playground/tsconfig.json
+++ b/playground/tsconfig.json
@@ -1,8 +1,3 @@
 {
-  "extends": "nitro/tsconfig",
-  "compilerOptions": {
-    "paths": {
-      "~/*": ["./*"]
-    }
-  }
+  "extends": "nitro/tsconfig"
 }


drewbaker:
@pi0 How does this relate to Node sub path imports defined in package.json? And the alias config in Nitro?

package.json:
```
    "imports": {
        "#server/*": "./server/*"
    }
```

nitro.config.ts
```
    alias: {
        "#server/*": resolve(__dirname, "./server/*"),
    },
```

And then the tsconfig.json:
```
        "paths": {
            "#server/*": [
                "./server/*"
            ]
        },
```

Do we need all of them? Can we use `~/` instead for all of them somehow? I thought Node sub-path needed to start with a `#` so seems maybe that is the best default to use instead of `~`?

I've been trying for days to get [Sidequest.js working with Nitro](https://github.com/sidequestjs/sidequest/issues/125) and having the import paths all work between Node and Nitro would just be so useful for that type of thing.

pi0:
@drewbaker subpath imports (by node convention) should start with `#` and should already work.

Tested on playground:

```json
"imports": {
    "#server/*": "./*"
}
```

drewbaker:
> @drewbaker subpath imports (by node convention) should start with `#` and should already work.
> 
> Tested on playground:
> 
> ```json
> "imports": {
>     "#server/*": "./*"
> }
> ```

Yeah they do work great, just seems like an overlap of the `~` convention, except `~` import's won't work in Node... So seemed to me that ideal use case would be that Nitro should just use `#server` (or maybe `#nitro`?) instead of `~/`

pi0:
It is in the end choice of users nitro v3 is not having any defaults. I do also prefer subpath imports which are more standard than tsconfig pathh…

drewbaker:
OK great, good to know. What do you use as the path? `#server/*`? `#nitro/*`? Just curious. I'll start using it instead of `~/` for all my imports now.

MickL:
I would love to see `~/utils`, `~/assets`, etc. working. That is the convention of Nuxt and old Nitro so to me it makes sense to keep using it.

pi0:
> What do you use as the path? #server/*? #nitro/*?

I think `#server/*` (if you prefer subpath) or `~/server` (if prefer typescript aliases) might make best sense. All paths relative to one root.

But anyway nitro won't be opinionated on this. All up to projects.

MickL:
Hey @pi0 , I just checked this. Shouldnt it be `"~/server/*": ["./server/*"]`, with a slash after `~`, in the examples? That is how it was working in Nuxt:

```ts
import { useSomething } from '~/utils/use-something';
```

instead of:

```ts
import { useSomething } from '~utils/use-something';
```

Ofc this is up to the user, but from my experience official docs and examples always determine code conventions broadly used.

pi0:
In starters it is more generic even https://github.com/nitrojs/starter/blob/7e9adcc03658ba4dee99f6030daa699acb8755bf/tsconfig.json#L5 

In example i just wanted to show interchangeable #server or ~server alias creation.

MickL:
Perfect for the starter !! :)

PR #3763, opened 2025-11-09
feat!: migrate to `oxc-transform` and `oxc-minify`

pi0:
As vite is moving to oxc, this helps better ecosystem consistency between different builders.


productdevbook:
🚀

PR #3769, opened 2025-11-10
feat: auto detect builder (optional rollup)

pi0:
Nitro v3 now has 4 options of builders: rollup/rolldown/vite/rolldown-vite

This PR makes rollup an optional peer dep to allow choosing between options without imposing additional install size.

### DIFF
--- a/src/config/resolvers/builder.ts
+++ b/src/config/resolvers/builder.ts
@@ -5,7 +5,8 @@ import { createRequire } from "node:module";
 const VALID_BUILDERS = ["rollup", "rolldown", "vite", "rolldown-vite"] as const;
 
 export async function resolveBuilder(options: NitroOptions) {
-  options.builder = (process.env.NITRO_BUILDER as any) || options.builder;
+  // NITRO_BUILDER environment variable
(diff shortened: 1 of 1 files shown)


src/config/resolvers/builder.ts
```diff
@@ -1,47 +1,81 @@
...
-        throw new Error(
-          `Builder "vite" is not available. Make sure to install "vite" package.`
-        );
-      }
+    return;
+  }
 
-      break;
+  // Auto-detect installed builder
+  for (const pkg of ["rolldown", "rollup", "vite"] as const) {
```
ProgramRipper, line 37: Maybe we should detect vite first? As vite should take precedence over rollup/rolldown, and vite depends on them. We may short-circuit at rollup/rolldown first before vite.

pi0: With vite, `vite dev|build` is used which explicitly sets builder itself: https://github.com/nitrojs/nitro/blob/4cf1b533435ab8b28a75cf20e817a3541216cff8/src/build/vite/plugin.ts#L361

PR #3775, opened 2025-11-10
chore: `vite-devtools` example

productdevbook:
https://vite-devtools.netlify.app/guide/

---

Vite version

CleanShot 2025-11-10 at 16 28 31@2x

vite + nitro (now dont working)

CleanShot 2025-11-10 at 16 31 22@2x

### ❓ Type of change

### 📝 Checklist


src/build/vite/dev.ts
```diff
@@ -207,7 +207,11 @@ export async function configureViteDevServer(
     next: () => void
   ) => {
     // Skip for vite internal requests or if already handled
-    if (/^\/@(?:vite|fs|id)\//.test(nodeReq.url!) || nodeReq._nitroHandled) {
+    if (
+      /^\/@(?:vite|fs|id)\//.test(nodeReq.url!) ||
+      nodeReq._nitroHandled ||
+      nodeReq.url!.startsWith("/.devtools-vite/")
```
pi0: /cc @antfu Do you think there is a better way?

pi0:
Dev now works with vite. For production, i'm not sure how to fix this:

image

src/build/vite/dev.ts
antfu: We are going to allow different integration to register their custom base URLs, so it can be anything (even if we don't, people could still add custom middlewares.

I don't think hard-coding is a good way. Is it possible to run through the real Vite middleware to see if they are getting handled or so?

pi0: Hmm that's a good idea we can exclude middleware with explicit bases i guess for phase 1 middleware (nitro middleware is conditionally first or last based on req)

antfu:
Production mode is not fully working and I need to work on it more

src/build/vite/dev.ts
pi0: https://github.com/nitrojs/nitro/pull/3805

pi0:
Closing as we have a poc that it works. Probably not worth to maintain an example where both are seperate things.

PR #3776, opened 2025-11-10
ci: use hash instead of version

iiio2:
I think i've read it using hash is always better practice for security reason in ci. That is why I have changed the actions to hash instead of versions. I thought maybe it might be better here.


pi0:
Thanks! While it makes changes more secure, for now eaiser maintenance and upgrade is more important.

PR #3778, opened 2025-11-11
chore: enable pnpm trust policy

iiio2:
since we're using pnpm 10.21, it has a very good option for security feature for enabling trust policy. maybe we can use it. (https://pnpm.io/settings#trustpolicy)


iiio2:
Thanks @pi0 . :blush:

PR #3787, opened 2025-11-14
feat(vercel): add support for skew protection

danielroe:
### 🔗 Linked issue

resolves https://github.com/nitrojs/nitro/issues/2311

### ❓ Type of change

### 📚 Description

seems like a relatively straightforward change

I'll open a similar one for [netlify](https://docs.netlify.com/build/frameworks/frameworks-api/#netlifyv1skew-protectionjson) as well

~~having looked at both implementations, I'm not sure we _need_ a [manifest](https://github.com/nitrojs/nitro/issues/2312) before merging this, but please do feel free to tell me if I've taken a wrong turn here.~~

### 📝 Checklist

### DIFF
--- a/src/presets/vercel/types.ts
+++ b/src/presets/vercel/types.ts
@@ -113,6 +113,14 @@ export interface VercelOptions {
   config: VercelBuildConfigV3;
 
   /**
+   * If you have enabled skew protection in the Vercel dashboard, it will
+   * be enabled by default.
+   *
+   * You can disable the Nitro integration by setting this option to `false`.
+   */
+  skewProtection?: boolean;
+
+  /**
    * If you are using `vercel-edge`, you can specify the region(s) for your edge function.
    * @see https://vercel.com/docs/concepts/functions/edge-functions#edge-function-regions
    */
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -165,7 +165,8 @@ function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
           return route;
         }),
       // Skew protection
-      ...(process.env.VERCEL_SKEW_PROTECTION_ENABLED && nitro.options.manifest?.deploymentId
+      ...(nitro.options.vercel?.skewProtection &&
+      nitro.options.manifest?.deploymentId
         ? [
             {
               src: "/.*",
--- a/src/presets/vercel/preset.ts
+++ b/src/presets/vercel/preset.ts
@@ -17,6 +17,9 @@ const vercel = defineNitroPreset(
     manifest: {
       deploymentId: process.env.VERCEL_DEPLOYMENT_ID,
     },
+    vercel: {
+      skewProtection: !!process.env.VERCEL_SKEW_PROTECTION_ENABLED,
+    },
     output: {
       dir: "{{ rootDir }}/.vercel/output",
       serverDir: "{{ output.dir }}/functions/__server.func",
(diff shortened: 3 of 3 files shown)


pi0:
> having looked at both implementations, I'm not sure we need a https://github.com/nitrojs/nitro/issues/2312 before merging this, but please do feel free to tell me if I've taken a wrong turn here.

Not (generated) manifest, but we need a configuration namespace `manifest: { deploymentId }` that stores `VERCEL_DEPLOYMENT_ID`.

Also can you please reopen on `main` branch? We can backport it later steps.

src/presets/vercel/utils.ts
```diff
@@ -187,6 +187,25 @@ function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
...
+              src: "/.*",
+              has: [
+                {
+                  type: "header",
+                  key: "Sec-Fetch-Dest",
+                  value: "document",
+                },
+              ],
+              headers: {
+                "Set-Cookie": `__vdpl=${process.env.VERCEL_DEPLOYMENT_ID}; Path=${nitro.options.baseURL}; SameSite=Strict; Secure; HttpOnly`,
```
pi0: To double check in conjunction with #3792, does it still work on vercel with manually overriden deploymentId?

src/presets/vercel/utils.ts
```diff
@@ -187,6 +187,25 @@ function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
           }
           return route;
         }),
+      // Skew protection
+      ...(process.env.VERCEL_SKEW_PROTECTION_ENABLED
```
pi0: Wondering if it should be configurable. We could have `vercel.skewProtection` with default set by `VERCEL_SKEW_PROTECTION_ENABLED` (later/soon agnostic top level `skewProtection`)

danielroe: afaict it wouldn't work if user modifies this (or deployment id), as skew protection is something that a user enables in the vercel dashboard ui

pi0: But the nitro+vercel feature (addition of cookie header rule) should be possible to opt out of at least.

danielroe: indeed! but not opt-in-able ...

pi0: > but not opt-in-able

Unless we implement agnostic feature ;)

danielroe: added ability to opt-out + manually configure skew protection....

agnostic feature would be amazing but requires storing assets from previous deployments or deployment URLs somewhere, right?

pi0: yes or hard reload i guess (in v3, we have access to client build too)

src/presets/vercel/types.ts
```diff
@@ -112,6 +112,14 @@ export interface VercelServerlessFunctionConfig {
 export interface VercelOptions {
   config: VercelBuildConfigV3;
 
+  /**
+   * If you have enabled skew protection in the Vercel dashboard, it will
+   * be enabled by default.
+   *
+   * You can disable the Nitro integration by setting this option to `false`.
+   */
+  skewProtection?: boolean;
```
pi0:
```suggestion
  skewProtection?: false;
```

danielroe: actually this is tricky - we rely on this being set to true by the env variable (and it might be useful also for user to read to see if it is enabled)

gerynugrh:
Hi @danielroe this is really interesting, is there a way for me to use this right now?

pi0:
https://nitro.build/guide/nightly

PR #3794, opened 2025-11-16
feat: `netlify.config` options namespace

danielroe:
### ❓ Type of change

### 📚 Description

this is a breaking change that I think would be good to get in before the coming of v3. with the [frameworks api](https://docs.netlify.com/build/frameworks/frameworks-api/#netlifyv1configjson) for netlify, `config.json` has a subset of the features that we might want to enable for netlify, so we can either flatten it out and manually extract keys we support (which has backwards compat) or move them into a namespaced `config`, as we do (for example) for vercel.

### 📝 Checklist

### DIFF
--- a/src/presets/netlify/preset.ts
+++ b/src/presets/netlify/preset.ts
@@ -46,7 +46,12 @@ const netlify = defineNitroPreset(
           generateNetlifyFunction(nitro)
         );
 
-        if (nitro.options.netlify?.config || nitro.options.netlify?.images) {
+        if (nitro.options.netlify?.images) {
+          nitro.options.netlify.config ||= {};
+          nitro.options.netlify.config.images ||= nitro.options.netlify?.images;
+        }
+
+        if (Object.keys(nitro.options.netlify?.config || {}).length > 0) {
           const configPath = join(
             nitro.options.output.dir,
             "../deploy/v1/config.json"
@@ -54,10 +59,7 @@ const netlify = defineNitroPreset(
           await fsp.mkdir(dirname(configPath), { recursive: true });
           await fsp.writeFile(
             configPath,
-            JSON.stringify({
-              images: nitro.options.netlify?.images,
-              ...nitro.options.netlify?.config,
-            }),
+            JSON.stringify(nitro.options.netlify?.config),
             "utf8"
           );
         }


src/presets/netlify/types.ts
```diff
@@ -2,11 +2,13 @@
  * Netlify options
  */
 export interface NetlifyOptions {
-  images?: {
```
pi0, line 5: We can keep top level keys as backward compatibility and merge them on write.

src/presets/netlify/types.ts
```diff
@@ -2,11 +2,13 @@
...
-    /**
-     * Permitted remote image sources. Array of regex strings.
-     * @see https://docs.netlify.com/image-cdn/overview/#remote-path
-     */
-    remote_images?: string[];
```
pi0, line 10: We can also keep the types I guess for top level `images`

test/presets/netlify.test.ts
```diff
@@ -12,8 +12,10 @@ describe("nitro:preset:netlify", async () => {
         publicDir: resolve(getPresetTmpDir("netlify"), "dist"),
       },
       netlify: {
-        images: {
-          remote_images: ["https://example.com/.*"],
```
pi0, line 16: We can keep tests as is to cover backward compat.

src/presets/netlify/types.ts
danielroe: I think it would be better and more intuitive not to support top level `images` config, as by doing so we can mirror the [netlify config](https://docs.netlify.com/build/frameworks/frameworks-api/#netlifyv1configjson), which includes blobs, skew protection etc.

the point of this PR is to make it more intuitive.

if we keep supporting `images`, we might as well not make this change at all

pi0: it is about backward compatibility with integrations (i suppose also nuxt image) that set this namespace. `@deprecated` tag can be used to make sure new users will use nested namespace.

danielroe: if we drop type support but keep handling it, integrations like nuxt/image will continue to work - they don't need the types

pi0: please keep it.

danielroe: done ✅ 

however, if it's not about integrations, I would love to know more about why we need to keep the types. it feels like v3 (which already brings a lot of breaking changes) is an opportunity to simplify the options here.

pi0: Types should match runtime logic. it is something nitro preset is still using and merging i don't think hiding it is best idea. We can remove in next majors instead.

PR #3802, opened 2025-11-18
feat: add `manifest.deploymentId`

danielroe:
### 🔗 Linked issue

backport of #3792

### ❓ Type of change

### 📚 Description

this backports #3792

### 📝 Checklist

### DIFF
--- a/src/presets/aws-amplify/preset.ts
+++ b/src/presets/aws-amplify/preset.ts
@@ -7,6 +7,10 @@ const awsAmplify = defineNitroPreset(
   {
     extends: "node-server",
     entry: "./runtime/aws-amplify",
+    manifest: {
+      // https://docs.aws.amazon.com/amplify/latest/userguide/environment-variables.html#amplify-console-environment-variables
+      deploymentId: process.env.AWS_JOB_ID,
+    },
     output: {
       dir: "{{ rootDir }}/.amplify-hosting",
       serverDir: "{{ output.dir }}/compute/default",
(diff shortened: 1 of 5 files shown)


pi0:
What would be benefit of this backport without agnostic skew detection support in v2?

I think would probably be more valuable if we add this flag to std-env (tracker: https://github.com/unjs/std-env/issues/59)

danielroe:
I would hope to also merge skew protection back into the vercel preset.

(and likewise for netlify!)

pi0:
> I would hope to also merge skew protection back into the vercel preset.

(as we discussed) Skew-protection will be a vendor+framework, specific feature if we backport to v2 (similar to ISR route rule) since Nitro v2 cannot have client-side logic.

If Nuxt does not migrate soon to Nitro v3, we need to find a temporary solution for Nuxt specifically to support skew protection (in 2 vendors)

---

Even if we want to backport a feature for Vercel/Netlify unblocking Nuxt v4, I think we can directly use the deployment ID hardcoded in them.

As for this PR, i appreciate if you have time to help porting it to std-env. It has no use for Nitro v2.

danielroe:
having investigated, I am not sure that `deploymentId` is implemented by enough providers to port to std-env.

I also agree (for this reason) that hard-coding the environment variables is a better solution for skew protection within nitro (as it's very specific to individual providers at the moment).

I would really like to see a generic solution but haven't yet figured out how to do it, with the primitives we have available in nitro.

pi0:
For the Nitro (v3) roadmap, a generic `deploymentId` (via std-env instead of hardcoding in nitro presets) and skew protection (via vite client utils) are sensible but i guess both won't directly benefit Nuxt.

danielroe:
what do you mean that these features won't benefit nuxt?

pi0:
client implementation.

PR #3816, opened 2025-11-24
docs: mention `bun.lock` for lockfile

iiio2:
Since bun.lock is also a lockfile, maybe we can include it here.


iiio2:
Thanks @pi0 . :blush:

PR #3825, opened 2025-12-01
chore(examples): add `qwik` example

pi0:
Investigating Qwik support with Nitro/vite plugin.

---

_(Moved from nitrojs/vite-examples#9 by @gioboa @wmertens ❤️)_


pi0x:
Rebased on latest `main` and pushed. Still a draft, I left that alone.

Conflicts were `pnpm-lock.yaml` and `pnpm-workspace.yaml`. I took main's version of both and re-ran `pnpm install`. The two `@netlify/*` entries this branch added to `trustPolicyExclude` are no longer needed, install passes without them.

One fix in an extra commit: the build failed because `@qwik.dev/router` no longer exports `getNotFound`. Switched to `getErrorHtml(404, "Not Found")`, which is what Qwik's own adapters use now. Also added the example to the root `tsconfig.json` exclude list, since Qwik JSX does not typecheck under the root React JSX settings.

Still broken, and it looks like the core open question: `/` returns 500 with Qwik error Q14. The `qwikPatches` plugin stubs `@qwik-client-manifest` as `{}`, so SSR cannot resolve event handlers. The example also still pins Vite 7 while the repo is on Vite 8, and has no README or docs page.

pi0:
closing for now for triage

PR #3830, opened 2025-12-02
chore(security): bump expressjs

iiio2:
@pi0 , there is a security issue in expressjs which published yesterday i guess. It needs to be updated. i have updated it and also deduped the lockfile.
https://github.com/expressjs/express/security/advisories/GHSA-pj86-cfqh-vqx6

### DIFF
--- a/examples/express/package.json
+++ b/examples/express/package.json
@@ -6,7 +6,7 @@
   },
   "devDependencies": {
     "@types/express": "^5.0.5",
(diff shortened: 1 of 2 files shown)


pi0:
I appreciate PR but it is a demo only dependency and it is not a security risk here.

iiio2:
Oh, I see. ok @pi0 . Thanks.

PR #3834, opened 2025-12-03
feat(cloudflare): support `exports.cloudflare.ts`

### LINKED ISSUE #3833: [v3]: Allow the `server.ts` entry to export other content.
### Describe the feature

For example, in Cloudflare workers, when combined with workflows, Cloudflare requires exporting the implementation class of `WorkerEntrypoint` in the entry file. Currently, Nitro 3 does not support exporting from anywhere.

We could create a new entry file and import `.output/server/index.mjs`, but this has a problem: if we use the same things in the new entry file as in Nitro 3 (e.g., ORM), the code will be packaged and deployed twice.

My workaround is to allow exporting content from `server.ts`. After a Nitro build, this will also be exported to `.output/server/index.mjs`.

---

I'd love to help with this, but I haven't found any documentation or anything related to the Nitro plugin. Of course, I don't want to rely on AI to do it.

### Additional information

medz:
### 🔗 Linked issue

resolves #3833

### ❓ Type of change

### 📚 Description

Support Cloudflare preset additional entry exports.

### 📝 Checklist

### DIFF
--- a/src/presets/cloudflare/entry-exports.ts
+++ b/src/presets/cloudflare/entry-exports.ts
@@ -1,44 +1,27 @@
 import type { Nitro } from "nitro/types";
-import { readFile } from "node:fs/promises";
-import { presetsDir } from "nitro/meta";
 import { resolveModulePath } from "exsolve";
-import { parseSync } from "oxc-parser";
-import { resolveNitroPath, prettyPath } from "../../utils/fs.ts";
+import { prettyPath } from "../../utils/fs.ts";
 
 const RESOLVE_EXTENSIONS = [".ts", ".js", ".mts", ".mjs"];
-const DEFAULT_DETECTD_EXPORTS_FILENAME = "exports.cloudflare";
 
 export async function setupEntryExports(nitro: Nitro) {
   const exportsEntry = resolveExportsEntry(nitro);
   if (!exportsEntry) return;
 
-  const exports = await resolveModuleExportNames(exportsEntry);
-  if (exports.includes("default")) {
-    throw new Error(
-      `Unsupported Cloudflare exports entry \`${prettyPath(exportsEntry)}\` exports default.`
-    );
-  }
+  const originalEntry = nitro.options.entry;
 
-  const serverEntry = resolveModulePath(nitro.options.entry, {
-    from: [presetsDir, nitro.options.rootDir, ...nitro.options.scanDirs],
-    extensions: RESOLVE_EXTENSIONS,
-  });
-
-  const serverEntryExports = await resolveModuleExportNames(serverEntry);
-  const id = (nitro.options.entry =
+  const virtualEntryId = (nitro.options.entry =
     "#nitro-internal-virtual/cloudflare-server-entry");
-  nitro.options.virtual[id] = `/* ts */
+  nitro.options.virtual[virtualEntryId] = /* ts */ `
       export * from "${exportsEntry}";
-      export { ${serverEntryExports.join(", ")} } from "${serverEntry}";
+      export * from "${originalEntry}";
+      export { default } from "${originalEntry}";
   `;
 }
 
 function resolveExportsEntry(nitro: Nitro) {
   const entry = resolveModulePath(
-    resolveNitroPath(
-      nitro.options.cloudflare?.exports ?? DEFAULT_DETECTD_EXPORTS_FILENAME,
-      nitro.options
-    ),
+    nitro.options.cloudflare?.exports || "./exports.cloudflare.ts",
     {
       from: nitro.options.rootDir,
       extensions: RESOLVE_EXTENSIONS,
@@ -57,14 +40,4 @@ function resolveExportsEntry(nitro: Nitro) {
   }
 
   return entry;
-}
-
-async function resolveModuleExportNames(path: string) {
-  const content = await readFile(path, "utf8");
-  const parsed = parseSync(path, content, { sourceType: "module" });
-  const exports = parsed.module.staticExports
-    .flatMap((exp) => exp.entries.map((e) => e.exportName))
-    .filter((e) => e.kind === "Default" || e.kind === "Name");
-
-  return exports.map((e) => e.name ?? "default");
 }


medz:
@coderabbitai generate docstrings

pi0:
Thanks for PR. It is a good idea and i was planning to work on it. But it should be via `server.[preset].ts` to explicitly mark preset we are targetting for (this way also project can still be deployed to multiple targets without need to change)

src/build/plugins/server-entry-exports.ts
```diff
@@ -0,0 +1,28 @@
+import type { Nitro } from "nitro/types";
+import type { Plugin } from "rollup";
+
+export function serverEntryExports(nitro: Nitro): Plugin {
+  return {
+    name: "nitro:server-entry-exports",
+    transform(code, id) {
+      // Only append if serverEntry is defined
+      if (!nitro.options.serverEntry) {
```
pi0: We should introduce a new entryExports[] config that can point to additional export file(s). (This feature should work standalone from server.ts support)
Presets can set a default to it, like cloudflare.ts for cloudflare presets.
An array allows composing multiple versions.

medz: Got it! 👍 This is a great choice.

src/build/plugins/server-entry-exports.ts
```diff
@@ -0,0 +1,28 @@
...
+      }
+
+      // Only process the preset entry file (need to add .mjs extension)
+      const entryWithExt = nitro.options.entry + ".mjs";
+      if (id !== entryWithExt) {
+        return null;
+      }
+
+      // Append re-export statement before code is bundled
+      const reExport = `\nexport * from "#nitro-internal-virtual/server-entry";\n`;
```
pi0: We need to create a new virtual that merges the main and user exports.

Then, we need to parse real `nitro.options.entry` and if there is a default export, also export it (otherwise, named exports and validate to avoid conflict)

`default` needs to be explicitly exported **if preset needs a `default` export**.


[this](https://github.com/unjs/obuild/blob/main/src/builders/transform.ts#L116-L129) might give you some ideas.

src/build/plugins/server-entry-exports.ts
medz: Furthermore, since server.ts does not handle platform-specific exports, it can easily implement platform-specific features such as workerd <=> vercel.

src/build/plugins.ts
```diff
@@ -42,6 +43,9 @@ export function baseBuildPlugins(nitro: Nitro, base: BaseBuildConfig) {
     plugins.push(routeMeta(nitro));
   }
 
+  // Server entry re-exports
+  plugins.push(serverEntryExports(nitro));
```
pi0: We should conditionally add plugin if `entryExports.lengh > 0`

medz:
```diff
- believe `entryExports` should have the following configuration functionality:

- ```ts
- export default defineConfig({
-   serverEntryExports: [
-     'exports.ts' // All preset
-     { preset: 'cloudflare', file: 'cf.exports.ts' }
-   ], // type `string | string[] | {preset: Preset, file: string} | {preset: Preset, file: string}[]`

- });

- ```

- For the default loaded server entry, the following rules should apply:

- ```
- server.exports.ts // All presets
- server.cloudflare.ts // Only workerd (pages/module)
- server.vercel.ts // Only Vercel
- ...
- ```

- @pi0 , what do you think of this?

---

- In this way, you can either configure it from `nitro.config.ts` or load the exports file using rules.
```

@pi0 It's okay, I misunderstood you. Haha, I understand now. You meant that other content is exported through `server.cloudflare.ts`, and if `server.cloudflare.ts` has an export default, then it should override `server.ts`, right?

pi0:
Having a filterable array of objects is nice, but it is also tricky to `1:1` map names of Nitro presets to the conditions, and TBH seems a little bit overengineering for current needs.

We could even move this feature under the cloudflare preset since it is very specific to this preset. I don't know any other deployment provider or runtime that even requires extra exports.

Thinking more, we could even move this feature under cloudflare preset to support `cloudflare.ts` (if it exists, we switch to virtual entry) and then generalize later if needed.

medz:
> Having a filterable array of objects is nice, but it is also tricky to `1:1` map names of Nitro presets to the conditions, and TBH seems a little bit overengineering for current needs.
> 
> We could even move this feature under cloudflare preset since it is very specific to this preset. I don't know any other deployment provider or runtime that even requires extra exports.
> 
> Thinking more, we could even move this feature under cloudflare preset to support `cloudflare.ts` (if it exists, we switch to virtual entry) and then generalize later if needed.

Yes, currently only Cloudflare among all presets requires this.

For example, bun/node directly start the server, so exporting content would be meaningless.

I think implementing this only for Cloudflare is sufficient. For instance, we could allow a `cloudflare.ts` file at the same level as `server.ts`, or allow a custom entry file in the Cloudflare entry in the Nitro config.

medz:
@pi0 So, I'll reimplement it and then add support for it in Cloudflare presets. We can generalize it as needed when adding other presets in the future. This avoids over-design that could lead to uncertainty in the results.

medz:
image
image
image

medz:
@pi0 I implemented the following features:

1. When preset to cloudflare, check the cloudflare.ts/mts/js/mjs files in rootDir.
2. Allow users to customize the `nitro.options.cloudflare.entrypoint` file.
3. Disable `export default` in cloudflare.ts.

src/presets/cloudflare/entry-exports.ts
```diff
@@ -0,0 +1,60 @@
...
+  `;
+}
+
+async function resolveEntrypoint(nitro: Nitro) {
+  const entry = resolveModulePath(
+    resolveNitroPath(
+      nitro.options.cloudflare?.entrypoint ?? "cloudflare",
+      nitro.options
+    ),
+    {
```
pi0, line 25: We need to set `try: true` flag otherwise it throws an error

src/presets/cloudflare/server-entry.ts
```diff
@@ -0,0 +1,60 @@
...
+  const exports = await resolveModuleExportNames(internalEntry);
+  const id = (nitro.options.entry =
+    "#nitro-internal-virtual/cloudflare-server-entry");
+  nitro.options.virtual[id] = `/* ts */
+      export { ${entryExports.join(", ")} } from "${entrypoint}";
+      export { ${exports.join(", ")} } from "${internalEntry}";
+  `;
+}
+
+async function resolveEntrypoint(nitro: Nitro) {
```
pi0: function does not need to be async

src/presets/cloudflare/server-entry.ts
```diff
@@ -0,0 +1,60 @@
...
+import { resolveNitroPath, prettyPath } from "../../utils/fs.ts";
+import { resolveModuleExportNames } from "mlly";
+
+const RESOLVE_EXTENSIONS = [".ts", ".js", ".mts", ".mjs"];
+
+export async function maybeServerEntry(nitro: Nitro, durable: boolean = false) {
+  const entrypoint = await resolveEntrypoint(nitro);
+  if (!entrypoint) return;
+
+  const entryExports = await resolveModuleExportNames(entrypoint);
```
pi0: We might use oxc-parser (sync) instead which is more reliable

src/presets/cloudflare/server-entry.ts
```diff
@@ -0,0 +1,60 @@
...
+    );
+  }
+
+  const internalEntry = resolveModulePath(
+    resolveNitroPath(nitro.options.entry, nitro.options),
+    {
+      from: [presetsDir, nitro.options.rootDir, ...nitro.options.scanDirs],
+      extensions: RESOLVE_EXTENSIONS,
+    }
+  )!;
```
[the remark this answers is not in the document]
medz: @coderabbitai Why? `nitro.options.entry` must exist.

src/presets/cloudflare/server-entry.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const entryExports = await resolveModuleExportNames(entrypoint);
+  if (entryExports.includes("default")) {
+    throw new Error(
+      `Unsupported Cloudflare entrypoint \`${prettyPath(entrypoint)}\` exports default.`
+    );
+  }
+
+  const internalEntry = resolveModulePath(
+    resolveNitroPath(nitro.options.entry, nitro.options),
```
pi0: `nitro.options.entry` should be already resolved.

medz:
CC @pi0 resolved

src/presets/cloudflare/entry-exports.ts
```diff
@@ -0,0 +1,69 @@
...
+      `Detected \`${prettyPath(entry)}\` as Cloudflare entrypoint.`
+    );
+  }
+
+  return entry;
+}
+
+async function resolveModuleExportNames(path: string) {
+  const content = await readFile(path, "utf8");
+  const parsed = parseSync(path, content, { sourceType: "module" });
```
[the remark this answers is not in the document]
medz: @pi0 Is it necessary to take action?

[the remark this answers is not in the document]
medz: The existence of the file can be determined during the initial path acquisition step, unless the current process lacks permission to read the file.

I didn't check this during previous usage (making runtime native suggestions more user-friendly) and oxc-parser suggestions more user-friendly.

cc @coderabbitai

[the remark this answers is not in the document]
medz: @coderabbitai
I prefer the latter; oxc's error output is inherently more user-friendly. node readfile's error output more clearly identifies the problem. Wrapping error messages separately might miss important information, preventing users from correctly understanding the issue.

medz:
@pi0 Okay, it's ready for review now. I just discovered a bug where there was no default export; using `export * from “xxx”` in `cloudflare.ts` wasn't supported.

These issues are now resolved.

medz:
@pi0 I supplemented the documentation and tests.

docs/2.deploy/20.providers/cloudflare.md
```diff
@@ -69,6 +69,34 @@ You can use [runtime hooks](/docs/plugins#nitro-runtime-hooks) below in order to
...
+
+You can also customize the entrypoint file location using the `cloudflare.entrypoint` option in your `nitro.config.ts`:
+
+```ts [nitro.config.ts]
+export default defineConfig({
+  cloudflare: {
+    entrypoint: "custom-entry.ts"
+  }
+})
+```
```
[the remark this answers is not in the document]
medz: No, you are wrong! In the latest v3 version, the configuration is done using the following code:

```ts
import { defineConfig } from 'nitro';
export default defineConfig({
    // ...
});
```

medz:
@pi0 🫣I've already started using it! It's fantastic! Finally, I can run complex, long-running logic elsewhere without binding it to additional workers.

I really hope it gets merged and released as version 3.0.1-alpha.2 😂

image

pi0:
Nice! As last part, i am wondering which name we should use for custom exports. `cloudflare.ts` _might_ conflict with an upcoming cloudflare convention (for example for their config).

Two other alternatives:
- `server.cloudflare.ts`
- `exports.cloudflare.ts`

(i will be sharing it in discord to see if there are any opinions)

docs/2.deploy/20.providers/cloudflare.md
```diff
@@ -69,6 +69,34 @@ You can use [runtime hooks](/docs/plugins#nitro-runtime-hooks) below in order to
...
+::warning
+The `cloudflare.ts` file must not have a default export.
+::
+
+You can also customize the entrypoint file location using the `cloudflare.entrypoint` option in your `nitro.config.ts`:
+
+```ts [nitro.config.ts]
+export default defineConfig({
+  cloudflare: {
+    entrypoint: "custom-entry.ts"
```
pi0: We might now rename config to `exports: `

test/cloudflare/cloudflare.test.ts
pi0, line 1: Would be better if we convert it to a lighter unit test (running without full nitro instantiation)

test/fixture/exports.cloudflare.ts
pi0, line 1: We can add `exports.cloudflare.ts` to main fixture and only test it in `presets/cloudflare` additional tests.

docs/2.deploy/20.providers/cloudflare.md
medz: It has been changed to exports

test/fixture/exports.cloudflare.ts
medz: It has been changed to the main test of preset + fixture.

test/cloudflare/cloudflare.test.ts
medz: It has been moved to preset/cloudflare for testing.

franciscohermida:
@medz you are a legend, was waiting for this for so long!

medz:
> @medz you are a legend, was waiting for this for so long!

image
@franciscohermida Thank you, I've been looking forward to this feature from Nitro. I finally decided to contribute it to Nitro. I'm currently using version `https://pkg.pr.new/nitrojs/nitro@8ae0c83` and have deployed it in production. ❤️

emretinaztepe:
@medz This is awesome!!! I have spent two months searching for an answer to this and have finally decided to simulate DO SQLite storage with a D1, which helped with the storage part only... All other cool features of Durable Objects were not available in my Nuxt app. I can't wait to see this in Nitro and Nuxt.

PR #3845, opened 2025-12-08
docs(vercel): fix ISR wording

fmoessle:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/2.deploy/20.providers/vercel.md
+++ b/docs/2.deploy/20.providers/vercel.md
@@ -86,7 +86,7 @@ To revalidate a page on demand:
 
 ### Fine-grained ISR config via route rules
 
-By default, query params are considered when caching ISR routes.
+By default, query params affect cache keys but are not passed to the route handler unless specified.
 
 You can pass an options object to `isr` route rule to configure caching behavior.
 


docs/2.deploy/20.providers/vercel.md
```diff
@@ -86,7 +86,7 @@ To revalidate a page on demand:
 
 ### Fine-grained ISR config via route rules
 
-By default, query paramas are ignored by cache.
+By default, query params are considered when caching ISR routes.
```
pi0: I think "ignore" might be more precise. Unless they are specified, ISR route does not uses them in rendering.

fmoessle: I see your point about rendering—the route doesn't use query params unless specified. But from a caching perspective, they're not ignored: ?a=1 and ?a=2 produce separate cache entries. The original wording "ignored by cache" could mislead users into thinking query params don't affect cache keys at all.
Maybe something like: "By default, query params affect cache keys but are not passed to the route handler unless specified"?

pi0: Makes better sense.

fmoessle: Updated—thanks for the clarification.

PR #3861, opened 2025-12-11
feat: mock runtime virtual imports

pi0:
Nitro runtime depends on generated virtual modules that only work during the build phase. 

This makes unit testing and writing shared code difficult when project code imports `nitro/runtime`, `nitro/storage`, etc (even if unused, unit testing needs manual mocks or a custom nitro env and shared code needs to be splitted)

This PR introduces stub implementations for all internal virtual modules using Node.js sub-path imports (`#nitro/virtual/*` used internally). This allows the runtime imports to work outside of the Nitro bundle.

```js
import { useStorage } from "nitro/storage"

// Normally a storage instance
// In Node.js and unit tests, this becomes a mocked empty storage instance
const storage = useStorage()
```

Outside of tests, a warning will be shown:

```
[warn] Nitro runtime imports detected without a builder or Nitro plugin. A stub implementation will be used.
```


package.json
```diff
@@ -24,11 +24,12 @@
...
     "./types": "./dist/types/index.mjs",
-    "./vite": "./dist/vite.mjs"
+    "./vite": "./dist/vite.mjs",
+    "./vite/runtime": "./dist/runtime/vite.mjs"
   },
   "imports": {
-    "#runtime/*": "./dist/runtime/internal/*.mjs",
-    "#vite-runtime": "./dist/runtime/vite-runtime.mjs"
+    "#nitro/runtime/*": "./dist/runtime/internal/*.mjs",
+    "#nitro/virtual/*": "./dist/runtime/virtual/*.mjs"
```
productdevbook, line 32: Very clean structure. ❤️

PR #3871, opened 2025-12-13
feat(cli): support opening in browser with dev command

huseeiin:
my first feature PR to nitro. might've made mistakes.
i tested and it opens `http://[::1]:3000/` in the browser instead of `http://localhost:3000`. might need changes?

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "jiti": "^2.6.1",
     "ofetch": "^2.0.0-alpha.3",
     "ohash": "^2.0.11",
+    "open": "^11.0.0",
     "oxc-minify": "^0.102.0",
     "oxc-transform": "^0.102.0",
     "srvx": "^0.9.8",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -33,6 +33,9 @@ importers:
       ohash:
         specifier: ^2.0.11
         version: 2.0.11
+      open:
+        specifier: ^11.0.0
+        version: 11.0.0
       oxc-minify:
         specifier: ^0.102.0
         version: 0.102.0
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
     "jiti": "^2.6.1",
     "ofetch": "^2.0.0-alpha.3",
     "ohash": "^2.0.11",
+    "open": "^11.0.0",
```
pi0, line 71: Ideally we should avoid additional dependencies or bundle them open adds [11 subdeps](https://npmgraph.js.org/?q=open#select=open%4011.0.0)

pi0: Wondering what vite CLI does use for this feature

huseeiin: open as well, i'm pretty sure

huseeiin: side idea: what if we base standalone nitro on vite? meaning like vite-plugin-nitro, but it only does backend and doesn't use index.html. this way we can inherit all of vite's support for `--open`, https, http2, etc.

it'd be like:
vite-plugin-nitro: plugin for nitro-based frameworks
standalone nitro: a *vite*-based framework in of itself

huseeiin: > Ideally we should avoid additional dependencies or bundle them open adds [11 subdeps](https://npmgraph.js.org/?q=open#select=open%4011.0.0)

maybe stupid question: what difference does bundling it make? the dependencies will still be there, just bundled

pi0:
Thanks for PR <3 

Closing as [open](https://npmgraph.js.org/?q=open#select=open%4011.0.0) adds lots of extra dependencies.

I love to see it reintroduced without deps. We can try to just simply invoke OS `open` command. One example is here: https://github.com/nitrojs/nitro/blob/main/src/cli/commands/docs.ts

Issue #3378, opened 2025-05-30
Support for Custom Cloudflare Durable Objects with `cloudflare_durable` preset


JhMaic:
### Describe the feature

Hi Nitro Team,

I'm exploring the use of Cloudflare Durable Objects with my Nuxt project and had a question regarding custom implementations.

**What I'm trying to achieve:**
I want to define a custom Durable Object class and access it from my API routes.

**What I've tried:**
1.  I've created a custom class, let's call it `Counter`, which extends `DurableObject` (from Cloudflare Workers types) within `server/durable-objects/Counter.ts`.
    ```typescript
    // server/durable-objects/Counter.ts
    export class Counter extends DurableObject {
      // ... implementation ...
      constructor(state: DurableObjectState, env: Env) { /* ... */ }
      async fetch(request: Request) { /* ... */ }
    }
    ```
2.  I've configured this in my `wrangler.jsonc` similar to how one would for a standard Workers project:
    ```jsonc
    // wrangler.jsonc (simplified)
    {
      "durable_objects": {
        "bindings": [
          { "name": "counter", "class_name": "Counter" }
        ]
      },
      "migrations": [
        { "tag": "v1", "new_sqlite_classes": ["Counter"] }
      ]
    }
    ```
3. So that I can asscess `counter` from another API route (e.g., `server/api/test.ts`):
    ```typescript
    // server/api/test.ts
    export default defineEventHandler(async (counter) => {
      // How to correctly get the Durable Object namespace and stub here?
      // const counter = event.context.cloudflare.env.counter;
      // const id = counter.idFromName("my-counter");
      // const stub = counter.get(id);
      // ...
    });
    ```

**Problem:**
When I attempt to deploy my application, I encounter the following error:
`Your Worker depends on the following Durable Objects, which are not exported in your entrypoint file: Counter.`.

**Question:**
1. Is it generally supported to define and use custom Durable Object classes (like my Counter example) with presets in Nitro?
2. If so, what is the correct way to ensure these custom DO classes are properly exported or recognized by Nitro's build process to avoid the deploy error?

Any pointers to relevant documentation or examples of this setup would be very helpful.
Thanks for your time and help!

Any guidance or clarification on this would be greatly appreciated!

Thanks.

### Additional information

- [ ] Would you be willing to help implement this feature?

they-cloned-me:
Similar question for me here #3398 
Following... 👀

valtlfelipe:
[...]

Issue #181, opened 2022-05-03
Universal `import.meta`


pi0:
We shall provide a universal implementation of `import.meta` with `meta.env` (alias to process.env mock) and `meta.url` (resolving to top-level entry url)

PR #3875, opened 2025-12-16
feat(cloudflare): add durable.bindingName config

onmax:
The `cloudflare-durable` preset currently requires its WebSocket Durable Object binding to be named `$DurableObject`. This adds `cloudflare.durable.bindingName` so applications can use a different Wrangler binding name. The default and exported class name remain `$DurableObject`.

```ts
// nitro.config.ts

export default defineConfig({
  preset: "cloudflare-durable",
  cloudflare: {
    deployConfig: true,
    durable: { bindingName: "MyDO" },
  },
});
```

With `cloudflare.deployConfig` enabled, Nitro generates the binding in the default and named Wrangler environments. It adds an initial SQLite migration when no migration history or Durable Object `exports` declaration exists. Existing lifecycle declarations are preserved, with validation left to Wrangler. Conflicting or remote bindings using the configured name are rejected. With generation disabled, the binding and lifecycle declarations must be configured manually.

Validation: `pnpm fmt` and the full `pnpm test` pipeline passed, including lint, build, type checking, and 1,002 tests per builder with Rollup and Rolldown. Miniflare tests exercise WebSocket connections with both default and custom binding names; configuration tests cover lifecycle declarations, named environments, and inherited presets.

Instance-name and resolver support is split into #4602. Custom Durable Object classes remain outside this PR's scope.

Related: nuxt-hub/core#722 and #3378. [Original patched reproduction](https://github.com/onmax/repros/tree/repro/nitro-3378/nitro-3378).

### DIFF
--- a/docs/2.deploy/20.providers/cloudflare.md
+++ b/docs/2.deploy/20.providers/cloudflare.md
@@ -117,7 +117,7 @@ No manual Wrangler configuration is needed. Nitro handles it for you.
 
 :read-more{title="Durable Objects" to="https://developers.cloudflare.com/durable-objects/"}
 
-This preset extends `cloudflare_module` and routes requests through a [Durable Object](https://developers.cloudflare.com/durable-objects/) instance, enabling stateful features such as WebSocket support (via [CrossWS](https://crossws.h3.dev/adapters/cloudflare#durable-objects)) and in-memory state that persists across requests.
+This preset extends `cloudflare_module` and routes WebSocket upgrades through a [Durable Object](https://developers.cloudflare.com/durable-objects/) instance using [CrossWS](https://crossws.h3.dev/adapters/cloudflare#durable-objects).
 
 ```ts [nitro.config.ts]
 import { defineConfig } from "nitro";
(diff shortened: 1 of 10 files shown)


franklin-tina:
Question about this (and maybe the current implementation of Durable Objects via the `cloudflare-durable` preset):
Does this mean we can only have one Durable Object in the Nuxt app?

Abdulla060:
The current implementation of the `cloudflare_durable` preset hard codes the durable object instance id to the value `server`.   The line in question is https://github.com/nitrojs/nitro/blob/ba5147a532ca76d5222bf23c5731a21cca164057/src/presets/cloudflare/runtime/cloudflare-durable.ts#L12

This causes the entire app to shares a single DurableObject instance. Cloudflare recommends a coordination server that manages multiple DurableObject instances since each DurableObject can only handle about 1000 requests/s [see](https://github.com/nitrojs/nitro/blob/ba5147a532ca76d5222bf23c5731a21cca164057/src/presets/cloudflare/runtime/cloudflare-durable.ts#L12). 

I believe there should also be a way to pass the instance name. In the config, the user should be able to enable/disable this feature.

Abdulla060:
@onmax Thank you very much for your work on this!

After reviewing the implementation, I have some concerns regarding how closely it aligns with the Nitro/Nuxt architecture. Specifically, I'm worried that consolidating all DO instance selection into a single file might become a bottleneck.

The following PR  [#181](https://github.com/h3js/crossws/pull/181) has just been made to `crossws` which will enable the functionality described in here.  This approach allows each route to handle its own DO selection independently and eliminates the need for external file loading, which feels a bit cleaner for this use case.

pi0x:
Rebased this onto latest `main` locally (I could not push the result, so the branch here is unchanged). It needs real conflict work: `main` removed the `cloudflare.dev` option, replaced `hasWebSocket` with `import.meta._websocket`, moved `WranglerConfig` to `@cloudflare/workers-utils`, and rewrote the Cloudflare docs page.

A few things worth addressing:

- The docs section duplicates the existing "Cloudflare Workers with Durable Objects" section on `main`, and uses the old `defineNitroConfig` / `nitro/config` import.
- The generated `durable_objects` binding has no matching `migrations` entry, so `wrangler deploy` still fails without one.
- The `cloudflare-durable` virtual template is registered in `src/build/virtual/_all.ts`, so every preset gets it; it would fit better on the preset itself. Same for re-exporting the resolver types from core `nitro/types`.
- The PR now adds `instanceName` and `resolver` too, so the title undersells it.

Also overlaps with #4341 — both touch Durable Objects; #4341 reads the same bindings for dev.

PR #3122, opened 2025-02-24
refactor: completely remove Zeabur preset

pan93412:
### 🔗 Linked issue

N/A

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

We have phased out the Zeabur Serverless mode[^1]. In Zeabur's "Containerized" mode,
`NITRO_PRESET=zeabur` introduces some unexpected behavior that prevents users
to build on our CI/CD.

[^1]: https://zeabur.com/changelogs/shared-cluster-pricing-update

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #3882, opened 2025-12-16
refactor(zeabur): update preset

pan93412:
### 🔗 Linked issue

A better implementation of #3122.

Fix ZEA-7651

### ❓ Type of change

### 📚 Description

Zeabur now uses the general "node-server" to build the Dockerfile. Therefore, we simply extend "node-server" and "static".

### 📝 Checklist

Checks:

- [x] Complete the tests of `zeabur` and `zeabur-static`.
- [x] "nuxt build" selects the "zeabur" provider and generates the same output as "node".
      ```
      $ ZEABUR=1 pnpm build
      ●  Nitro preset: zeabur
      [success] [nitro] You can preview this build using `node .output/server/index.mjs`
      ```
- [x] "nuxt generate" selects the ~~"zeabur-static"~~ "zeabur" provider and generates the same output as "static" (?)
      ```
      $ ZEABUR=1 pnpm generate
      ●  Nitro preset: zeabur
	  [success] [nitro] Generated public .output/public
      [success] [nitro] You can preview this build using `node .output/server/index.mjs`
      ```

### DIFF
--- a/src/presets/zeabur/preset.ts
+++ b/src/presets/zeabur/preset.ts
@@ -1,54 +1,8 @@
-import fsp from "node:fs/promises";
-import { writeFile } from "../_utils/fs.ts";
 import { defineNitroPreset } from "../_utils/preset.ts";
-import type { Nitro } from "nitro/types";
-import { dirname, relative, resolve } from "pathe";
-
-// https://zeabur.com/docs/advanced/serverless-output-format
 
 const zeabur = defineNitroPreset(
   {
-    entry: "./zeabur/runtime/zeabur",
-    output: {
-      dir: "{{ rootDir }}/.zeabur/output",
-      serverDir: "{{ output.dir }}/functions/__nitro.func",
-      publicDir: "{{ output.dir }}/static",
-    },
-    hooks: {
-      async compiled(nitro: Nitro) {
-        const buildConfigPath = resolve(
-          nitro.options.output.dir,
-          "config.json"
-        );
-        const cfg = {
-          containerized: false,
-          routes: [{ src: ".*", dest: "/__nitro" }],
-        };
-        await writeFile(buildConfigPath, JSON.stringify(cfg, null, 2));
-
-        // Write ISR functions
-        for (const [key, value] of Object.entries(nitro.options.routeRules)) {
-          if (!value.isr) {
-            continue;
-          }
-          const funcPrefix = resolve(
-            nitro.options.output.serverDir,
-            ".." + key
-          );
-          await fsp.mkdir(dirname(funcPrefix), { recursive: true });
-          await fsp.symlink(
-            "./" +
-              relative(dirname(funcPrefix), nitro.options.output.serverDir),
-            funcPrefix + ".func",
-            "junction"
-          );
-          await writeFile(
-            funcPrefix + ".prerender-config.json",
-            JSON.stringify({ type: "Prerender" })
-          );
-        }
-      },
-    },
+    extends: "node-server",
   },
   {
     name: "zeabur" as const,
@@ -59,17 +13,9 @@ const zeabur = defineNitroPreset(
 const zeaburStatic = defineNitroPreset(
   {
     extends: "static",
-    output: {
-      dir: "{{ rootDir }}/.zeabur/output",
-      publicDir: "{{ output.dir }}/static",
-    },
-    commands: {
-      preview: "npx serve ./static",
-    },
   },
   {
     name: "zeabur-static" as const,
-
     static: true,
   }
 );
--- a/test/presets/zeabur.test.ts
+++ b/test/presets/zeabur.test.ts
@@ -1,0 +1,35 @@
+import { existsSync } from "node:fs";
+import fsp from "node:fs/promises";
+import { resolve } from "pathe";
+import { describe, expect, it } from "vitest";
+import { setupTest } from "../tests.ts";
+
+describe("nitro:preset:zeabur-static", async () => {
+  const ctx = await setupTest("zeabur-static");
+
+  it("should not generate a server folder", async () => {
+    const contents = await fsp.readdir(resolve(ctx.outDir));
+    expect(contents).toMatchInlineSnapshot(`
+      [
+        "nitro.json",
+        "public",
+      ]
+    `);
+  });
+
+  it("output has public directory", async () => {
+    // make output directory aligned with `nuxt generate`'s
+    // output directory, which is ".output/public"
+    expect(existsSync(resolve(ctx.outDir, "public", "favicon.ico"))).toBe(true);
+  });
+});
+
+describe("nitro:preset:zeabur", async () => {
+  const ctx = await setupTest("zeabur");
+
+  it("output directory is .output/server and has index.mjs", async () => {
+    // Zeabur sets ".output/server/index.mjs" as the entry point
+    // https://github.com/zeabur/zbpack/blob/b8d76b5758fed32203cbdbf0456a9bc5c948dfc0/internal/nodejs/plan.go#L883
+    expect(existsSync(resolve(ctx.outDir, "server", "index.mjs"))).toBe(true);
+  });
+});


src/presets/zeabur/preset.ts
```diff
@@ -1,54 +1,8 @@
...
-          nitro.options.output.dir,
-          "config.json"
-        );
-        const cfg = {
-          containerized: false,
-          routes: [{ src: ".*", dest: "/__nitro" }],
-        };
-        await writeFile(buildConfigPath, JSON.stringify(cfg, null, 2));
-
-        // Write ISR functions
```
pi0, line 26: Whay happened to ISR support?

src/presets/zeabur/preset.ts
```diff
@@ -59,17 +13,9 @@ const zeabur = defineNitroPreset(
 const zeaburStatic = defineNitroPreset(
   {
     extends: "static",
-    output: {
-      dir: "{{ rootDir }}/.zeabur/output",
```
pi0, line 56: What happened to build output API? Can you confirm if zeabur **regardless** of detecting "nuxt" or not will be able to pick `.output/` artifcats? (nitro or other nitro based meta-frameworks)

pan93412:
Since both of your review comments point to the same underlying issue, I'd like to provide some additional context here to address them together.

Last year, after evaluating stability and necessity, Zeabur decided to deprecate serverless runtime support, as containerization effectively covers those use cases. Currently, all projects on Zeabur are wrapped in a Dockerfile and executed as containers. As a result, the current role of `zbpack` is limited to:

* Determining if the project is a Nitro project (currently hardcoded via `package.json`):

    https://github.com/zeabur/zbpack/blob/b8d76b5758fed32203cbdbf0456a9bc5c948dfc0/internal/nodejs/plan.go#L339-L347

* If it is, it simply starts Nitro's own `node-server`:

    https://github.com/zeabur/zbpack/blob/b8d76b5758fed32203cbdbf0456a9bc5c948dfc0/internal/nodejs/plan.go#L713-L719

* The `node-server` then handles SSR and ISR. Users can choose to fully utilize SSG by specifying [`ZBPACK_OUTPUT_DIR`](https://zeabur.com/docs/en-US/guides/nodejs#static-website-deployment) and using `nuxt generate` as the [`ZBPACK_BUILD_COMMAND`](https://zeabur.com/docs/en-US/guides/nodejs#modifying-build-and-start-commands). We do not interfere with the build artifacts further.

You might also notice that we currently hardcode the `NITRO_PRESET`. Consequently, the specific build artifacts generated by the `zeabur` preset are actually no longer being used by our platform internally. However, this preset still impacts users who choose to write their own Dockerfiles. For example:

```dockerfile
FROM node:25-slim AS builder
WORKDIR /src
COPY . .
RUN npm install -g pnpm && pnpm install --prod --frozen-lockfile
RUN pnpm run generate

FROM zeabur/caddy-static AS server
EXPOSE 8080
COPY --from=builder /src/.output/public /usr/share/caddy

```

Because Zeabur automatically injects `ZEABUR=1` during CI, `std-env` defaults to the Zeabur preset. However, since the output path for the Zeabur preset differs from standard expectations (the artifacts end up in `.zeabur/output` instead of `.output/public`), the Dockerfile code above fails.

This has caused confusion for many users who wonder, "Why does my Dockerfile build fine locally (using `node-server` or `static` presets) but fail in the Zeabur environment?" Therefore, my goal is to align the `zeabur` preset behavior more closely with our current containerized environment.

If you have any better suggestions or thoughts on how to handle this, I'd love to hear them!

pan93412:
@pi0 sorry for pinging – what's your opinion?

src/presets/zeabur/preset.ts
pan93412: yes, we don't use the proprietary folder `.zeabur` now

src/presets/zeabur/preset.ts
pan93412: handled by `node-server`

pi0:
Thanks for the explanations, @pan93412 ❤️. It makes sense if the platform is simplified by switching to a `node-server`–style preset that also works with Dockerfile templates.

My concern is that framework detection can easily fail. In [`plan.go`](https://github.com/zeabur/zbpack/blob/b8d76b5758fed32203cbdbf0456a9bc5c948dfc0/internal/nodejs/plan.go#L339-L347), the `"nitro"` (v3) NPM package is not detected (this PR is also targeting nitro v3 btw. there is not even an official Nuxt version supporting it yet, not sure how you could test). There are meta-frameworks and tools (like [undocs](https://github.com/unjs/undocs)) that use Nitro as an internal dependency, and those will fail as well. Nuxt is not the only direct dep we should be checking for.

With the current preset implementation, when a Nitro build runs in Zeabur CI, it produces a predictable output that the deployment platform can consume. I am fine if the output directory becomes `.output` or something else, but it should be detectable from build artifacts rather than relying on dependency-name detection for special behavior.

gxres042:
Vercel CLI is also detect framework by matching package:

https://github.com/vercel/vercel/blob/d4490e7ffbf47a6efb87fe3b66b6368fb0ac9d3b/packages/frameworks/src/frameworks.ts#L2365-L2395
https://github.com/vercel/vercel/blob/d4490e7ffbf47a6efb87fe3b66b6368fb0ac9d3b/packages/frameworks/src/frameworks.ts#L1520-L1588
https://github.com/vercel/vercel/blob/d4490e7ffbf47a6efb87fe3b66b6368fb0ac9d3b/packages/frameworks/src/frameworks.ts#L1908-L1944

https://github.com/vercel/vercel/blob/d4490e7ffbf47a6efb87fe3b66b6368fb0ac9d3b/packages/fs-detectors/src/detect-framework.ts

pan93412:
> My concern is that framework detection can easily fail. In [`plan.go`](https://github.com/zeabur/zbpack/blob/b8d76b5758fed32203cbdbf0456a9bc5c948dfc0/internal/nodejs/plan.go#L339-L347), the `"nitro"` (v3) NPM package is not detected (this PR is also targeting nitro v3 btw. there is not even an official Nuxt version supporting it yet, not sure how you could test). There are meta-frameworks and tools (like [undocs](https://github.com/unjs/undocs?rgh-link-date=2025-12-19T12%3A57%3A38Z)) that use Nitro as an internal dependency, and those will fail as well. Nuxt is not the only direct dep we should be checking for.

Currently, we can only list all frameworks based on Nitro and apply the same logic to them. Most zero-config application builders use this approach, including [Vercel](https://github.com/vercel/vercel/blob/d4490e7ffbf47a6efb87fe3b66b6368fb0ac9d3b/packages/fs-detectors/src/detect-framework.ts) and [Railway's Railpack](https://github.com/railwayapp/railpack/blob/4046c5db2ce79c90d389f2699ba365a9d77e2d2f/core/providers/node/node.go#L591). I think it is fine at the moment.

Nitro v3 will be implemented after I have cleaned up the logic in zbpack.

> With the current preset implementation, when a Nitro build runs in Zeabur CI, it produces a predictable output that the deployment platform can consume. I am fine if the output directory becomes `.output` or something else, but it should be detectable from build artifacts rather than relying on dependency-name detection for special behavior.

Yes. Once we declare that a framework (e.g., Nuxt, Undocs, TanStack Start) is based on Nitro, they will all follow the same bundle logic in `zbpack`, as their outputs are predictable. The only thing we need to maintain is the list of Nitro frameworks.

pi0:
Some providers also detect "some" Nitro based frameworks however zero config detection build output and docs are based on assumption that platform's framework detection is not a hard requirement.

Maintaining a list of known pkg names per each provider is not desired nor practical honestly. The fact that official "nitro" package itself(!) is missing today (and our e2e deploys failed without notice) is a proof of this. Anyone should be able to make a Nitro based tool that just works everywhere without waiting.

Regardless, i will properly review platform support as soon as could and likely defaulting to something guaranteed to work on zeabur even if detection fails.

Feel free to reach me out in discord (`@pi0`) to chat more.

pi0x:
I rebased this onto latest `main` locally. The conflict in `src/presets/zeabur/preset.ts` was only reformatting on `main`, so I kept your version. Typecheck and lint pass, and `test/presets/zeabur.test.ts` passes (3 tests). I could not push, so the branch here is unchanged.

One leftover: dropping `entry` leaves `src/presets/zeabur/runtime/zeabur.ts` unreferenced. Nothing else imports it, so I deleted it.

The docs page does not mention the old `.zeabur/output` layout, so no docs change is needed.

Nothing else looked wrong to me. The open question is still the one @pi0 raised about zbpack detection, which is a product decision rather than a code issue.

PR #3886, opened 2025-12-18
fix: use `module` condition and main fields for bundling

pi0:
Before runtimes adopting ESM, bundlers and package authors made up a convention to use top level `module` field to indicate bundler-friendly-ESM entry. This convention then confused and added to the `exports` object as an export condition.

Neither of the export condition or field is respected by any runtime but only by bundlers. Standard is the `exports` field and `import` condition.

This PR adds `module` to default conditions to be consistent with rolldown and esbuild (https://github.com/rolldown/rolldown/pull/4703) and fixing bundling issues.

### DIFF
--- a/src/config/resolvers/export-conditions.ts
+++ b/src/config/resolvers/export-conditions.ts
@@ -43,7 +43,7 @@ function _resolveExportConditions(
   }
 
   // 5. Add default conditions
-  // "module" is NOT A STANDARD CONDITION but some pOpUlAr packages like tslib use it and can help fixing bundling issues.
+  // "module" is NOT A STANDARD CONDITION but widely used in the ecosystem adding helps with compatibility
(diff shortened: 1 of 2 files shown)


src/config/resolvers/export-conditions.ts
```diff
@@ -43,7 +43,8 @@ function _resolveExportConditions(
   }
 
   // 5. Add default conditions
-  resolvedConditions.push("import", "default");
+  // "module" is NOT AND STANDARD CONDITION but some pOpUlAr packages like tslib use it and can help fixing bundling issues.
```
danielroe:
```suggestion
  // "module" is NOT A STANDARD CONDITION but some pOpUlAr packages like tslib use it and can help fixing bundling issues.
```

PR #3904, opened 2025-12-30
feat(cloudflare): always augment request with cf context

emretinaztepe:
Currently, there is no way to access or pass Cloudflare env and bindings for websockets. 

This PR passes the Cloudflare environment and context to the CrossWS peer, making bindings accessible within WS handlers.

### DIFF
--- a/src/presets/cloudflare/runtime/_module-handler.ts
+++ b/src/presets/cloudflare/runtime/_module-handler.ts
@@ -115,10 +115,7 @@ export function augmentReq(
   ctx: NonNullable<ServerRuntimeContext["cloudflare"]>
 ) {
   const req = cfReq as ServerRequest;
-  req.waitUntil = ctx.context?.waitUntil.bind(ctx.context);
   req.runtime ??= { name: "cloudflare" };
-  req.runtime.cloudflare = {
-    ...req.runtime.cloudflare,
-    ...ctx,
-  };
+  req.runtime.cloudflare = { ...req.runtime.cloudflare, ...ctx };
+  req.waitUntil = ctx.context?.waitUntil.bind(ctx.context);
 }


pi0:
Thanks. This looks like a nice enhancenment, however,  it seems we can always inject env and context in [shared fetch handler](https://github.com/emretinaztepe/nitro/blob/8211f4af76fe40a16b3df0c4d227da7b2156cfc5/src/presets/cloudflare/runtime/_module-handler.ts#L24) to request object so it works both in ws upgrades and nornmal paths (we seem missing it in v3)

emretinaztepe:
@pi0 Any plans to add it to the shared fetch handler soon? I am currently using a local pnpm patch to work around it and would love to remove it as soon as it is officially supported.

pi0:
I think you can refactor your PR to do this so we can merge/release it asap!

emretinaztepe:
@pi0 I believe it should be fixed for all handlers to be available for all Cloudflare Worker entrypoints such as email, scheduled, etc. What do you think?

pi0:
We could make an internal util like `augmentContext(req, env, ctx)` perhaps and call it in all?

emretinaztepe:
@pi0 That makes sense, but I am afraid I am not familiar with the implementation of other presets. I will push a refactored version. Please feel free to modify it if it doesn't make sense.

src/presets/cloudflare/runtime/_module-handler.ts
```diff
@@ -1,13 +1,46 @@
...
+  req.runtime ??= { name: "cloudflare" };
+  (req.runtime as any).cloudflare = {
+    ...(req.runtime as any).cloudflare,
+    env,
+    context,
+  };
+  req.waitUntil = context.waitUntil.bind(context);
+
+  // crossws compatibility: request.context is forwarded to peer.context
+  let reqContext = (req as any).context;
```
pi0: `req` object is always accessable in crossws peers we can simplify to use single context instead

pi0 (approved):
Thanks! Made few adjustments, after this, `req.runtime.` and `peer.req.runtime.` both should work always.

We might need some follow-up fixes in crossws too. When a durable instance is hydrated, crossws makes a new fallback req [here](https://github.com/h3js/crossws/blob/440a0884e8ec6785555afe7e644be2399e093d2b/src/adapters/cloudflare.ts#L280) (since the original is only on first wake) and we need to inject context to it.

emretinaztepe:
@pi0 Thank you! Not sure if it is the right place to share it, but another major issue that is on more than a dozen forums, unsolved, is that Nitro does not provide any mechanism to export Durable Objects. This prevents the use of Durable Objects in Nitro and Nuxt in dev mode. Are there any plans to address this issue? Cloudflare Workers expect DOs to be exported, and the below workaround makes it work in Nuxt preview/build mode, but after two months of research, I had no luck getting it to work in dev mode. 

Below is the only workaround I was able to make it work for preview/build mode by setting wrangler's main property to "server/worker.ts"
[worker.ts](https://github.com/user-attachments/files/24524872/worker.ts)

Any update or future roadmap info would be useful.

pi0:
We recently supported `exports.cludflare.ts` entry (see https://github.com/nitrojs/nitro/pull/3834)

PR #3919, opened 2026-01-05
feat(vite): experimental rsc plugin support

hi-ogawa:
### 🔗 Linked issue

- Closes https://github.com/vitejs/vite-plugin-react/issues/876
- Closes https://github.com/nitrojs/nitro/pull/3736 (supersedes)

`@vitejs/plugin-rsc` now assumes `__VITE_ENVIRONMENT_RUNNER_IMPORT__` global as convention for `import.meta.viteRsc.loadModule` implementation (see https://github.com/vitejs/vite-plugin-react/pull/1037). This PR defines the global inside nitro dev worker so RSC plugin works out of the box. (NOTE: This is not released yet, so using pkg.pr.new in the demo. After this approach is confirmed with Nitro, I can release new version)

I noticed `services` config is removed, so I added to `experimental.vite.services`. RSC plugin still assumes `rollupOptions.input.index`, so adjusted `createServiceEnvironment` too.

I added `examples/vite-rsc` for the demo.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/test/examples.test.ts
+++ b/test/examples.test.ts
@@ -8,15 +8,17 @@ import type { ViteDevServer } from "vite";
 
 const examplesDir = fileURLToPath(new URL("../examples", import.meta.url));
 
-const { createServer, createBuilder } = (await import(
+const { createServer, createBuilder, rolldownVersion } = (await import(
   process.env.NITRO_VITE_PKG || "vite"
 )) as typeof import("vite");
+
+const isRolldown = !!rolldownVersion;
 
 const skip = new Set<string>(["websocket"]);
 
 const skipDev = new Set<string>(["auto-imports", "cached-handler"]);
 
-const skipProd = new Set<string>();
+const skipProd = new Set<string>(isRolldown ? [] : ["vite-rsc"]);
 
 for (const example of await readdir(examplesDir)) {
   if (example.startsWith("_")) continue;


hi-ogawa:
It looks like vite-rsc example is failing with rollup. I'm not sure how to test this locally. :thinking: 
(`pnpm test:rollup test/examples.test.ts -t vite-rsc` is passing locally)

src/build/vite/env.ts
```diff
@@ -68,7 +68,7 @@ export function createServiceEnvironment(
   return {
     consumer: "server",
     build: {
-      rollupOptions: { input: serviceConfig.entry },
+      rollupOptions: { input: { index: serviceConfig.entry } },
```
pi0, line 71: it would be nice if plugins support both `string` and `{ index: string }` formats..

test/examples.test.ts
```diff
@@ -8,15 +8,17 @@ import type { ViteDevServer } from "vite";
...
 )) as typeof import("vite");
 
+const isRolldown = !!rolldownVersion;
+
 const skip = new Set<string>(["websocket"]);
 
 const skipDev = new Set<string>(["auto-imports", "cached-handler"]);
 
-const skipProd = new Set<string>();
+const skipProd = new Set<string>(isRolldown ? [] : ["vite-rsc"]);
```
pi0, line 21: @hi-ogawa i have excluded example (prod only) from matrix for now.

Locally, you can use 

```sh
NITRO_VITE_PKG=vite7  pnpm vitest test/examples.test.ts -t vite-rsc
```

i cannot reproduce issue outside vitest...

examples/vite-rsc/app/root.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <ul className="read-the-docs">
+        <li>
+          Edit <code>src/client.tsx</code> to test client HMR.
+        </li>
+        <li>
+          Edit <code>src/root.tsx</code> to test server HMR.
+        </li>
+        {/* <li>
+          Visit{" "}
+          <a href="?__rsc" target="_blank">
```
pi0, line 62: This seems not implemented

pi0:
Merging for faster iteration! LMK or feel free to make a followup PR to upgrade plugin to stable version.

examples/vite-rsc/app/root.tsx
hi-ogawa: Oh good catch. I fixed my template https://github.com/vitejs/vite-plugin-react/pull/1041. Will fix it here in next PR on nitro.

src/build/vite/env.ts
hi-ogawa: This should be now possible with https://github.com/vitejs/vite-plugin-react/pull/1040, but I don't think enforcing string form on Nitro is a good direction (though I'm not sure whether hard-coding `index` is good either, so maybe nitro specific entry name?). 

If Nitro has need for this to be single entry, then RSC plugin and rsc example can adjust it, but not sure if framework can adjust it easily. Please let me know if reverting it back `{ input: serviceConfig.entry }` is necessary for Nitro.

pi0: Nitro supports both.

PR #3926, opened 2026-01-07
fix(dev): apply headers from route rules for static assets

typed-sigterm:
from #2814

for triggering pkg.pr.new release

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/core/dev-server/server.ts
+++ b/src/core/dev-server/server.ts
@@ -247,7 +247,8 @@ class DevServer {
               );
               if (rules.headers) {
                 for (const [k, v] of Object.entries(rules.headers)) {
-                  if (k !== 'cache-control') // avoid long caching dev assets
+                  if (k !== "cache-control")
+                    // avoid long caching dev assets
                     res.appendHeader(k, v);
                 }
               }


src/core/dev-server/server.ts
```diff
@@ -227,6 +238,18 @@ class DevServer {
...
+              if (!res.req.url) {
+                return;
+              }
+              const rules: NitroRouteRules = defu(
+                {},
+                ...routeRulesMatcher.matchAll(res.req.url).reverse()
+              );
+              if (rules.headers) {
+                for (const [k, v] of Object.entries(rules.headers)) {
+                  res.appendHeader(k, v);
```
pi0: Lef over from #2814, we should filter `cache-control` route rule header to avoid long caching dev assets

typed-sigterm: fixed. Should we do `k.toLowerCase()`? I'm not sure if radix handled this

src/core/dev-server/server.ts
```diff
@@ -227,6 +238,20 @@ class DevServer {
               if (path.endsWith(".gz")) {
                 res.setHeader("Content-Encoding", "gzip");
               }
+              if (!res.req.url) {
```
pi0, line 241: url always exists

src/core/dev-server/server.ts
pi0: yeah lets explictly lower-case when comparing

PR #3927, opened 2026-01-07
fix(vercel): avoid adding o11y links for prerendered paths

pi0:
(reported by @atinux and observed in docus ([repro](https://github.com/nuxt-content/docus/blob/c4485833c1630c8823de006af4308fbbccefcda4/layer/nuxt.config.ts#L58-L59))

We should avoid making o11y symlinks for paths matching (exactly) function name to avoid bypassing filesystem handler.


atinux (approved):
❤️

PR #3928, opened 2026-01-07
feat(vite): auto-register server consumer environments as services

schiller-manuel:


src/build/vite/plugin.ts
```diff
@@ -383,6 +383,32 @@ async function setupNitroContext(
     ctx.nitro!.routing.sync();
   }
 
+  // Auto-register server consumer environments as services
+  for (const [envName, envConfig] of Object.entries(
+    userConfig.environments || {}
+  )) {
+    if (
+      envName === "ssr" ||
+      envName === "client" ||
```
pi0: I think we don't need to hardcode "client". only "nitro" to be excluded and all `consumer === "server"` envs with entry register as service.

src/build/vite/plugin.ts
```diff
@@ -383,6 +383,32 @@ async function setupNitroContext(
...
+    const entry = getEntry(envConfig.build?.rollupOptions?.input);
+    if (typeof entry === "string") {
+      const resolvedEntry =
+        resolveModulePath(entry, {
+          from: [ctx.nitro.options.rootDir, ...ctx.nitro.options.scanDirs],
+          extensions: DEFAULT_EXTENSIONS,
+          suffixes: ["", "/index"],
+          try: true,
+        }) || entry;
+      ctx.services[envName] = { entry: resolvedEntry };
```
pi0: We should skip already registered services. This way we can also remove explicit `"ssr"` exclude above

PR #3940, opened 2026-01-11
fix(prerender): skip writing redirect responses to disk

Flo0806:
### 🔗 Linked issue

nuxt/nuxt#33924

### ❓ Type of change

### 📚 Description

While investigating nuxt/nuxt#33924 together with @danielroe, we found that redirect responses were being written to disk as extensionless files during prerendering.

**The problem:** When a route returns a redirect (e.g., due to `baseURL` handling), Nitro writes a file like `/admin-dashboard` with just `"Redirecting..."` as content. Later, when the actual HTML page is generated with `autoSubfolderIndex`, Nitro tries to create `/admin-dashboard/index.html` - but `mkdir("/admin-dashboard")` fails because a file with that name already exists.

**The fix:** Simply skip writing redirect responses to disk. They don't serve any purpose as static files anyway - redirects on static hosting are handled through server configuration (`_redirects`, `.htaccess`, etc.), not HTML files.

### 📝 Checklist

### DIFF
--- a/src/core/prerender/prerender.ts
+++ b/src/core/prerender/prerender.ts
@@ -245,10 +245,16 @@ export async function prerender(nitro: Nitro) {
 
     // https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
     const redirectCodes = [301, 302, 303, 304, 307, 308];
+    const isRedirect = redirectCodes.includes(res.status);
     if (![200, ...redirectCodes].includes(res.status)) {
       _route.error = new Error(`[${res.status}] ${res.statusText}`) as any;
       _route.error!.statusCode = res.status;
       _route.error!.statusMessage = res.statusText;
+    }
+
+    // Skip writing redirect responses to disk - they would conflict with actual HTML files
+    if (isRedirect) {
+      _route.skip = true;
     }
 
     // Measure actual time taken for generating route


src/core/prerender/prerender.ts
```diff
@@ -245,12 +245,18 @@ export async function prerender(nitro: Nitro) {
...
     // https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
     const redirectCodes = [301, 302, 303, 304, 307, 308];
+    const isRedirect = redirectCodes.includes(res.status);
     if (![200, ...redirectCodes].includes(res.status)) {
       _route.error = new Error(`[${res.status}] ${res.statusText}`) as any;
       _route.error!.statusCode = res.status;
       _route.error!.statusMessage = res.statusText;
     }
 
+    // Skip writing redirect responses to disk - they would conflict with actual HTML files
```
danielroe, line 255: we probably do want to write redirects to disk - it's a feature

what we want to avoid is writing redirects that are triggered because of a base url (because they will then be moved into a sub folder and overlap with the base url-prefixed routes)

can you reproduce the issue with nitro alone (not nuxt)?

Flo0806: Ok... understand. And I found the error. Nitro is fine - the problem comes 100% from `ufo`. I fixed the ` withoutBase` method, but I forgot the `withBase`.... I will close this issue.

PR #3941, opened 2026-01-12
feat(types): speed up typed routes

dannote:
## Summary
This switches typed route generation + matching to a nested route tree, so `$fetch` types walk path segments instead of scoring every route key. It keeps method narrowing intact while avoiding the deep type instantiation that slows or breaks TypeScript on larger APIs.

## What changed
- Generate a nested `InternalApi` with a `$` method bucket at each leaf (`src/build/types.ts`).
- Replace score-based matching with a segment-walk matcher that normalizes paths and resolves exact -> param -> catch-all (`src/types/fetch/_match.ts`).
- Recompute `NitroFetchRequest` as a union of nested route paths and use matched method maps for `$fetch` narrowing (`src/types/fetch/fetch.ts`).
- Add type-level fixtures for exact, param, and catch-all cases (`test/fixture/typed-routes.types.ts`).

## Why this helps
The previous flat route map required comparing every route on each `$fetch` call (O(n x path depth)). The nested tree resolves in O(path depth) and dramatically reduces TypeScript instantiations.

## Related issues
- https://github.com/nitrojs/nitro/issues/470
- https://github.com/nitrojs/nitro/issues/2758

## Elysia/Eden inspiration
The nested route tree idea is inspired by Elysia's `CreateEden` and Eden's path mapping utilities:
- https://github.com/elysiajs/elysia/blob/main/src/types.ts#L2090-L2113
- https://github.com/elysiajs/eden/blob/main/src/types.ts#L87-L100

## Tests
- `pnpm lint:fix`
- `pnpm test:types`
- `pnpm test` (fails in `test:rollup` with Vitest unhandled `ECONNRESET` / "Server is not running"; `test:rolldown` did not run)


Tenrys:
Super interested in this, is there a way I could test this in advance in a current project of mine?

huseeiin:
> Super interested in this, is there a way I could test this in advance in a current project of mine?

@pi0 add pkg.pr.new support this repository?

Tenrys:
I'm seeing now that https://github.com/unjs/fetchdts is being developed as a replacement for the current system, does that invalidate this pull request entirely? How long can we expect to wait until it's ready?

I would really like to have an idea of what to expect and what's the best course of action at the moment, because as of recent comments on the related issue, it seems like it's standard to flat out disallow generation of nitro typed routes in large projects. While I can get behind that, for performance's sake, I do use `FetchResult` and rely on types coming from `$fetch` to typecheck my app... 

@pi0

timhanlon:
FWIW, I just pointed Claude Code at this PR and asked it to patch Nitro using pnpm, then validate it against a project of mine which was hitting the instantiation depth limits. Worked beautifully, here's its summary:

```
Here’s a summary of what was done:

Patch created: patches/nitropack@2.13.1.patch

Applied the nested route tree technique from nitrojs/nitro#3941 to nitropack@2.13.1. Three files patched:

dist/core/index.mjs — Changed route generation from flat '/api/:orgId/entity': { get: ..., patch: ... } to a nested tree: api: { ':orgId': { entity: { $: { get: ..., patch: ... } } } }. Routes now share path segments.

dist/types/index.d.ts + index.d.mts — Replaced the O(n × path depth) score-based matcher (CalcMatchScore, MaxTuple, etc.) with an O(path depth) segment-walk matcher that resolves exact → param → catch-all. Also changed NitroFetchRequest to compute route paths from the nested tree.

All ~150 routes are now included in the generated types (1248 lines) without hitting TS instantiation depth limits.
```

Happy to help move this forward in any way possible.

Tenrys:
I went ahead and did pretty much the same thing as @timhanlon, only asking OpenCode (using some Claude model, with my Copilot subscription) to do it instead. I hadn't thought of doing this, but it worked wonders for me!

pi0:
I appreciate the PR. Nitro v3 is removing the typed fetch from core (https://github.com/nitrojs/nitro/pull/4572) and Nuxt using new better fetchdts base for typed fetch.

PR #3942, opened 2026-01-13
chore: add `vite-better-auth` example

pi0:


pi0x:
Rebased on latest `main` and pushed. Only conflict was `pnpm-lock.yaml`; took main's copy and re-ran `pnpm install`. Still a draft, I left that alone.

It builds, and sign-up works once the database is migrated. Fixed in one extra commit: the SSR head rendered `<link type="modulepreload">` instead of `rel="modulepreload"`, so nothing was actually preloaded. Also added the example to the root `tsconfig.json` exclude list, because the `@/*` alias is not visible to the root typecheck (same as `import-alias`), and ran `pnpm fmt`.

Left for you, since it is your draft:

- `pnpm auth:migrate` fails. The CLI reads `auth.ts`, sees `export default { fetch: auth.handler }` and rejects it. It works if the auth instance lives in its own file.
- `secret: "supersecretkey"` is hardcoded, and better-auth warns about it at startup.
- `vite.config.ts` sets `serverDir: "./server"` but there is no `server/` folder.
- No README and no docs page yet, and no `vite` in the example dependencies.

PR #3952, opened 2026-01-15
feat(vercel, netlify, cloudflare): support `req.ip`

huseeiin:
add more ip support :)


src/presets/vercel/runtime/vercel.node.ts
```diff
@@ -12,6 +12,11 @@ export default function nodeHandler(
   req: NodeServerRequest,
   res: NodeServerResponse
 ) {
+  // replace req.socket.remoteAddress with x-forwarded-for because req.socket.remoteAddress is not accurate and req.socket.remoteAddress is how the srvx node adapter reads the ip. this is subject to debate
+  Object.defineProperty(req.socket, "remoteAddress", {
+    value: req.headers["x-forwarded-for"],
```
pi0: /cc @QuiiBz do you think of any better way for Node.js style handler? (can't we rely on original `remoteAddress`? i see it is currently always `127.0.0.1`)

src/presets/vercel/runtime/vercel.web.ts
```diff
@@ -26,11 +26,15 @@ export default {
...
-    // srvx compatibility
-    req.runtime ??= { name: "vercel" };
+    req.runtime = {
+      name: "vercel",
+      // @ts-expect-error (add to srvx types)
+      vercel: { context },
+    };
+
+    // there's also x-vercel-forwarded-for, x-vercel-proxied-for, x-real-ip
+    req.ip = req.headers.get("x-forwarded-for") || undefined;
```
pi0: `x-forwarded-for` (by-spec) might be comma seperated and we need to only pick first

pi0 (approved):

PR #3956, opened 2026-01-17
fix netlify caching based on pathname instead of pathname + search

huseeiin:
fix for a major bug i found in nitro v2 where netlify caches based on PATHNAME only so:
when you request `/user?name=hussein` you get hussein
when you request `/user?name=pooya` you get hussein
when you request `/user` you get hussein
because it cached the first request you made.


huseeiin:
fix didn't work and didn't find another way. closing because too hard

huseeiin:
if anyone wants to continue this, it can be fixed by `Netlify-Vary: query`

PR #3962, opened 2026-01-20
refactor: include name of hoisted libs in chunk name

pi0:
rolldown sometimes prefers to create a group of multiple libraries in a single chunk (hoisting)

This PR makes the chunk name clearer by adding all names concatenated by `+`. 

To avoid potential fs limits, after hitting 30 chars, we switch to hash

image

### DIFF
--- a/build.config.ts
+++ b/build.config.ts
@@ -108,7 +108,7 @@ export default defineBuildConfig({
                     )?.groups?.package
                 )
                 .filter(Boolean)
-                .map((name) => name!.split("/").pop()!)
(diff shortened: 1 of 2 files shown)


productdevbook:
What does the rolldown team think about this?

pi0:
It is a configurable behavior up to consumer in this context :)

productdevbook:
So it wasn't a bug then, right?

pi0:
No, hoisting is actually a very important feature of rolldown :D (just names are more readable after this)

Issue #3669, opened 2025-10-21
Autoimports' `typeFrom` property not used to generate declaration files


8ctavio:
### Environment

Nitro: 2.12.7
Node: v20.12.1

```ts
export default defineNitroConfig({
  srcDir: "server",
  compatibilityDate: "2025-10-21",
  imports: {
    imports: [
      {
        name: "foo",
        from: resolve(import.meta.dirname, "./pkg/src/index.js"),
        typeFrom: resolve(import.meta.dirname, "./pkg/dts/index.d.ts"),
      },
    ],
  },
});
```

### Reproduction

[Reproduction link](https://codesandbox.io/p/github/nitrojs/nitro/csb-x892rv/bug/generate-declarations?file=%2Fnitro.config.ts)

- Run `pnpm prepare`.
- Inspect `.nitro/types/nitro-imports.d.ts`.
- Look for a `foo` const; its type is `string`, but should be `string | number | boolean` (see `pkg/dts/index.d.ts`).

### Describe the bug

Nitro does not use the `typeFrom` option of registered autoimports to generate declaration files, which may cause type declarations to not be found.

Nitro uses unimport's `generateTypeDeclarations` to generate declaration files' content, which supports the `typeFrom` option. However, a custom resolver that is [missing `typeFrom` support](https://github.com/nitrojs/nitro/blob/d1eea57b1bdc596de415d4bfdd5515ff7390c895/src/build/types.ts#L97) is provided.

### Additional context

Nuxt had the same issue for app autoimports: https://github.com/nuxt/nuxt/pull/33373.

### Logs

```sh

```

PR #3963, opened 2026-01-20
fix: add `typeFrom` support for autoimports exports generation

8ctavio:
### 🔗 Linked issue

#3669 part 2

### ❓ Type of change

### 📚 Description

PR #3670 added imports' `typeForm` support for global declarations; this PR adds `typeFrom` support for autoimports' `export` statements, so that explicit imports from `#imports` may also find type declarations.

Inside the `writeTypes` function, `toExports` call is updated to use imports' `typeFrom` property when available (see https://github.com/unjs/unimport/pull/477).

### 📝 Checklist

### DIFF
--- a/src/core/build/types.ts
+++ b/src/core/build/types.ts
@@ -50,10 +50,9 @@ export async function writeTypes(nitro: Nitro) {
 
     const allImports = await nitro.unimport.getImports();
 
-    autoImportExports = toExports(allImports, undefined, false, { declaration: true }).replace(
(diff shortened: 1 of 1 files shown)


8ctavio:
I think changing `toExports`'s third parameter (`includeType`) to `true` should be considered as well. Currently, the generated `nitro-imports.d.ts` only re-exports values. With `includeType` set to `true`, types are re-exported too, and it would be possible to import them from `#imports`.

PR #3964, opened 2026-01-21
fix: use static replacement for websocket feature

pi0:
We use static flags from a virtual import + `optimization.inlineConst` + `minify: dce-only` to enable rolldown / rolldown-vite to tree-shake unused logic.

However it seems not effective on entrypoints where we conditionally import crossws for opt-in websocket feature (/cc @sapphi-red it is reprodicable with both rolldown and vite8 with any nitro nightly before this)

This PR worksaround issue by using a replacement via `import.meta.`.


sapphi-red:
@pi0 Is there an example I can take a look?

Issue #2973, opened 2025-01-07
Preset support tracker


pi0:
This issue is a unified tracker for requested deployment presets.

- Please upvote 👍🏼 original (closed) issues to support and comment if you would like to work on them
- Please be patient with WIP items. PRs often need redo and other works for long-term maintenance
- Please comment (here) if you would like to propose a new deployment preset (new platform) support
 
| Name | Original Issue | Documentation Link | Draft PR |
|  - |  - |  - |  - |
| Spin / Fermyon Cloud | #2844 | [Docs](https://developer.fermyon.com/spin/v2/javascriptcomponents) |  - |
| appwrite.io | #2778 | [Docs](https://appwrite.io/docs/products/functions) |  - |
| open-runtimes | #2449 | [Docs](https://github.com/open-runtimes/open-runtimes) | -
| azure -app -service | #2231 |  - |  - |
| akash network | #1877 | [Docs](https://akash.network/) |  - |
| wasmer edge | #1866 | [Docs](https://wasmer.io/products/edge) |  - |
| Plesk | #1819 |  - |  - |
| Spiderfire | #1781 |  - |  - |
| azion | #1414 | [Docs](https://www.azion.com/en/) |  - |
| docker | #54 |  - |  - |
| fastly edge | #1426 | [Docs](https://www.fastly.com/documentation/guides/compute/javascript/) |  - |
| scaleway functions | #1261 | [Docs](https://www.scaleway.com/en/serverless-functions/) |  - |
| edgegap | #156 | [Docs](https://edgegap.com/) |  - |
| AWS Lambda behind Load Balancer (ALB) | #2072 |  - | #2914 |
| coolify | #2762 |  - | #2767 |
| AWS CDK | #1387 |  - | #240 (stalled) |
| AWS lambda edge | #79 | [Docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-edge.html) | #1557 (stalled) |
| SST | #133 - #2236 | -
| Clever Cloud | -  | [Docs](https://www.clever-cloud.com/developers/doc/applications/javascript/nodejs/) | -
| Tencent’s EdgeOne | #3234 | [Docs](https://edgeone.ai/) | -
| Bare | [-](https://github.com/wintercg/runtime-keys/issues/22) | [Docs](https://bare.pears.com/) | -
[...]

PR #3970, opened 2026-01-23
feat: add Scaleway serverless function preset

Siilwyn:
### 🔗 Linked issue

#2973 

### ❓ Type of change

### 📚 Description

Add Scaleway serverless functions preset.

### 📝 Checklist

## Other
Open to all feedback, I'm figuring out what to add further to the documentation and preset config so deploys go as easy as possible.


Siilwyn:
Moved from draft so CodeRabbit picks it up.

src/presets/scaleway/runtime/scaleway-serverless.ts
```diff
@@ -0,0 +1,31 @@
...
+      event.path
+    ),
+    event.queryStringParameters!
+  );
+
+  const request = new Request(url, {
+    method: event.httpMethod,
+    headers,
+    body: event.httpMethod === "GET" ? undefined : event.body,
+  });
```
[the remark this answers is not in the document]
Siilwyn: Since all tests pass I don't think this is needed?
- `event.queryStringParameters` can be `null` because `withQuery` handles this.
- The expected headers are set by the Scaleway runtime afaik, so `X-Forwarded-Proto` should always be in this casing.
- Maybe there are no tests for `isBase64Encoded`, is this expected to work?
- Are `HEAD` requests without a body also not tested?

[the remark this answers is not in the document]
Siilwyn: Fixed the query handling and header casing.
The base 64 encoding is indeed in the type and documentation but I don't see a way to configure this.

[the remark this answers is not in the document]
Siilwyn: Suggestions are applied, headers are not case sensitive and base 64 encoding is handled.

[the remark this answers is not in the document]
Siilwyn: @coderabbitai resume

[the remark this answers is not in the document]
Siilwyn: @coderabbitai full review

src/presets/scaleway/preset.ts
```diff
@@ -0,0 +1,18 @@
...
+    entry: "./scaleway/runtime/scaleway-serverless",
+    rollupConfig: {
+      output: {
+        entryFileNames: "index.mjs",
+        format: "esm",
+      },
+    },
+  },
+  {
+    name: "scaleway-serverless" as const,
```
pi0: We might choose a more clearname for this, perhaps We might choose a better name perhaps `scaleway-functions`?

(since we also have Scaleway [serverless containers](https://www.scaleway.com/en/serverless-containers/) product)

src/presets/scaleway/runtime/scaleway-serverless.ts
```diff
@@ -0,0 +1,34 @@
...
+    Object.fromEntries(Object.entries(event.headers!).map(([key, value]) => [key, String(value)]))
+  );
+
+  const url = withQuery(
+    joinURL(
+      headers.get("X-Forwarded-Proto") === "http" ? "http://" : "https://",
+      headers.get("Host")!,
+      event.path
+    ),
+    event.queryStringParameters ?? {}
```
pi0, line 22: Can you please refactor to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams to stringify `queryStringParameters`? (also don't we have access to original request string?)

src/presets/scaleway/runtime/scaleway-serverless.ts
```diff
@@ -0,0 +1,34 @@
+import "#nitro/virtual/polyfills";
+import { useNitroApp } from "nitro/app";
+import { joinURL, withQuery } from "ufo";
+import type { serveHandler } from "@scaleway/serverless-functions";
+import { Buffer } from "node:buffer";
```
pi0: `Buffer` should be available globally

src/presets/scaleway/preset.ts
```diff
@@ -0,0 +1,18 @@
+import { defineNitroPreset } from "../_utils/preset.ts";
+
+const scalewayServerless = defineNitroPreset(
+  {
```
pi0, line 4: we might enable serve static handler (with inlined contents) or explain it in docs (containers have better support for static assets i guess also)

src/presets/scaleway/preset.ts
Siilwyn: Yes makes sense, done in fb41dc04fa95339e686501357ab3a40916be9fcc

src/presets/scaleway/runtime/scaleway-serverless.ts
Siilwyn: I like adding these instead of relying on globals but all good, removed in 32ca17a7

pi0x:
I rebased this onto latest `main` locally (conflicts in `package.json`, `pnpm-lock.yaml` and the generated preset files). I could not push, so the branch here is unchanged.

Two blockers found:

- `pnpm install` now fails. `main` enabled `trustPolicy: no-downgrade`, and `@scaleway/serverless-functions` pulls `pino@9.14.0` and `undici-types@6.21.0`, which both trip it. It needs `trustPolicyExclude` entries in `pnpm-workspace.yaml`, or a test that does not need this package.
- After the rebase, 5 shared preset tests fail (route-rule CORS, encoded `%2f` in proxy paths, custom status text, `QUERY` method). Some look like limits of the local emulator, but the `%2f` one is real: rebuilding the URL from `event.path` loses the original encoding. That is the same point @pi0 raised about using the original request string.

Also still open: @pi0's asks about `URLSearchParams` and static assets. The docs page has no deploy steps yet.

PR #3974, opened 2026-01-24
presets(aws-lambda): avoid modifying response headers in streaming mode

pi0:
(In an investigation with @sommeeeer and https://github.com/h3js/h3/pull/1287)

AWS lambda (streaming) preset tries to modify web response headers which might be frozen.

This PR simply updated implementation to add missing `transfer-encoding` only on the final lambda headers object.


sommeeeer:
Thank a lot @pi0 for the collabration on this. I'll keep you updated if this fixes the frozen header problem on our side!

Update: Error is gone and it seems to work fine now!

PR #3978, opened 2026-01-25
docs(cloudflare): remove invalid docs and add note about wrangler env

RihanArfan:
### ❓ Type of change

### 📚 Description

1. Cloudflare dev emulation via `getPlatformProxy()` only accepts a file path (wrangler.<toml/json/jsonc>), and not via `nitro.cloudflare.wrangler`.
2. 
https://github.com/nitrojs/nitro/blob/9c6abf1790d9743504ea244b5856e171cc4c3ee1/src/presets/cloudflare/runtime/plugin.dev.ts#L55-L83

3. Adds a note about the environment property in dev to use a different Wrangler environment.

### 📝 Checklist

### DIFF
--- a/docs/2.deploy/20.providers/cloudflare.md
+++ b/docs/2.deploy/20.providers/cloudflare.md
@@ -280,7 +280,7 @@ defineHandler(async (event) => {
 
 ### Access to the bindings in local dev
 
-To access bindings in dev mode, we first define them. You can do this in a `wrangler.toml`/`wrangler.json` file, or directly in your Nitro config under `cloudflare.wrangler` (accepts the same type as `wrangler.json`).
+To access bindings in dev mode, we first define them. You can do this in a `wrangler.jsonc`/`wrangler.json`/`wrangler.toml` file
 
 For example, to define a variable and a KV namespace in `wrangler.toml`:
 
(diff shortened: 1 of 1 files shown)


docs/2.deploy/20.providers/cloudflare.md
```diff
@@ -311,32 +311,21 @@ id = "xxx"
...
+If you have multiple Wrangler environments, you can specify which Wrangler environment to use during Cloudflare dev emulation:
 
-
-```js [nitro.config.js]
+```ts [nitro.config.ts]
 import { defineNitroConfig } from "nitro/config";
 
 export default defineNitroConfig({
-    cloudflare: {
-      wrangler: {
```
pi0, line 322: It is for prod and valid. We merge user config here: https://github.com/nitrojs/nitro/blob/9c6abf1790d9743504ea244b5856e171cc4c3ee1/src/presets/cloudflare/utils.ts#L246

RihanArfan: For prod that works for the generated deployConfig wrangler.json but I don't think it works during local development?

Here it's under the Access to the bindings in local dev section. 

image

We have cloudflare.wrangler: {} documented with deployConfig 

image


https://github.com/RihanArfan/nitro/blob/bd5d54492a3d6037c11d0e76eeb2bf3b98a7a420/docs/2.deploy/20.providers/cloudflare.md?plain=1#L37

PR #3991, opened 2026-01-30
docs: add code splitting section

onmax:
Code splitting is a key Nitro feature but wasn't documented. Added section explaining how route handlers are chunked and how to disable it with `inlineDynamicImports`.

### DIFF
--- a/docs/1.docs/5.routing.md
+++ b/docs/1.docs/5.routing.md
@@ -320,17 +320,7 @@ This behaviour can be overridden by some request properties (e.g.: `Accept` or `
 
 Nitro creates a separate chunk for each route handler. Chunks load on-demand when first requested, so `/api/users` doesn't load code for `/api/posts`.
 
-To bundle everything into a single file:
-
-```ts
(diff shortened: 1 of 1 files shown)


docs/1.docs/5.routing.md
```diff
@@ -316,6 +316,22 @@ The way errors are sent back to the client depends on the environment. In develo
...
 
+## Code splitting
+
+Nitro creates a separate chunk for each route handler. Chunks load on-demand when first requested, so `/api/users` doesn't load code for `/api/posts`.
+
+To bundle everything into a single file:
+
+```ts
+export default defineNitroConfig({
+  inlineDynamicImports: true
```
pi0: We could just link it to configuration section for changing behavior. Some deployment presets are _forced_ to do this (not as benefit really, it increases cold start actually!), not in any normal conditions it is good idea to enable flag.

onmax: Addressed in ef3e460

PR #3993, opened 2026-01-31
chore: add ai config in gitignore

iiio2:
Recently I've learned about skills. Suppose we're using skills. Then we may create those 2 folders. maybe we need to put that in gitignore.


iiio2:
Thanks. :blush:

PR #3994, opened 2026-02-01
fix(vite): remove extension check to support dotted API routes

### LINKED ISSUE #3990: infinite redirect with baseUrl + nested tRPC procedure
### Environment

nitro: latest
node: 24

### Reproduction

https://stackblitz.com/edit/nitrojs-starter-jczkd8ob?file=vite.config.ts

You'll see the infinite redirect happen when clicking the increment button.

### Describe the bug

When configuring nitro/vite to serve from a baseUrl and a call is made to a nested tRPC proc an infinite redirect happens. 

Nested tRPC procs craft a url with a `.` in it, e.g. `/trpc/nestedProc.get`. I _think_ when the url contains this character, the code introduced in https://github.com/nitrojs/nitro/pull/3817 is hit making nitro think the request is for a static asset. This only seems to happen when `baseURL` is configured however.

A workaround we've found is to prefix the nested procedure with a character (`_` works) that breaks the regex above: 

```typescript
export const router = createTRPCRouter({
  nestedProc: {
    _get: get,
  },
});
```

### Additional context

_No response_

### Logs

```sh

```

kricsleo:
### 🔗 Linked issue

resolves https://github.com/nitrojs/nitro/issues/3990

### ❓ Type of change

### 📚 Description

Removed overly broad extension check that broke routes like `/api/foo.get`. Added `404` fallback to Vite instead.

### 📝 Checklist

### DIFF
--- a/src/build/vite/dev.ts
+++ b/src/build/vite/dev.ts
@@ -222,6 +222,11 @@ export async function configureViteDevServer(ctx: NitroPluginContext, server: Vi
       if (nodeRes.writableEnded || nodeRes.headersSent) {
         return;
       }
+      if (envRes.status === 404) {
+        // Allow Vite to try
+        nodeReq._nitroHandled = false;
+        return next();
+      }
       return await sendNodeResponse(nodeRes, envRes);
     } catch (error) {
       return next(error);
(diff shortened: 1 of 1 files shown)


pi0:
Appreciate the PR. It should be fixed on main (#4108)

PR #4000, opened 2026-02-04
fix: support PORT=0 for random port assignment

wadefletch:
### 🔗 Linked issue

N/A - Minor fix discovered while using the library.

### ❓ Type of change

### 📚 Description

The current port parsing pattern in all runtime files:

```typescript
const port = Number.parseInt(process.env.NITRO_PORT || process.env.PORT || "") || 3000;
```

Has two issues with `PORT=0`:

1. `Number.parseInt("0")` returns `0`
2. `0 || 3000` evaluates to `3000` because `0` is falsy in JavaScript

This prevents users from using `PORT=0` for random port assignment (a common pattern where the OS assigns an available port).

**Solution:** Replace the double-`||` pattern with explicit `NaN` checking:

```typescript
const _parsedPort = Number.parseInt(process.env.NITRO_PORT ?? process.env.PORT ?? "");
const port = Number.isNaN(_parsedPort) ? 3000 : _parsedPort;
```

This correctly handles:

- `PORT=0` → port is `0` (random assignment)
- `PORT` not set → port is `3000` (default)
- `PORT="8080"` → port is `8080`
- `PORT="abc"` → port is `3000` (invalid input fallback)

### 📝 Checklist

### DIFF
--- a/src/presets/bun/runtime/bun.ts
+++ b/src/presets/bun/runtime/bun.ts
@@ -8,9 +8,7 @@ import { startScheduleRunner } from "#nitro/runtime/task";
 import { trapUnhandledErrors } from "#nitro/runtime/error/hooks";
 import { resolveWebsocketHooks } from "#nitro/runtime/app";
 
-const _parsedPort = Number.parseInt(
-  process.env.NITRO_PORT ?? process.env.PORT ?? ""
-);
+const _parsedPort = Number.parseInt(process.env.NITRO_PORT ?? process.env.PORT ?? "");
 const port = Number.isNaN(_parsedPort) ? 3000 : _parsedPort;
 const host = process.env.NITRO_HOST || process.env.HOST;
 const cert = process.env.NITRO_SSL_CERT;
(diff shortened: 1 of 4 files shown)


src/presets/bun/runtime/bun.ts
```diff
@@ -8,7 +8,8 @@ import { startScheduleRunner } from "#nitro/runtime/task";
 import { trapUnhandledErrors } from "#nitro/runtime/error/hooks";
 import { resolveWebsocketHooks } from "#nitro/runtime/app";
 
-const port = Number.parseInt(process.env.NITRO_PORT || process.env.PORT || "") || 3000;
+const _parsedPort = Number.parseInt(process.env.NITRO_PORT ?? process.env.PORT ?? "");
+const port = Number.isNaN(_parsedPort) ? 3000 : _parsedPort;
```
pi0, line 12: I guess we can simplify to this:

```ts
const port = Number(process.env.NITRO_PORT ?? process.env.PORT ?? 3000);
```

When `"0"` is passed it will be preserved before 3000 fallback

wadefletch: This would result in `port = NaN` for `PORT=true` or `PORT=123abc`. 

Is that desirable or do we want to handle NaN's explicitly?

pi0 (approved):

PR #4002, opened 2026-02-05
refactor: add `rolldown` as dependency

pi0:
Currently nitro has optional peer dependency on rolldown + oxc-transform (for OpenAPI route meta) and oxc-minify for legacy.

Vite v8 will be having dependency on rolldown, by adding rolldown dependency to nitro:
- Nitro + Vite => The install size will be practically reduced as we share rolldown dep
- Nitro CLI for API => Nitro has built-in rolldown dependency which increases DX!

> [!IMPORTANT]
> Current PR depends on `rolldown/experimental` imports. Although i tried to make use of dynamic imports to reduce chance of fatal issues in case of breaking changes, we will be waiting and immediately update to better `rolldown/` subpath export when available.

/cc @TheAlexLichter @nitrojs/vite

### DIFF
--- a/test/examples.test.ts
+++ b/test/examples.test.ts
@@ -14,7 +14,15 @@ const { createServer, createBuilder, rolldownVersion } = (await import(
 
 const isRolldown = !!rolldownVersion;
 
-const skip = new Set<string>(["websocket", ...(isRolldown ? [] : ["vite-rsc"])]);
+const skip = new Set<string>([
+  "websocket",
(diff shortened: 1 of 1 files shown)


sandros94:
EDIT: Fixed in d3b1f7b

@pi0 shouldn't both `output` options be optional?

https://github.com/nitrojs/nitro/blob/db30877da38571bd81875328b136dad013a7b991/src/types/build.ts#L13-L19

PR #4003, opened 2026-02-05
perf: optimize duplicate handler filtering

OskarLebuda:
### 🔗 Linked issue

n/a

### ❓ Type of change

### 📚 Description

The current implementation uses `findIndex` inside `filter`, which scans the entire array for each handler - resulting in O(n2) complexity.

Replaced with a Set-based approach for O(1) lookups instead of repeatedly scanning the array. Behavior stays the same - keeps first occurrence, removes duplicates.

For large applications with many routes, this should provide noticeable performance improvement during startup.

### 📝 Checklist


pi0 (approved):
thnx!

PR #4005, opened 2026-02-06
fix(vite): set `copyPublicDir: false` by default

sandros94:
### 🔗 Linked issue

Discussed in https://github.com/nitrojs/nitro/issues/3868#issuecomment-3856956699

### ❓ Type of change

### 📚 Description

By default Vite copies over the `public` dir to the `outDir` (`.output/server`), resulting in duplicate content, for example:

```
// minimal test in nitro@3.0.1-alpha.2

>> ls .output/public
assets nitro.svg robots.txt vite.svg

>> ls .output/server
_libs _routes index.mjs nitro.svg robots.txt vite.svg 
```

Solution should be as simple as setting `build.copyPublicDir: false` while still allowing overrides

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "ohash": "^2.0.11",
     "rolldown": "^1.0.0-rc.3",
     "srvx": "^0.11.2",
-    "undici": "^7.20.0",
     "unenv": "^2.0.0-rc.24",
     "unstorage": "^2.0.0-alpha.5"
   },
@@ -189,7 +188,6 @@
   },
   "resolutions": {
     "nitro": "link:.",
-    "undici": "^7.19.2",
     "vite": "8.0.0-beta.13"
   },
   "engines": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,7 +6,6 @@ settings:
 
 overrides:
   nitro: link:.
-  undici: ^7.19.2
   vite: 8.0.0-beta.13
 
 importers:
@@ -43,9 +42,6 @@ importers:
       srvx:
         specifier: ^0.11.2
         version: 0.11.2
-      undici:
-        specifier: ^7.19.2
-        version: 7.20.0
       unenv:
         specifier: ^2.0.0-rc.24
         version: 2.0.0-rc.24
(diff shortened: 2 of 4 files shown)


src/build/vite/plugin.ts
```diff
@@ -117,6 +117,7 @@ function nitroEnv(ctx: NitroPluginContext): VitePlugin {
         debug("[env]  Configuring client environment", name === "client" ? "" : ` (${name})`);
         config.build!.emptyOutDir = false;
         config.build!.outDir = useNitro(ctx).options.output.publicDir;
+        config.build!.copyPublicDir ??= false;
```
sandros94, line 120: Oh wait, @pi0 I just read your https://github.com/nitrojs/nitro/issues/3868#issuecomment-3859856536

> I think we should disable it for SSR and Nitro vite env configs (only client consumers)

What is the implication of doing this in client consumers like here? Should we revert this line?

pi0 (commented):
thnx!

src/build/vite/plugin.ts
pi0: actually you were correct i guess. nitro already does copy public dir!

sandros94: ah no, ok I see. My brain was interpreting the comment backwards 😅

sandros94: > actually you were correct i guess. nitro already does copy public dir!

yes exactly, and as I was doing some tests this always resulted in a `.output` folder ready for production distribution anyway (regardless of monolitic or via CDN/Proxy)

darklight9811:
I don't know if this is kinda of a regression, but it stopped copying for client too, breaking all of my apps that relied on the asset being served, I dont know if that was intentional

sandros94:
> I don't know if this is kinda of a regression, but it stopped copying for client too, breaking all of my apps that relied on the asset being served, I dont know if that was intentional

@darklight9811 could you provide a bit more context in your situation? Which build preset are you using? Which version of Nitro?

darklight9811:
@sandros94 I'm using nitro alongside tanstack start with bun preset (only config change that I do to nitro), with vite I change publicDir to `../../public` since I'm on a monorepo.

When I build, its not copying the content of public dir in anyway, so every public asset is lost when I try to run the built project (returns not found).

version that it works: `3.0.1-20260202-124820-1954b824`
issue still persists on version: `3.0.260311-beta`

steps to reproduce:
- use tanstack start with nitro on a monorepo
- add a public dir that is outside of the app
- point to it using vite publicDir
- build and run the project

pi0:
> point to it using vite publicDir

You should add extra public dirs to `publicAssets` nitro config (docs: https://nitro.build/config#publicassets)

darklight9811:
@pi0 but why was it working and doesn't anymore? I want to understand why the change in the behavior because it seems logical to get that info from vite config

pi0:
We were wrongly double copying public assets from vite's public dir before i guess.

darklight9811:
@pi0 but when the publicDir location is changed, you are not copying it, the link you sent says it already copies if detects the /public, but its not reading the config file to update the reference to the default public dir. Should I open it as a bug?

pi0:
Yes please if you can open a minimal reproduction (nitro alone)

Issue #1129, opened 2023-04-12
UNIX socket support


uroybd:
### Describe the feature

Allow users to provide an UNIX socket path instead of host and port. Extremely useful because UNIX sockets are in general more performant.

### Additional information

- [X] Would you be willing to help implement this feature?

PR #4017, opened 2026-02-10
fix: restore runtime `close` hook and implement graceful shutdown env vars

### LINKED ISSUE #1243: Document graceful shutdown
https://github.com/unjs/nitro/pull/1227

Since this feature is mostly for node, it could be in https://nitro.unjs.io/deploy/node

### LINKED ISSUE #1740: UNIX socket support for bun preset.
### Describe the feature

Nitro already supports `NITRO_UNIX_SOCKET` for `node-server` preset. This can be used to bring the UNIX socket support for bun preset too.

### Additional information

### LINKED ISSUE #2376: Outdated/wrong documentation about `NITRO_SHUTDOWN` env variable
### Environment

- nitropack@2.9.6
- node@18.17.1

### Reproduction

N/A

### Describe the bug

The documentation about env variables is outdated/wrong for the following one: `NITRO_SHUTDOWN`: https://nitro.unjs.io/deploy/runtimes/node#environment-variables

Indeed, in the codebase, `NITRO_SHUTDOWN_DISABLED` is used (see [this line of code](https://github.com/unjs/nitro/blob/main/src/runtime/shutdown.ts#L7)).

Also, I find the naming a bit confusing, especially when trying to understand this line of code:
```ts
export function getGracefulShutdownConfig() {
  return {
    disabled: !!process.env.NITRO_SHUTDOWN_DISABLED,
```

IMO it'd be easier to understand with something like:
```ts
export function getGracefulShutdownConfig() {
  return {
    enabled: (process.env.NITRO_SHUTDOWN_ENABLED || false),
```

I can totally open another issue for this, if needed.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #2566: Close hook not executed with 'bun' build preset
### Environment

| Name | Version |
| :--- | ---: |
| Nitro | 2.9.6 |
| Node.js | 20.11.0 |
| Bun | 1.1.7 |
| OS | macOS 14.4.1 |

## Validate that, when built with the `node-server` preset, the `close` hook runs as expected under both Node.js and Bun runtimes:

1. Install and build using the default `node-server` preset:
```shell
git clone https://github.com/unjs/nitro.git
cd nitro/examples/graceful-shutdown/
pnpm install --ignore-workspace
pnpm exec nitro build
```

2. Run the Nitro server with `node`:
```shell
$ node .output/server/index.mjs
Listening on http://[::]:3000
```

3. Press Ctrl + C to send `SIGINT` to the process and observe the expected output:
```shell
^CDisconnecting database...
Database is disconnected!
```

4. Run the Nitro server with `bun`:
```shell
bun .output/server/index.mjs
Listening on http://[::]:3000
```

5. Press Ctrl + C to send `SIGINT` to the process and observe the expected output:
```shell
^CDisconnecting database...
Database is disconnected!
```

The `close` hook runs whether the runtime is `node` or `bun` given that the build used the `node-server` preset.

## Reproduce that the `close` hook fails to run when built with the `bun` preset.

1. Remove the `.nitro` and `.output` directories:
```shell
rm -rf ./nitro ./output
```

2. Build using the `bun` preset:
```shell
pnpm exec nitro build --preset bun
```

3. Run the Nitro server:

```shell
$ bun .output/server/index.mjs
Listening on http://localhost:3000...
```

4. Press Ctrl + C to send `SIGINT` to the process and observe the output:
```shell
^C
```

The code specified in the `close` hook never runs.

### Describe the bug

When using the Nitro web server built with the `bun` preset, the `close` hook fails to run upon receiving a SIGINT signal (Ctrl + C). This behavior contrasts with the expected functionality observed when using the `node-server` preset, where the `close` hook executes properly under both Node.js and Bun runtimes.

Specifically, when built with the `node-server` preset, pressing Ctrl + C results in the expected output:
```
Disconnecting database...
Database is disconnected!
```

However, when built with the `bun` preset and run using Bun, pressing Ctrl + C simply terminates the process without executing the `close` hook, resulting in no output.

This issue prevents proper graceful shutdown procedures from being executed when using the `bun` preset, which could lead to resource leaks or incomplete cleanup operations in production environments.

Note: The root cause of this issue has not been isolated, and therefore, I am unable to submit a pull request with a fix at this time.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #2735: Close Hook not awaited
### Environment

"nuxt": "^3.13.0"

### Reproduction

plugin
```ts
import { writeFile } from 'node:fs/promises';

  nitroApp.hooks.hook('close', async () => {
    console.log('Shutting down');
    await new Promise((res) => setTimeout(() => res(1), 10000)); 
    await WireGuard.Shutdown();
  });
});
```

### Describe the bug

If I stop the nuxt server the File wont be created

Edit: while I don't see logs when stopping as I am using Docker. In dev mode when reloading i get thise

```
Force closing worker after 3 seconds
```

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #4015: Node server preset never calls `nitro.hooks.callHook("close")` on shutdown
### Environment

- Nitro (nightly): 3.0.1-20260202-124820-1954b824
- srvx: 0.10.1
- Node.js: 22
- Runtime: Docker on ECS/Fargate (receives SIGTERM from ECS task stop / rolling deployment)

### Reproduction

1. Create a Nitro plugin that registers a `close` hook:

```ts
// server/plugins/shutdown.ts
export default defineNitroPlugin((nitro) => {
  nitro.hooks.hook("close", async () => {
    console.log("[shutdown] Starting cleanup...");
    await flushTelemetry();
    await stopJobQueue();
    console.log("[shutdown] Cleanup complete");
  });
});
```

2. Build and run with the Node server preset (`node-server`).
3. Send SIGTERM (e.g., `docker stop`, `kill -TERM`, Kubernetes pod eviction, ECS task stop).
4. Observe that neither "[shutdown] Starting cleanup..." nor "[shutdown] Cleanup complete" appear in logs.

### Describe the bug

When a Nitro application receives SIGTERM/SIGINT in production, there is no way for Nitro plugins to run async cleanup before the process exits. The `close` hook on the runtime `NitroApp` instance is never triggered during shutdown.

The node server entry point (`src/presets/node/runtime/node-server.ts`) creates a srvx server and passes Nitro's fetch handler, but establishes no lifecycle connection between srvx's shutdown and Nitro's hook system:

```js
const server = serve({
  port,
  hostname: host,
  fetch: nitroApp.fetch,  // <-- only connection between srvx and Nitro
});
```

When srvx receives SIGTERM, its `gracefulShutdownPlugin` catches the signal, calls `server.close()`, and exits. At no point does anything call `nitroApp.hooks.callHook("close")`.

The `close` hook **does** exist and work at build time -- `nitro.close()` in the build-time Nitro instance calls `nitro.hooks.callHook("close")`. But the **runtime** `useNitroApp()` instance that runs in production has no equivalent bridge.

This makes it impossible to:
- Flush telemetry/tracing pipelines (OpenTelemetry, Datadog dd-trace)
- Stop background job processors (pg-boss, BullMQ)
- Drain database connection pools
- Flush buffered log transports (pino, winston)

#### Observed behavior (Docker)

With a plugin that registers both a `close` hook and a direct `process.on("SIGTERM")` workaround:

**Test 1: Relying on the `close` hook (no workaround)**

```
$ docker stop app
# srvx output: "Shutting down server in 25s... Server closed."
# ExitCode=0
# Plugin shutdown logs: NONE -- close hook never called
```

**Test 2: Direct `process.on("SIGTERM")` workaround**

```
$ docker stop app
# srvx output: "Shutting down server in 25s... Server closed."
# ExitCode=0
# Plugin logs: "[shutdown] Shutdown initiated" -- appears
# Plugin logs: "[shutdown] Shutdown complete" -- MISSING
```

The workaround fires the handler, but srvx's `process.exit(0)` (h3js/srvx#178) kills the process before async cleanup completes. Both issues must be fixed for graceful shutdown to work end-to-end.

### Additional context

**Related issues:**
- #2735 -- "Close Hook not awaited" (open, Sep 2024). Same root cause.
- #2566 -- "Close hook not executed with 'bun' build preset" (open, Jun 2024). Same family of issues.
- #4013 -- PR wiring `NITRO_SHUTDOWN_*` env vars to srvx options. Notes the `process.exit(0)` problem.
- h3js/srvx#178 -- Filed companion issue for the srvx side (`process.exit(0)` preventing async cleanup).

**Proposed fix:**

The node server entry point needs to bridge srvx's shutdown to Nitro's close hook. Depending on how h3js/srvx#178 is resolved:

**If srvx removes `process.exit(0)`** -- Nitro registers its own signal handlers:

```js
// In node-server entry
for (const sig of ["SIGTERM", "SIGINT"]) {
  process.on(sig, () => {
    nitroApp.hooks.callHook("close");
  });
}
```

**If srvx adds an `onShutdown` hook** -- Nitro passes it through:

```js
const server = serve({
  fetch: nitroApp.fetch,
  onShutdown: () => nitroApp.hooks.callHook("close"),
});
```

### Logs

```sh
# Expected (close hook fires and completes):
[shutdown] Starting cleanup...
[shutdown] Cleanup complete

# Actual (close hook never called):
# (nothing)
```

wadefletch:
### 🔗 Linked issue

- Resolves #4015
- Resolves #2735
- Resolves #2566
- Resolves #1740
- Resolves #1243
- Resolves #2376
- Related to #1129
- Upstream: h3js/srvx#178 / h3js/srvx#179 (`process.exit(0)` after graceful shutdown prevents application-level async cleanup from completing)

### ❓ Type of change

### 📚 Description

The runtime `close` hook stopped firing after the srvx migration in #3705 removed the old `setupGracefulShutdown` machinery. srvx handles HTTP-level shutdown (connection draining) but never calls Nitro's application-level `close` hook, making it impossible for plugins to run async cleanup on shutdown (flush telemetry, drain database connections, stop job queues, etc.).

**Root cause:** `src/runtime/internal/shutdown.ts` and its `setupGracefulShutdown()` function (which listened for SIGTERM/SIGINT and called `useNitroHooks().callHook("close")`) were deleted in #3705 when the presets were rewritten to use `srvx`. The HTTP shutdown was delegated to srvx's `gracefulShutdownPlugin`, but the bridge from OS signals to Nitro's hook system was never recreated.

This PR restores the application-level close hook AND wires through the documented-but-unimplemented graceful shutdown env vars to srvx.

**New `src/runtime/internal/shutdown.ts` module**

- `setupShutdownHooks()` registers SIGTERM/SIGINT handlers that `await` the Nitro `close` hook, with `try`/`catch` error logging. Follows the same pattern as the existing `trapUnhandledErrors()` utility.
- `resolveGracefulShutdownConfig()` reads `NITRO_SHUTDOWN_DISABLED` and `NITRO_SHUTDOWN_TIMEOUT` env vars and returns the srvx-compatible `gracefulShutdown` config. Typed with `ServerOptions["gracefulShutdown"]` from srvx.
- Both are imported by all long-running server preset entries (`node-server`, `node-cluster`, `bun`, `deno-server`), eliminating duplicated env var parsing.

**Implement env vars across runtimes**

Wires documented-but-unimplemented env vars through to srvx `serve()` options in the Node, Bun, and Deno server presets:

- **`NITRO_UNIX_SOCKET`**: Passes socket path to srvx via `node: { path }` (Node.js) and `bun: { unix }` (Bun). Not available for Deno as `Deno.serve` does not support UNIX sockets.
- **`NITRO_SHUTDOWN_DISABLED`**: Maps to `gracefulShutdown: false` when set (any truthy value).
- **`NITRO_SHUTDOWN_TIMEOUT`**: Maps to `gracefulShutdown: { gracefulTimeout }` (converted from ms to seconds for srvx).

**Remove env vars that cannot be configured through srvx**

Two env vars were previously documented but cannot be implemented through srvx's current API, so they are removed from the docs rather than left as misleading:

- `NITRO_SHUTDOWN_SIGNALS` -- srvx hardcodes SIGINT/SIGTERM
- `NITRO_SHUTDOWN_FORCE` -- srvx always calls `process.exit(0)`

**Documentation updates**

- Added `"close"` to the available hooks list in plugin docs, expanded graceful shutdown section with example, clarified that the close hook is only available on long-running server presets
- Fixed `"response"` hook signature in plugin docs (`(response, event)` not `(event, { body })`)
- Added env var documentation for Bun and Deno runtime pages

**Alignment with srvx**

Verified against srvx v0.11.2 (which nitro is pinned to via `^0.11.2`). The graceful shutdown plugin was refactored in h3js/srvx@54e7074:

- `gracefulTimeout` is the only config option the implementation reads
- Default timeout is 5s -- docs updated accordingly
- Signals remain hardcoded to SIGINT/SIGTERM
- `process.exit(0)` is always called after close

**Important:** srvx currently calls `process.exit(0)` immediately after graceful server close (h3js/srvx#178), which kills the process before application-level async cleanup (like the `close` hook handlers registered by `setupShutdownHooks()`) can complete. h3js/srvx#179 fixes this by replacing `process.exit(0)` with `return` on the happy path so the event loop drains naturally. Until that lands, the `close` hook will fire but async handlers may be interrupted before finishing.

### 📝 Checklist


src/runtime/internal/shutdown.ts
```diff
@@ -0,0 +1,32 @@
...
+  } catch (error) {
+    console.error("[nitro] Error running close hook:", error);
+  }
+}
+
+export function setupShutdownHooks() {
+  if (typeof process !== "undefined" && process.on) {
+    process.on("SIGTERM", _shutdownHandler);
+    process.on("SIGINT", _shutdownHandler);
+  }
```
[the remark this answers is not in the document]
wadefletch: I declined to change this to follow the pattern from the pre-existing error handling hooks. 

This is likely a good safety feature, though double-closes are in most cases no-op and/or can be handled in the application, depending on how abstracted/magic the hooks system should be–preferably less, imo, since hooks are by nature kind of a break through the abstraction wall.

src/runtime/internal/shutdown.ts
```diff
@@ -0,0 +1,26 @@
+import { useNitroApp } from "../app.ts";
+import type { ServerOptions } from "srvx";
+
+export function resolveGracefulShutdownConfig(): ServerOptions["gracefulShutdown"] {
+  if (process.env.NITRO_SHUTDOWN_DISABLED === "true") {
+    return false;
+  }
+
+  const timeoutMs = Number.parseInt(process.env.NITRO_SHUTDOWN_TIMEOUT ?? "", 10);
```
pi0, line 9: I think it is best that we don't introduce new `NITRO_` env vars where srvx already supports `SERVER_SHUTDOWN_TIMEOUT`

src/presets/node/runtime/node-server.ts
```diff
@@ -13,7 +14,8 @@ const port = Number.isNaN(_parsedPort) ? 3000 : _parsedPort;
 const host = process.env.NITRO_HOST || process.env.HOST;
 const cert = process.env.NITRO_SSL_CERT;
 const key = process.env.NITRO_SSL_KEY;
-// const socketPath = process.env.NITRO_UNIX_SOCKET; // TODO
+const socketPath = process.env.NITRO_UNIX_SOCKET;
```
pi0, line 17: We could introduce this also in srvx directly via an env without `NITRO_`.

halo951:
This PR is great — I tried your version locally.

But based on my testing, currently only build mode can trigger the `close` hook. Using `npm run dev`, the hook cannot be triggered.

Is there a way to solve this?

PR #4024, opened 2026-02-11
feat: `nitro preview`

pi0:
This PR adds new `nitro preview` command to preview the production build using srvx/loader or wrangler for cloudflare.

(extracted from vite plugin to also reuse for a temporary prerendering solution in next steps)

TODO:
- [x] preview command works with default targets
- [x] Cloudflare wait-for-port issue
- [x] Migrate vite plugin to new preview utils

### DIFF
--- a/src/preview.ts
+++ b/src/preview.ts
@@ -160,18 +160,9 @@ async function runPreviewCommand(opts: {
 
   await waitForPort(randomPort, { retries: 20, delay: 500, host: "localhost" });
 
-  const url = `http://localhost:${randomPort}`;
-
   return {
-    async fetch(req: ServerRequest) {
-      const reqURL = new URL(req.url);
-      const targetURL = url + reqURL.pathname + reqURL.search;
-      const res = await proxyFetch({ port: randomPort, host: "localhost" }, targetURL, {
-        method: req.method,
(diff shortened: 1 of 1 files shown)


src/preview.ts
```diff
@@ -0,0 +1,168 @@
+import type { ServerHandler, ServerRequest } from "srvx";
+import { spawn } from "node:child_process";
```
pi0:
```suggestion
import type { LoadOptions } from "srvx/loader";
import { spawn } from "node:child_process";
```

src/preview.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  if (buildInfo.serverEntry) {
+    for (const [key, val] of dotEnvEntries) {
+      if (!process.env[key]) {
+        process.env[key] = val;
+      }
+    }
+    const { loadServerEntry } = await import("srvx/loader");
+    const entryPath = resolve(outputDir, buildInfo.serverEntry);
+    const entry = await loadServerEntry({ entry: entryPath });
```
pi0:
```suggestion
    const entry = await loadServerEntry({ entry: entryPath, ...opts.loader });
```

PR #4030, opened 2026-02-12
presets(vercel): integrate with scheduled tasks

RihanArfan:
### 🔗 Linked issue

Related https://github.com/nitrojs/nitro/issues/1974 - we should update the checklist

### ❓ Type of change

### 📚 Description

Support Vercel Cron within Nitro Tasks. Automatically creates Vercel Cron configuration based on Nitro scheduled tasks, so fully zero-config when deploying to Vercel.

### 📝 Checklist

### DIFF
--- a/src/build/rollup/prod.ts
+++ b/src/build/rollup/prod.ts
@@ -14,7 +14,6 @@ export async function buildProduction(nitro: Nitro, rollupConfig: RollupConfig) 
   const buildStartTime = Date.now();
 
   await scanHandlers(nitro);
-  nitro.routing.sync();
   await writeTypes(nitro);
 
   let output: RollupOutput | undefined;
--- a/src/presets/vercel/runtime/cron-handler.ts
+++ b/src/presets/vercel/runtime/cron-handler.ts
@@ -20,12 +20,12 @@ export default defineHandler(async (event) => {
     throw new HTTPError("Missing x-vercel-cron-schedule header", { status: 400 });
   }
 
-  const result = await runCronTasks(cron, {
+  await runCronTasks(cron, {
     context: {},
     payload: {
       scheduledTime: Date.now(),
     },
   });
 
-  return { cron, tasks: result };
+  return { success: true };
 });
--- a/src/build/rolldown/prod.ts
+++ b/src/build/rolldown/prod.ts
@@ -15,7 +15,6 @@ export async function buildProduction(nitro: Nitro, config: RolldownOptions) {
   const buildStartTime = Date.now();
 
   await scanHandlers(nitro);
-  nitro.routing.sync();
   await writeTypes(nitro);
 
   let output: RolldownOutput | undefined;


src/presets/vercel/runtime/cron-handler.ts
```diff
@@ -0,0 +1,27 @@
+import { H3, HTTPError } from "h3";
+import { runCronTasks } from "#nitro/runtime/task";
+
+export default new H3().get("/_nitro/tasks/vercel", async (event) => {
+  // Validate CRON_SECRET if set - https://vercel.com/docs/cron-jobs/manage-cron-jobs#securing-cron-jobs
+  const cronSecret = process.env.CRON_SECRET;
+  if (cronSecret) {
+    const authHeader = event.req.headers.get("authorization");
+    if (authHeader !== `Bearer ${cronSecret}`) {
```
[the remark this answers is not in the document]
RihanArfan: Sorted

src/presets/vercel/runtime/cron-handler.ts
```diff
@@ -0,0 +1,31 @@
+import { timingSafeEqual } from "node:crypto";
+import { H3, HTTPError } from "h3";
+import { runCronTasks } from "#nitro/runtime/task";
+
+export default new H3().get("/_nitro/tasks/vercel", async (event) => {
```
pi0: We don't need H3 instance. A simple event handler is enough

src/presets/vercel/preset.ts
```diff
@@ -50,6 +52,18 @@ const vercel = defineNitroPreset(
...
         logger.info(`Using \`${serverFormat}\` entry format.`);
         nitro.options.entry = nitro.options.entry.replace("{format}", serverFormat);
+
+        // Cron tasks handler
+        if (
+          nitro.options.experimental.tasks &&
+          Object.keys(nitro.options.scheduledTasks || {}).length > 0
+        ) {
+          nitro.options.handlers.push({
+            route: "/_nitro/tasks/vercel",
```
pi0: Would be nice if we support a config to make endpoint configurable. Also doesn't needs to be `_nitro` prefixed IMO, could be simply `/_vercel/cron` (or something similar to next.js)

src/presets/vercel/runtime/cron-handler.ts
RihanArfan: Sorted

src/presets/vercel/preset.ts
RihanArfan: Changed the default, and introduced `nitro.options.vercel.cronHandlerPath`

src/presets/vercel/runtime/cron-handler.ts
```diff
@@ -0,0 +1,31 @@
...
+    const b = Buffer.from(expected);
+    if (a.length !== b.length || !timingSafeEqual(a, b)) {
+      throw new HTTPError("Unauthorized", { status: 401 });
+    }
+  }
+
+  const cron = event.req.headers.get("x-vercel-cron-schedule");
+  if (!cron) {
+    throw new HTTPError("Missing x-vercel-cron-schedule header", { status: 400 });
+  }
```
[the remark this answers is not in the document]
RihanArfan: As of today that header is now passed. Docs on Vercel are coming soon.

src/build/rollup/prod.ts
```diff
@@ -14,6 +14,7 @@ export async function buildProduction(nitro: Nitro, rollupConfig: RollupConfig)
   const buildStartTime = Date.now();
 
   await scanHandlers(nitro);
+  nitro.routing.sync();
```
pi0: We might do it in scanHandlers.

Please make seperate PR.

src/presets/vercel/runtime/cron-handler.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  const result = await runCronTasks(cron, {
+    context: {},
+    payload: {
+      scheduledTime: Date.now(),
+    },
+  });
+
+  return { cron, tasks: result };
```
pi0: Is this object consumed by vercel? Why this specific shape?

src/presets/vercel/preset.ts
```diff
@@ -50,6 +53,18 @@ const vercel = defineNitroPreset(
...
         logger.info(`Using \`${serverFormat}\` entry format.`);
         nitro.options.entry = nitro.options.entry.replace("{format}", serverFormat);
+
+        // Cron tasks handler
+        if (
+          nitro.options.experimental.tasks &&
+          Object.keys(nitro.options.scheduledTasks || {}).length > 0
+        ) {
+          nitro.options.handlers.push({
+            route: nitro.options.vercel!.cronHandlerPath!,
```
pi0: I think we can put a default here (because otherwise an external module _can_ potentially modify it to undefined before we hit here

src/presets/vercel/types.ts
```diff
@@ -132,6 +132,18 @@ export interface VercelOptions {
...
+   * The route path for the Vercel cron handler endpoint.
+   *
+   * When `experimental.tasks` and `scheduledTasks` are configured,
+   * Nitro registers a cron handler at this path that Vercel invokes
+   * on each scheduled cron trigger.
+   *
+   * @default "/_vercel/cron"
+   * @see https://vercel.com/docs/cron-jobs
+   */
+  cronHandlerPath?: string;
```
pi0: Maybe call it `cronHandlerRoute`?

src/presets/vercel/preset.ts
RihanArfan: Added a default

src/presets/vercel/types.ts
RihanArfan: Renamed

src/presets/vercel/runtime/cron-handler.ts
RihanArfan: No particular reason. I've double checked and Vercel only shows things that were logged but not the response of this endpoint, so I'll update it so we return nothing. Should we log anything? Like Running cron schedule * * * * *

pi0: We mighjt like success: true for example. My main concern is to not introduce a reponse and break it later only

RihanArfan: Sorted

PR #4037, opened 2026-02-16
feat(tasks): expose req and waitUntil in context

RihanArfan:
### ❓ Type of change

### 📝 Checklist

# TODO
- [ ] Docs
- [ ] https://github.com/h3js/srvx/pull/183

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -9,7 +9,7 @@
     "zod": "^4.3.6"
   },
   "devDependencies": {
-    "shaders": "^2.2.48",
+    "shaders": "^2.3.63",
     "undocs": "^0.4.16"
   }
 }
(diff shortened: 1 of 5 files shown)


src/runtime/internal/routes/dev-tasks.ts
```diff
@@ -23,5 +23,11 @@ export default new H3()
       ...Object.fromEntries(event.url.searchParams.entries()),
       ...body,
     };
-    return await runTask(name!, { payload });
+    return await runTask(name!, {
+      context: {
+        req: event.req,
```
pi0: This would make behavior inconsistent (req available in dev and not prod) lets remove it

src/runtime/internal/task.ts
```diff
@@ -51,7 +51,7 @@ export async function runTask<RT = unknown>(
 }
 
 /** @experimental */
-export function startScheduleRunner() {
+export function startScheduleRunner(waitUntil: (promise: Promise<void>) => void) {
```
pi0: We can make it an interface with `waitUntil` key (so thjat later if anything else had to be passed we won't break api or need second arg

Issue #4020, opened 2026-02-10
Server HMR broken: ModuleRunner evaluatedModules cache not cleared on reload


franzwilhelm:
## Description

When using Nitro 3.0.1-alpha.2 as a Vite plugin (e.g. with TanStack Start), server-side HMR doesn't work — changes to server files aren't picked up until the dev server is manually restarted.

## Root Cause

In `dist/runtime/internal/vite/node-runner.mjs`, the `EnvRunner.reload()` method calls `this.runner.import(this.entryPath)` without first clearing the `ModuleRunner`'s `evaluatedModules` cache. Since the cache still holds the old module evaluation, `import()` returns stale code.

The underlying issue is that the `full-reload` HMR message never reaches the `ModuleRunner`'s internal handler (which would call `evaluatedModules.clear()`). This is because the transport filter on line 29 only passes `type === "custom"` messages:

```js
// node-runner.mjs line 29
parentPort.on("message", (payload) => {
  if (payload?.type === "custom" && payload.viteEnv === name) {
    onMessage(payload);  // Only "custom" messages reach ModuleRunner
  }
});
```

A `full-reload` message has `type: "full-reload"`, so it's filtered out. The separate `parentPort` listener at line 211 *does* catch `full-reload` and calls `reload()`, but `reload()` calls `runner.import()` against a stale cache.

For reference, Vite's own `full-reload` handler in `ModuleRunner` (vite/dist/node/module-runner.js) calls `runner.evaluatedModules.clear()` before re-importing — this is the step that's being skipped.

## Message flow

```
1. File changes
2. Vite hotUpdate fires → env.hot.send({ type: "full-reload" })
3. Transport wraps → worker.postMessage({ type: "full-reload", viteEnv: "nitro" })
4. Worker parentPort receives message:
   a. ModuleRunner transport listener: type === "custom"? NO → skipped (bug)
   b. Direct listener: type === "full-reload"? YES → reload()
5. reload() → runner.import(entryPath) → returns CACHED old module ❌
```

## Fix

Add `this.runner.evaluatedModules.clear()` before re-importing in `EnvRunner.reload()`:

```js
async reload() {
  try {
    this.runner.evaluatedModules.clear();  // Clear stale cache
    this.entry = await this.runner.import(this.entryPath);
    this.entryError = undefined;
  } catch (error) {
    console.error(error);
    this.entryError = error;
  }
}
```

[...]

PR #4044, opened 2026-02-19
fix(vite): invalidate shared modules in SSR environment on HMR

### LINKED ISSUE #4043: bug(vite): shared modules not invalidated in SSR environment on HMR
### Environment

- nitro — 3.0.1-alpha.2 (also confirmed on 3.0.1-20251124-135409-0c8b5392)
- vite — 8.0.0-beta.15
- @tanstack/react-start — 1.140.5
- Bun — 1.3.6

### Reproduction

1. Create a TanStack Start (or any SSR framework using Nitro) project with Vite dev mode
2. Edit any React component that is rendered server-side
3. Observe: browser console reports a hydration mismatch — the server HTML
reflects the old component while the client JS reflects the new one

### Describe the bug

When a file is updated during development, Nitro's `hotUpdate` hook skips
server-side invalidation for any module that also exists in a client environment
(e.g. shared React components used for SSR). The SSR module cache is never
updated, so the server continues rendering stale HTML while the client receives
the fresh JS — causing hydration mismatches like:

> A tree hydrated but some attributes of the server rendered HTML didn't match
> the client properties.

I'll create a PR for this.

### Additional context

_No response_

### Logs

```sh

```

Nazaire:
Modules present in both client and server environments were skipped during server-side invalidation due to an overly strict guard. This caused SSR to render stale HTML after HMR updates, leading to React hydration mismatches.

### 🔗 Linked issue

Fixes #4043
Related to #4020

### ❓ Type of change

### 📚 Description

In `src/build/vite/plugin.ts`, the `hotUpdate` hook skipped server-side
invalidation for any module that also exists in a client environment:

```ts
if (mod.id && !clientEnvs.some((env) => env.moduleGraph.getModuleById(mod.id!))) {
  hasServerOnlyModule = true;
  env.moduleGraph.invalidateModule(mod, invalidated, timestamp, false);
}
```

The !clientEnvs.some(...) guard means: if a module is present in a client
environment, don't invalidate it in the server environment. The intent was
presumably to defer to Vite's normal client HMR — but the server-side copy of
that module still needs invalidation for SSR to re-render correctly.

The fix removes the exclusion guard so that all modules with an id are
invalidated in the server module graph on HMR, regardless of whether they also
appear in a client environment.
### 📝 Checklist

### DIFF
--- a/src/build/vite/plugin.ts
+++ b/src/build/vite/plugin.ts
@@ -251,15 +251,14 @@ function nitroMain(ctx: NitroPluginContext): VitePlugin {
       const clientEnvs = Object.values(server.environments).filter(
         (env) => env.config.consumer === "client"
       );
-      let hasServerOnlyModule = false;
+      let hasServerModule = false;
       const invalidated = new Set<EnvironmentModuleNode>();
       for (const mod of modules) {
-        if (mod.id && !clientEnvs.some((env) => env.moduleGraph.getModuleById(mod.id!))) {
-          hasServerOnlyModule = true;
-          env.moduleGraph.invalidateModule(mod, invalidated, timestamp, false);
-        }
+        if (!mod.id) continue;
+        hasServerModule = true;
+        env.moduleGraph.invalidateModule(mod, invalidated, timestamp, false);
       }
-      if (hasServerOnlyModule) {
+      if (hasServerModule) {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR ❤️ Sorry forgot to comment here

PR #4067, opened 2026-03-02
fix: support `**` wildcard in query parameters for route rules

cernymatej:
### 📚 Description

This fixes an issue where using the `**` wildcard in query parameters for route rule redirects did not work. Previously, only ending route slugs were supported.

Now, it is possible to do the following:
```ts
routeRules: {
  '/source/**': { redirect: { to: "/target?param=**" } },
}
```

----
I'm not very familiar with the route redirect presets for providers like Vercel, etc. I implemented that part based on the existing code, so please let me know if there are any issues 🙏

### DIFF
--- a/src/presets/netlify/utils.ts
+++ b/src/presets/netlify/utils.ts
@@ -28,10 +28,9 @@ export async function writeRedirects(nitro: Nitro) {
       code = 301;
     }
     contents =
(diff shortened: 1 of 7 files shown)


pi0:
I appreciate PR ❤️ since it is stalled and nitro now using h3 for handling route rules, preset half is fixed in main https://github.com/nitrojs/nitro/commit/066510fd1c0d6b015b817c662b5edfd58da6d905

PR #4079, opened 2026-03-06
feat(cache): add `allowQuery` option to filter query params in cache key

fmoessle:
## Summary
- Adds `allowQuery` option to `defineCachedEventHandler` and route rules `cache` config
- When set, only the listed query parameter names are included in the cache key
- If set to an empty array `[]`, all query parameters are ignored (only pathname is used)
- If `undefined` (default), all query parameters are included — no behavior change
- Useful for ignoring tracking params like `utm_source`, `fbclid`, etc.

Closes nuxt/nuxt#33728

## Test plan
- [x] Added test fixture `api/cached-allow-query.ts` with `allowQuery: ["q"]`
- [x] Added route rule `/rules/allow-query/**` with `allowQuery: ["q"]`
- [x] Added test: same `q` param with different unlisted params hits cache
- [x] Added test: different `q` param gets different cache entry
- [x] Tests pass on multiple presets (node, cloudflare-pages, cloudflare-module, etc.)
- [x] Updated cache docs with `allowQuery` field

### DIFF
--- a/docs/1.guide/6.cache.md
+++ b/docs/1.guide/6.cache.md
@@ -267,6 +267,10 @@ The `cachedEventHandler` and `cachedFunction` functions accept the following opt
   ::field{name="varies" type="string[]"}
     An array of request headers to be considered for the cache, [learn more](https://github.com/nitrojs/nitro/issues/1031). If utilizing in a multi-tenant environment, you may want to pass `['host', 'x-forwarded-host']` to ensure these headers are not discarded and that the cache is unique per tenant.
   ::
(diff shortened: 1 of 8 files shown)


pi0:
Support for `allowQuery` landed in https://github.com/unjs/ocache/pull/29 (in next v3 beta of nitro)

fmoessle:
Will this be merged into nitro v2 too?

pi0:
probably not but nuxt (v5 nightly) already uses nitro v3.

Issue #4093, opened 2026-03-10
Nitro build does not add native externals that are required.


dake3601:
### Environment

Node v24 and Nitro v3.0.1-alpha.2

### Reproduction

Reproduction in test of linked PR.: https://github.com/dake3601/nitro-cjs-trace-repro

### Describe the bug

Importing cjs package that requires a native external will not be in the bundled output.

### Additional context

See #4094 for more info.

### Logs

```sh

```

dake3601:
@pi0 added reproduction to the issue.

pi0:
Thanks @dake3601. Can you please try also with latest beta if still happening?

dake3601:
@pi0 Updated the repro with latest beta and corresponding outputs, the bug is still there.

garylake-calvium:
Hey folks, I struggling with an issue that might be related to this one, so don't want to open anything new up if I can just watch this one. I've done some digging and this feels like it could be it...

Repro here: https://github.com/garylake-calvium/nuxt-unovis-bug

Specifically, it's a Nuxt 4 app and it's the Unovis chart library that's failing to find dependencies when I do a build and run that (it's fine in dev)

I've been working around the issue by doing this in nuxt.config.js:
```
build: {
    transpile: ["@unovis/ts", "@unovis/vue", /@emotion\//, "stylis"],
  },
```

PR #4094, opened 2026-03-10
fix(externals): trace CJS require() dependencies in production builds

dake3601:
### 🔗 Linked issue

#4093 

### ❓ Type of change

### 📚 Description

The resolveId handler had `if (rOpts.custom?.["node-resolve"]) return null;` which blanket-skipped CJS require resolutions, preventing packages like `bufferutil` (in NodeNativePackages) from being externalized and traced even when matched by the include filter.

Fix: check `opts.trace` before skipping. In dev mode (no tracing), skip as before. In production, resolve directly via `tryResolve` to avoid recursion and externalize if the resolved path matches the include filter.

Includes test fixtures (nitro-cjs-untrace → @fixture/nitro-native-mock) and a node preset test that verifies CJS require deps are traced to the output node_modules.

### 📝 Checklist

### DIFF
--- a/src/build/plugins/externals.ts
+++ b/src/build/plugins/externals.ts
@@ -71,21 +71,33 @@ export function externals(opts: ExternalsOptions): Plugin {
           };
         }
 
-        // Skip nested rollup-node resolutions
+        let resolved;
+
         if (rOpts.custom?.["node-resolve"]) {
-          return null;
-        }
-
-        // Resolve by other resolvers
-        let resolved = await this.resolve(id, importer, rOpts);
-
-        // Skip rolldown-plugin-commonjs resolver for externals
-        const cjsResolved = resolved?.meta?.commonjs?.resolved;
-        if (cjsResolved) {
-          if (!filter(cjsResolved.id)) {
-            return resolved; // Bundled and wrapped by CJS plugin
+          // For nested node-resolve resolutions (e.g., CJS require()),
+          // resolve directly to avoid recursion but still externalize
+          // modules that match the include filter (e.g., native deps).
+          // Only applies when tracing is active (production builds).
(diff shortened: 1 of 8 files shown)


pi0:
Please provide a reproduction first 🙏🏼

dake3601:
Done! https://github.com/dake3601/nitro-cjs-trace-repro 
also added the repro to the github issue.

giggo1604:
@dake3601 would that solve https://github.com/nitrojs/nitro/discussions/4068?

dake3601:
@giggo1604 I don't think so, this just helps trace cjs deps.

dake3601:
@pi0 could I please get a review?

pi0x:
Rebased this onto latest `main` locally (one trivial conflict; I could not push the result, so the branch here is unchanged).

The main problem is the test: it does not fail without the fix. I reverted `src/build/plugins/externals.ts` to `main` and re-ran `test/presets/node.test.ts` — "should trace CJS require() dependencies" still passes, and `_libs/nitro-cjs-untrace.mjs` is byte-identical either way. The fixture uses the `@fixture` scope, which the test fixture already lists in `traceDeps`, and nft copies the package anyway because the emitted `__require("...")` has a literal specifier. Could you port the fixture from your repro (a package that actually gets bundled today) and assert on the emitted chunk rather than on the file existing?

Also `src/presets/_all.gen.ts` is generated — the reordering in the diff is stale; `pnpm gen-presets` restores `_nitro`/`_static` first, and their order affects preset resolution.

The fix itself looks reasonable and well scoped, but I could not confirm it without a test that fails first.

pi0:
Appreciate the PR ❤️ Merging via #4598

PR #4109, opened 2026-03-14
refactor: use built-in escapeRegExp util

ofershap:
tiny-escape has the same API as escape-string-regexp. Ships ESM + CJS with built-in TypeScript types. 192 bytes gzipped, zero deps.

https://npmgraph.js.org/?q=tiny-escape


pi0:
nice find! We actually have builtin utils wondering might be best to just refactor and use them https://github.com/nitrojs/nitro/blob/main/src/utils/regex.ts

pi0 (approved):

PR #4124, opened 2026-03-17
feat(vercel): allow overriding function config by route

### LINKED ISSUE #3815: split bundle across multiple vercel functions
### Describe the feature

Would be neat to be able to have the build output multiple vercel functions to be able to have custom configuration for each. For example, I might have one endpoint where I wanna increase the `maxDuration`. Currently I have to do that globally for each route which could be costly.

### Additional information

RihanArfan:
### 🔗 Linked issue

Closes #3815

### ❓ Type of change

### 📚 Description

Allows setting Vercel function configuration (https://vercel.com/docs/build-output-api/primitives#serverless-function-configuration) on a per-function basis. It continues to symlink the existing to __server.func besides .vc-config.json for functions which match the routes which has custom config.

This unblocks future work for supporting Vercel Queues within Nitro as we'll be able to add triggers to a new queue handler route.

### 📝 Checklist

### DIFF
--- a/src/presets/vercel/types.ts
+++ b/src/presets/vercel/types.ts
@@ -156,7 +156,7 @@ export interface VercelOptions {
    *
    * @example
    * ```ts
-   * routeFunctionConfig: {
+   * functionRules: {
    *   '/api/my-slow-routes/**': { maxDuration: 3600 },
    *   '/api/queues/fulfill-order': {
    *     experimentalTriggers: [{ type: 'queue/v2beta', topic: 'orders' }],
@@ -164,7 +164,7 @@ export interface VercelOptions {
    * }
    * ```
    */
-  routeFunctionConfig?: Record<string, VercelServerlessFunctionConfig>;
+  functionRules?: Record<string, VercelServerlessFunctionConfig>;
 }
 
 /**
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -64,20 +64,20 @@ export async function generateFunctionFiles(nitro: Nitro) {
     nitro.logger.warn(
       "`experimentalTriggers` on the base `vercel.functions` config applies to the catch-all function and is likely not what you want. " +
         "Routes with queue triggers are not accesible on the web." +
-        "Use `vercel.routeFunctionConfig` to attach triggers to specific routes instead."
+        "Use `vercel.functionRules` to attach triggers to specific routes instead."
     );
   }
 
   const functionConfigPath = resolve(nitro.options.output.serverDir, ".vc-config.json");
   await writeFile(functionConfigPath, JSON.stringify(baseFunctionConfig, null, 2));
 
-  const routeFunctionConfig = nitro.options.vercel?.routeFunctionConfig;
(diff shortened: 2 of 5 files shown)


src/presets/vercel/types.ts
```diff
@@ -147,6 +147,24 @@ export interface VercelOptions {
...
+   * ```ts
+   * routeFunctionConfig: {
+   *   '/api/my-slow-routes/**': { maxDuration: 3600 },
+   *   '/api/queues/fulfill-order': {
+   *     experimentalTriggers: [{ type: 'queue/v2beta', topic: 'orders' }],
+   *   },
+   * }
+   * ```
+   */
+  routeFunctionConfig?: Record<string, VercelServerlessFunctionConfig>;
```
RihanArfan: @pi0 what should we call the key? I'm not satisfied with this, just used it as a placeholder for now. Also as it accepts a list of routes perhaps the key should be plural with `routeFunctionsConfig` if we decide not to change it to something else altogether?

src/presets/vercel/utils.ts
```diff
@@ -512,6 +617,38 @@ function normalizeRouteDest(route: string) {
   );
 }
 
+async function createFunctionDirWithCustomConfig(
```
pi0, line 631: Have you tried hardlinks?

test/presets/vercel.test.ts
```diff
@@ -420,9 +432,47 @@ describe("nitro:preset:vercel:web", async () => {
             "functions/_vercel",
             "functions/api/cached.func (symlink)",
             "functions/api/db.func (symlink)",
-            "functions/api/echo.func (symlink)",
+            "functions/api/echo.func/.vc-config.json",
```
pi0: These should be omited from snapshot

src/presets/vercel/utils.ts
```diff
@@ -35,6 +36,29 @@ const ISR_SUFFIX = "-isr"; // Avoid using . as it can conflict with routing
 
 const SAFE_FS_CHAR_RE = /[^a-zA-Z0-9_.[\]/]/g;
 
+/**
+ * Encodes a function path into a consumer name for queue/v2beta triggers.
+ * Mirrors the encoding from @vercel/build-utils sanitizeConsumerName().
+ * @see https://github.com/vercel/vercel/blob/main/packages/build-utils/src/lambda.ts
+ */
+function sanitizeConsumerName(functionPath: string): string {
```
pi0: Please move new internal helpers to the end

src/presets/vercel/utils.ts
RihanArfan: Briefly but had complexities with symlinks in node_modules etc. since only files can be hard linked it seems. I don't think many functions would have overrides and copying reduces a lot of complexity. I may revisit getting symlinks working in Vercel CI later again though.

test/presets/vercel.test.ts
RihanArfan: Fixed

src/presets/vercel/utils.ts
RihanArfan: Fixed

PR #4127, opened 2026-03-18
feat(vercel): queues

RihanArfan:
### 🔗 Linked issue

Related https://github.com/nitrojs/nitro/issues/1974

### ❓ Type of change

### 📚 Description

Support Vercel Queues triggers through Nitro hooks, and documents using it with Nitro Tasks.

### 📝 Checklist

### DIFF
--- a/docs/2.deploy/20.providers/vercel.md
+++ b/docs/2.deploy/20.providers/vercel.md
@@ -182,7 +182,7 @@ export default defineNitroPlugin((nitro) => {
 
 ### Running tasks from queue messages
 
-You can use queue messages to trigger [Nitro tasks](/tasks):
+You can use queue messages to trigger [Nitro tasks](/docs/tasks):
 
 ```ts [server/plugins/queues.ts]
 import { runTask } from "nitro/task";


examples/vercel-queues/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "module",
+  "scripts": {
+    "dev": "vite dev",
+    "build": "vite build"
+  },
+  "dependencies": {
+    "@vercel/queue": "latest"
```
pi0: It should be also a devDep (we bundle)

src/presets/vercel/runtime/queue-handler.ts
```diff
@@ -0,0 +1,11 @@
+import { handleCallback } from "@vercel/queue";
+import { defineHandler } from "nitro";
+import { useNitroHooks } from "nitro/app";
+
+const handler = handleCallback(async (message, metadata) => {
```
pi0: Checking pkg implementation, `handleCallback` itself returns a Promise, it will be at least a dangling promise! We should either move to each request or at least use defineLazyHander or avoid non handled promise

pi0: From claude:

```ts
import { handleCallback } from "@vercel/queue";
import { defineHandler } from "nitro";
import { useNitroHooks } from "nitro/app";
import { createClient } from "@vercel/queue"; // or however the client is constructed

const client = createClient({ token: process.env.QUEUE_TOKEN });

export default defineHandler(async (event) => {
  const request = event.req as Request;

  // Parse the callback request body that Vercel sends
  const body = await request.json(); // { queueName, consumerGroup, messageId, ... }

  await handleCallback(
    async (message, metadata) => {
      await useNitroHooks().callHook("vercel:queue", { message, metadata });
    },
    body,          // the parsed callback request
    { client }     // required options
  );

  return new Response("OK", { status: 200 });
});
```

src/presets/vercel/runtime/queue-handler.ts
```diff
@@ -0,0 +1,11 @@
...
+  await useNitroHooks().callHook("vercel:queue", { message, metadata });
```
pi0: Since we anyway require `@vercel/queue`, i was thinking maybe we can also add `send` to the context?

src/presets/vercel/runtime/queue-handler.ts
RihanArfan: The @vercel/queue handleCallback returns a promise for a Response (`(req) => Promise<Response>`). Anything returned by Nitro gets passed up to it and then returned. Do we still need this? 

We could move the handler inside the request handler.
> A lazily-created default client auto-detects the region from VERCEL_REGION, falling back to iad1. https://vercel.com/docs/queues/sdk#top-level-exports
```ts
import { handleCallback } from "@vercel/queue";
import { defineHandler } from "nitro";
import { useNitroHooks } from "nitro/app";

export default defineHandler((event) => {
  return handleCallback(async (message, metadata) => {
    await useNitroHooks().callHook("vercel:queue", { message, metadata });
  })(event.req as Request);
});
```

src/presets/vercel/runtime/queue-handler.ts
RihanArfan: Hmm how would it look?

```ts
useNitroHooks().callHook("vercel:queue:send", 'orders', payload, {
  region: 'sfo1',
  retentionSeconds: 3600,
  delaySeconds: 60,
  idempotencyKey: 'order-123',
  headers: { 'x-trace-id': 'abc-123' },
}));
```

src/presets/vercel/runtime/queue-handler.ts
pi0: Please check actual implementation in dist. From what i see it was immediately calling.

src/presets/vercel/runtime/queue-handler.ts
pi0: No we basically just provide send function in context so users don’t have it manually import it.

src/presets/vercel/runtime/queue-handler.ts
```diff
@@ -0,0 +1,9 @@
+import { handleCallback } from "@vercel/queue";
+import { defineHandler } from "nitro";
+import { useNitroHooks } from "nitro/app";
+
+export default defineHandler((event) => {
+  return handleCallback(async (message, metadata) => {
+    await useNitroHooks().callHook("vercel:queue", { message, metadata });
```
pi0: Can you confirm how error handling currently works? We might wrap it to console.error + use `nito.captureError` (so other plugins can see errors) and _if_ workflows has special handling for errors rethrow or otherwise do not.

RihanArfan: > if your queue handler throws an exception or returns an HTTP 5xx status code (including 500), Vercel Queues treats this as a failure and will automatically schedule the message for a retry. The platform also retries on other error status codes such as 408 (Request Timeout), 409 (Conflict), and 429 (Too Many Requests), but not for status codes outside this set. Retries are not triggered for successful (2xx) responses or for other error codes like 400 (Bad Request)⁠.
⁠


We now console error, nitro.captureError and rethrow

src/presets/vercel/preset.ts
```diff
@@ -65,6 +66,36 @@ const vercel = defineNitroPreset(
...
+            route: handlerRoute,
+            lazy: true,
+            handler: join(presetsDir, "vercel/runtime/queue-handler"),
+          });
+
+          nitro.options.vercel!.functionRules = {
+            ...nitro.options.vercel!.functionRules,
+            [handlerRoute]: {
+              ...nitro.options.vercel!.functionRules?.[handlerRoute],
+              experimentalTriggers: queues.triggers.map(({ topic, ...opts }) => ({
```
pi0: we could also deep merge `experimentalTriggers` (users might have their custom ones?)

test/presets/vercel.test.ts
```diff
@@ -584,6 +584,59 @@ describe("nitro:preset:vercel:bun", async () => {
   });
 });
 
+describe("nitro:preset:vercel:queues", async () => {
```
pi0: It would be best if we merge it with main matrix to avoid CI/Test time overhead

PR #4133, opened 2026-03-19
fix: guard against `createRequire(import.meta.url)` in bundled output

### LINKED ISSUE #4132: [cloudflare-module] Guard against createRequire(import.meta.url) in bundled output (crashes on Workers)
### Description

When using Nitro with the `cloudflare-module` preset and `rolldown-vite` as the builder, the server bundle contains:

```javascript
import { createRequire } from "node:module";
var __require = /* @__PURE__ */ createRequire(import.meta.url);
```

This crashes at module load time on Cloudflare Workers because `import.meta.url` is `undefined` in the `workerd` runtime. Every route returns 500.

### Root cause (upstream)

The real bug is in rolldown-vite / Vite, which defaults all server SSR builds to `platform: 'node'` even when the target is a Worker. This causes rolldown to emit `createRequire`-based CJS interop that is incompatible with `workerd`. We've filed this upstream at vitejs/vite#21962.

### Why Nitro should also address this

Even after an upstream fix, Nitro's cloudflare presets would benefit from a safety net because:

1. **Users may be on older rolldown-vite versions** where the upstream fix doesn't exist
2. **Nitro knows the target runtime** — the preset is specifically for Workers
3. **Nitro already uses unenv for Node polyfills** — guarding `createRequire` is consistent with that approach
4. **The pattern may reappear** via other bundler changes or dependency updates

### Environment

- Nitro `~3.0.260311-beta`
- rolldown-vite `^7.3.1` (latest — the bug exists in every 7.x version)
- `cloudflare-module` preset
- Workers with `nodejs_compat` + compatibility_date `2025-09-02`

### Reproduction

1. Use Nitro with `cloudflare-module` preset and rolldown-vite
2. Install any dependency that transitively pulls in a CJS package using Node built-ins (e.g., `supports-color` → `node:os`, `debug` → `node:util`)
3. Build with `vite build`
4. Test: `wrangler dev .output/server/index.mjs` — all routes return 500

### Suggested approach

**Option A**: Add a rollup output plugin to the `cloudflare-module` and `cloudflare-pages` presets:

```typescript
// In nitro/src/presets/cloudflare/preset.ts (or equivalent)
{
  name: "nitro:cloudflare-guard-createRequire",
  generateBundle(_options, bundle) {
    for (const chunk of Object.values(bundle)) {
      if (chunk.type === "chunk" && chunk.code?.includes("createRequire(import.meta.url)")) {
        chunk.code = chunk.code.replace(
          /createRequire\(import\.meta\.url\)/g,
          'createRequire(import.meta.url || "file:///")',
        );
      }
    }
  },
}
```

This is safe because `createRequire` only needs a valid base URL for resolving relative paths, and in a fully bundled output the exact URL is irrelevant. On Node.js, the fallback never triggers.

**Option B**: Signal the correct platform to the builder. If Nitro's cloudflare presets set `ssr.target: 'webworker'` or otherwise communicate to Vite/rolldown-vite that the platform is not Node, the bundler would stop emitting `createRequire` in the first place. This is the cleaner fix if Nitro has that lever.

### Current user workaround

```typescript
nitro({
  preset: "cloudflare-module",
  rollupConfig: {
    plugins: [{
      name: "patch-createRequire-workers",
      generateBundle(_options, bundle) {
        for (const chunk of Object.values(bundle)) {
          if (chunk.type === "chunk" && chunk.code?.includes("createRequire(import.meta.url)")) {
            chunk.code = chunk.code.replace(
              /createRequire\(import\.meta\.url\)/g,
              'createRequire(import.meta.url || "file:///app/index.mjs")',
            );
          }
        }
      },
    }],
  },
})
```

### Related issues

- vitejs/vite#21962 — The upstream platform-selection bug (just filed)
- vitejs/vite#20983 — Vite 7.1.11 introduced the `createRequire` pattern (no fix released)
- rolldown/rolldown#8097 — `Could not resolve 'node:module' in rolldown:runtime`
- better-auth/better-auth#6690 — Same crash on Workers (suggests wrangler.toml alias)

jmalmo:
## Problem

When `rolldown-vite` (or other bundlers) bundles SSR apps for Cloudflare Workers, it emits `createRequire(import.meta.url)` in shared chunks. On the Workers runtime, `import.meta.url` is `undefined`, so `createRequire` throws `TypeError: The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received undefined` — crashing every route.

See https://github.com/vitejs/vite/issues/21962 for the upstream Vite issue.

Fixes #4132

## Solution

Adds a rollup output plugin (`nitro:cloudflare-guard-createRequire`) to the **cloudflare-module**, **cloudflare-pages**, and **cloudflare-durable** presets that rewrites:

```js
createRequire(import.meta.url)
```

to:

```js
createRequire(import.meta.url || "file:///")
```

in the `generateBundle` hook, after all other transforms have completed.

The `"file:///"` fallback allows `createRequire` to construct a valid resolver. Any subsequent `require()` calls go through the normal Node.js compat layer provided by the Workers runtime (with the `nodejs_compat` flag).

## Why defense-in-depth

This is a defense-in-depth fix on the Nitro side. The root cause is in rolldown-vite (vitejs/vite#21962), but since Nitro controls the final output and knows the target is Cloudflare Workers, it can safely guard against this pattern regardless of which bundler is used upstream.

## Affected presets

- `cloudflare-module`
- `cloudflare-pages`
- `cloudflare-durable` (inherits from `cloudflare-module`)

### DIFF
--- a/src/presets/cloudflare/preset.ts
+++ b/src/presets/cloudflare/preset.ts
@@ -26,13 +26,10 @@ function guardCreateRequire(): Plugin {
     name: "nitro:cloudflare-guard-createRequire",
     generateBundle(_options, bundle) {
       for (const chunk of Object.values(bundle)) {
-        if (
-          chunk.type === "chunk" &&
-          chunk.code?.includes("createRequire(import.meta.url)")
-        ) {
+        if (chunk.type === "chunk" && chunk.code?.includes("createRequire(import.meta.url)")) {
           chunk.code = chunk.code.replace(
             /createRequire\(import\.meta\.url\)/g,
-            'createRequire(import.meta.url || "file:///")',
+            'createRequire(import.meta.url || "file:///")'
           );
         }
       }


hi-ogawa:
Is it possible for Nitro preset side to set `rollupOptions.platform: "neutral"` (or now `rolldownOptions` on Vite 8)? If Nitro can set `ssr.target`, then probably `platform` options should be preferred now.

pi0:
Yes we can and we used to. But the thing is Workerd has good Node.js compatibility support for most of built-ins. We reduce compatibility by changing the platform for running libs like mongodb. 
 special code is generated in rolldown runtime chunk is mainly the issue

hi-ogawa:
Right, makes sense. So from that point of view, you don't want `ssr.target: "webworker"` to mean `platform: "neutral"` either? The context of issue seems a bit hard to follow.

pi0:
Current workaround fixes `pg` on clodudflare workers (as it supports createRequire). Setting platform: neutral does not works since internal requires are not transformed.

image

PR #4141, opened 2026-03-22
fix: default export

huseeiin:
in dev nitro doesn't complain if there is no default export in `server.ts`, this fixes it in prod.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/build/virtual/routing.ts
+++ b/src/build/virtual/routing.ts
@@ -32,7 +32,7 @@ ${allHandlers
   .filter((h) => !h.lazy)
   .map(
     (h) =>
-      /* js */ `const { default: ${h._importHash} } = await import("${h.handler}").catch(() => {})`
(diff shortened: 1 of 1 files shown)


huseeiin:
@pi0 coderabbitai is right, this ignores all errors. what should we do?

pi0x:
Rebased this on current `main` — it still applies cleanly. I know it is a draft and left the draft status alone.

On your question about swallowing errors: swallowing is the wrong direction. A missing default export should fail loudly, ideally at build time with the file name, not become an `undefined` handler that crashes later in `h3.toEventHandler`.

One side effect worth knowing: I built `test/minimal` with and without this patch. Switching to `await import()` makes the bundler emit every non-lazy handler as its own chunk (`_routes/*.mjs`) instead of inlining it, and adds top-level await to the server entry. That changes output layout for all presets.

pi0:
Closing as introduicing TLA requirement might be risky for all presets. We might find an alternative solution later. Thanks for PR!

PR #4147, opened 2026-03-24
chore(examples): add Arrow.js SSR example

pi0:
## Summary

- Add `vite-ssr-arrow` example with [Arrow.js](https://github.com/standardagents/arrow-js) (~3KB reactive UI)
- Todo app showcasing reactive state, components with props, keyed lists, computed filtering, and watch effects
- SSR via `@arrow-js/ssr` with client-side takeover for interactivity

## Notes

- `@arrow-js/vite-plugin-arrow` is workspace-only (aliases to monorepo paths), so we configure `ssr.noExternal` directly
- Hydration (`@arrow-js/hydrate`) doesn't adopt event listeners yet — using client-side mount as workaround

### DIFF
--- a/examples/vite-ssr-arrow/.gitignore
+++ b/examples/vite-ssr-arrow/.gitignore
@@ -1,0 +1,2 @@
+.vercel
+.env*.local
--- a/examples/vite-ssr-arrow/nitro.config.ts
+++ b/examples/vite-ssr-arrow/nitro.config.ts
@@ -1,0 +1,5 @@
+import { defineConfig } from "nitro";
+
+export default defineConfig({
+  traceDeps: ["jsdom"],
+});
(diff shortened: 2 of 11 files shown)


pi0x:
Rebased on latest `main` and pushed. Only conflict was `pnpm-lock.yaml`; took main's copy and re-ran `pnpm install`. Still a draft, I left that alone.

Builds and serves the SSR page fine. Two fixes in one extra commit:

- `pnpm typecheck` failed on `src/entry-client.ts` (`HTMLElement` vs Cloudflare `ParentNode`). Added it to the root `tsconfig.json` exclude list, same as the Solid example does.
- Ran `pnpm fmt`, which reformatted `app.ts`, `entry-server.ts` and removed a trailing comma in the example `tsconfig.json`.

Still missing before it can land: a `README.md` and a `docs/4.examples/vite-ssr-arrow.md` page, which every other example has.

PR #4156, opened 2026-03-28
fix(cache): sanitize cache key segments to prevent ENOTDIR on fs drivers

### LINKED ISSUE #4142: Caching routes with `fs` or `fs-lite` fails when a route is both a path and a prefix of another path
### Environment

 Node.js v25.6.1
  nitropack 2.13.2                                                                                                  
  macOS      

### Reproduction

https://stackblitz.com/github/maxmalysh/nitro-fs-lite-enotdir 

### Describe the bug

It's not possible to cache both `/foo` and `/foo/bar` when using `fs` or `fs-lite` as the storage driver:
```
nitro: {
  storage: {
      // Broken because of the bug described below
    cache: { driver: 'fs-lite', base: '.cache/nitro' }
  }
}
```
Caching `/foo` creates a **file** on disk. Then caching `/foo/bar` needs `/foo` to be a directory — fails with `ENOTDIR`.     

This makes `fs` or `fs-lite` unusable for route caching whenever a route can be both a leaf and a prefix of another route (e.g. `/en` and `/en/about`, or `/items/123` and `/items/123/details`).
                                                                                                                    
  **Steps to reproduce:**                                                                                           
  1. Open the StackBlitz reproduction (or clone https://github.com/maxmalysh/nitro-fs-lite-enotdir)
  2. Visit `/foo` — payload is cached successfully                                                                  
  3. Visit `/foo/bar` — crashes with ENOTDIR     

  **Expected:** Both routes cache their payloads successfully. 

  **Actual:**
  ```json
  {                                                                                                                 
    "error": "ENOTDIR: not a directory, open '.../.cache/nitro/foo/bar'",
    "code": "ENOTDIR"                                                                                               
  }                                                               

  Note: unstorage's key-to-path mapping (: → /) is by design. The question is whether nitro should normalize/hash   
  keys for filesystem drivers, or document that fs/fs-lite are unsafe for route-based caching.
  ```                                                                                         

### Additional context

**Real-world impact:** 
- Nuxt 4.4.0 introduced runtime payload caching which stores payloads using route paths as storage keys: 
nuxt/nuxt#34410
- This causes `ENOTDIR` when a route has optional subroutes: 
nuxt/nuxt#34547.  
- @danielroe's fix adds a custom hash-based cache driver wrapping `fs-lite`, but only for `devStorage`: 
nuxt/nuxt#34569
- He suggests raising the issue upstream in nitro, since production `storage` with `fs-lite` is still affected:
https://github.com/nuxt/nuxt/pull/34569#issuecomment-4106493728

### Logs

```sh
ℹ Error: ENOTDIR: not a directory, open './.nitro/cache/foo/bar'

 ⁃ (ENOTDIR: not a directory, open '.nitro/cache/foo/bar':undefined:undefined)
 ⁃ (at Error: ENOTDIR: not a directory, open '.nitro/cache/foo/bar':undefined:undefined:undefined:undefined)

[CAUSE]
Error {
  code: 'ENOTDIR',
  errno: -20,
  path: './.nitro/cache/foo/bar',
  syscall: 'open',
  message: "ENOTDIR: not a directory, open 
  './.nitro/cache/foo/bar'",
  stack: "ENOTDIR: not a directory, open './.nitro/cache/foo/bar'\n" +
  "at Error: ENOTDIR: not a directory, open './.nitro/cache/foo/bar':undefined:undefined)",
}
```

claygeo:
## Problem

Fixes #4142

When using `fs` or `fs-lite` as the storage driver, caching routes that share path prefixes crashes with `ENOTDIR`. For example:

1. Cache `/foo` → creates a **file** at `.cache/nitro/route-rules/**/foo/foo.abc123.json`
2. Cache `/foo/bar` → needs `foo` to be a **directory** → `ENOTDIR: not a directory`

This breaks any Nitro/Nuxt app where a route can be both a leaf and a prefix of another route (e.g. `/en` and `/en/about`, `/items/123` and `/items/123/details`).

**Real-world impact:** Nuxt 4.4.0's runtime payload caching (nuxt/nuxt#34410) triggers this for any site with nested routes, as reported in nuxt/nuxt#34547. @danielroe's Nuxt-side workaround (nuxt/nuxt#34569) covers `devStorage` but not production `storage` with `fs-lite`, and he suggested fixing this upstream in Nitro.

## Root Cause

Cache `group` and `name` values contain `/` characters:
- Groups: `"nitro/route-rules"`, `"nitro/handlers"`, `"nitro/functions"`
- Names: `"/**:/foo"` (route path keys)

Unstorage's fs drivers convert `:` (key separator) to `/` (path separator). But the `/` characters already embedded in group/name values also become path separators, creating unpredictable directory structures where a path segment can be both a file and a directory prefix.

```
ocache _buildCacheKey: "/cache:nitro/route-rules:/**:/foo:key.json"
                                  ↑                 ↑
                        "/" in group          "/" in name
                                  ↓                 ↓
unstorage fs-lite:     .cache/nitro/route-rules/**/foo/key.json
                                                    ↑
                                              "foo" is a FILE
                                              but /foo/bar needs
                                              it as a DIRECTORY
```

## Fix

Three layers of defense:

1. **Group names** — Replace `/` with `-` in all cache group constants (`nitro/functions` → `nitro-functions`, etc.)
2. **Route-rules name** — Sanitize route path keys by replacing `/` with `_` before passing to ocache
3. **Public API** — Apply `sanitizeCacheKey()` to user-provided `name` values in `defineCachedFunction` and `defineCachedHandler`, protecting any custom cache names from the same issue

No collision risk: ocache's `getKey` already appends a hash of the full path to the storage key, so the `name` is only a namespace prefix, not the unique identifier.

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -13,6 +13,16 @@ import type { CacheOptions, CachedEventHandlerOptions } from "nitro/types";
 
 let _storageReady = false;
 
+/**
+ * Sanitize a cache group or name so it is safe for filesystem-based storage
+ * drivers (fs, fs-lite). Unstorage maps ":" to "/" on disk, so any "/" already
+ * present in the value would create unexpected directory nesting and cause
+ * ENOTDIR errors when a path is both a file and a prefix of another path.
+ */
+function sanitizeCacheKey(value: string | undefined): string | undefined {
+  return value?.replace(/\//g, "_");
+}
+
 function ensureStorage() {
   if (_storageReady) {
     return;
@@ -37,9 +47,10 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
 ): (...args: ArgsT) => Promise<T> {
   ensureStorage();
   return _defineCachedFunction(fn, {
-    group: "nitro/functions",
+    group: "nitro-functions",
     onError: defaultOnError,
     ...opts,
+    name: sanitizeCacheKey(opts.name),
   });
 }
 
@@ -49,12 +60,13 @@ export function defineCachedHandler(
 ): EventHandler {
   ensureStorage();
   const ocacheHandler = _defineCachedHandler(handler as any, {
-    group: "nitro/handlers",
+    group: "nitro-handlers",
     onError: defaultOnError,
     toResponse: (value, event) => toResponse(value, event as H3Event),
     createResponse: (body, init) => new FastResponse(body, init),
     handleCacheHeaders: (event, conditions) => handleCacheHeaders(event as H3Event, conditions),
     ...opts,
+    name: sanitizeCacheKey(opts.name),
   });
   return defineHandler((event) => ocacheHandler(event as any));
 }
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -47,9 +47,9 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
 ): (...args: ArgsT) => Promise<T> {
   ensureStorage();
   return _defineCachedFunction(fn, {
-    group: "nitro-functions",
     onError: defaultOnError,
     ...opts,
+    group: sanitizeCacheKey(opts.group || "nitro-functions"),
     name: sanitizeCacheKey(opts.name),
   });
 }
@@ -60,12 +60,12 @@ export function defineCachedHandler(
 ): EventHandler {
   ensureStorage();
   const ocacheHandler = _defineCachedHandler(handler as any, {
-    group: "nitro-handlers",
     onError: defaultOnError,
     toResponse: (value, event) => toResponse(value, event as H3Event),
     createResponse: (body, init) => new FastResponse(body, init),
     handleCacheHeaders: (event, conditions) => handleCacheHeaders(event as H3Event, conditions),
     ...opts,
+    group: sanitizeCacheKey(opts.group || "nitro-handlers"),
     name: sanitizeCacheKey(opts.name),
   });
   return defineHandler((event) => ocacheHandler(event as any));
--- a/src/runtime/internal/route-rules.ts
+++ b/src/runtime/internal/route-rules.ts
@@ -70,9 +70,14 @@ export const cache: RouteRuleCtor<"cache"> = ((m) =>
     const key = `${m.route}:${route}`;
     let cachedHandler = cachedHandlers.get(key);
     if (!cachedHandler) {
+      // Sanitize the name to avoid filesystem path conflicts.
+      // Route paths contain "/" which unstorage's fs drivers interpret as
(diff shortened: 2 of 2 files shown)


pi0:
Thank you for the PR ❤️ Keeping nested namespaces is a delibrate feature it allows faster lookups, easier invalidation and namespaced mounting for using different cache backends per prefix. 

I haven't looked deeply into this issue but we can either workaround it in [ocache](https://github.com/unjs/ocache) key generation for `/foo` + /foo/bar` (by adding a suffix to the end of HTTP endpoints) or add a special flag to unstorage's `fs` driver to disable namespacing when mounted (since this is very specific to fs backend)

I will keep PR as draft until we move it forward.

claygeo:
Thanks for the context — that makes total sense, keeping the namespace structure for lookups and invalidation is clearly the right call.

The ocache approach sounds cleaner to me since it keeps the fix at the key-generation layer rather than adding a flag that only applies to one driver. Happy to take a look at that path if you can point me at the relevant spot in ocache, or I can wait if you'd rather land a different fix there first.

Either way I'm happy to update this PR or close it in favour of whatever direction fits best.

claygeo:
Thanks @pi0 — that context makes sense. Keeping the hierarchical namespacing in ocache for performance/invalidation is the right call.

The issue is specific to the `fs` driver: when `foo` and `foo/bar` are both cache keys, the fs driver tries to create both a file at `.../foo` and a directory at `.../foo/`, which collide. 

Happy to move the fix upstream to the `unstorage` fs driver (adding a suffix like `:` to HTTP endpoint keys to avoid the file/directory collision) if that direction works for you, or I can look at the ocache key generation approach. Let me know which you'd prefer and I'll update accordingly.

pi0:
I think we can first start by tackling unstorage fs driver issues. thanks!

claygeo:
@pi0 Done — opened unjs/unstorage#767 with the fix at the fs driver level.

Adds a `dataSuffix` option to both `fs` and `fs-lite` drivers. When set (e.g. `dataSuffix: ".data"`), key files get the suffix appended on disk, so `foo` → `foo.data` and `foo:bar` → `foo/bar.data`, avoiding the file/directory collision. Defaults to `undefined` for backward compatibility.

Once that lands, nitro can set `dataSuffix` in its cache driver config and this issue goes away without touching the namespace structure.

claygeo:
Closing in favor of unjs/unstorage#767.

pi0 clarified above that nested namespaces are intentional and the fix belongs in the unstorage `fs` driver rather than in Nitro's cache-key sanitization. On top of that, Nitro 3.x's current `ocache`-based cache path doesn't reproduce the collision (the generated entry key always gets a `.json` extension, and the built-in HTTP key is sanitized + hashed — details in nitrojs/nitro#4142), so this Nitro-side sanitization is no longer the right change.

The driver-level fix (an opt-in `dataSuffix` option) lives in unjs/unstorage#767.

PR #4159, opened 2026-03-28
fix: respect compress settings in dev server

hellozyemlya:
### ❓ Type of change

### 📚 Description
Implements handling of "compressPublicAssets" option from nitro.config.ts in development mode.

### 📝 Checklist

### DIFF
--- a/src/dev/app.ts
+++ b/src/dev/app.ts
@@ -10,6 +10,7 @@ import { stat } from "node:fs/promises";
 import { createReadStream } from "node:fs";
 import { createGzip, createBrotliCompress } from "node:zlib";
 import { createVFSHandler } from "./vfs.ts";
+import type { CompressOptions } from "../types/config.ts";
 
 import devErrorHandler, {
   defaultHandler as devErrorHandlerInternal,
@@ -70,6 +71,7 @@ export class NitroDevApp {
           dir: asset.dir,
           base: assetBase,
           fallthrough: asset.fallthrough,
+          compress: this.nitro.options.compressPublicAssets,
         })
       );
     }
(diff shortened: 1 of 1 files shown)


jantimon:
I applied the patch locally for our setup and it works as expected:

| `compressPublicAssets` | dev behavior with PR              |
|------------------------|-----------------------------------|
| `true`                 | ✅ gzip + br on the fly (as today)   |
| `false`                | ✅ no compression                    |
| `{ gzip: true }`  | ✅ only enabled gzip            |
| unset                  | (new) no compression                    |

@pi0 the @coderabbitai remarks of `Accept-Encoding: br;q=0` and `.includes()` are correct but are preexisting bugs - can we please fix that in a follow up pr?

pi0x:
- Rebased onto main. One small conflict on the import line, easy to fix. `pnpm fmt` and `pnpm typecheck` pass and the dev preset tests pass. Push is blocked in my environment, so nothing was pushed.
- The logic looks right, and thanks for the clear behaviour table in the thread.
- Worth calling out: `compressPublicAssets` defaults to `false`, so after this the dev server stops compressing public assets for everyone by default. That is the intended fix, but it is a visible change and should be in the release notes.
- No regression test. To add one you need a second dev context with `compressPublicAssets: false`, which needs dev tests to use a random port instead of 3000. #4549 makes exactly that change, so it may be easiest to land after it.
- Small note: dev still never serves zstd, so `{ zstd: true }` alone now means no compression at all.
- I also moved `CompressOptions` to come from `nitro/types` instead of `../types/config.ts`, to match the rest of the file.

Issue #2136, opened 2024-02-05
getRouterParams(event) is always empty in nitro middlewares


cosbgn:
### Environment

nuxt@latest

### Reproduction

https://stackblitz.com/edit/github-fxp3q3?file=middleware%2Fexample.js

### Describe the bug

const params = getRouterParams(event)
is always empty in nitro middlewares

manniL:
To me, this behavior makes sense as the middleware runs before the route handlers, meaning that the middleware might not have any knowledge about possible router params.

cosbgn:
@manniL I'm not sure, in theory I agree, but we do have access to the URL, even the nuxt example is:

```js
export default defineEventHandler((event) => {
  console.log('New request: ' + getRequestURL(event))
})
```

I have a lot of use cases where if I have `[_book_id]` in the URL I want to make sure the user has access to the the book, the middleware would be a perfect place to do this, so I can be sure any route which has `[_book_id]` is automatically validated.

If you do have access to the URL, you should have also access to the params, no?

manniL:
Access to the URL makes totally sense, but as middleware is triggered *before* any router matching logic happens (AFAIK), the event is not aware of any route parameters yet (when using global middleware)

CHIBX:
So would there be a problem if the middleware was given knowledge of the URL that made it to run, and of course the previous one?

> Access to the URL makes totally sense, but as middleware is triggered _before_ any router matching logic happens (AFAIK), the event is not aware of any route parameters yet (when using global middleware)

manniL:
The middleware already knows the URL of the request :)

CHIBX:
@manniL Should this be resolved, and router params are able to be read from the event parameter in nitro middlewares, would it impose any security vulnerability?

sameignatovich:
Any updates?

xiao-e-yun:
This fix was provided by Claude Code.

## For v3 version
~~I have already submitted the relevant PR. (#4178)~~ (it is working by default)

> This PR fixes getRouterParams(event) always returning an empty object {} when called inside Nitro middlewares.
> 
> Root cause: In src/runtime/internal/app.ts, the ~findRoute hook was assigned as a one-liner that returned the matched route but never populated event.context.params. H3 sets event.context.params from the route match result only inside its own request handling flow — but since ~findRoute is called before middleware runs, params were never available to middleware handlers.
> 
[...]

PR #4178, opened 2026-04-03
fix: getRouterParams always empty in middlewares

xiao-e-yun:
This is #2136 fix for v3 Nitro.

Since I am not very familiar with the relevant Nitro code, I may need your help to check for any potential problems.

### 🔗 Linked issue

#2136

### ❓ Type of change

### 📚 Description

This PR fixes `getRouterParams(event)` always returning an empty object `{}` when called inside Nitro middlewares.

**Root cause:** In `src/runtime/internal/app.ts`, the `~findRoute` hook was assigned as a one-liner that returned the matched route but never populated `event.context.params`. H3 sets `event.context.params` from the route match result only inside its own request handling flow — but since `~findRoute` is called before middleware runs, params were never available to middleware handlers.

**Fix:** Expand the `~findRoute` assignment to merge `match.params` into `event.context.params` immediately when a route is matched, making router params available to all middleware that runs before (and after) the route handler.

### 📝 Checklist

This fix was provided by Claude Code.

### DIFF
--- a/src/runtime/internal/app.ts
+++ b/src/runtime/internal/app.ts
@@ -175,7 +175,13 @@ function createH3App(config: H3Config) {
   const h3App = new H3Core(config);
 
   // Compiled route matching
-  hasRoutes && (h3App["~findRoute"] = (event) => findRoute(event.req.method, event.url.pathname));
+  hasRoutes && (h3App["~findRoute"] = (event) => {
+    const match = findRoute(event.req.method, event.url.pathname);
+    if (match) {
+      event.context.params = { ...event.context.params, ...match.params };
+    }
+    return match;
+  });
 
   hasGlobalMiddleware && h3App["~middleware"].push(...globalMiddleware);
 


xiao-e-yun:
## For v2 version
Required changes

1. packages/nitropack — src/runtime/internal/app.ts (or equivalent)

In onRequest (the h3 app-level lifecycle hook that runs before all handlers), perform an early route resolution using the router's __resolve__ method and pre-populate event.context.params:
```ts
const resolved = await nitroApp.router?.handler?.__resolve__?.(event.path);
if (resolved?.params) {
  event.context.params = { ...event.context.params, ...resolved.params };
}
```
This runs before any middleware, so getRouterParams(event) returns correct values throughout the request lifecycle.

2. packages/h3 (peer dependency, v1.x) — src/router.ts

The __resolve__ method on router.handler currently returns { route, handler } but omits params. The radix3 match object has .params available at that point but it is discarded. It needs to be included in the
return value:
```ts
// before
return { route: match.matched.path, handler: match.handler };

// after
return { route: match.matched.path, handler: match.handler, params: match.matched.params || {} };
```
Without this change, the onRequest hook in nitropack cannot access the matched params, since __resolve__ is the only public API for early route resolution without re-running the full handler.

Why h3 needs to change too

The h3 change is necessary because matchHandler is a private closure inside createRouter — it is not exposed on the router object. The only public surface for pre-resolving a route is
router.handler.__resolve__, so that is the correct place to surface params.

Scope summary

| File | Package | Change |
| :--- | :--- | :--- |
| `src/runtime/internal/app.ts` | nitropack | Call `__resolve__` in `onRequest`, merge params into `event.context.params` |
| `src/router.ts` | h3 (v1.x) | Return params from `router.handler.__resolve__` |

Both changes are small and surgical — no architectural impact. The h3 change is strictly additive (new field in return value), so it is fully backwards-compatible.

This fix was provided by Claude Code.

pi0:
thanks for pr but please first provide minimal nitro reproduction.

xiao-e-yun:
Sorry!!! That is working for v3 without this PR.

This issue is limited to v2 only.
nitro v2 bad
nuxt v4 bad (because using nitro v2)

PR #4193, opened 2026-04-08
feat: add version meta

RihanArfan:
### ❓ Type of change

### 📚 Description

Adds meta to the Nitro app with version details.

v2 backport: #4194

### DIFF
--- a/src/nitro.ts
+++ b/src/nitro.ts
@@ -1,6 +1,7 @@
 import { consola } from "consola";
 import { Hookable, createDebugger } from "hookable";
 import type { LoadConfigOptions, Nitro, NitroConfig, NitroDynamicConfig } from "nitro/types";
+import { version as nitroVersion } from "nitro/meta";
 import { loadOptions } from "./config/loader.ts";
 import { updateNitroConfig } from "./config/update.ts";
 import { installModules } from "./module.ts";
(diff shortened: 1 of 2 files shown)


src/nitro.ts
```diff
@@ -17,6 +18,11 @@ export async function createNitro(
 
   // Create nitro context
   const nitro: Nitro = {
+    meta: {
+      version: nitroVersion,
+      majorVersion: Number.parseInt(nitroVersion.split(".")[0]),
+      [nitroVersion.split(".")[0]]: {},
```
pi0: Why adding empty object?

src/nitro.ts
```diff
@@ -17,6 +18,11 @@ export async function createNitro(
 
   // Create nitro context
   const nitro: Nitro = {
+    meta: {
+      version: nitroVersion,
+      majorVersion: Number.parseInt(nitroVersion.split(".")[0]),
```
pi0: We could hardcode it to `3` to reduce runtime cost (or have it also in nitro/meta) it is not going to change in main

src/nitro.ts
RihanArfan: Based on what we discussed although I'll remove it for now since it's empty then.
image

src/nitro.ts
RihanArfan: Hard coded it

pi0 (approved):

src/nitro.ts
pi0: haha sorry i used shell syntax to show it is an object with majorVersion!

pi0 (approved):

Issue #2590, opened 2024-07-04
Returning Web Response in Azure SWA is not working


targetlucked69:
### Environment

Nuxt: 3.12.3
Node: 20

### Reproduction

Deploy a basic Nuxt app with an api that returns a web response like

```ts
export default function eventHandler((event) => {
  return new Response(JSON.stringify({ username: 'test' }), { headers: { 'Content-Type': 'application/json' } })
})
```

### Describe the bug

Instead of the actual json, Im getting an empty object returned when deployed to Azure SWA:

```json
// /api/user
{}
```

itpropro:
I was able to reproduce this. But there are alternative options to construct and send a response that work with SWA and are more the "Nitro like":

Directly returning the object (Nitro automatically sets the headers):

```
export default defineEventHandler(() => {
  return { username: 'test' }
})
```

Using the send function:

```
export default defineEventHandler((event) => {
  return send(event, { username: 'test' }, 'application/json')
  // also possible
  // return send(event, JSON.stringify({ username: 'test' }), 'application/json')
})
```

Both are tested and confirm to work with Static Web Apps.

Do you want to keep this issue open @pi0 or would you agree not to use the native Response object and instead use the Nitro utilities?

itpropro:
> Returning a new `Response` should be supported in handlers. We should investigate why [`body`](https://github.com/unjs/nitro/blob/v2/src/presets/azure/runtime/azure-swa.ts#L39) (which can be a stream) is not handled, it is a bug.

I verified that this is also a bug in the `azure-function` preset.
It is definitely cause by Nitro handling the response binding for Azure Functions

```
  context.res = {
    status,
    cookies: getAzureParsedCookiesFromHeaders(headers),
    headers: normalizeLambdaOutgoingHeaders(headers, true),
    body,
  };
```

We should address that with the update to Azure Functions runtime 4.

pi0:
thanks for investigation @itpropro. Does runtime v4 support accept `ReadableStream` as valid `body`?

itpropro:
> thanks for investigation @itpropro. Does runtime v4 support accept `ReadableStream` as valid `body`?

No, we can also release the bug fix beforehand independently of the v4 update.

jvpelt:
Hi @pi0, 

[...]

PR #4195, opened 2026-04-09
fix(azure): prevent azure-swa runtime errors, and response body returning {}

Xiang-CH:
### 🔗 Linked issue

#2590

### ❓ Type of change

### 📚 Description

This PR improves the azure-swa preset so Nitro runs reliably on Azure Static Web Apps

**What was failing**

- Some responses were returned as {} in Azure SWA instead of the expected payload. (#2590)
- Request is built with relative URL, causing runtime error

**How it is fixed**

- **Outgoing body**: Replace passing `response.body` (a stream) with `Buffer.from(await response.arrayBuffer())` when a body exists, so Azure Functions gets bytes instead of a stream object that could surface as `{} `or otherwise fail at runtime.
- **Incoming headers**: Pass `headers: new Headers(req.headers)` so the Fetch `Request` receives the same headers as the Azure HTTP trigger
- **Request URL**: Build the Request with an absolute URL via new `URL(url, resolveBaseUrl(req))` instead of a relative URL string. resolveBaseUrl prefers `x-forwarded-proto / x-forwarded-host` (or host), then falls back to the origin from x-ms-`original-url`, then `http://localhost`.

### 📝 Checklist

### DIFF
--- a/docs/2.deploy/20.providers/azure.md
+++ b/docs/2.deploy/20.providers/azure.md
@@ -29,7 +29,7 @@ npx @azure/static-web-apps-cli start .output/public --api-location .output/serve
 
 Azure Static Web Apps are [configured](https://learn.microsoft.com/en-us/azure/static-web-apps/configuration) using the `staticwebapp.config.json` file.
 
-Nitro automatically generates this configuration file whenever the application is built with the `azure` preset.
+Nitro automatically generates this configuration file whenever the application is built with the `azure-swa` preset.
 
 Nitro will automatically add the following properties based on the following criteria:
 
--- a/docs/2.deploy/20.providers/azure.md
+++ b/docs/2.deploy/20.providers/azure.md
@@ -6,13 +6,13 @@
 
 **Preset:** `azure-swa`
 
-:read-more{title="Azure Static Web Apps" to="https://azure.microsoft.com/en-us/products/app-service/static"}
+:read-more{title="Azure Static Web Apps" to="[https://azure.microsoft.com/en-us/products/app-service/static"}](https://azure.microsoft.com/en-us/products/app-service/static"})
 
 ::note
 Integration with this provider is possible with [zero configuration](/deploy/#zero-config-providers).
 ::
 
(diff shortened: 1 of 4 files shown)


derkoe:
@pi0 could you please have a look at that? I've tried the fix and it fixes my issues with Tanstack using nitor on Azure SWA.

icecog:
After spending at least 10 hours over 3 days trying to get this to work I found @Xiang-CH 's post (https://cxiang.site/blog/deploy-full-stack-nitro-apps-to-azure-swa) and I finally got it to work! Thank you, @Xiang-CH!

@pi0, it would be magical to have it work right out of the box.
p.s. Your Azure SWA deploy doesn't work right now (https://github.com/nitrojs/nitro-deploys) (https://delightful-mushroom-0f662f303.4.azurestaticapps.net/base/).

In addition to Xiang's stuff I have a copy:local script in package.json that I run after the build to move a local.settings.json file from the root to the server folder.

```json
"scripts": {
"copy:local": "node -e \"require('fs').copyFileSync('local.settings.json', '.output/server/local.settings.json')\"",
"build:azure": "set DEPLOY_TARGET=azure&& vite build",
...
}
```
local.settings.json file is just the standard Azure Function settings file:
```json
{ 
    "IsEncrypted": false, 
    "Values": { 
        "FUNCTIONS_WORKER_RUNTIME": "node", 
     } 
}
```

and then in my Github Action I have to move the staticwebapp.config.json created in the project folder to the .output/public folder, otherwise the Azure/static-web-apps-deploy@v1 step fails.

```yml
- name: Prepare Azure Configuration
      run: |
        # 1. Move the file from the root to your output folder
        mv staticwebapp.config.json .output/public/
```

pi0x:
- Rebased onto latest `main` and force-pushed. The conflict was in `docs/2.deploy/20.providers/azure.md`: `main` has since rewritten that page and already fixed the preset name and the Node version table. So I dropped the docs commits and kept `main`'s version. The PR is now the runtime fix only.
- The three runtime fixes look right to me. Dropping the request headers and passing a stream as the Azure response body were both real bugs.
- I added unit tests for `resolveBaseUrl` in `test/unit/azure.utils.test.ts`, since the fix had no test.
- One question: the path comes from `x-ms-original-url`, but the origin comes from `host` first. If those two disagree the URL is a mix. Would it be safer to take the origin from `x-ms-original-url` first when it is present?
- Small note: a comma-separated `x-forwarded-proto` (like `https,http`) makes `new URL` throw, so it warns and falls through. Probably worth taking the first value.

pi0 (approved):

PR #4216, opened 2026-04-20
docs(cache): invalidation using ocache

### LINKED ISSUE #3969: Programmatic cache invalidation online
### Describe the feature

**Motivation**
I want to let my editor choose when to invalidate the cache after content updates.

**Problem**
The current implementation sets maxAge, so the file may get's cached by the browser itself.
I have therefor no tool to revoke the browser cache or tell the browser to invalidate the cache.

**Caveat**
Please correct me if I am wrong or point me to the right direction if I missed something.

**Solution**
https://github.com/nitrojs/nitro/blob/main/src/runtime/internal/cache.ts#L283

If we would allow custom cacheControl we could customize the behavior as we want.
it could also be a function to be more flexible.

```
const cacheControl = [];
if (opts.customCacheControl) {
cacheControl.push(opts.customCacheControl)
} else if (opts.swr) {

}
```

**Benefit**
I could set customCacheControl to no-store and have always served the cached files until I regenerate them by a webhook call.

### Additional information

RihanArfan:
### 🔗 Linked issue

Closes #3969 

### ❓ Type of change

### 📚 Description

Updates cache invalidation to use ocache's new invalidation methods

### 📝 Checklist

### DIFF
--- a/docs/1.docs/7.cache.md
+++ b/docs/1.docs/7.cache.md
@@ -328,7 +328,68 @@ When `swr` is disabled and a cached value has expired:
 1. The stale entry is cleared.
 2. The client waits for the function/handler to resolve with a fresh value.
 
-## Cache keys and invalidation
+## Cache invalidation
+
+Cached entries can be invalidated programmatically at runtime (for example from a webhook when the underlying data changes) without waiting for `maxAge` to expire.
+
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thnx!

PR #4229, opened 2026-04-25
fix(config): prepend bun export condition under bun runtime

### LINKED ISSUE #4228: vite: dev server breaks under bun --bun runtime due to "node" export condition
### Environment

- nitro (nitro-nightly): 3.0.1-20260425-084511-feebdc1e
- bun: 1.3.13
- vite: 8.0.10
- OS: macOS 25.3.0 (arm64)

### Reproduction

Minimal Nitro + Vite app with one task, dev script `bun --bun vite dev`, then:

```sh
curl -i http://localhost:3000/_nitro/tasks/hello
```

### Describe the bug

Under `bun --bun vite dev`, every request that reaches the Nitro dev worker (`/_nitro/*` devtools endpoints, plus any user route handled by the `nitro` Vite environment) fails with `500 Parse Error` from `node:_http_client:230`. Under plain `bunx vite dev` (Node runtime), the same setup works.

### Root cause (verified)

`_resolveExportConditions` in `nitro/dist/_chunks/nitro.mjs:154` always pushes `"node"` to Vite's resolve conditions, regardless of the actual dev runtime:

```js
if (opts.node) conditions.push("node");
```

Under `bun --bun`, this causes Vite's `ModuleRunner` (inside the dev worker) to load `srvx/dist/adapters/node.mjs` and `h3/dist/_entries/node.mjs`. The Node adapter's `NodeResponse` wrapper class is then returned from h3's response pipeline. `Bun.serve` (the worker's outer server, which used the *Bun* srvx adapter at startup) rejects the `NodeResponse` with `error: Expected a Response object, but received 'NodeResponse {...}'` and falls back to its default landing page, which Node's HTTP parser then fails to parse.

### Verified fix

```diff
-	if (opts.node) conditions.push("node");
+	if (opts.node) {
+		if (typeof globalThis.Bun !== "undefined") conditions.push("bun");
+		else conditions.push("node");
+	}
```

I patched this locally and rebuilt — `/_nitro/tasks`, `/_nitro/tasks/hello` (GET and POST) all return `200 {"result":"ok"}` under `bun --bun vite dev`. Node-runtime users see no change (the `else` branch preserves current behavior).

### Logs

```
error: Expected a Response object, but received 'NodeResponse {
  status: [Getter],
  statusText: [Getter],
  ...
}'
[vite] Internal server error: Parse Error
      at unknown
      at <anonymous> (node:_http_client:230:56)
      at processTicksAndRejections (native:7:39)
```

Modules loaded by ModuleRunner (excerpt):

```
[module runner] fetching file:///.../node_modules/srvx/dist/adapters/node.mjs
[module runner] fetching file:///.../node_modules/h3/dist/_entries/node.mjs
```

sotasan:
### Linked Issue

Closes #4228

### Description

`_resolveExportConditions` only pushed `"node"` to Vite's resolve conditions, regardless of the actual dev runtime. Under `bun --bun vite dev`, this caused Vite's `ModuleRunner` (inside the dev worker) to load `srvx/dist/adapters/node.mjs` and `h3/dist/_entries/node.mjs`. The Node adapter's `NodeResponse` wrapper class is then returned from h3's response pipeline to `Bun.serve` — which rejects it (`error: Expected a Response object, but received 'NodeResponse {...}'`) and falls back to its default response, breaking every `/_nitro/*` request with a `Parse Error` from `node:_http_client:230`.

### Change

Detect Bun via `"Bun" in globalThis` and prepend `"bun"` to the conditions list when running under Bun. `"node"` is still pushed unconditionally, so packages that only declare a `"node"` export condition still resolve correctly. For packages declaring both (srvx, h3) the order of the resolve conditions causes `"bun"` to win — picking the Bun adapter as intended.

Pattern matches the existing `"Bun" in globalThis` check in `src/presets/vercel/utils.ts:438`.

### Verified

- Patched the bundled `dist/_chunks/nitro.mjs` in a real Nitro + Vite app — `GET /_nitro/tasks`, `GET/POST /_nitro/tasks/:name`, and the main `GET /` route all return 200 under `bun --bun vite dev`. Scheduled tasks continue to fire.
- `pnpm lint` passes.
- `pnpm typecheck` passes for the modified file (other unrelated type errors are pre-existing on `main`).
- `pnpm build` produces a clean bundle with the expected logic.
- Pure Node setups: behavior is identical to before (the `"Bun" in globalThis` check is false, so the only conditions pushed are the same as `main`).

### DIFF
--- a/src/config/resolvers/export-conditions.ts
+++ b/src/config/resolvers/export-conditions.ts
@@ -21,6 +21,9 @@ function _resolveExportConditions(
   }
 
   if (opts.node) {
+    if ("Bun" in globalThis) {
+      conditions.push("bun");
+    }
     conditions.push("node");
   }
 


pi0:
Thanks for PR! Same effect should be happening with https://github.com/nitrojs/nitro/pull/4397 now in vite dev (for prod, bun and deno presets push their own conditions)

PR #4246, opened 2026-05-07
feat!: update minimum Node.js version to 22

gameroman:
Node.js 20 is EOL


pi0:
Thnx for PR. Although we should eventually upgrader engines field there is no reason to hard enforce it early as ecosystem needs more time to upgrade. I checked rolldown v1 is still minimum v20 too btw

PR #4258, opened 2026-05-11
fix(cloudflare): auto-add .wrangler/state/v3 to .gitignore on default persistDir

jel-massih:
In Cloudflare dev emulation, Nitro tries to auto-append `.wrangler/state/v3` to the project `.gitignore` so Wrangler emulator state doesn't get committed.

The guard compares a resolved absolute path against the relative literal `".wrangler/state/v3"`, so it never matches and the entry is never added. Switching to `!config.persistDir` makes the check fire when the user is on the default location.

(Also removed dead commented code but can keep that if preffered.)

### DIFF
--- a/src/presets/cloudflare/dev.ts
+++ b/src/presets/cloudflare/dev.ts
@@ -53,12 +53,10 @@ export async function cloudflareDevModule(nitro: Nitro) {
     startingFrom: nitro.options.rootDir,
   }).catch(() => undefined);
 
-  // let addedToGitIgnore = false;
-  if (gitIgnorePath && persistDir === ".wrangler/state/v3") {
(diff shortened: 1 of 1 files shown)


Montiwa11:
src/presets/cloudflare/dev.

pi0 (approved):
Thanks!

pi0x:
Superseded by #4338, which replaced the wrangler-based Cloudflare local dev with miniflare/env-runner and removed `src/presets/cloudflare/dev.ts` entirely. The `.gitignore`/`persistDir` handling this PR fixed no longer exists — miniflare now manages its own `.wrangler` state — so there's no remaining code path for the fix. Closing as obsolete.

PR #4263, opened 2026-05-12
fix(security): prevent SSRF (CWE-918) in proxy route rules and path traversal bypass (CWE-22)

absalonCRC:
## Summary

Fixes two HIGH severity security vulnerabilities discovered during a code audit of the Nitro runtime.

### 1. SSRF via Unvalidated Proxy Route Rules (CWE-918)

**File:** `src/runtime/internal/route-rules.ts`, lines 105-128

The `proxy` route rule handler passes the user-configured `target` URL directly to `proxyRequest()` without any validation. A compromised Nitro configuration or misconfigured preset could proxy traffic to internal hosts (cloud metadata endpoints, internal services, localhost).

**Fix:**
- Added `PRIVATE_IPV4_RANGES` blocklist covering all RFC 1918 ranges (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`), loopback (`127.0.0.0/8`), link-local (`169.254.0.0/16`), and `0.0.0.0`
- Added `isPrivateHost()` to detect private/internal hostnames
- Added `assertSafeProxyTarget()` that throws HTTP 400 when a proxy target resolves to a private host
- Gracefully handles relative URLs (which can only target same-origin paths)
- Called before every `proxyRequest()` invocation

### 2. Path Traversal via Double-Encoding Bypass (CWE-22)

**File:** `src/runtime/internal/route-rules.ts`, lines 175-189

The `isPathInScope()` function (introduced to fix GHSA-5w89-w975-hf9q) pre-decodes `%2f` → `/` and `%5c` → `\` before checking path containment. However, it only performs a single pass, making it vulnerable to double-encoding attacks. An attacker could send `..%252f..%252f` which decodes to `..%2f..%2f` (passing the single-pass check), then a downstream decoder normalizes it to `../..`.

**Fix:**
- Replaced single-pass replacement with recursive decoding
- The loop continues until no more `%2f` or `%5c` sequences remain
- This prevents double-encoding bypasses while maintaining backward compatibility

## Verification

Both fixes are additive and backward-compatible:
- Valid proxy targets continue to work normally
- Valid path names within the scope continue to pass the check
- Only unsafe targets and path traversal attempts are blocked

## Related

- CWE-918: Server-Side Request Forgery (SSRF)
- CWE-22: Improper Limitation of a Pathname to a Restricted Directory
- GHSA-5w89-w975-hf9q: Previously addressed path traversal in `/**` route rules

### DIFF
--- a/src/prerender/utils.ts
+++ b/src/prerender/utils.ts
@@ -39,7 +39,27 @@ export async function extractLinks(html: string, from: string, res: Response, cr
 
   // Extract from x-nitro-prerender headers
   const header = res.headers.get("x-nitro-prerender") || "";
-  _links.push(...header.split(",").map((i) => decodeURIComponent(i.trim())));
+  const prerenderLinks = header
+    .split(",")
+    .map((i) => decodeURIComponent(i.trim()))
+    .filter((link) => {
+      // Validate prerender header links to prevent injection of
+      // malicious paths, traversal sequences, or external URLs
+      if (!link || link.includes("..") || link.includes("//")) {
+        return false;
+      }
+      try {
+        const parsed = new URL(link, "http://localhost");
+        // Reject absolute URLs (external redirects) and URLs not starting with /
+        if (parsed.protocol !== "http:" || parsed.host !== "localhost") {
+          return false;
+        }
+        return parsed.pathname.startsWith("/");
+      } catch {
+        return false;
+      }
+    });
+  _links.push(...prerenderLinks);
 
   for (const link of _links.filter(Boolean)) {
     const _link = parseURL(link);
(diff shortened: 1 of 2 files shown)


absalonCRC:
@pi0 @danielroe This PR fixes 3 security vulnerabilities including 2 HIGH severity (SSRF CWE-918 + Path Traversal CWE-22). Would appreciate a maintainer review. These were found during a security audit of the Nitro runtime.

pi0:
Being able to proxy requests to a (fixed/trusted) internal address is a delibrate feature of Nitro to use it as a reverse bff proxy.

According to RFC 3986, percent-decoding should be applied exactly once, double decoding violates it and break legitimate URLs containing %25 sequences

Side notes: Please consider reporting security vulnerabilities according to the [Security Policy](https://github.com/nitrojs/nitro?tab=security-ov-file#readme) of project.

Issue #4241, opened 2026-04-30
Regression - Sec-fetch-dest: image header causes 404 on API-served images (regression in 3.0.260429-beta)


TheUniqueMathemagician:
### Runtime
- node 24.7.0
- npm 11.5.1
- bun 1.3.3

### Nitro
- nitro 3.0.260429-beta

### Reproduction example repository
https://github.com/TheUniqueMathemagician/nitro-issue

### To reproduce
Make a request to a TanStack Start API endpoint (e.g. `/api/images/$id/file`) that serves images dynamically, and include the `sec-fetch-dest: image` header, or simply use an `<img>` tag pointing to that endpoint, as browsers (e.g. Brave, Chrome, Edge, ...) add this header automatically for image requests.

### Describe the bug

After upgrading from `3.0.260415-beta` to `3.0.260429-beta`, dynamically served images return `404 Not Found` when the request includes the `Sec-Fetch-Dest: image` header.
This header is automatically added by browsers whenever an `<img>` tag loads a resource. The images are not static files in a `/public` folder, they are served through TanStack Router's API route at `/api/images/$id/file` (in my case).

The regression appears to be introduced by [PR #4238](https://github.com/nitrojs/nitro/pull/4238/changes/b15f297208849123f72368916e4393ada0159369), which added sec-fetch-dest-aware logic that seems to intercept requests with sec-fetch-dest: image before they reach the API handler, returning a 404 instead.

All other request methods (direct browser URL, cURL, Postman) work correctly when the `Sec-Fetch-Dest` header is absent.

### Additional context

The following Nitro config options were attempted to disable static asset serving, but none had any effect on the issue — the interception appears to happen upstream of the asset configuration:

```typescript
nitro({
    apiBaseURL: "/api",
    noPublicDir: true,
    serveStatic: false,
    // publicAssets: [],
    publicAssets: [{
        baseURL: "/",
        dir: resolve(import.meta.dirname, "publics"),
        fallthrough: true,
        ignore: ["**/*"],
        maxAge: 60 * 60 * 24 * 7,
    }],
})
```

### Logs

```sh
(no additional logs — the only observable symptom is a 404 response on the image endpoint when Sec-Fetch-Dest: image is present)
```

pi0:
Can you please make a minimal (nitro only) reproduction?

TheUniqueMathemagician:
I created two additional branches on https://github.com/TheUniqueMathemagician/nitro-issue to help you narrow down the issue:

[...]

Issue #4234, opened 2026-04-28
Accessing via a non-localhost URL breaks vite dev server


dunklesToast:
### Environment

node v24.10.0
nitro 3.0.260415-beta

### Reproduction

https://github.com/dunklesToast/tanstack-nitro-vite-allowedhost

### Describe the bug

it seems that if you use a `allowedHost` domain which just resolves to your machine / `127.0.0.1` the vite nitro
addon breaks the vite dev mode and returns the content of the splat page instead of the vite modules. 

I suspect  https://github.com/nitrojs/nitro/pull/4108 to be the source of this issue as it worked in 3.0.260311-beta and  when patching nitro to log the `isNitroRoute` boolean, it will be true when accessing via localhost:3000 but false when using any other domain.

(more details can be found in the repro-repos README.md)

### Additional context

Sorry for not using the provided template but Stackblitz was just stuck at "Cloning repo from GitHub" for 15min

### Logs

```sh

```

dunklesToast:
Thank you very much for the quick fix, @pi0 - I've already tested it via `pnpm install https://pkg.pr.new/nitro@6016153` and then with the repro - works 🚀

pi0:
Awesome! Thanks for confirming the fix works.

PR #4266, opened 2026-05-14
fix(dev): route specific Nitro paths even when URL has asset extension

### LINKED ISSUE #4252: Vite dev middleware swallows asset-extensioned URLs
Vite dev middleware swallows asset-extensioned URLs (`.jpg`/`.png`/…) even when a Nitro route matches — `req._nitroHandled` is set unconditionally in `nitroDevMiddlewarePre`.

### Environment

- `nitro@3.0.260429-beta`
- `vite@8.0.11`
- `h3@2.0.1-rc.22`
- Node 24 / macOS

### Summary

When the Nitro Vite plugin is in use, the dev server intercepts any URL whose extension matches the built-in `ASSET_EXT_RE` (jsx/tsx/png/**jpg**/svg/webp/avif/woff2/…) and routes it to Vite's static middleware *even if a user-defined Nitro route matches the URL*. The intercept also poisons the late-stage Nitro fallback by setting `req._nitroHandled = true`, so the request never gets a chance to reach the user's route handler. Vite's `serve-static@2.x` then emits the canonical `404 + Content-Security-Policy: default-src 'none' + X-Content-Type-Options: nosniff`.

URLs without an extension, or with extensions not in `ASSET_EXT_RE` (e.g. `.txt`), reach the route handler normally. This produces a confusing asymmetry where `/api/photos/12345.jpg` 404s but `/api/photos/12345.txt` works.

### Reproduction

Minimal vanilla repro (no TanStack Start / Nuxt / etc. needed):

`package.json`
```json
{
  "name": "nitro-static-demo",
  "private": true,
  "type": "module",
  "scripts": { "dev": "vite" },
  "devDependencies": {
    "h3": "2.0.1-rc.22",
    "nitro": "3.0.260429-beta",
    "vite": "^8.0.10"
  }
}
```

`vite.config.ts`
```ts
import { defineConfig } from "vite";
import { nitro } from "nitro/vite";
export default defineConfig({ plugins: [nitro()] });
```

`nitro.config.ts`
```ts
import { defineConfig } from "nitro/config";
export default defineConfig({ scanDirs: ["."] });
```

`routes/api/photos/[...].ts`
```ts
import { defineEventHandler } from "h3";
export default defineEventHandler((event) => ({
  handlerReached: true,
  url: event.url?.pathname ?? event.path,
  params: event.context.params,
}));
```

Then:
```sh
pnpm install && pnpm dev
curl -si http://localhost:3000/api/photos/12345        # 200, handler reached
curl -si http://localhost:3000/api/photos/12345.txt    # 200, handler reached
curl -si http://localhost:3000/api/photos/12345.jpg    # 404 with `Content-Security-Policy: default-src 'none'`
```

### Bisect

Same project, only the `nitro` version changed:

| `nitro` version | `/api/photos/12345` | `/api/photos/12345.txt` | `/api/photos/12345.jpg` |
| --- | --- | --- | --- |
| `3.0.1-alpha.2` | 200 | 200 | **200** |
| `3.0.260429-beta` | 200 | 200 | **404 (serve-static@2.x)** |

Vite version held constant at 8.0.11. Reverting Vite alone does not fix it; reverting Nitro alone does. So this is a Nitro regression introduced between `3.0.1-alpha.2` and `3.0.260429-beta`.

### Root cause

`nitro/dist/vite.mjs`, in the `nitroDevMiddlewarePre` middleware:

```js
// nitro/dist/vite.mjs (3.0.260429-beta)
const ASSET_EXT_RE = /^(?:[jt]sx?|mjs|cjs|css|s[ac]ss|less|styl|vue|svelte|astro|mdx?|map|wasm|png|jpe?g|gif|svg|webp|avif|ico|bmp|woff2?|ttf|otf|eot|mp[34]|webm|wav|ogg|m4a)$/i;
// …
server.middlewares.use(function nitroDevMiddlewarePre(req, res, next) {
  const fetchDest = req.headers["sec-fetch-dest"];
  const accept = req.headers["accept"];
  const ext = req.url.match(/\.([a-z0-9]+)(?:[?#]|$)/i)?.[1];
  const isNitroRoute = ext
    ? !!nitro.routing.routes.match(req.method || "", new URL(withBase(req.url, nitro.options.baseURL), "http://localhost").pathname)
    : false;
  const isDocumentLike = fetchDest
    ? /^(document|iframe|frame|empty)$/.test(fetchDest)
    : !(ext && ASSET_EXT_RE.test(ext) && !(typeof accept === "string" && /\btext\/html\b/.test(accept)));
  res.setHeader("vary", "sec-fetch-dest, accept");
  if (isDocumentLike && (!ext || isNitroRoute) && !/^\/(?:__|@)/.test(req.url)) {
    nitroDevMiddleware(req, res, next);
  } else {
    if (!isDocumentLike) req._nitroHandled = true;   // <-- regression
    next();
  }
});
```

Two problems:

1. The condition `if (isDocumentLike && (!ext || isNitroRoute) && …)` already computes `isNitroRoute`, but only consults it under the `isDocumentLike` gate. Asset-extension URLs always fail `isDocumentLike`, so a matching Nitro route can never be reached for them.
2. The `else` branch sets `req._nitroHandled = true` whenever `!isDocumentLike`, regardless of whether a Nitro route exists. The late-stage Nitro fallback registered by the same plugin (`return () => { server.middlewares.use(nitroDevMiddleware); }`) checks this flag and bails out, so even if Vite's static middleware would have fallen through, Nitro never sees the request.

For comparison, in `3.0.1-alpha.2` the same middleware looked like:

```js
server.middlewares.use(function nitroDevMiddlewarePre(req, res, next) {
  const fetchDest = req.headers["sec-fetch-dest"];
  res.setHeader("vary", "sec-fetch-dest");
  if (
    (!fetchDest || /^(document|iframe|frame|empty)$/.test(fetchDest)) &&
    !req.url.match(/\.([a-z0-9]+)(?:[?#]|$)/i)?.[1] &&
    !/^\/(?:__|@)/.test(req.url)
  ) nitroDevMiddleware(req, res, next);
  else next();
});
```

No `_nitroHandled` poisoning, so the late-stage fallback got the request whenever Vite's static middleware fell through.

### Suggested fix

Let a matching Nitro route override the asset-extension shortcut, and don't poison the fallback when one exists:

```js
if ((isDocumentLike || isNitroRoute) && !/^\/(?:__|@)/.test(req.url)) {
  nitroDevMiddleware(req, res, next);
} else {
  if (!isDocumentLike && !isNitroRoute) req._nitroHandled = true;
  next();
}
```

This preserves the optimization (avoid a Nitro round-trip for genuine static-asset URLs) while ensuring user-defined routes always win when they actually match.

### Smoking-gun headers

When the bug fires, the 404 response carries:

```
Content-Security-Policy: default-src 'none'
X-Content-Type-Options: nosniff
Content-Type: text/html; charset=utf-8
```

These are the canonical `serve-static@2.x` "not found" headers — handy for confirming whether a 404 came from Vite's static middleware vs Nitro's own routing layer.

harshagarwalnyu:
Asset-extensioned URLs (e.g. `/api/photos/12345.jpg`) were routed to Vite's static middleware
instead of the matching Nitro handler. The `ASSET_EXT_RE` guard in `nitroDevMiddlewarePre`
applied unconditionally to any matched route, including specific prefixed catch-alls like `/api/photos/**`.

The guard was introduced (#4234) to prevent a root-level renderer or user catch-all (`/**`) from
swallowing Vite's own `<script>`/`<link>` serves on plain-HTTP non-loopback origins where
`Sec-Fetch-Dest` is absent.

**Fix**: narrow the guard so the asset-extension filter only applies when the matched Nitro route
pattern is exactly `/**`. Specific routes (`/api/**`, `/api/photos/**`) are never registered on
paths Vite owns and should always win — restoring the behaviour from `nitro@3.0.1-alpha.2`.

## Test plan

- [ ] Route `routes/api/photos/[...].ts` → `curl http://localhost:3000/api/photos/12345.jpg` returns 200 from handler (was 404)
- [ ] Root-level renderer still handles page navigations normally
- [ ] `<script src="/src/app.ts">` still served by Vite (not swallowed by renderer catch-all)
- [ ] Requests with `Sec-Fetch-Dest: image` for explicit Nitro routes still reach Nitro (#4241)

Fixes #4252

### DIFF
--- a/src/build/vite/dev.ts
+++ b/src/build/vite/dev.ts
@@ -245,10 +245,9 @@ export async function configureViteDevServer(ctx: NitroPluginContext, server: Vi
     const fetchDest = req.headers["sec-fetch-dest"];
     const accept = req.headers["accept"];
     const ext = req.url!.split(/[?#]/, 1)[0].match(/\.([a-z0-9]+)$/i)?.[1];
-    const isNitroRoute = !!nitro.routing.routes.match(
-      req.method || "",
-      new URL(withBase(req.url!, nitro.options.baseURL), "http://localhost").pathname
-    );
+    const reqPathname = new URL(withBase(req.url!, nitro.options.baseURL), "http://localhost").pathname;
+    const nitroRouteMatch = nitro.routing.routes.match(req.method || "", reqPathname);
+    const isNitroRoute = !!nitroRouteMatch;
     // Sec-Fetch-* is only sent on "potentially trustworthy" origins, so on plain-HTTP non-loopback (e.g. http://10.0.0.x) it's absent and a splat Nitro route may swallow browser asset loads (#4234). When the header is missing, treat known asset extensions without `text/html` in Accept as asset loads and let Vite handle them.
     const isAssetByDest =
       typeof fetchDest === "string" && !/^(document|iframe|frame|empty)$/.test(fetchDest);
@@ -256,8 +255,17 @@ export async function configureViteDevServer(ctx: NitroPluginContext, server: Vi
     const acceptsHTML = typeof accept === "string" && /\btext\/html\b/.test(accept);
     const treatAsAsset = isAssetByDest || (!fetchDest && isAssetByExt && !acceptsHTML);
     res.setHeader("vary", "sec-fetch-dest, accept");
-    // An explicit Nitro route reaches Nitro even when the request is tagged as an asset (e.g. `<img src="/api/image">` with `sec-fetch-dest: image`, #4241), UNLESS the URL also has an asset-like extension — in that case Vite stays the definitive handler so a splat doesn't swallow `<script src=".../entry-client.ts">` (#4234).
-    const nitroWins = isNitroRoute && !(isAssetByExt && treatAsAsset);
+    // An explicit Nitro route reaches Nitro even when the request is tagged as an asset (#4241).
+    // The asset-extension guard only applies to root-level wildcards (/**) to prevent a renderer
+    // or user catch-all from swallowing Vite's own <script>/<link> serves (#4234). Specific
+    // prefixed routes (e.g. /api/photos/**) are never wildcards on Vite-managed paths (#4252).
+    const matchedRoutePattern = Array.isArray(nitroRouteMatch)
+      ? nitroRouteMatch[0]?.route
+      : nitroRouteMatch?.route;
+    const isRootWildcard =
+      matchedRoutePattern === "/**" ||
+      matchedRoutePattern?.startsWith("/**:");
+    const nitroWins = isNitroRoute && (!isRootWildcard || !(isAssetByExt && treatAsAsset));
     // Fallback for unknown URLs: extensionless, non-asset requests default to Nitro (page navigation, SSR catch-all).
     const documentFallback = !ext && !treatAsAsset;
     const routeToNitro =
--- a/test/vite/baseurl-dotted-param-fixture/routes/[...path].ts
+++ b/test/vite/baseurl-dotted-param-fixture/routes/[...path].ts
@@ -1,0 +1,6 @@
+import type { EventHandler } from "h3";
+
+export default ((event) => {
+  const path = event.context.params?.path ?? "";
+  return `root-wildcard:${path}`;
+}) satisfies EventHandler;
--- a/test/vite/baseurl-dotted-param.test.ts
+++ b/test/vite/baseurl-dotted-param.test.ts
@@ -29,7 +29,6 @@ describe("vite:baseURL dotted params", { sequential: true }, () => {
   });
 
   test("serves Nitro API routes with dotted params under baseURL without redirecting", async () => {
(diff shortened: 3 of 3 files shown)


harshagarwalnyu:
@coderabbitai review:
- Removed inline comments throughout the test file
- Fixed handler typing (`satisfies EventHandler`)
- Moved the two stray tests (`prefixed splat routes win...` and `root-level wildcards do not swallow...`) inside the `describe` block — they were accidentally at top-level scope, meaning they had no closure access to `serverURL` or `server`. This was a real correctness issue, not just a style nit.

harshagarwalnyu:
Squashed all prior commits into one (bbea608). The comment referencing commit 3a48bb9 above is now part of that single commit — no logic changes beyond the squash.

harshagarwalnyu:
@coderabbitai review

pi0:
Thanks a lot for this fix and the clear analysis ❤️ I went with an alternative approach in #4272 that builds on the same root-vs-explicit distinction — it also covers the named root catch-all (/**:slug) case you raised here 🙏

PR #4271, opened 2026-05-18
docs: correct tsconfig paths alias in `vite-ssr-tsr-react` example

ZDSJdeJT:
The `@/*` path alias in `examples/vite-ssr-tsr-react/tsconfig.json` pointed to `sec/*`, which does not exist. This causes TypeScript to fail resolving modules imported via `@/`.

Changed `"sec/*"` to `"./src/*"` to match the actual project structure and align with the convention used in `examples/vite-ssr-tss-react/tsconfig.json`.


pi0 (approved):
thnx!

PR #4274, opened 2026-05-20
fix(vite): respect tanstack start route manifest in dev

pi0:
Fixes the TanStack Start regression reported in [TanStack/router#7403](https://github.com/TanStack/router/issues/7403) (specifically [this comment](https://github.com/TanStack/router/issues/7403#issuecomment-4492301829)): `` returns `404 text/html` in dev, even though hitting the URL directly works.

## Why #4272 was not enough

#4272 made the dev middleware consult `nitro.routing.routes` to decide whether a request is an explicit user route. But TanStack Start API routes live at `src/routes/api/**.ts` and register via `viteDevServer.middlewares.use(...)`, so they never appear in `nitro.routing.routes` — that table only contains the SSR catch-all `/**`. As a result, `isExplicitRoute` is always `false` for Start API routes, and any image-context request (`Sec-Fetch-Dest: image`) falls into the asset branch and is claimed by Vite.

## Fix

- During dev, consult `globalThis.TSS_ROUTES_MANIFEST` (published by `@tanstack/start-plugin-core`'s `routes-manifest-plugin` for sibling Vite plugins) so URLs matching a TanStack Start filesystem route are treated as explicit.
- These requests then flow through nitro's dev middleware (where Start's `devApp` gets first shot) instead of being classified by the asset/page heuristic.
- Bumps `@tanstack/react-router`, `@tanstack/react-router-devtools`, and `@tanstack/react-start` in the `vite-ssr-tss-react` example to versions that publish the manifest (separate commit).

## Verification

Verified locally against `examples/vite-ssr-tss-react` (Start app whose API routes are not visible to nitro), with the fix toggled on/off:

| Request | Without fix | With fix |
|---|---|---|
| `GET /api/test` (plain) | 200 json | 200 json |
| `GET /api/test` + `Sec-Fetch-Dest: image` | **404 html** | 200 json |
| `GET /api/files/42` + `Sec-Fetch-Dest: image` (parametrized `\$id`) | **404 html** | 200 json |
| `GET /api/files/foo.png` + `Sec-Fetch-Dest: image` (`.png` in path param) | **404 html** | 200 json |
| `GET /api/nope/x` + image (unmatched) | 404 | 404 |
| `GET /favicon.ico` + image (real asset) | 404 | 404 |

## Caveat

The probe relies on the undocumented `globalThis.TSS_ROUTES_MANIFEST` contract. If Start renames/removes it, the fallback silently disables and we regress to the #4272 behavior. Worth coordinating with the Start team so the contract is intentional.

### DIFF
--- a/docs/4.examples/vite-ssr-tss-react.md
+++ b/docs/4.examples/vite-ssr-tss-react.md
@@ -20,9 +20,9 @@ icon: i-simple-icons-tanstack
     "start": "node .output/server/index.mjs"
   },
   "dependencies": {
-    "@tanstack/react-router": "^1.168.8",
-    "@tanstack/react-router-devtools": "^1.166.11",
-    "@tanstack/react-start": "^1.167.13",
+    "@tanstack/react-router": "^1.170.4",
+    "@tanstack/react-router-devtools": "^1.167.0",
+    "@tanstack/react-start": "^1.168.6",
     "nitro": "latest",
     "react": "^19.2.4",
     "react-dom": "^19.2.4",
--- a/examples/vite-ssr-tss-react/package.json
+++ b/examples/vite-ssr-tss-react/package.json
@@ -6,9 +6,9 @@
     "start": "node .output/server/index.mjs"
   },
   "dependencies": {
-    "@tanstack/react-router": "^1.168.8",
-    "@tanstack/react-router-devtools": "^1.166.11",
-    "@tanstack/react-start": "^1.167.13",
+    "@tanstack/react-router": "^1.170.4",
+    "@tanstack/react-router-devtools": "^1.167.0",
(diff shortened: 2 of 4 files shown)


bhouston:
I ran into this issue myself: https://github.com/bhouston/tanstack-start-image-nitro-conflict

And I have tested out this PR and it does fix the issue: https://github.com/bhouston/tanstack-start-image-nitro-conflict/pull/1

Looking forward to seeing this merged.

jimmy-chiang:
I've encountered this issue too. I tried "nitro": "^3.0.260610-beta", but it still hasn't been solved. Hope this will be merged in the near future.

pi0:
should be handled better on main now

PR #4291, opened 2026-05-29
fix(vite): only skip vite-* WebSocket upgrades destined for Vite's HMR base

pi0x:
## Summary

In dev mode, Nitro's Vite integration installed an `upgrade` listener that skipped **every** WebSocket upgrade whose `Sec-WebSocket-Protocol` started with `vite-`. The intent — don't fight Vite's own HMR socket — was right, but the condition was too broad: it keyed off the subprotocol alone and ignored the request path.

As a result, a Nitro route that reverse-proxies an upstream Vite dev server never received the upstream's proxied HMR upgrade (`vite-hmr` on a non-HMR path): Nitro skipped it and outer Vite ignored it (wrong path), so the socket hung in `pending` forever.

## Fix

Mirror Vite's own claim condition in `configureViteDevServer` (`src/build/vite/dev.ts`): only defer to Vite when the subprotocol is `vite-hmr`/`vite-ping` **and** the request path equals Vite's resolved HMR base. Also account for `server.hmr.{server,port,path}` — when HMR runs on a separate server/port, Vite attaches no listener here, so Nitro must not defer.

## Tests

`test/unit/vite-ws.test.ts` spins up a real Nitro+Vite dev server with a WebSocket handler on a non-HMR path and performs a raw `vite-hmr` upgrade handshake. It fails (hangs to timeout) on the old code and passes with the fix; a plain upgrade is covered as a control. `pnpm typecheck` and `pnpm fmt` are clean.

### DIFF
--- a/src/build/vite/dev.ts
+++ b/src/build/vite/dev.ts
@@ -11,7 +11,7 @@ import { watch as chokidarWatch } from "chokidar";
 import { watch as fsWatch } from "node:fs";
 import { join } from "pathe";
 import { debounce } from "perfect-debounce";
-import { withBase } from "ufo";
+import { joinURL, withBase } from "ufo";
 import { scanHandlers } from "../../scan.ts";
 import { getEnvRunner } from "./env.ts";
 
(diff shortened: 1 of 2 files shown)


pi0x:
- I know this is a draft; I left the draft status alone.
- Rebased onto main. One small conflict: the `ufo` import line, since main added `withoutBase`. Nothing else moved. Push is blocked in my environment, so nothing was pushed.
- I checked the new condition against Vite 8's own code and it matches: `["vite-hmr","vite-ping"].includes(protocol) && pathname === hmrBase`.
- The test is a real regression test: it fails on main and passes with the fix.
- One edge case: if a user sets `server.hmr.server` to the *same* http server, Vite does listen here, but we would return `undefined` and take the upgrade away from it. Rare, but worth a guard.
- Small nit: `test/vite/` already holds vite dev-server tests with checked-in `*-fixture` dirs. This test would fit better there than in `test/unit/` with a `mkdtemp` inside the repo (stray `.tmp-vite-ws-*` dirs are left behind if it crashes, and they are not gitignored).

PR #4313, opened 2026-06-04
fix: try to resolve server entry also from server dir

pi0x:


src/config/resolvers/paths.ts
```diff
@@ -92,7 +92,10 @@ export async function resolvePathOptions(options: NitroOptions) {
     } else {
       const detected = resolveModulePath("./server", {
         try: true,
-        from: options.rootDir,
+        from:
+          options.serverDir && options.serverDir !== options.rootDir
+            ? [options.rootDir, options.serverDir]
```
pi0:
```suggestion
            ? [options.serverDir, options.rootDir]
```

PR #4315, opened 2026-06-05
feat: add task concurrency control

Teages:
### 🔗 Linked issue

#1974

### ❓ Type of change

### 📚 Description

The PR added a `concurrency` option to task, allow developer to control task concurrency behavior.

By default it is `dedupe` and `key: (event) => ohash(event.payload)`.

### 📝 Checklist


pi0:
Thanks for PR. This is nice to add control but i suggest to keep it simple. Main thing we should do (by default) is to hash task params and make sure name+params is unique (i made pr draft to update it)

Teages:
This stems from our actual needs: some data synchronization tasks can be dedupeed, but some update tasks should run in serial.

A custom key function is also necessary: ​​I usually add some extra parameters to params, which only used in logging.

@pi0 I have update the default key to `ohash(event.payload)`. It is a breaking change now, but it is experimental feature so that should be fine.

pi0x:
Rebased this on latest main locally. Only the docs concurrency section conflicted, and it was easy to resolve.

- Blocker: `ohash` is not a Nitro dependency, so `pnpm typecheck` fails with `Cannot find module 'ohash'`. Please add it to `dependencies` and to `tracePkgs` in `build.config.ts`, next to `croner`. With that added, all 11 of your tests pass.
- Docs nit: in the example, `payload.tenant` is `unknown`, so the `key` function needs `String(payload.tenant)` to typecheck.
- Heads up: #4416 and #3237 also change `runTask` / `startScheduleRunner`. Whichever lands first, the others will need a rebase. Of the two API proposals, yours reads closer to how Nitro does things.

PR #4345, opened 2026-06-12
fix(task): add request timeout to devFetch so a stalled dev server rejects instead of hanging

### LINKED ISSUE #4292: fix(task): add per-request timeout to `devFetch` http.request() call to prevent indefinite hang
## Summary

The `devFetch` helper defined in `src/task.ts` — used by both `runTask()` and `listTasks()` — wraps `http.request()` without setting any timeout. If the dev server worker is alive (so `_pidIsRunning` passes) but stalled or not accepting connections on the socket, the `Promise` returned by `devFetch` never settles. The caller — typically a CLI command or a script in a CI pipeline — hangs indefinitely with no error, no diagnostic, and no way to recover short of a job-level kill signal.

## What I observed

**File:** `src/task.ts` — introduced in PR #3268 (merged 2025-12-16).

```ts
// src/task.ts  lines 73-104
const request = http.request(
  url,
  {
    socketPath,
    method: options?.method,
    headers: {
      Accept: "application/json",
      "Content-Type": "application/json",
    },
    // ← no `timeout` option here
  },
  (response) => { /* … */ }
);

request.on("error", (e) => reject(e));
// ← no request.setTimeout() call, no AbortSignal
```

`http.request()` supports a `timeout` field in its options object (Node.js ≥ 8). When omitted, the socket waits until the OS TCP stack times out (often minutes, sometimes never on a Unix domain socket). There is no fallback rejection path.

## Suggested fix

Add a `timeout` option directly to the `http.request()` options, and destroy the socket on the `'timeout'` event so the connection is actually aborted:

```ts
const TIMEOUT_MS = 30_000; // 30 s — configurable via opts if desired

const request = http.request(
  url,
  {
    socketPath,
    method: options?.method,
    timeout: TIMEOUT_MS,
    headers: {
      Accept: "application/json",
      "Content-Type": "application/json",
    },
  },
  (response) => { /* … unchanged … */ }
);

request.on("timeout", () => {
  request.destroy(
    new Error(`devFetch timed out after ${TIMEOUT_MS}ms ${_devHint}`)
  );
});
request.on("error", (e) => reject(e));
```

Calling `request.destroy(err)` on `'timeout'` ensures the `'error'` handler fires with a meaningful message, so the existing `reject(e)` path surfaces to the caller. The timeout value (30 s) is a reasonable default; it could optionally be exposed on `TaskRunnerOptions` for users who run unusually slow tasks.

## Notes

- The `_pidIsRunning` guard (line 113) only checks whether the process is alive via `kill(pid, 0)` — it does not verify that the socket is accepting connections. So a zombie or newly-restarting worker can pass the guard while the socket is not yet ready, making this timeout the only backstop.
- PR #3268 replaced `ofetch` with bare `http.request` specifically to support `socketPath`; the timeout was not addressed in that refactor.
- A `TaskRunnerOptions.timeout?: number` extension would allow callers to set a tighter limit for CI use, but a safe default is the minimum viable fix.

spokodev:
Fixes #4292

### Problem

`devFetch` in `src/task.ts` (used by `runTask()` and `listTasks()`) wraps `http.request()` without a timeout. The `_pidIsRunning` guard only checks that the dev server process is alive via `kill(pid, 0)` — it does not verify the socket is accepting or answering requests. When the worker is alive but stalled (or restarting and not yet listening on the unix socket), the returned promise never settles, so `nitro task run` in a CI pipeline hangs until a job-level kill with no diagnostic.

### Fix

Set the `timeout` option on `http.request()` (default 30s) and destroy the request with a descriptive error on the `timeout` event so the existing `error` → `reject` path surfaces it to the caller. `TaskRunnerOptions` gains an optional `timeout` so callers (e.g. CI scripts) can set a tighter limit.

### Test

`test/unit/task.test.ts` spins an HTTP server on a unix socket that accepts connections but never responds, points `nitro.dev.json` at it with a live pid, and asserts `listTasks({ cwd, timeout: 200 })` rejects with the timeout error. On `main` the promise never settles and the test fails by timing out; with the fix it rejects in ~200ms.

`pnpm vitest run test/unit/` — the new test passes; the 4 pre-existing cloudflare preset failures reproduce identically on a clean checkout of `main` and are unrelated. `pnpm fmt` and `pnpm typecheck` clean.

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -37,6 +37,7 @@ const _devHint = `(is dev server running?)`;
 async function _getTasksContext(opts?: TaskRunnerOptions) {
   const cwd = resolve(process.cwd(), opts?.cwd || ".");
   const buildDir = resolve(cwd, opts?.buildDir || "node_modules/.nitro");
+  const timeout = opts?.timeout ?? 30_000;
 
   const buildInfoPath = resolve(buildDir, "nitro.dev.json");
   if (!existsSync(buildInfoPath)) {
@@ -75,6 +76,7 @@ async function _getTasksContext(opts?: TaskRunnerOptions) {
         {
           socketPath,
           method: options?.method,
+          timeout,
           headers: {
             Accept: "application/json",
             "Content-Type": "application/json",
@@ -95,6 +97,11 @@ async function _getTasksContext(opts?: TaskRunnerOptions) {
         }
       );
 
+      request.on("timeout", () => {
+        // destroy with an error so the `error` handler rejects instead of
+        // leaving the promise pending forever on a stalled socket
+        request.destroy(new Error(`Request timed out after ${timeout}ms ${_devHint}`));
+      });
       request.on("error", (e) => reject(e));
 
       if (options?.body) {
--- a/src/task.ts
+++ b/src/task.ts
@@ -13,6 +13,8 @@ export async function runTask(
   taskEvent: TaskEvent,
   opts?: TaskRunnerOptions
 ): Promise<{ result: unknown }> {
+  // No default timeout: the socket stays idle while the task runs, so any
+  // default would cap how long a task may take. Opt in via `opts.timeout`.
   const ctx = await _getTasksContext(opts);
   const result = await ctx.devFetch(`/_nitro/tasks/${taskEvent.name}`, {
     method: "POST",
@@ -23,7 +25,9 @@ export async function runTask(
 
 /** @experimental */
 export async function listTasks(opts?: TaskRunnerOptions) {
-  const ctx = await _getTasksContext(opts);
+  // Listing is metadata the dev server answers immediately, so a default is
+  // safe here and stops a stalled socket from hanging the caller forever.
+  const ctx = await _getTasksContext(opts, 30_000);
   const res = (await ctx.devFetch("/_nitro/tasks")) as {
     tasks: Record<string, { meta: { description: string } }>;
   };
@@ -34,10 +38,10 @@ export async function listTasks(opts?: TaskRunnerOptions) {
 
 const _devHint = `(is dev server running?)`;
 
-async function _getTasksContext(opts?: TaskRunnerOptions) {
+async function _getTasksContext(opts?: TaskRunnerOptions, defaultTimeout?: number) {
   const cwd = resolve(process.cwd(), opts?.cwd || ".");
   const buildDir = resolve(cwd, opts?.buildDir || "node_modules/.nitro");
-  const timeout = opts?.timeout ?? 30_000;
+  const timeout = opts?.timeout ?? defaultTimeout;
 
   const buildInfoPath = resolve(buildDir, "nitro.dev.json");
   if (!existsSync(buildInfoPath)) {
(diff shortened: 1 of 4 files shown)


pi0x:
Rebased this on latest main locally. It applies cleanly, `pnpm typecheck` passes and your test passes. The fix itself looks right.

One concern about the default. Node's `timeout` counts socket idle time, and the dev server holds the connection open until the task finishes. So a flat 30s would kill any task that runs longer than 30s, which migrations and seeds often do. Making it opt-in, or keeping a short default only for `listTasks`, would avoid that. The same point is on #4292, so a maintainer may want to pick.

Also note #4416 adds a `test/unit/task.test.ts` too, so one of you will need to rename.

spokodev:
You were right, and testing it turned up a second problem underneath.

`listTasks()` keeps a 30s default, since the dev server answers it immediately. `runTask()` now has no default and only takes one through `TaskRunnerOptions.timeout`, so a migration or seed is never capped by us.

While verifying that, the naive version of this still killed `runTask` — at 5s, not 30. Passing `timeout: undefined` to `http.request` is not the same as omitting it: the socket still reports the peer's own idle close as a `timeout` event, and the handler then rejected with `Request timed out after undefinedms`. Plain Node, no Nitro:

```
{}                   → 'timeout' event at 5008ms
{"timeout": undefined} → 'timeout' event at 5004ms
{"timeout": 0}       → nothing after 8000ms
```

The 5s is the stub server's own `keepAliveTimeout`. So both the option and the handler are now attached only when a timeout is actually configured, which leaves `runTask` on exactly today's code path when the caller opts out.

Measured against a socket that accepts and never answers:

| call | result |
|---|---|
| `listTasks({ cwd })` | rejects at 30011ms, `Request timed out after 30000ms` |
| `runTask(event, { cwd })` | still pending at 40s |
| either, with `{ timeout: 200 }` | rejects at ~200ms |

Tests: the regression case is kept and a second one covers the opt-in path on `runTask`. Asserting the *absence* of a default would mean a 30s test, so that side is held by construction rather than by a test — say the word if you would rather have it.

Renamed mine to `test/unit/task-runner.test.ts` and left `test/unit/task.test.ts` to #4416, since that one covers the runtime scheduler in `src/runtime/internal/task.ts` while this covers the client in `src/task.ts`.

`pnpm fmt` clean. `pnpm typecheck` reports one error in `src/task.ts`, the pre-existing `Cannot find module 'nitro/types'` on line 9 that shows up on a clean checkout too (590 of them without a build), nothing from this change.

spokodev:
Good point — I had written that side off as untestable because a 30s wait is too slow, but reading the request options is the obvious way round that.

Added `defaults the timeout for listTasks but not for runTask`: it spies on `http.request`, asserts `listTasks({ cwd })` passes `timeout: 30_000`, that `runTask(event, { cwd })` passes no `timeout` key at all, and that an explicit `{ timeout: 200 }` still wins. It runs in about 100ms.

Checked that it actually discriminates rather than just passing: giving `runTask` a default back fails it, and taking the default away from `listTasks` fails it too. 3 passing, `pnpm fmt` clean.

PR #4352, opened 2026-06-15
fix(build,azure-swa): prefer index.mjs serverEntry; fix Azure SWA request URL and body

tonoizer:
## Problem

Three bugs that compound when deploying a Nitro fullstack app with Vite Module Federation `exposes` (and when using the Azure Static Web Apps preset).

### Reproduction demo

**Demo repo:** https://github.com/tonoizer/demo-nitro-vite-module-federation-serverEntry-bug

Minimal repro for the `serverEntry` / `vite preview` issue:

```bash
pnpm dlx create-nitro-app my-app
cd my-app
pnpm add -D @module-federation/vite
# add federation plugin + expose (see demo repo vite.config.ts / module-federation.config.ts)
pnpm build
pnpm check          # fails: serverEntry is server/_chunks/app.mjs
pnpm preview        # broken: 404 on / and /api/*
pnpm preview:workaround   # works: node .output/server/index.mjs
```

`pnpm dev` works — only production preview is broken because it relies on `.output/nitro.json`.

---

**1. Wrong `serverEntry` in `nitro.json`**

`writeBuildInfo` used `.find()` to pick the first `isEntry` chunk from Rollup output. Vite Module Federation marks expose chunks as entry points too, so when they appear before `server/index.mjs` in the output array, `nitro.json` points at e.g. `server/_chunks/app.mjs` instead of the server bundle. `vite preview` then loads a file with no `fetch` handler → 404 everywhere.

**2. Azure SWA: relative URL passed to `new Request()`**

`azure-swa.ts` constructed a relative path (`/api/...` or `pathname + search`) and passed it directly to `new Request(url, ...)`. Node.js requires an absolute URL — this throws in production and breaks all `/api/*` routes.

**3. Azure SWA: response body handling**

`context.res.body` was set incorrectly for Azure Functions v3 (which expects `string | Buffer`). Text responses could be empty/wrong; binary responses were corrupted when unconditionally calling `.text()`.

## Fix

- **`resolveNitroServerEntryFileName`** — when multiple `isEntry` chunks exist, prefer the one matching `/(^|\/)index\.mjs$/` (Nitro's server entry convention) instead of taking array-first.
- **`resolveAzureSwaRequestUrl`** — use `x-ms-original-url` as-is (it's already an absolute URL) or construct `new URL(path, "http://nitro.local").href` for direct `/api/*` calls.
- **`azureResponseBody`** — inspect `content-type` and return UTF-8 string for text types or `Buffer` for binary (same pattern as AWS Lambda preset).
- Unit tests for extracted helpers.

## Test plan

- [x] `pnpm vitest run test/unit/build-info.test.ts test/unit/azure-swa.test.ts`
- [ ] Demo repo: `pnpm build && pnpm check` passes; `pnpm preview` serves SPA + `/api/hello`
- [ ] Azure SWA preset: `/api/*` routes return real JSON/HTML; binary routes return correct bytes

### DIFF
--- a/src/build/info.ts
+++ b/src/build/info.ts
@@ -1,5 +1,5 @@
 import type { Nitro, NitroBuildInfo, WorkerAddress } from "nitro/types";
-import { join, relative, resolve } from "pathe";
+import { extname, join, relative, resolve } from "pathe";
 import { version as nitroVersion } from "nitro/meta";
 import { presetsWithConfig } from "../presets/_types.gen.ts";
 import { writeFile } from "../utils/fs.ts";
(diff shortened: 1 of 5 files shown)


pi0:
Appreciate the PR and sorry for late reply. Closing in favor of #4195

PR #4355, opened 2026-06-16
presets(vercel): export tracing channels messages as otlp spans

RihanArfan:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/pull/4001

### ❓ Type of change

### 📚 Description

Export tracing channel messages as OpenTelemetry spans when deployed on Vercel. This adds a Vercel-preset runtime plugin that turns those events into OpenTelemetry spans and reports them to Vercel's tracing pipeline, so app-level spans show up in Vercel session traces with no OpenTelemetry SDK bundled.

See https://vercel.com/docs/tracing/session-tracing for Vercel usage

### 📝 Checklist

# POC notes:
- Spans are parented flat under the platform root (siblings, not nested) and tagged internal
- One span per IPC message; per-request batching (like @vercel/otel's BatchSpanProcessor + waitUntil) is a TODO for high-span-count routes.
- Only channels created after the plugin runs are captured (no API to enumerate pre-existing ones); current producers create theirs lazily, so they're covered.


src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,141 @@
...
+  // The runtime correlates spans to the request by trace id and drops the rest
+  if (!root) return;
+
+  // Respect the platform sampling decision (bit 0 of traceFlags).
+  if (root.traceFlags !== undefined && (root.traceFlags & 1) === 0) return;
+
+  // The Vercel runtime deserializes spans against a strict OTLP schema, so
+  // every field must be present even when empty (omitting e.g. `events` or
+  // `attributes` makes the runtime drop the whole batch).
+  const span = {
```
pi0: I guess it might be move performant to have a `Span` class with explicit prototype (also more typesafe)

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+    ],
+  });
+}
+
+export default definePlugin(() => {
+  const diagnostics: any = globalThis.process?.getBuiltinModule?.("node:diagnostics_channel");
+  if (!diagnostics?.tracingChannel) return;
+  // Avoid double-wrapping if the plugin is initialised more than once.
+  if (diagnostics.__nitroTracingWrapped) return;
```
pi0: this is a dangrous side-effect on node lib! we can just use a global `let` variable in same plugin

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const createTracingChannel = diagnostics.tracingChannel;
```
pi0: style: we could destructure on L93 without avoiding to put side-effects on node builtin

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,141 @@
...
+      error() {},
+      asyncEnd(data: any) {
+        try {
+          const start = starts.get(data);
+          if (start === undefined) return;
+          starts.delete(data);
+          // This reports one span per IPC message. TODO: for high span-count
+          // requests, batch spans per request and flush once (via `waitUntil`) to
+          // cut IPC overhead as `@vercel/otel` does with its BatchSpanProcessor.
+          reportSpan(name, start, data.error);
```
pi0: I guess we should be able to pass more context in spans, specially http request info?

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,177 @@
...
+  if (root.traceFlags !== undefined && (root.traceFlags & 1) === 0) return;
+
+  const span = new Span(
+    root.traceId,
+    randomSpanId(),
+    root.spanId,
+    channel,
+    SPAN_KIND_INTERNAL,
+    startTimeUnixNano,
+    nowUnixNano(),
```
pi0: nowUnixNano(), randomSpanId() could be simplified to class itself

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,177 @@
...
+      },
+    ],
+  });
+}
+
+// Patch `tracingChannel` once, even if the plugin initialises more than once.
+let patched = false;
+
+export default definePlugin(() => {
+  const diagnostics: any = globalThis.process?.getBuiltinModule?.("node:diagnostics_channel");
```
pi0: Why are we typing it to any? (could be typeof import(...)

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,194 @@
+import { definePlugin } from "nitro";
+import type { SpanContext } from "@opentelemetry/api";
+import type {
+  IExportTraceServiceRequest,
+  ISpan,
+} from "@opentelemetry/otlp-transformer/build/src/trace/internal-types.js";
```
pi0: lets just inline types where needed instead of extra massive external deps only for it

RihanArfan: Vendored them

src/presets/vercel/runtime/telemetry.ts
RihanArfan: Now passing additional context

src/presets/vercel/runtime/telemetry.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  // OTLP's default kind: we have not established a remote relationship.
+  return { name: channel, kind: SPAN_KIND_INTERNAL, attributes: [attr("nitro.channel", channel)] };
+}
+
+function describeH3Request(channel: string, data: Record<string, unknown>): SpanInfo {
+  const attributes = [attr("nitro.channel", channel)];
+  const event = asRecord(data.event);
+  const method = asString(asRecord(event?.req)?.method);
+  const path = asString(asRecord(event?.url)?.pathname);
```
pi0: We we really need overhead of utils like asString? `pathname` is alsways a string from url for example!

I think also `push` is unnecessary (it is faster to make an array with known length vs modify it)

RihanArfan: Made changes

pi0 (approved):

PR #4356, opened 2026-06-17
docs: fix misleading middleware example path in routing docs (#4354)

### LINKED ISSUE #4354: docs: misleading directory path in route-scoped middleware handlers example
Howdy,
In the [[Routing Documentation](https://nitro.build/docs/routing#route-scoped-middleware)](https://nitro.build/docs/routing#route-scoped-middleware) under the **Route-Scoped Middleware** section, the example given for registering a scoped middleware via the `handlers` config array contains a misleading directory structure path.

The snippet demonstrates this setup:

```typescript
// nitro.config.ts

export default defineConfig({
  handlers: [
    {
      route: "/api/**",
      handler: "./server/middleware/api-auth.ts", // <--- Causes a naming conflict/redundancy
      middleware: true,
    },
  ],
});
```

### Why this is (a bit) misleading

The text just below explicitly states that: 
> *"Unlike global middleware (registered in the middleware/ directory which match `/**`), route-scoped middleware only run for requests matching the specified pattern."*

So this puts the reader into a paradox like "Wait, since the scripts inside `server/middleware/` runs for all upcoming request, why are we again running the same middleware check by specifying it explicitly?" 

More precisely:
1. It triggers a double-registration loop where the file executes globally on all requests, and then again specifically for `/api/**` routes.
2. It makes the manual `route: "/api/**"` configuration rule completely redundant since the file's physical directory forces it to run globally regardless of your configuration array.

### Suggested Correction

To prevent this "reader confusion" we could just change the middleware directory script into some folder which is not the global `/middleware/`

```typescript
// nitro.config.ts

export default defineConfig({
  handlers: [
    {
      route: "/api/**",
      handler: "./server/server-utils/api-auth.ts", 
      middleware: true,
    },
  ],
});
```

And additionally we could add a text warning advising developers to keep their manually registered scoped config files *outside* of the native `server/middleware/` folder.

rafaumeu:
Fixes #4354

Changed the route-scoped middleware example to use `./server/utils/api-auth.ts` instead of `./server/middleware/api-auth.ts` to avoid confusion with global middleware behavior.

Also added a warning note clarifying that files in `server/middleware/` run globally regardless of the `route` pattern.

### DIFF
--- a/docs/1.docs/5.routing.md
+++ b/docs/1.docs/5.routing.md
@@ -271,12 +271,15 @@ export default defineConfig({
   handlers: [
     {
       route: "/api/**",
-      handler: "./server/middleware/api-auth.ts",
+      handler: "./server/utils/api-auth.ts",
       middleware: true,
     },
   ],
 });
 ```
+
+> [!WARNING]
+> When using `handlers` for route-scoped middleware, place handler files outside the `server/middleware/` directory. Files in `server/middleware/` run globally for all requests regardless of the `route` pattern.
 
 Each handler entry supports the following options:
 
@@ -408,7 +411,7 @@ export default defineConfig({
   handlers: [
     {
       route: "/api/**",
-      handler: "./server/middleware/api-auth.ts",
+      handler: "./server/utils/api-auth.ts",
       middleware: true,
     },
   ],


pi0:
Thanks for PR but example is correctly using relavant `middleware` dir since `serverDir` that enables scanning, is opt-in.

PR #4360, opened 2026-06-18
fix(hmr): add hmr tests when reload hangs

martinszeltins:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/issues/4359

### 📚 Description

This patch seems to fix the issue, but I am not 100% it is the correct fix.

```diff
diff --git a/src/core/dev-server/server.ts b/src/core/dev-server/server.ts
index 3d46c8ac..7da6df9d 100644
--- a/src/core/dev-server/server.ts
+++ b/src/core/dev-server/server.ts
@@ -85,6 +85,9 @@ class DevServer {
     nitro.hooks.hook("dev:start", () => {
       this.building = true;
       this.buildError = undefined;
+      for (const worker of this.workers) {
+        worker.close();
+      }
     });
 
     nitro.hooks.hook("dev:reload", () => {
@@ -164,12 +167,17 @@ class DevServer {
   async getWorker() {
     let retry = 0;
     const maxRetries = isTest || isCI ? 100 : 10;
+    const getWorkerStart = Date.now();
+    const maxBuildingWaitMs = 120_000;
     while (this.building || ++retry < maxRetries) {
+      if (Date.now() - getWorkerStart > maxBuildingWaitMs) {
+        return;
+      }
       if ((this.workers.length === 0 || this.buildError) && !this.building) {
         return;
       }
       const activeWorker = this.workers.find((w) => w.ready);
-      if (activeWorker) {
+      if (activeWorker && !this.building) {
         return activeWorker;
       }
       await new Promise((resolve) => setTimeout(resolve, 600));
```

### DIFF
--- a/test/dev/dev-reload-race.test.ts
+++ b/test/dev/dev-reload-race.test.ts
@@ -1,0 +1,154 @@
+import { mkdirSync, mkdtempSync, writeFileSync } from "node:fs";
+import { promises as fsp } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "pathe";
+import { isWindows } from "std-env";
+import { describe, expect, it } from "vitest";
+import { type Context, describeIf, setupTest } from "../tests";
+
+const HANG_TIMEOUT_MS = 10_000;
+const REQUEST_ASSIGN_DELAY_MS = 100;
+
+async function waitForDevWorker(
+  context: Context,
+  timeoutMilliseconds = 30_000
+) {
+  const start = Date.now();
+  while (Date.now() - start < timeoutMilliseconds) {
+    try {
+      await context.fetch("/api/hello");
+      return;
+    } catch {
+      // Retry until the dev worker is ready.
+    }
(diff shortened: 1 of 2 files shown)


suzunn:
I think the hook-based case currently exercises a request queued during a rebuild rather than one already assigned to the old worker. `dev:start` sets `building = true` before `context.fetch()` runs, so `getWorker()` cannot return an active worker until `dev:reload` clears that flag. That means the test can pass without covering the reported in-flight race.

I would start the slow request while the worker is ready, wait until the handler has definitely begun (a deferred test hook or observable signal is more deterministic than 100 ms), and only then call `dev:start` followed by `dev:reload`. An assertion that the request remains bound to the original worker would make this a stronger regression test and avoid timing-dependent passes on slower CI runners.

martinszeltins:
@danielroe @pi0 Hey guys, any chance we can get this annoying bug fixed? 🙏🏻

PR #4365, opened 2026-06-21
fix(build): add experimental flag to rewrite leaked runtime require of bundled deps (#4171)

### LINKED ISSUE #4171: Cannot find module 'react' in Docker runner: use-sync-external-store/shim leaves __require('react') unresolved in SSR chunk
## Bug Report

When using Nitro (via TanStack Start) in a **pnpm monorepo** with `@auth0/auth0-react` and a workspace package that uses `use-sync-external-store/shim`, the built SSR bundle contains an unresolved `__require("react")` call. This fails at runtime in any Docker runner stage where `node_modules` is not present.

## Reproduction

https://github.com/psachs-alto/nitro-rolldown-repro

```bash
git clone https://github.com/psachs-alto/nitro-rolldown-repro
cd nitro-rolldown-repro
docker build -t repro .
docker run -p 3000:3000 repro
curl http://localhost:3000/
```

## Error

```
Error: Cannot find module 'react'
Require stack:
- /app/.output/server/_ssr/app-<hash>.mjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1456:15)
    ...
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/app/.output/server/_ssr/app-<hash>.mjs' ]
```

## Root Cause

Rolldown splits the SSR bundle into multiple chunks. When `@auth0/auth0-react` is present, Auth0 + React are bundled together into `_libs/auth0__auth0-react+react.mjs`, which exports an inlined `require_react`. However, `use-sync-external-store/shim` (coming from a **separate workspace package**) ends up in a different SSR chunk (`_ssr/app-*.mjs`) that does **not** have access to the inlined React. Instead it contains:

```js
var require_use_sync_external_store_shim_production = /* @__PURE__ */ __commonJSMin(((exports) => {
    var React = __require("react"); // __require = createRequire(import.meta.url)
    ...
}));
```

`__require` is `createRequire(import.meta.url)` — a real Node.js `require()`. The Nitro runner stage only copies `.output/` (not `node_modules`), so `require("react")` has nothing to resolve and throws.

Note: `nf3` already handles this for `tslib` by copying it into `.output/server/node_modules`. `react` needs the same treatment when it appears as an implicit CJS external in `@__PURE__` factory wrappers.

## Workaround

After `pnpm run build`, copy React into the server output:

```bash
node -e "
  const fs = require('fs'), path = require('path');
  const r = path.dirname(require.resolve('react/package.json'));
  fs.cpSync(r, '.output/server/node_modules/react', { recursive: true, dereference: true });
"
```

## Environment

| | |
|---|---|
| nitro | 3.0.1-alpha.2 |
| react | 19.2.3 |
| @tanstack/react-start | 1.167.9 |
| @auth0/auth0-react | ^2.16.0 |
| use-sync-external-store | ^1.5.0 |
| vite (Rolldown) | ^8.0.3 |
| pnpm | 10.28.0 |
| Node.js | 24 |

pi0x:
[ai assited] resolves #4171 via a workaround until upstream issue is fixed

Try:

- Use `"nitro": "https://pkg.pr.new/nitro@4365"` in `package.json`
- Inside `nitro.config.ts` add `experimental: { cjsRequireRewrite: "react" }`

---

Vite's SSR build lowers a `require("react")` inside vendored CommonJS (e.g. `use-sync-external-store`'s shim, as shipped inside base-ui / recharts / reactflow) to a runtime `__require("react")` (`createRequire`), because React is an SSR-external. Once Nitro bundles React for a self-contained `.output/`, that require is left dangling:

- **`Cannot find module 'react'`** when `node_modules` isn't shipped (Docker runner, serverless), or
- **`Invalid hook call` / null dispatcher** when it *is* — `__require` loads a second React instance.

Root cause is the Vite SSR lowering (pure Rolldown links the same nested require correctly); this is a downstream workaround.

This PR adds an opt-in `experimental.cjsRequireRewrite` flag — a `generateBundle` pass that rewrites leaked `__require("x")` to the copy of `x` already bundled in the output, so a single instance is used and the output stays self-contained.

- `"react"` — targets the React family (`react`, `react-dom`, `react/jsx-runtime`, …).
- `true` — covers any bundled dependency that leaked a runtime require.
- unset (default) — disabled, no behavior change.

Unresolved leaks emit a warning instead of shipping silently.

### DIFF
--- a/src/build/plugins.ts
+++ b/src/build/plugins.ts
@@ -11,6 +11,7 @@ import { virtual, virtualDeps } from "./plugins/virtual.ts";
 import { sourcemapMinify } from "./plugins/sourcemap-min.ts";
 import { raw } from "./plugins/raw.ts";
 import { externals } from "./plugins/externals.ts";
+import { cjsRequire } from "./plugins/cjs-require.ts";
 
 export async function baseBuildPlugins(nitro: Nitro, base: BaseBuildConfig) {
   const plugins: Plugin[] = [];
@@ -67,6 +68,12 @@ export async function baseBuildPlugins(nitro: Nitro, base: BaseBuildConfig) {
             },
       })
     );
+  }
+
+  // Rewrite runtime `require()` of bundled deps to their bundled copy (#4171)
+  const cjsRewrite = nitro.options.experimental.cjsRequireRewrite;
+  if (cjsRewrite) {
+    plugins.push(cjsRequire(nitro, cjsRewrite === true ? "all" : "react"));
   }
 
   // Sourcemap minify
--- a/src/build/plugins/cjs-require.ts
+++ b/src/build/plugins/cjs-require.ts
@@ -1,0 +1,195 @@
+import { dirname, relative } from "pathe";
+import type { Nitro } from "nitro/types";
+import type { OutputChunk, Plugin } from "rollup";
+
+// Some dependencies vendor their own pre-bundled CommonJS (e.g.
+// `use-sync-external-store`'s shim shipped inside base-ui / recharts / reactflow).
+// When such code does `require("react")` and React is an SSR-external, Vite's SSR
+// build lowers it to a runtime `__require("react")` (createRequire). Once a
+// downstream bundler (Nitro, for a self-contained `.output/`) bundles React, that
+// runtime require is left dangling — it either fails (`Cannot find module 'react'`)
+// or loads a *second* React instance with a null dispatcher ("Invalid hook call").
+// See nitrojs/nitro#4171.
+//
+// This plugin rewrites such leaked `__require("x")` calls to the copy of `x` already
+// bundled in the output, so a single instance is used and the output stays
+// self-contained.
+
+// Maps a specifier to the CJS initializer variable Rolldown generates for it.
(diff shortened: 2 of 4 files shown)


pi0x:
Rebased onto latest `main` locally (one conflict: `main` replaced `sourcemapMinify()` with the new `sourcemap()` plugin). Left the draft status alone and did not push.

Two things to consider:

- `generateBundle` mutates `chunk.code` after sourcemaps are generated. Prepending an import shifts every line, so the chunk's sourcemap ends up off by one.
- Detection depends on rolldown's exact codegen (`__require(...)`, `var require_x = ... __commonJS`). With `minify: true`, or under the rollup builder, those identifiers differ, the regexes never match, and the plugin silently does nothing — not even the unresolved warning.

On the relation to #4094: they are complementary, not overlapping. #4094 makes a required package external and traced; this one handles the case where the package is already bundled. Worth noting in the issue that simply tracing `react` would fix `Cannot find module` but not the duplicate-instance "Invalid hook call" — the rewrite fixes both.

PR #4370, opened 2026-06-21
perf: skip duplicate routing sync

TheAlexLichter:
This PR removes a duplicate routing sync introduced by [`0fe0b835`](https://github.com/nitrojs/nitro/commit/0fe0b835), which added `nitro.routing.sync()` after the initial handler scan. Later, [`7857f44e`](https://github.com/nitrojs/nitro/commit/7857f44e) moved that responsibility into `scanHandlers()`, causing routing to be rebuilt twice during initialization. 

Removing the second call keeps behavior while reducing unnecessary startup work.

### DIFF
--- a/src/nitro.ts
+++ b/src/nitro.ts
@@ -78,8 +78,5 @@ export async function createNitro(
   // Ensure initial handlers are populated
   await scanHandlers(nitro);
 
-  // Sync routers
-  nitro.routing.sync();
-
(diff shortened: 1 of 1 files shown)


pi0:
Thnx for PR but this is for assurance (sync is really light)

TheAlexLichter:
@pi0 should we add a comment at least?

PR #4371, opened 2026-06-21
perf(build): only cross-resolve internal aliases

TheAlexLichter:
This PR skips package aliases that cannot participate in Nitro’s internal alias resolution, reducing unnecessary nested comparisons. It preserves `~`, `@`, and `#` alias behavior while leaving package aliases unchanged, with regression coverage. Benchmarks show improvements from roughly 3× on small alias sets to 48× on large sets.


pi0 (approved):
Thnx! Only fixed to keep supporting `@@`

PR #4373, opened 2026-06-23
presets(vercel): use reflinks for custom function dirs

RihanArfan:
### ❓ Type of change

### 📚 Description

When a function needs a modified `.vc-config.json`, we previously copied the entire `__server.func` directory and then overwrote the config. We now copy it with copy-on-write (`COPYFILE_FICLONE`) reflinks, which share the underlying data blocks on supported filesystems (and transparently fall back to a regular copy when unsupported), and skip copying the base `.vc-config.json` since a merged one is written right after.

We avoid symlinking entries within a function dir because Vercel's build container cannot resolve symlinks that point outside the `.func` directory at runtime.

### 📝 Checklist

### DIFF
--- a/src/presets/_utils/fs.ts
+++ b/src/presets/_utils/fs.ts
@@ -1,5 +1,5 @@
 import fsp from "node:fs/promises";
-import { relative, dirname } from "pathe";
+import { relative, dirname, join } from "pathe";
 import consola from "consola";
 import { colors } from "consola/utils";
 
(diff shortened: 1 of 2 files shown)


src/presets/_utils/fs.ts
```diff
@@ -23,3 +23,45 @@ export async function isDirectory(path: string) {
...
+ * see them as regular files while local disk usage stays flat.
+ *
+ * Directories cannot be hard-linked, so the tree is recreated and each
+ * file is linked individually. Symlinks are preserved as symlinks.
+ * Falls back to `copyFile` when crossing filesystems (`EXDEV`) or when
+ * the platform refuses the link operation (`EPERM`).
+ *
+ * Entries listed in `skip` are ignored at the top level only.
+ */
+export async function hardLinkDir(src: string, dest: string, options: { skip?: Set<string> } = {}) {
```
pi0: Have you tried `fs.cp` (https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options) with `COPYFILE_FICLONE` flag? (https://nodejs.org/api/fs.html#fspromisescopyfilesrc-dest-mode) (it should have same CoW effect but hopefully faster since we don't do recursion ourselve.)

RihanArfan: We could use this however it's filesystem dependant. It works on macOS, Windows Dev Drives (although not NTFS), and several other Linux ones like ZFS, BTRFS but not ext4 which I'm guessing CI will use.

RihanArfan: Also while it might be faster since on disk it's not copying again on supported FS fools like `du` can't see that so people still may perceive output as larger than hard linking. 

Building the test fixture and measuring with du -sh:
- main 3.4M
- CoW 3.4M
- hard linking 1.0M

pi0: in generals du in filesystems with cow is an estimation only. `COPYFILE_FICLONE` uses reflinks btw which is also supported usually in ext4.

image

src/presets/vercel/utils.ts
```diff
@@ -636,11 +637,11 @@ async function createFunctionDirWithCustomConfig(
...
 ) {
-  // Copy the entire server directory instead of symlinking individual
-  // entries. Vercel's build container preserves symlinks in the Lambda
-  // zip, but symlinks pointing outside the .func directory break at
-  // runtime because the target path doesn't exist on Lambda.
-  await fsp.cp(serverDir, funcDir, { recursive: true });
+  await fsp.cp(serverDir, funcDir, {
+    recursive: true,
+    mode: constants.COPYFILE_FICLONE,
+    filter: (src) => basename(src) !== ".vc-config.json",
```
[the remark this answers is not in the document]
pi0x: Dismissing: the CoW reflink behavior is intentional, not an oversight. This thread reads the (now-updated) "hard link" wording, but the implementation deliberately moved away from hard-linking per the maintainer suggestion to use `fs.cp` + `COPYFILE_FICLONE`. Reflinks share data blocks on supported filesystems and fall back to a regular copy otherwise — strictly no worse than the previous plain `fsp.cp`. The `.vc-config.json` filter is retained. No change needed.

pi0 (approved):

PR #4386, opened 2026-06-26
fix(cache): respect original h3's decoded path in swr handler

Tofandel:
### 🔗 Linked issue
https://github.com/nuxt/nuxt/issues/35447

### ❓ Type of change

### 📚 Description

We copy the original h3 event's path to the newly created event, as h3 internally in it's handlers decodes the req.url, but using createEvent does not do that, meaning the new event has an encoded url while the old event has a decoded url

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -364,6 +364,7 @@ export function defineCachedEventHandler<
 
       // Call handler
       const event = createEvent(reqProxy, resProxy);
+      event._path = incomingEvent._path;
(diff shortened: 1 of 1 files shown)


pi0 (approved):

Tofandel:
What's the release schedule of nitro like for v2? When can I expect a release, to know if I should wait or if I need to do a patch package?

letoast:
Hi @pi0 any release schedule for v2? Double url encoding is causing quite a headache with some urls :/

PR #4391, opened 2026-06-28
fix(externals): force-trace named traceDeps to fix pnpm nested deps

### LINKED ISSUE #4372: `traceDeps` fails to trace nested dependencies in .pnpm folder
## Versions 
- node: v24.13.0
- nitro: 3.0.260610-beta
- pnpm: 10.30

## Config
```ts
{
  traceDeps: ["sharp", "bcrypt"]
}
```

### Reproduction

1. Use pnpm as package manager
2. Install and import dependency `xyz` that internally depends on a non-bundleable dependency such as `sharp` or `bcrypt`
3. Add `sharp` / `bcrypt` to nitro's `traceDeps` array
4. pnpm build

### Describe the bug

Nitro / nf3 fail to trace nested dependencies of dependencies inside `node_modules/.pnpm`.

## Resolve Aliasing
I first tried to fix this issue, by adding an alias for the failing traces like below. Said alias fixed the result of rolldown's resolve at https://github.com/nitrojs/nitro/blob/6de81eef256aa765e9be3c9cb37beefe23e6b150/src/build/plugins/externals.ts#L91, but nitro still failed to trace the dependency later down the line at https://github.com/nitrojs/nitro/blob/6de81eef256aa765e9be3c9cb37beefe23e6b150/src/build/plugins/externals.ts#L119.

```
{
   rolldownConfig: {
      resolve: {
        alias: {
          "sharp": "path/to/node_modules/.pnpm/sharp-xyz"
        }
      }
    }
}
``` 

## Functioning workaround
By adding the failing dependencies to pnpms `publicHoistPattern`, they will be installed directly in `node_modules` and can be traced my nitro / nf3

```yaml
# pnpm-workspace.yaml:
publicHoistPattern:
  - "sharp"
  - "bcrypt"

```

### Logs

```sh

```

pi0x:
Closes #4372

## Problem

Under pnpm, transitive dependencies are not hoisted to the top-level `node_modules` — a native, non-bundleable dependency (e.g. `sharp` / `bcrypt`) that is only an *indirect* dependency lives exclusively under `node_modules/.pnpm/...`. nft can't statically detect such packages when they're loaded dynamically (native bindings), so they're dropped from the trace and the build fails / ships incomplete output. The known workaround was pnpm's `publicHoistPattern`.

## Fix

`resolveTraceDeps()` now also returns `traceInclude` — the resolved **named** deps (builtins + user `traceDeps`, negations removed, RegExp entries excluded since they can't be resolved by name). The `buildEnd` handler forwards them to nf3's `traceNodeModules` via its new `traceInclude` option.

nf3 resolves each name from `rootDir` **and** from the roots of traced packages that declare it as a dependency — so a pnpm nested dep resolves from its dependent's real `.pnpm` location and gets force-traced (native binaries included).

## Depends on

- unjs/nf3#50 — adds the `traceInclude` option to `traceNodeModules` (resolved dependency-driven, so forwarding the full builtin allowlist stays cheap).

**Draft** until that nf3 change is released and the `nf3` dependency here is bumped to a version that includes it (otherwise CI typecheck fails against the published types).

## Tests

`test/unit/trace-deps.test.ts` — added coverage for the `traceInclude` output (builtins + user, RegExp excluded, negation handling). All unit tests pass.


katywings:
@pi0 It might be related to the pkg.pr preview 🤔, but I am getting the following error when building Nitropage with this preview:

```
error during build:
Build failed with 1 error:

[plugin nitro:externals]
SyntaxError: The requested module '@vercel/nft' does not provide an export named 'nodeFileTrace'
file:///path/to/project/node_modules/.pnpm/nf3@0.3.18/node_modules/nf3/dist/_chunks/trace.mjs:6
import { nodeFileTrace } from "@vercel/nft";
```

pi0:
Very strange @katywings bundled @vercel/nft is not even ESM, it is CJS, it seems something is trying to bundle nitro > nf3 > vercel/nft itself inside output!! Do you have a branch i can test?

pi0x:
Full fix landed (`ca0a6171`) and validated end-to-end against the real nitropage + SolidStart app (sharp/bcrypt under pnpm, **`publicHoistPattern` disabled** — the true failing scenario).

Testing that app surfaced two gaps beyond the initial `traceIncludeRoots` wiring, plus one in nf3:

1. **Bundler couldn't resolve `sharp`** — imported from the generated `.nitro/vite/services/ssr` entry, outside the declaring package's resolution scope, so the build failed before tracing ran. Now an include-matched native dep is externalized **by name** when unresolvable from the importer (it's force-copied into the output by the trace).
2. **The declarer (`nitropage`) is bundled in an upstream vite environment**, so it never appears in the nitro env's module graph → no `traceIncludeRoots`. Now those roots are also seeded from the app's **direct dependencies** (resolvable from `rootDir` even when bundled).
3. **nf3 dropped `sharp`** because it resolved `traceInclude` names under nitro's minimal `["node"]` conditions, and `sharp`'s `exports` have no top-level `default` (only `import`/`require`). Fixed in **nf3 0.3.21** (unjs/nf3#57); this branch now requires it.

Result: build succeeds, **56 deps traced** including `sharp`, `bcrypt`, and `@img/sharp-linux-x64` / `@img/sharp-libvips-linux-x64` with the real `.node` binaries; both `require("sharp")` and `bcrypt.hashSync` work from `.output/server`.

@<tester> a fresh pkg.pr.new off `ca0a6171` should now build your project cleanly with the workaround removed — another run would be much appreciated. 🙏

pi0 (approved):

PR #4394, opened 2026-06-29
fix(build): escape dynamic route segments in Rollup chunk names

enlorik:
## What

`getChunkName` passes the output of `routeToFsPath` directly as a Rollup `chunkFileNames` pattern. `routeToFsPath` encodes dynamic params as `[id]`, `[...]`, etc., but Rollup interprets `[xxx]` as a placeholder token and rejects any name that contains brackets.

Before this fix, routes with dynamic segments (e.g. `/api/users/:id`) would cause a Rollup error at build time:

```
[rolldown:chunk_filenames] Invalid pattern in `output.chunkFileNames`
```

## How

Replace `[` → `%5B` and `]` → `%5D` using `.replace(/\[([^\]]*)\]/g, "%5B$1%5D")` on the `routeToFsPath` output before it is used as a chunk name.

`%` is never emitted by `routeToFsPath` for literal path segments (it maps non-alphanumeric characters to `_`), so percent-encoded brackets are guaranteed to be in a distinct namespace from any literal segment. This avoids the collision where `/api/:id` and `/api/_id_` would previously both emit `_routes/api/_id_.mjs`.

## Tests

- Updated existing dynamic-route test to expect `%5Bid%5D`
- Added nested dynamic route test (`/api/applications/:id/context`)
- Added explicit collision regression test: asserts `/api/:id` → `_routes/api/%5Bid%5D.mjs` and `/api/_id_` → `_routes/api/_id_.mjs` are not equal

### DIFF
--- a/src/build/chunks.ts
+++ b/src/build/chunks.ts
@@ -78,7 +78,7 @@ export function getChunkName(chunk: { name: string; moduleIds: string[] }, nitro
       .flatMap((h) => h.data)
       .find((h) => h.handler === mainId);
     if (routeHandler?.route) {
-      return `_routes/${routeToFsPath(routeHandler.route)}.mjs`;
+      return `_routes/${routeToFsPath(routeHandler.route).replace(/\[([^\]]*)\]/g, "_$1_")}.mjs`;
     }
 
     const taskHandler = Object.entries(nitro.options.tasks).find(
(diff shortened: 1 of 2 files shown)


pi0:
Can you please provide a reproduction (failing nitro v3 repo)?

PR #4401, opened 2026-07-02
feat(cache): warn when recompute is cheaper than cache retrieval

### LINKED ISSUE #2559: Warn when cache computation is cheaper than cache retrival
(From question by @InstantlyMoist in https://github.com/unjs/nitro/discussions/2558)

In development mode, we can detect situations when cache computation is cheaper than storing and retrieving it to warn users about it. It is not a precise prediction though because in production (when cache might be stored externally to disk or a remote server and database/server load is heavier than single user local/staging, the cache would make sense so maybe if we do it, we should do it for dev and prod)

raminjafary:
### 🔗 Linked issue

Closes #2559

### ❓ Type of change

### 📚 Description

`cachedEventHandler` / `cachedFunction` can be slower than the code they cache when the value is cheap to compute — the store/retrieve round-trip (and JSON (de)serialization) costs more than just recomputing it (see #2558).

This adds a **dev-only** warning to surface that. At the storage layer we measure, per cache:
- **retrieval time** — the duration of a hit `getItem`, and
- **recompute time** — the interval from a miss to its following `setItem`.

When median recompute is cheaper than median retrieval, we log a single warning suggesting the cache may be adding overhead.

Notes:
- Gated behind `import.meta.dev`, so the whole path is tree-shaken out of production builds.
- The user function is **not wrapped**, so ocache's integrity hash (used to invalidate on code edits) is unchanged and production behavior is identical.
- Warns at most once per cache; per-key tracking is bounded (256 keys) to cap dev memory.

Tested with unit tests covering the warn / no-warn / threshold / warn-once cases, plus a regression guard that the function reaches ocache unwrapped. Full suite passes on both the rollup and rolldown builders.

### 📝 Checklist

### DIFF
--- a/src/runtime/internal/cache.ts
+++ b/src/runtime/internal/cache.ts
@@ -40,6 +40,8 @@ export function defineCachedFunction<T, ArgsT extends unknown[] = any[]>(
   return _defineCachedFunction(fn, {
     group: "nitro/functions",
     onError: defaultOnError,
+    // Warn in dev when caching a value costs more than recomputing it (#2559).
+    warnWhenSlower: import.meta.dev,
     ...opts,
   });
 }
@@ -52,6 +54,7 @@ export function defineCachedHandler(
   const ocacheHandler = _defineCachedHandler(handler as any, {
     group: "nitro/handlers",
     onError: defaultOnError,
+    warnWhenSlower: import.meta.dev,
     toResponse: (value, event) => toResponse(value, event as H3Event),
     createResponse: (body, init) => new FastResponse(body, init),
     handleCacheHeaders: (event, conditions) => handleCacheHeaders(event as H3Event, conditions),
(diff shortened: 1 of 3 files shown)


pi0:
Do you think we can port this to https://github.com/unjs/ocache ?

raminjafary:
Yeah, I will port it to `ocache`. By doing so, I think the measurement would be more accurate as well.

raminjafary:
Moved the detection logic into ocache as an opt-in `warnWhenSlower` option, so ocache measures recompute vs. retrieval directly instead of Nitro instrumenting the storage adapter. 

Nitro just enables it globally in dev via import.meta.dev (stripped in prod, overridable per cache).

**Blocked on the ocache side:** ocache PR must be merged and its version bumped in nitro.

pi0:
lets track in https://github.com/unjs/ocache/pull/28

PR #4404, opened 2026-07-06
Fix Deno dev-server crash: reconstruct Response in host realm

### LINKED ISSUE #4390: Vite dev server: responses crossing the dev worker-thread bridge fail Deno's native Response realm check
### Environment
- nitro: 3.0.260610-beta
- vite: 8.1.0
- hono: 4.12.27
- deno: 2.9.0
- OS: Linux (EndeavourOS, kernel 7.0.10-arch1-1), x86_64
- Package manager: npm
- vite.config.ts preset: "deno-server"

### Describe the bug
When the Vite dev process itself is launched under Deno (`deno task dev` → runs `vite`), any response that has to cross back from Nitro's isolated dev worker thread to the host process throws on the Deno side:

TypeError: Return value from serve handler must be a Response constructed via the Response constructor in this realm

at finishOrReturnNative (ext:deno_http/00_serve.ts:1:17128)

at ext:deno_http/00_serve.ts:1:19630

This is **not** caused by application code returning a non-Response value. I confirmed this by reproducing the identical error on two completely unrelated code paths in the same project:

1. A Hono app mounted as a custom `serverEntry` (`format: "web"`), where every branch explicitly returns `c.json(...)` / a native `Response`.
2. After removing the custom `serverEntry`, the **default Nitro/Vite HTML renderer** (no app code involved at all) — same exact error.

The error only appears when the dev orchestrator process is run under Deno. The same code is unaffected in two other configurations:
- `npm run dev` (Vite/Nitro dev server launched under Node) — works fine.
- Production build run natively under Deno (`vite build` then `deno run -A .output/server/index.mjs`, same `deno-server` preset) — works fine.

This narrows it down to Nitro's Vite-plugin dev-time worker thread / `dispatchFetch`→`NodeEnvRunner` bridge specifically misbehaving when the *host* process is Deno — most likely in how a Response (or the data representing it) is reconstructed in the main thread after coming back from the worker.

### Reproduction steps (minimal)
```bash
npm create vite@latest nitro-deno-repro -- --template vanilla-ts
cd nitro-deno-repro
npm install
npm install nitro hono
```

`vite.config.ts`:
```ts
import { nitro } from "nitro/vite"
import { defineConfig } from "vite"

export default defineConfig({
  plugins: [
    nitro({
      preset: "deno-server",
      serverEntry: {
        handler: "./src/server.ts",
        format: "web"
      }
    })
  ]
})
```

`src/server.ts`:
```ts
import { Hono } from "hono"

const app = new Hono()
app.get("/", (c) => c.json({ ok: true }))

export default { fetch: app.fetch }
```

`deno.json`:
```jsonc
{
  "tasks": {
    "dev": "vite",
    "build": "vite build",
    "start": "deno run -A .output/server/index.mjs"
  },
  "nodeModulesDir": "auto"
}
```

Then:
```bash
deno task dev
# request the dev URL (curl or browser) and watch the terminal
```

> Note: I simplified this down from a larger reproduction in a real app — please sanity-check it reproduces as-is before relying on it; if it doesn't, I'm happy to share a trimmed clone of the actual repo instead.

### Expected behavior
Dev server behaves the same regardless of whether `vite`/Nitro's dev process is launched via Node or Deno, since the production build of the same preset works correctly under Deno.

### Actual behavior
`Deno.serve()`'s realm check rejects the Response coming back through Nitro's dev worker-thread bridge, even for code paths (the default renderer) that never touch user code.

tsushanth:
Fixes #4390

## Problem

Nitro's Vite dev server crashes under Deno with:
```
TypeError: Return value from serve handler must be a Response constructed via the Response constructor in this realm
```

The `catchAllHandler` in `src/dev/server.ts` calls `this.#manager.fetch(req)`, which goes through `httpxy.proxyFetch`. That builds its `Response` using undici's `Response` class. `Deno.serve` performs a native V8 realm check and rejects any `Response` not constructed by Deno's own `Response` constructor. Node's HTTP stack doesn't do this check, so the bug only surfaces in Deno dev.

## Fix

Wrap the proxied response in `new Response(...)` before returning it, so whichever runtime's global constructor is invoked:

```diff
- return response;
+ return new Response(response.body, {
+   status: response.status,
+   statusText: response.statusText,
+   headers: new Headers(response.headers),
+ });
```

`response.body` is a `ReadableStream` and is safely transferable across realm boundaries. This fix is a no-op on Node and Bun (their constructors accept the same arguments) and resolves the realm mismatch on Deno.

1 file, 4 lines.

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -50,7 +50,14 @@ export class NitroDevServer extends NitroDevApp implements RunnerRPCHooks {
       if (response.status === 503 && !this.#manager.ready) {
         return this.#generateError();
       }
-      return response;
+      // Reconstruct in the current realm so runtime-native instanceof checks pass.
+      // The worker bridge returns an undici Response; runtimes like Deno reject it
+      // because it was not constructed by their own Response constructor.
+      return new Response(response.body, {
+        status: response.status,
+        statusText: response.statusText,
+        headers: new Headers(response.headers),
+      });
     });
 
     // Bind all methods to `this`


pi0:
have you tested against last nightly if having still issues? (see https://github.com/nitrojs/nitro/pull/4397)

tsushanth:
Thanks for the pointer! #4397 fixes export conditions so Deno resolves runtime-native modules in dev — that's a different layer from this fix. #4397 would cause `httpxy` to use Deno's own `fetch`, which would produce a Deno-realm `Response`, making this PR potentially redundant for the specific case where `httpxy` has a Deno export condition.

However, this change is still defensive: if any code path through the worker bridge returns a non-native `Response` (e.g. when Deno export conditions aren't available in a dep), the realm reconstruction here ensures it won't crash.

If you can confirm #4397 fully resolves #4390 in the nightly, happy to close this in favor of that.

PR #4410, opened 2026-07-08
fix(runtime): harden routeRules proxy path normalization to keep wildcard scope

dnukumamras:
## What

Harden `routeRules` proxy and redirect path handling for `/**` wildcard rules so the forwarded upstream path stays inside the configured base, regardless of how the incoming path is shaped.

Previously the scope check ran only on the incoming request path. Equivalent inputs (repeated/leading slashes, `/./`, empty segments, mixed-case or double-encoded separators) could normalize differently before vs. after the remainder was appended, letting a request escape the intended base once the upstream decoded it.

The fix validates the **final resolved target** (the bytes actually forwarded), not just the incoming path. Proxy and redirect now share one `resolveWildcardTarget` helper.

## Flow

Rule: `"/api/orders/**": { proxy: { to: "http://upstream/orders/**" } }`

Happy path

- `GET /api/orders/list.json` stays in scope, forwarded to `http://upstream/orders/list.json`.
- `GET /api/orders/foo%2f..%2fbar` resolves within `/orders`, forwarded unchanged.
- Query strings are preserved.

Unhappy path (rejected with 400, upstream never contacted)

- `GET /api/orders/..%2fadmin%2fconfig.json`
- `GET /api/orders//..%2fadmin%2fconfig.json` (doubled slash)
- `GET /api/orders/..%2Fadmin` (mixed-case)
- `GET /api/orders//..%252fadmin` (double-encoded)

## Tests

- Unit tests for `resolveWildcardTarget`: benign passthrough, every equivalent traversal shape blocked, and a regression assert that a `/**` target can never resolve outside the base.
- Runtime test asserting encoded traversal is rejected before the upstream is reached.

### DIFF
--- a/src/runtime/internal/route-rules.ts
+++ b/src/runtime/internal/route-rules.ts
@@ -5,9 +5,9 @@ import {
   requireBasicAuth,
   resolveDotSegments,
 } from "h3";
-import type { BasicAuthOptions, EventHandler, Middleware } from "h3";
+import type { BasicAuthOptions, EventHandler, H3Event, Middleware } from "h3";
 import type { MatchedRouteRule, NitroRouteRules } from "nitro/types";
-import { joinURL, withQuery, withoutBase } from "ufo";
+import { joinURL, parseURL, withQuery, withoutBase } from "ufo";
 import { defineCachedHandler } from "./cache.ts";
 
 // Note: Remember to update RuntimeRouteRules in src/build/virtual/routing.ts when adding new route rules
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for the PR/report. It should have been fixed in latest [nightly](https://nitro.build/docs/nightly) channel (`"nitro": "npm:nitro-nightly@3.0.1-20260712-095235-05bf1c38"`) via migration to [h3-rules](https://github.com/h3js/h3-rules).

`/proxy/..%2fsecret`, `/proxy//..%2fsecret`, `/proxy/..%252fsecret`, `/proxy//..%252fsecret`, `/proxy/..%2f..%2fsecret`, `/proxy/%2e%2e%2fsecret` returning 400 now.

PR #4412, opened 2026-07-09
fix(vercel): strip trailing slash from prerendered route overrides

### LINKED ISSUE #4392: Vercel preset emits trailing-slash override paths
## Environment

- Nitro: `2.13.4` (via Nuxt `4.4.8`)
- Node: `24.18.0`
- Preset: `vercel`

## Reproduction

- **Repo:** https://github.com/JonathanXDR/repro-nuxt-vercel-trailing-slash-override
- **Live demo:** [`/slash/`](https://repro-nuxt-vercel-trailing-slash-override-4ufpl8wq9.vercel.app/slash/) (broken) vs [`/noslash`](https://repro-nuxt-vercel-trailing-slash-override-4ufpl8wq9.vercel.app/noslash) (works)

`nuxt.config.ts` (no Nuxt modules needed):

```ts
  nitro: { prerender: { crawlLinks: false, routes: ['/', '/slash/', '/noslash'] } },
})
```

```bash
NITRO_PRESET=vercel npm run build
node -e "console.log(require('./.vercel/output/config.json').overrides)"
npx vercel deploy --prebuilt   # then GET /slash/ and /noslash a few times
```

> StackBlitz/WebContainers can't run the Vercel build/runtime, so use a real `vercel build` + deploy to observe it.

## Describe the bug

A prerendered route with a **trailing slash** gets a Build Output `overrides` entry whose `path` keeps the slash, and Vercel does not serve the static file from such a path. The request falls through to the SSR function.

Generated `overrides`:

```jsonc
{
  "index.html":         { "path": "" },        // static ✓
  "noslash/index.html": { "path": "noslash" }, // static ✓
  "slash/index.html":   { "path": "slash/" }   // function ❌, trailing slash
}
```

Each page renders a server-side `useState(new Date().toISOString())` (frozen if static, fresh per request if the function runs):

| Route | override `path` | `x-vercel-cache` | timestamp over 3 requests | served by |
|---|---|---|---|---|
| `/slash/`  | `slash/`  | `MISS` | **changes each request** | **function** ❌ |
| `/noslash` | `noslash` | `HIT`  | frozen | static ✓ |

Changing **only** `slash`'s override from `slash` to `slash/` on the same build flips it from static to function, so the trailing slash is the sole cause.

### Root cause

`src/presets/vercel/utils.ts`, `generateBuildConfig()`:

```ts
{ path: route.replace(/^\//, '') }   // strips the leading slash, KEEPS the trailing slash
```

Prerender routes are stored verbatim, so a `/slash/` route becomes `{ path: "slash/" }`. Per the [Build Output API v3 spec](https://vercel.com/docs/build-output-api/v3/configuration) an override `path` is always slash-free (`blog.html → { path: "blog" }`). Trailing-slash behavior is expressed via 308 redirect `routes`, not the override `path`. So `{ path: "slash/" }` is off-spec.

### Suggested fix

Make the override `path` slash-free, either:

1. **Omit the override** for a `<dir>/index.html` (Vercel's directory-index already serves it at both `/dir` and `/dir/`), or
2. emit a clean `path` **plus** 308 trailing-slash redirect `routes` (e.g. `@vercel/routing-utils` `getTransformedRoutes({ trailingSlash })`).

Note: rewriting to `{ path: "slash" }` *alone* would leave the canonical `/slash/` unserved, since the preset emits no redirect routes. Likely also fixes #4242. Happy to open a PR.

## Additional context

Real-world trigger: a Nuxt site using `@nuxtjs/i18n` + `nuxt-site-config` with `trailingSlash: true` produces `{ "de/index.html": { "path": "de/" } }` and serves `/de/` from the function.

Related:

- #4242 (open, root `/` variant)
- #1651 / PR #500 (original "serve prerendered routes statically" fix)

## Logs

```sh
# /slash/  (override "slash/")  -> timestamp changes = function
req1 x-vercel-cache=MISS renderedAt=2026-06-28T14:20:44.124Z
req2 x-vercel-cache=MISS renderedAt=2026-06-28T14:20:44.563Z
req3 x-vercel-cache=MISS renderedAt=2026-06-28T14:20:45.019Z

# /noslash (override "noslash") -> timestamp frozen = static
req1-3 x-vercel-cache=HIT renderedAt=2026-06-28T14:20:06.367Z
```

thribhuvan003:
### 🔗 Linked issue

resolves #4392

### ❓ Type of change

### 📚 Description

prerendering a route with a trailing slash (e.g. `/slash/`) emits a Build Output override whose `path` keeps the slash — `"slash/index.html": { "path": "slash/" }` — and vercel doesn't match override paths like that, so the route falls through to the server function instead of being served statically (repro + live deploy in #4392).

this strips the trailing slash the same way the leading one is already stripped, so the override becomes `"path": "slash"`. the root route is unaffected (`/` still maps to `""`). extended the vercel:web preset test with a prerendered `/slash/` route.

the same line exists on the v2 branch (`src/presets/vercel/utils.ts:166`) — happy to open a backport if useful.

### 📝 Checklist


pi0x:
Reviewed this and verified it end-to-end on real Vercel deployments rather than just the snapshots. **The approach is correct and it does fix #4392.** I pushed one follow-up commit for a regression it introduced (see below).

### Verification

Playground with prerendered `/`, `/noslash`, `/slash/`, `/deep/nested/` and a JSON `/json/`. Each response stamps a render timestamp, so a frozen build-time value means the static file was served and a fresh one means the SSR function ran. Five prebuilt deployments, one per config variant:

| Path | `main` | this PR | keeping overrides, only stripping the trailing slash |
|---|---|---|---|
| `/` | static | static | static |
| `/noslash` | static | static | static |
| `/slash` | **function** ❌ | **static** ✅ | static ✅ |
| `/slash/` | **function** ❌ | **static** ✅ | static ✅ |
| `/deep/nested/` | **function** ❌ | **static** ✅ | static ✅ |
| `/json/` | **function** ❌ | **static** ✅ | static ✅ |
| `/slash/index.html` | **404** | 200 | 404 |

Two things worth recording, since neither is documented anywhere:

- Vercel really does resolve directory indexes for Build Output static files — `<dir>/index.html` **and** extensionless `<dir>/index` both serve at `/dir` and `/dir/`. So dropping the overrides is safe, and the optional `.html` in `INDEX_FILE_RE` is right. I initially read ["neither their contents, nor their file name or extension will be modified in any way"](https://vercel.com/docs/build-output-api/primitives) as ruling this out; it doesn't — that sentence is about upload paths, not the CDN lookup.
- `main` is worse than the issue reported: `/slash/index.html` returns **404** there. The trailing-slash override doesn't merely fail to match, it moves the file to an unmatchable path, so the prerendered content is unreachable at *any* URL.

### Follow-up commit: content type for extensionless files

One real regression, now fixed in 2f302a4:

A non-HTML route ending in `/` prerenders to an extensionless file (`/json/` → `json/index`, via `routeWithIndex` in `src/prerender/prerender.ts`). With the override gone, Vercel infers the content type from the file name and has nothing to go on, so it served `application/octet-stream` — a download instead of JSON. On `main` this route fell through to the function and got the right type, so the PR as it stood traded "SSR but correct" for "static but broken".

Fix is a `contentType`-only override. Without a `path` the file stays put, so the directory index keeps working:

```jsonc
"overrides": { "json/index": { "contentType": "application/json;charset=UTF-8" } }
```

Verified on a deployment: `/json` and `/json/` now both serve statically as `application/json`. The condition is deliberately narrow — only when `mime.getType(fileName)` is null, i.e. the file name cannot convey a type at all. Routes whose extension merely *disagrees* with the response header (`prerender-custom.html` served as `text/plain`) are left alone; that mismatch predates this PR and is a separate question.

The vercel snapshot picks up six of these, all extensionless prerendered files that were being served as downloads.

### Minor, not blocking

- `/noslash/index.html` now returns 200 alongside `/noslash`. On `main` and with the narrower fix it 404s, because the override moves the file rather than aliasing it. Two URLs serving identical content is an SEO/canonicalization wrinkle — worth a line in the PR description, if nothing else.
- `route` carries `baseURL` while `fileName` is `withoutBase(...)`, so with a non-root base neither skip branch can ever fire and the override is emitted regardless. Pre-existing asymmetry, but the new equality checks make it load-bearing.
- Nit: `test/unit/vercel-overrides.test.ts` vs the existing `azure.utils.test.ts` / `cloudflare.utils.test.ts` naming.

### Separate bug this turned up (not caused by this PR)

My first two deployments had *every* route SSR'ing regardless of overrides. Cause: with `compatibilityDate >= 2025-07-15`, `getObservabilityRoutes` emits a `.func` symlink per declared handler, and those functions shadow same-path static files during the filesystem phase. So an app with explicit `server/routes/*` handlers gets no benefit from prerendering on Vercel at all — on `main` and on this branch alike. Dropping the compat date below that threshold was the only change that made static serving work.

Nuxt-style apps (single catch-all renderer) are unaffected, which is why #4392 was reproducible in the first place. I'll open a separate issue.

Thanks for digging into this one — the analysis in the description was on the money.

pi0 (approved):

PR #4418, opened 2026-07-10
fix(aws-lambda): terminate streaming responses that have no body

### LINKED ISSUE #4417: aws-lambda streaming preset: bodyless responses (redirects, 204) never terminate the stream → ~20s hang behind CloudFront
# aws-lambda streaming preset: bodyless responses (redirects, 204) never terminate the stream → ~20s hang behind CloudFront

### Environment

- Nitro: `3.0.260429-beta` (pinned via `@tanstack/react-start`). **Confirmed identical handler in the latest published beta `3.0.260610-beta` and current `nitro-nightly@3.0.1-20260707-232134-a13bd4ee`.**
- Preset: `aws-lambda` with `awsLambda: { streaming: true }`
- Deploy: SST → Lambda **Function URL** (`InvokeMode: RESPONSE_STREAM`), fronted by **CloudFront**. Node 24 runtime. Function timeout: 20s.
- TanStack Start (server `redirect()` in a route `beforeLoad`).

### Describe the bug

Any SSR response with an **empty body** — a TanStack `throw redirect()` (307), or equivalently `new Response(null, { status: 204 })` — **never terminates its response stream** in the `aws-lambda` streaming handler. The status + headers are flushed immediately, but the stream is never closed.

- **Via CloudFront:** CloudFront waits for the origin stream to end, which only happens when the **Lambda 20s function timeout** reaps the invocation. The client receives the 307 after **~20s**.
- **Direct to the Function URL:** the response headers arrive in ~0.2s but the stream still never closes (a `curl -o /dev/null` runs to its timeout). **Browsers appear fine here only because they act on the redirect `Location` header without waiting for the body to close** — which is why this looks "fixed" when eyeballed in a browser but is not.
- **Rendered routes (non-empty body) are unaffected** — the stream ends naturally with content.

This is the same problem as **#3829** (TanStack Start redirect timeout). That issue was closed as completed after the reporter said `nitro-nightly@latest` fixed it — but (a) a second reporter (@j-dominguezp) noted it still reproduces via CloudFront while the direct Function URL looks immediate (matching the browser-masking above), and (b) the nightly handler is **byte-identical** to the broken one, so it cannot have been fixed. Please consider reopening / linking.

### Root cause

`presets/aws-lambda/runtime/aws-lambda-streaming.mjs` (identical across published betas and current nightly):

```js
const httpResponseMetadata = { statusCode: response.status, ...awsResponseHeaders(response) };
if (!httpResponseMetadata.headers["transfer-encoding"]) {
  httpResponseMetadata.headers["transfer-encoding"] = "chunked";        // forced even for bodyless responses
}
const body = response.body ?? new ReadableStream({ start(controller) {
  controller.enqueue("");                                              // zero-length chunk
  controller.close();
}});
const writer = awslambda.HttpResponseStream.from(responseStream, httpResponseMetadata);
const reader = body.getReader();
await streamToNodeStream(reader, responseStream);                       // writes "" then end()
writer.end();
```

For a bodyless response the handler forces `transfer-encoding: chunked` **and** streams a single zero-length chunk (`enqueue("")`). In HTTP chunked encoding a zero-length chunk *is* the terminator, so `writer.write("")` followed by `writer.end()` appears to emit a malformed/ambiguous terminator that CloudFront never accepts as a complete response — it waits until the Lambda times out. (Hypothesis on the exact wire behavior; the empirical facts below are solid.)

Note the stable `3.0.0` handler guards with `if (response.body) { … }` and writes nothing for empty bodies — but that path never calls `HttpResponseStream.from`, so it would drop the status code and `Location` header for a redirect. So neither shape is correct; a fix needs to **write the prelude (status + headers) and cleanly terminate** for bodyless responses.

### Evidence (`curl -w`, admin app, unauthenticated `/` → 307 to /login)

| Target | Path | Result |
|---|---|---|
| Direct Function URL | `/` (redirect, 0-byte 307) | `TTFB=0.13s` … stream never closes (`total` hits curl `--max-time`) |
| Direct Function URL | `/login` (renders, 17 KB) | `TTFB=0.25s total=0.35s` ✅ closes |
| Via CloudFront | `/` (redirect) | `TTFB=20.3s` (== Lambda timeout) |

CloudWatch for the redirect invocation: `Init 234ms`, `Duration 1350ms` — the handler *completes* in ~1.35s; only the response stream fails to terminate. So it is not the framework, DB, cold start, or event-loop drain.

### Reproduction

Same as #3829's repro (https://github.com/jtaavola/tanstack-sst-redirect): a TanStack Start route whose `beforeLoad` does `throw redirect(...)`, `awsLambda: { streaming: true }`, deployed to a Lambda Function URL **behind CloudFront**. A plain `new Response(null, { status: 204 })` server route reproduces it too. **Please verify any fix via CloudFront (or `curl` against the Function URL), not a browser hitting the Function URL directly** — the browser masks it for redirects.

### Suggested fix

For responses with no body (or bodyless status codes 204/304/3xx): do **not** force `transfer-encoding: chunked`, and do **not** enqueue an empty chunk — write the prelude via `HttpResponseStream.from` (preserving status + headers) and end the stream immediately (or set `content-length: 0`). Happy to open a PR if that direction sounds right.

### Workaround

`awsLambda: { streaming: false }` (buffered) returns bodyless responses correctly. Trade-off: loses response streaming (progressive SSR; 6 MB response cap).

thribhuvan003:
bodyless responses (redirects, 204, `Response(null)`) on the aws-lambda streaming preset were hanging the stream behind CloudFront until the Lambda timeout. the handler forced `transfer-encoding: chunked` and wrote a zero-length chunk for empty bodies, which never cleanly closed the stream.

when `response.body` is null we now write the status/headers prelude and end immediately (`content-length: 0`), and only use the chunked streaming path when there is a real body. unit tests cover the emit plan (3/3).

Fixes #4417.

### DIFF
--- a/src/presets/aws-lambda/runtime/aws-lambda-streaming.ts
+++ b/src/presets/aws-lambda/runtime/aws-lambda-streaming.ts
@@ -1,6 +1,7 @@
 import "#nitro/virtual/polyfills";
 import { useNitroApp } from "nitro/app";
 import { awsRequest, awsResponseHeaders } from "./_utils.ts";
+import { planAwsLambdaStreamingEmit } from "./streaming-body.ts";
 
 import type { StreamingResponse } from "@netlify/functions";
 import type { Readable } from "node:stream";
@@ -19,22 +20,36 @@ export const handler = awslambda.streamifyResponse(
       ...awsResponseHeaders(response),
     };
 
-    if (!httpResponseMetadata.headers!["transfer-encoding"]) {
-      httpResponseMetadata.headers!["transfer-encoding"] = "chunked";
+    const plan = planAwsLambdaStreamingEmit(response);
+
+    // Bodyless responses (redirects, 204/304, empty handlers): write the
+    // status+headers prelude and end immediately. Forcing chunked TE and
+    // enqueueing a zero-length chunk hangs the stream behind CloudFront
+    // (~Lambda timeout) even though the handler itself finishes quickly.
+    if (plan.kind === "bodyless") {
+      if (
+        !httpResponseMetadata.headers!["content-length"] &&
+        !httpResponseMetadata.headers!["Content-Length"]
+      ) {
+        httpResponseMetadata.headers!["content-length"] = plan.contentLength;
+      }
+      const writer = awslambda.HttpResponseStream.from(
+        responseStream,
+        httpResponseMetadata
+      );
+      writer.end();
+      return;
     }
 
-    const body =
-      response.body ??
-      new ReadableStream<string>({
-        start(controller) {
-          controller.enqueue("");
-          controller.close();
-        },
-      });
+    if (plan.transferEncoding) {
+      if (!httpResponseMetadata.headers!["transfer-encoding"]) {
+        httpResponseMetadata.headers!["transfer-encoding"] = plan.transferEncoding;
+      }
+    }
 
     const writer = awslambda.HttpResponseStream.from(responseStream, httpResponseMetadata);
 
-    const reader = body.getReader();
+    const reader = response.body!.getReader();
     await streamToNodeStream(reader, responseStream);
     writer.end();
   }
--- a/src/presets/aws-lambda/runtime/streaming-body.ts
+++ b/src/presets/aws-lambda/runtime/streaming-body.ts
@@ -1,0 +1,27 @@
+/**
+ * Decide how the aws-lambda streaming handler should emit a fetch Response.
+ *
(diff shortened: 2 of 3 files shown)


pi0:
Thanks for PR ❤️ 

Nitro is now using srvx aws adapter and fix is added for it https://github.com/h3js/srvx/commit/9abe7201ff07983a1454ebcb4620311159bd5820, https://github.com/h3js/srvx/commit/175d1cf2f943d0d1be98a99a63f687e36dffe6d7

PR #4422, opened 2026-07-11
fix(vercel): use `SameSite=Lax` for skew protection cookie

### LINKED ISSUE #4414: Vercel skew protection: SameSite=Strict on the __vdpl cookie causes asset 404s after cross-site navigations
### Environment

nitropack 2.13.4 (via Nuxt 4), `vercel` preset, Node 24. Vercel project with Skew Protection enabled and prerendered routes.

### Reproduction

Needs a live Vercel deployment (the cookie routing happens on Vercel's edge), so a StackBlitz cannot show it. Observable with curl against any Nitro app deployed twice with skew protection on. `dpl_OLD` is the previous deployment's ID, `NEWCHUNK.js` is a hashed asset that only exists in the current build:

```sh
curl -s -o /dev/null -w "%{http_code}" -b "__vdpl=dpl_OLD" https://example.app/_nuxt/NEWCHUNK.js   # 404
curl -s -o /dev/null -w "%{http_code}" https://example.app/_nuxt/NEWCHUNK.js                       # 200
```

### Describe the bug

The preset emits the skew protection cookie with `SameSite=Strict` (`src/presets/vercel/utils.ts`):

```
set-cookie: __vdpl=dpl_...; Path=/; SameSite=Strict; Secure; HttpOnly
```

Strict cookies are omitted on cross-site top-level navigations but sent on the page's same-site subresource requests. A user who last visited before a deploy and returns via any cross-site link (search, chat, even the Vercel dashboard) therefore gets the newest HTML, while their `/_nuxt/*` requests carry the stale pin and are routed to the old deployment, which lacks the new chunks. Every script 404s and the page fails to boot. Since the cookie is an HttpOnly session cookie, a hard reload does not recover; only clearing cookies does.

Suggested fix: `SameSite=Lax`, which is also sent on cross-site top-level GET navigations, so the document and its assets always pin to the same deployment. The cookie only selects a deployment version, so Strict adds no security value here. Happy to send a PR.

### Additional context

Prerendered documents served as Vercel edge cache HITs never refresh the pin (the Set-Cookie only appears on function responses), so a stale pin can persist for a whole browser session. With Lax this becomes harmless (consistent old-version pages). Feature introduced in #3787 / #3850.

thribhuvan003:
### 🔗 Linked issue

fixes #4414

### ❓ Type of change

### 📚 Description

skew pin cookie was SameSite=Strict, so after coming in from another site the html refreshed but /_nuxt assets still had the old __vdpl and 404'd. switched to Lax so the pin updates with the document. still Secure + HttpOnly.

unit test on the generated set-cookie — fails with Strict, passes with Lax.

### 📝 Checklist

- [x] I have updated the tests accordingly.


pi0 (approved):
LGTM, however this is diverging from sveltekit and vercel docs. keeping open little bit more. /cc @danielroe

danielroe (approved):
❤️

i would remove the llm generated test as i'm not sure it's that helpful, btw

danielroe:
i think Lax is the more correct option

cc: @elliott-with-the-longest-name-on-github

src/presets/vercel/utils.ts
```diff
@@ -211,7 +211,7 @@ export async function generateStaticFiles(nitro: Nitro) {
   await writeFile(buildConfigPath, JSON.stringify(buildConfig, null, 2));
 }
 
-function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
+export function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
```
pi0:
```suggestion
function generateBuildConfig(nitro: Nitro, o11Routes?: ObservabilityRoute[]) {
```

PR #4423, opened 2026-07-11
fix(bun): restore idle timeout environment variable

### LINKED ISSUE #3454: set custom `Bun.serve` `idleTimeout`
### Describe the feature

add `server.bun.idleTimeout` to nitro config to set https://bun.sh/docs/api/http#idletimeout
Image

### Additional information

b-mounir-dev:
Restores `NITRO_BUN_IDLE_TIMEOUT` support for the Bun preset after it was lost during the move to `srvx`.

Values from `0` to `255` are forwarded to `Bun.serve`, including `0` to disable the timeout. Invalid values are ignored.

Includes unit coverage, a Bun preset regression test, and updated runtime documentation.

Fixes #3454

### DIFF
--- a/docs/2.deploy/10.runtimes/bun.md
+++ b/docs/2.deploy/10.runtimes/bun.md
@@ -16,4 +16,12 @@ After building with bun preset using `bun` as preset, you can run server in prod
 bun run ./.output/server/index.mjs
 ```
 
+## Idle Timeout
+
+Set `NITRO_BUN_IDLE_TIMEOUT` to change Bun's server idle timeout. The value is in seconds and must be between `0` and `255`. Setting it to `0` disables the idle timeout.
+
+```bash
+NITRO_BUN_IDLE_TIMEOUT=30 bun run ./.output/server/index.mjs
+```
+
 :read-more{to="https://bun.sh"}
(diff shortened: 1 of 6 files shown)


b-mounir-dev:
Hi @pi0 , any updates on this plz ? We are currently using a small patch, but it would be more convinient if we could rely on an official fix upstream. Thanks

pi0x:
- Rebased onto latest `main` and force-pushed. Two conflicts: `main` already added `setupCloseHooks` to `bun.ts` and already reworked the test's process cleanup. I kept `main`'s shape and re-applied your parts.
- The fix looks correct. srvx spreads `options.bun` into `Bun.serve`, so the value reaches Bun. Supporting `0` is better than Nitro 2 was.
- I changed two small things: the test now passes the env var only to the spawned Bun process, not globally, and uses `42` instead of `1`, because a 1 second idle timeout can drop keep-alive connections and make other tests flaky. I also moved the docs under `## Environment Variables` to match the Node page.
- Tests pass locally. Note that #3454 also asked for a `server.bun.idleTimeout` config option, which this does not add.

pi0x:
Thanks! With #4620 the server entry can pass options straight to srvx, so this can be done from `server.ts` without a dedicated env var: `export default defineServerEntry({ fetch, bun: { idleTimeout: 30 } })`. Unless there is a strong case for the env var itself, I'd suggest going with that once #4620 lands.

_Triaged with AI assistance._

PR #4436, opened 2026-07-16
fix(preset): keep lazy handlers code-split in dev

### LINKED ISSUE #4435: Dev: import-time error in a lazy route reported as unrelated `Cannot access '<x>' before initialization`
### Environment

- Nitro (nitropack): **2.13.4**
- Node: **22.17.1**
- OS: macOS (Darwin 25.5.0)
- rollup 4.62.2, h3 1.15.11
- Preset: `nitro-dev` (default dev server)

Also reproducible through Nuxt 4.4.8 (which uses Nitro 2.13.4).

### Reproduction

Minimal repro (no Nuxt, no third-party code — two route files):
https://github.com/hiranuma/nitro-dev-import-error-repro

```bash
npm install
npx nitro dev
# in another shell, request the HEALTHY, UNRELATED route:
curl http://localhost:3000/ok
```

```ts
// routes/ok.get.ts
export default defineEventHandler(() => ({ ok: true }));
```
```ts
// routes/boom.get.ts — throws while this module is being *evaluated* (import time)
throw new Error("boom_import_time_error");
export default defineEventHandler(() => ({ unreachable: true }));
```

### Describe the bug

When a lazy route handler throws at **import (module-evaluation) time**, the dev server
returns 500 on **every** request — including unrelated routes — with a misleading error that
names an internal symbol instead of the module that actually threw:

```
GET /ok  ->  500  Cannot access 'ok_get$1' before initialization
```

`/ok` is unrelated to `/boom`, yet it fails. The reported symbol (`ok_get$1`, or `renderer`
in a Nuxt app) is just whatever `const` Rollup happened to emit after the throw. The **real**
error (`boom_import_time_error`) is printed only **once**, at startup, as an
`[uncaughtException]`, and never again.

**Cause.** The `nitro-dev` preset sets `inlineDynamicImports: true`
([`src/presets/_nitro/nitro-dev.ts`](https://github.com/nitrojs/nitro/blob/v2/src/presets/_nitro/nitro-dev.ts),
comment `// externals plugin limitation`). `handlers.ts` emits lazy handlers as real dynamic
imports (`const _lazy_x = () => import('<handler>')`), but with inlining Rollup rewrites them
to `() => Promise.resolve().then(() => ns)` and inlines each handler body into the single
`.nitro/dev/index.mjs`, so every handler module is evaluated **eagerly** at that bundle's top
level. Rollup places those inlined bodies *after* the entry code, giving this order:

| line (repro) | statement |
|---|---|
| … | `const _lazy_… = () => Promise.resolve().then(() => renderer)` |
| ~1016 | `server.listen(...)` ← worker binds the port and reports ready |
| ~1292 | `throw new Error("boom…")` ← inlined handler body |
| later | `const renderer = …` / `const ok_get$1 = …` ← **never initialized** |

So evaluation starts the server, then aborts at the `throw`, leaving later `const`s
permanently in the temporal dead zone. `trapUnhandledNodeErrors()` logs and swallows the real
error, and the already-listening worker serves the TDZ error on every request.

**Supporting evidence for the ordering claim:** a **middleware** that throws at import time
already behaves correctly today — it is a non-lazy static import, so it throws *before*
`server.listen()`, the worker never boots, and the dev server reports the real error with its
source location. Only **lazy** handlers hit this. That asymmetry is exactly what the ordering
above predicts.

I believe this is the underlying mechanism behind a long tail of reports that were closed as
"circular dependency in user code" — in each, the actual trigger is a module that throws
during evaluation, and the reported symbol is unrelated: #1670, nuxt/nuxt#20576,
sidebase/nuxt-auth#271 / #420.

### Possible fixes (happy to open a PR)

**A. Keep lazy handlers code-split in dev** (`inlineDynamicImports: false` for `nitro-dev`).
I have this working locally with a regression test (fails before / passes after): `/ok` → 200,
`/boom` → 500 with the real error at its source. The full test suite is unchanged
(**643 passed, 0 failures**, identical to baseline). I'm aware of the guidance in #1670 against
touching `inlineDynamicImports`, and the `// externals plugin limitation` comment predates me —
so rather than assume: **is that limitation still real?** The current test fixture doesn't
reproduce it, but it may not cover whatever motivated the comment.

**B. Don't change bundling; make the init failure fatal.** Defer the worker's "ready" message
until after the bundle has fully evaluated, and if evaluation threw, report *that* error to the
dev server (which already renders build/worker errors) instead of serving requests against a
half-initialized module.

If either direction is acceptable I'll open a PR — A is ready; for B I'd want maintainer
direction first.

### Logs

```sh
# startup — note the server is already listening (➜ Local) BEFORE the real error:
  ➜ Local:    http://localhost:3000/
[nitro] ✔ Nitro Server built in 325ms
[uncaughtException] Error: boom_import_time_error
    at file:///.../.nitro/dev/index.mjs:1292:7
    at ModuleJob.run (node:internal/modules/esm/module_job:329:25)

# then, on EVERY request:
 ERROR  [request error] [unhandled] [GET] http://localhost:3000/ok
 Error: Cannot access 'ok_get$1' before initialization
   (.nitro/dev/index.mjs:1020:65)
```

jibin7jose:
### Description

This PR fixes an issue where an import-time error in a single lazy route handler takes down the entire dev server and breaks unrelated routes with a misleading `Cannot access '<x>' before initialization` error.

**Cause:** 
The `nitro-dev` preset was setting `inlineDynamicImports: true`. This caused Rollup to inline all lazy handler bodies into the single `index.mjs` startup bundle. An error evaluating one route would abort the execution of the bundle, leaving subsequent routes permanently in the Temporal Dead Zone (TDZ).

**Fix:** 
We've set `inlineDynamicImports: false` in the `nitro-dev` preset. This ensures lazy handlers remain code-split during local development, so an import-time failure in one module will not disrupt the evaluation of the rest of the server or other routes. When a broken route is accessed, the dev server will now accurately report the original initialization error.

Fixes #4435

### DIFF
--- a/src/presets/_nitro/nitro-dev.ts
+++ b/src/presets/_nitro/nitro-dev.ts
@@ -19,7 +19,7 @@ const nitroDev = defineNitroPreset(
     ],
     externals: { noTrace: true },
     serveStatic: true,
-    inlineDynamicImports: true, // externals plugin limitation
+    inlineDynamicImports: false,
     sourcemap: true,
   },
   {
--- a/src/presets/_nitro/nitro-dev.ts
+++ b/src/presets/_nitro/nitro-dev.ts
@@ -1,30 +1,30 @@
-import { runtimeDir } from "nitro/meta";
-import { defineNitroPreset } from "../_utils/preset.ts";
-import { join } from "pathe";
+import { defineNitroPreset } from "nitropack/kit";
 
 const nitroDev = defineNitroPreset(
   {
-    entry: "./_nitro/runtime/nitro-dev",
+    entry: "./runtime/nitro-dev",
     output: {
       dir: "{{ buildDir }}/dev",
       serverDir: "{{ buildDir }}/dev",
       publicDir: "{{ buildDir }}/dev",
     },
-    handlers: [
-      {
-        route: "/_nitro/tasks/**",
-        lazy: true,
-        handler: join(runtimeDir, "internal/routes/dev-tasks"),
-      },
-    ],
-    externals: { noTrace: true },
+    externals: { trace: false },
     serveStatic: true,
+    // Keep lazy handlers as real, code-split dynamic imports in dev. When this
+    // is `true`, Rollup inlines every lazy handler into a single `index.mjs`
+    // and evaluates each server module eagerly at that module's top level, so
+    // one module throwing at import time aborts the whole bundle's evaluation
+    // and leaves unrelated `const`s (e.g. `renderer`) in the temporal dead
+    // zone. Every request then fails with a misleading
+    // `Cannot access '<x>' before initialization` instead of the real error.
+    // See nitrojs/nitro#1670, nuxt/nuxt#20576.
     inlineDynamicImports: false,
-    sourcemap: true,
+    sourceMap: true,
   },
   {
     name: "nitro-dev" as const,
     dev: true,
+    url: import.meta.url,
   }
 );
 
--- a/test/fixtures/dev-import-error/routes/boom.get.ts
+++ b/test/fixtures/dev-import-error/routes/boom.get.ts
@@ -1,0 +1,8 @@
+// This module throws while it is being *evaluated* (import time), not inside
+// the handler. Regression guard for the dev server surfacing the real error
+// instead of an unrelated `Cannot access '<x>' before initialization` TDZ.
+throw new Error("boom_import_time_error");
+
+export default defineEventHandler(() => {
+  return { unreachable: true };
+});
--- a/test/fixtures/dev-import-error/routes/ok.get.ts
+++ b/test/fixtures/dev-import-error/routes/ok.get.ts
@@ -1,0 +1,5 @@
+// A perfectly healthy route that has nothing to do with the throwing one.
+// It must keep working even when a sibling route throws at import time.
+export default defineEventHandler(() => {
+  return { ok: true };
+});
--- a/test/unit/dev-import-error.test.ts
+++ b/test/unit/dev-import-error.test.ts
@@ -1,0 +1,67 @@
+import { fileURLToPath } from "node:url";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import type { Listener } from "listhen";
+import { build, createDevServer, createNitro, prepare } from "nitropack/core";
+import { fetch } from "ofetch";
+import { joinURL } from "ufo";
+
+// Regression test for the dev server reporting a misleading
+// `Cannot access '<x>' before initialization` when one server module throws at
+// import (evaluation) time. See nitrojs/nitro#1670, nuxt/nuxt#20576.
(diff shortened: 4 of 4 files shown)


hiranuma:
Heads-up after verifying this against `main` — I don't think `inlineDynamicImports: false` is the right change for v3, and I'd hate for it to land as a no-op or a regression.

I built `main` and ran a regression test (a healthy `/ok` route + a `/boom` route that throws at import time) across both dev builders:

| builder | `inlineDynamicImports` | result |
|---|---|---|
| **rolldown** (v3 default dev builder) | `true` (current) | ✅ `/ok` → 200 (isolated), `/boom` → real error, **no TDZ** |
| **rolldown** | `false` (this PR) | ✅ identical — the flag changes nothing |
| **rollup** | `true` (current) | ⚠️ worker init fails fast with the real error → 503 for all routes (real error in logs, no TDZ) |
| **rollup** | `false` (this PR) | ❌ **build crashes**: `TypeError: Cannot read properties of null (reading 'getVariableForExportName')` |

So on v3 the misleading `Cannot access '<x>' before initialization` symptom doesn't actually reproduce — v3's dev architecture already surfaces the real error. This PR is therefore a no-op on rolldown, and on the rollup builder it breaks the build (which looks like exactly the `// externals plugin limitation` the comment warns about — code-splitting isn't supported there).

Where the bug *does* reproduce is **v2**, which is what current Nuxt ships (`@nuxt/nitro-server` → `nitropack@^2.13.4`). I opened #4458 against the `v2` branch with the same one-line change plus a regression test that fails-before/passes-after; the existing suite is unchanged (641→641, 0 failures).

Also — the test and the preset comment in this PR are copied from my v2 branch, so they still import from `nitropack/*` (`nitropack/core`, `nitropack/kit`), which won't resolve on v3, and the v3-specific preset structure (the `/_nitro/tasks/**` handler, `externals: { noTrace: true }`, entry path) got overwritten. Worth reverting those if this PR continues.

Thanks for pushing on this either way — happy to help get the v2 side landed.

pi0x:
Triaged this against current `main`. Nothing pushed to your branch.

- I reproduced what @hiranuma reported: with `inlineDynamicImports: false` the `nitro-dev` preset fails to build under the rollup builder — `TypeError: Cannot read properties of null (reading 'getVariableForExportName')`. With `true`, `test/presets/nitro-dev.test.ts` passes (56 tests). That is the `// externals plugin limitation` the existing comment refers to.
- The diff also replaces the v3 `nitro-dev.ts` with the v2 one: it imports `nitropack/kit`, restores the old entry path and `externals: { trace: false }`, and drops the `/_nitro/tasks/**` dev task handler. `test/unit/dev-import-error.test.ts` imports `nitropack/core`, which does not exist in v3.

So this can't land on v3 as-is. #4458 against `v2` looks like the right home for the fix.

PR #4437, opened 2026-07-16
feat(vite): add `vite:before:close` hook

### LINKED ISSUE #4428: provide new hook to be triggered before closing nitro
### Describe the feature

Some integrations requires write/build some assets after copying public assets and/or prerender process. My example is PWA integration at tanstack start where we're using nitro v3 vite plugin.

Right now, nitro prerender process can be skipped even being triggered, missing hook about cancelling, for context: since any nitro module can remove all the prerender routes (without crawling), the prerender process is silently removed from the pipeline, but some prerender hooks triggered.

My integration at tanstack start uses nitro `compiled` hook, but nitro has been closed, and so I have no idea if we can have some side effects.

I guess we should add a new hook `before:close` to be triggered after `prerender:done` but before closing nitro here https://github.com/nitrojs/nitro/blob/main/src/build/vite/prod.ts#L110

For more context: at nuxt integration (nuxt <5), we use this logic in the nuxt module:

<details>
<summary>Nuxt PWA module and nitro hooks to build the service worker</summary>

```ts
  if (!nuxt.options.dev) {
    if (semver.gte(ctx.nuxt.nuxtVersion, '3.8.0')) {
      nuxt.hook('nitro:build:public-assets', async () => {
        await buildPwaAssets(ctx as unknown as any)
      })
    }
    else {
      nuxt.hook('nitro:init', (nitro) => {
        nitro.hooks.hook('rollup:before', async () => {
          await buildPwaAssets(ctx as unknown as any)
        })
      })
      if (nuxt.options.nitro.static || (nuxt.options as any)._generate /* TODO: remove in future */) {
        nuxt.hook('close', async () => {
          await buildPwaAssets(ctx as unknown as any)
        })
      }
    }
  }
```
</details>

You can see nuxt `nitro:build:public-assets` hook here (not sure yet how nuxt integrates nitro v3): https://github.com/nuxt/nuxt/blob/c8d7662e549177da82d6b50501b24bf5c08a9155/packages/nitro-server/src/legacy.ts#L17

### Additional information

userquin:
### 🔗 Linked issue

resolves #4428

### ❓ Type of change

### 📚 Description

Vite builder differs a lot from Rolldown/Rollup builders, the new hook is only for Vite builder and for build.

Once Nitro is closed, any asset written to the output public folder won't be added to the routes (404 when requesting assets generated at `compiled` hook)

We should add a custom entry for hooks like Nuxt docs, the config entry only mentons hookable and links to lifecycle that seems unrelated to Nitro hooks.

### 📝 Checklist

### DIFF
--- a/src/build/vite/prod.ts
+++ b/src/build/vite/prod.ts
@@ -103,6 +103,9 @@ export async function buildEnvironments(ctx: NitroPluginContext, builder: ViteBu
   // Prerender routes if configured
   await prerender(nitro);
 
+  // Call vite:before:close hook
+  await nitro.hooks.callHook("vite:before:close", nitro);
+
   // Build the Nitro server bundle
   const output = (await builder.build(builder.environments.nitro)) as RolldownOutput;
 
--- a/src/types/hooks.ts
+++ b/src/types/hooks.ts
@@ -10,6 +10,10 @@ export interface NitroHooks {
   "types:extend": (types: NitroTypes) => HookResult;
   "build:before": (nitro: Nitro) => HookResult;
   "rollup:before": (nitro: Nitro, config: RollupConfig) => HookResult;
+  /**
+   * Called before Nitro instance is closed during Vite build but after prerender process.
+   */
+  "vite:before:close": (nitro: Nitro) => HookResult;
   compiled: (nitro: Nitro) => HookResult;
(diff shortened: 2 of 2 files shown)


userquin:
Closing, I need to update my fork, I'm using v3 branch intead main, will send a new PR later 🤦‍♂️

PR #4444, opened 2026-07-17
docs: fix link navigation

iiio2:
I think when we click on `Docs for AI` in the home page, we do not get anything. But the link should move to `/llms.txt`. I think what I have done fix the issue.

https://nitro.build/


pi0:
it copies to clipboard.

PR #4460, opened 2026-07-22
docs: clarify noExternals path patterns

### LINKED ISSUE #3269: More docs about externals inlining
### Environment

Nitro 2.11.8

### Reproduction

https://github.com/Gwynerva/nitro-alias-bug

```
pnpm i
cd packages/nitro-app
pnpm nitro dev
```

### Describe the bug

Nitro does not to recognize aliases defined in `nitro.config.ts` in files from external dependencies.
When it meets aliases in such external files, it treats them as external dependencies, throwing an error.

@pi0 [explained](https://github.com/nuxt/nuxt/issues/31616#issuecomment-2768967400) the context:

> Any code that needs import aliases to work needs to be bundled and not externalized. By default, Nitro externalizes all node modules for performance reasons. We need to manually enlist these patterns.

### Workaround

The only way to solve this issue for now is to manually tell Nitro to inline modules that use imports with aliases:

```ts
// nitro.config.ts

import { resolve } from "node:path";

  compatibilityDate: "2025-03-01",
  srcDir: "server",
  alias: {
    "@shared": resolve(__dirname, "./shared"),
  },
  externals: {
      inline: [
        /packages\/lib/
      ],
  },
});
```

### Additional context

Original issue in Nuxt repo:

https://github.com/nuxt/nuxt/issues/31616

vittorioexp:
## Summary

Closes #3269.

Documents how `noExternals` array patterns work: they match resolved absolute module paths, not bare import specifiers. Adds monorepo and `node_modules` examples and a Nitro 2 `externals.inline` migration note.

Also updates the `noExternals` JSDoc in `src/types/config.ts` to match.

## Test plan

- [x] Verified against `src/build/config.ts` and `src/build/plugins/externals.ts`
- [x] Docs-only change (plus JSDoc)

### DIFF
--- a/docs/3.config/0.index.md
+++ b/docs/3.config/0.index.md
@@ -1065,13 +1065,45 @@ export default defineConfig({
 
 - Default: `false`{lang=ts}
 
-Prevent specific packages from being externalized. Set to `true` to bundle all dependencies, or pass an array of package names/patterns.
+By default, Nitro externalizes dependencies from `node_modules` for faster builds. Inlined modules are bundled into the output and can use your `alias` configuration.
+
+Set `noExternals` to `true` to bundle **all** dependencies, or pass an array of patterns to inline specific modules only.
 
 ```ts
 export default defineConfig({
   noExternals: true, // bundle all dependencies
 });
 ```
+
+#### Inlining specific packages
+
+When you pass an array, each entry is converted to a path regular expression and matched against the **resolved absolute module path** (the `id` Rollup sees after resolution). Patterns are **not** matched against bare import specifiers such as `@scope/pkg` or `lodash`.
+
+Use this when:
+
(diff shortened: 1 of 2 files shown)


pi0x:
- Rebased onto `main`, no conflicts.
- I tested this against v3 and two claims were wrong, so I fixed them in an extra commit:
  - Patterns **are** matched against bare specifiers. `noExternals: [/^my-lib$/]` (which can never match a file path) still inlines the package. They are matched against the import specifier and against the resolved path.
  - v3 production builds do not externalize `node_modules` by default; they bundle almost everything. Only native/non-bundleable packages and `traceDeps` stay external. The old behaviour still applies in dev, which is where this option matters.
- Rest of the PR was correct and I kept it: RegExp used as-is, strings escaped as substrings, `[\\/]` for Windows, and the Nitro 2 `externals.inline` note.

pi0 (approved):

PR #4461, opened 2026-07-22
docs: clarify prerender and isr route rules

### LINKED ISSUE #4041: Using `prerender` + `isr` route rules
## Description

When using `isr` or `swr` in `routeRules`, Nitro sets CDN-level caching (via `s-maxage` / platform-specific headers), but does **not** set an explicit browser-level `Cache-Control` header on the HTML response.

This means the browser falls back to **heuristic caching** — it looks at `Last-Modified` or `Date` and guesses a TTL (up to 10% of the resource's age per [RFC 7234 §4.2.2](https://httpwg.org/specs/rfc7234.html#heuristic.freshness)). In practice, this causes browsers to serve stale HTML after new deployments, even after a hard refresh (`Ctrl+Shift+R`).

## Steps to reproduce

```js
// nuxt.config.ts (deployed to Vercel)
  routeRules: {
    '/**': {
      prerender: true,
      isr: 3600,
    },
  },
})
```

1. Deploy the app
2. Visit the site in a browser
3. Deploy a new version with visible changes
4. Hard refresh (`Ctrl+Shift+R`) — **stale content persists**
5. Open in incognito — new content appears
6. In the original browser, DevTools → Application → Clear site data — new content finally appears

## Expected behavior

Nitro should set `Cache-Control: public, max-age=0, must-revalidate` (or equivalent) on HTML responses served via ISR/SWR, so browsers always revalidate with the server/CDN instead of heuristically caching stale pages.

For reference, Next.js on Vercel sends `Cache-Control: public, max-age=0, must-revalidate` by default on ISR pages.

## Current workaround

Manually add browser cache headers to the route rule:

```js
routeRules: {
  '/**': {
    prerender: true,
    isr: 3600,
    headers: {
      'Cache-Control': 'public, max-age=0, must-revalidate',
    },
  },
},
```

## Additional context

- Platform: Vercel
- Nuxt 3 / Nitro
- The ISR CDN caching works correctly — the issue is only about the **browser-facing** `Cache-Control` header being absent

vittorioexp:
## Summary

Closes #4041.

Documents that `prerender` and `isr` must not be combined on the same route — prerendered pages are generated at build time and ISR is ignored. Adds routing guide warnings and updates the `isr` JSDoc.

## Test plan

- [x] Docs-only change (plus JSDoc)
- [x] Wording aligned with maintainer clarification on the issue

### DIFF
--- a/docs/1.docs/5.routing.md
+++ b/docs/1.docs/5.routing.md
@@ -660,9 +660,19 @@ export default defineConfig({
 });
 ```
 
+Prerendered routes are generated once at build time and served as static files. They are not revalidated at runtime.
+
+::warning
+Do not combine `prerender` and `isr` on the same route. Prerendering takes precedence, so `isr` is ignored and the page will not be regenerated after deployment. Use `isr` alone when you need incremental regeneration on supported platforms.
+::
+
 ### ISR (Vercel)
 
 Configure Incremental Static Regeneration for Vercel deployments:
+
+::note
+ISR generates pages on demand and revalidates them according to the expiration you set. It is an alternative to build-time `prerender`, not something you combine with it on the same route.
+::
 
 ```ts [nitro.config.ts]
 import { defineConfig } from "nitro";
--- a/docs/1.docs/5.routing.md
+++ b/docs/1.docs/5.routing.md
@@ -655,9 +655,19 @@ export default defineConfig({
 });
 ```
 
+Prerendered routes are generated once at build time and served as static files. They are not revalidated at runtime.
+
+::warning
+Do not combine `prerender` and `isr` on the same route. Prerendering takes precedence, so `isr` is ignored and the page will not be regenerated after deployment. Use `isr` alone when you need incremental regeneration on supported platforms.
+::
+
 ### ISR (Vercel)
 
 Configure Incremental Static Regeneration for Vercel deployments:
+
+::note
+ISR generates pages on demand and revalidates them according to the expiration you set. It is an alternative to build-time `prerender`, not something you combine with it on the same route.
+::
 
 ```ts [nitro.config.ts]
 import { defineConfig } from "nitro";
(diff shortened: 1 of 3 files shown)


pi0x:
Thanks! Confirmed this isn't documented anywhere today — neither the prerendering guide, the routing guide, nor the Vercel provider page mention the interaction.

The behavior is accurate: in `src/presets/vercel/utils.ts` the `{ handle: "filesystem" }` entry is emitted before the ISR rewrites, so a prerendered file matches first and the ISR function is never invoked.

Two notes before merging:

1. Please add the same warning to `docs/2.deploy/20.providers/vercel.md` (the "On-demand incremental static regeneration (ISR)" section) — that's where most users read about ISR, and it currently says nothing about `prerender`.
2. The `::note` currently sits between the `### ISR (Vercel)` heading and its code block. Moving it below the example reads better and keeps the section opening with the config it documents.

pi0x:
Rebased this PR onto latest `main`. There were no conflicts, and the original change is intact.

**I checked the claim in the code, and it is correct.**

In `src/presets/vercel/utils.ts` the generated `config.routes` array puts `{ handle: "filesystem" }` before the ISR rewrites are pushed. A prerendered route is written as a static file, so Vercel serves that file first and the ISR function is never called. The `overrides` map from `getPrerenderOverrides()` also points those paths at the static output. So `prerender` really does win over `isr` on the same route.

**I also applied the two things @pi0x asked for**, in one extra commit on top of yours:

1. Moved the `::note` in `docs/1.docs/5.routing.md` to below the ISR code example, so the section starts with the config it documents.
2. Added the same warning to `docs/2.deploy/20.providers/vercel.md`, in the "On-demand incremental static regeneration (ISR)" section.

Nothing else needed changing. `pnpm fmt` and `pnpm typecheck` both pass.

One note for maintainers, not for the author: CodeRabbit flagged that issue #4041 also asked for browser revalidation headers. That is a code change and out of scope here. The Netlify preset already sets `Cache-Control: public, max-age=0, must-revalidate` for ISR routes (`src/presets/netlify/runtime/netlify.ts`), so if that behaviour is wanted on Vercel too it should be a separate PR.

pi0 (approved):

PR #4464, opened 2026-07-22
docs: document custom storage drivers

### LINKED ISSUE #3847: Custom drivers
Hello Nitro team!

### Expected behavior

I would expect Nitro to support **custom driver functions or objects** in the storage configuration, not just string identifiers.

Or clarify the documentation that only string identifiers for built-in drivers are supported in `nuxt.config.ts`: https://unstorage.unjs.io/guide/custom-driver

---

### Describe the bug

When configuring a custom storage driver for Nitro in `nuxt.config.ts` using `defineDriver`, Nitro tries to resolve the driver function or object as a package. This results in an error like:

```
// with driver: upstashCustomDriver()
Cannot find package [object Object]' imported from node_modules/.../nitropack/dist/core/index.mjs
```
```
// with driver: upstashCustomDriver
Cannot find package '(opts = {}) => { … }' imported from node_modules/.../nitropack/dist/core/index.mjs
```

Currently, only string identifiers for built-in drivers (e.g. 'redis', 'memory') work in the config.

---

### Environment

* Nuxt / Nitro version: 3.x / 2.x
* Deployment: Vercel serverless
* Storage: Custom Unstorage driver based on Upstash

---

### Reproduction

```ts
// drivers/upstashCustomDriver.ts
import { defineDriver } from 'unstorage';
import { upstashDriver } from 'unstorage/drivers';

export const upstashCustomDriver = defineDriver((opts = {}) => {
  const driver = upstashDriver(opts);
  return {
    ...driver,
    async setItem(key: string, value: any, opts?: { ttl?: number }) {
      const modifiedOpts = opts?.ttl ? { ...opts, ttl: Math.ceil(opts.ttl / 1000) } : opts;
      return driver.setItem!(key, value, modifiedOpts);
    }
  };
});
````

```ts
// nuxt.config.ts
import { upstashCustomDriver } from './drivers/upstashCustomDriver';

  nitro: {
    storage: {
      upstash: {
        driver: upstashCustomDriver(), // ❌ causes error on Vercel
        //driver: upstashCustomDriver, // ❌ causes error too
        url: process.env.KV_REST_API_URL,
        token: process.env.KV_REST_API_TOKEN
      }
    }
  }
});
```

vittorioexp:
## Summary

Closes #3847.

Documents how to register custom unstorage drivers via a file path in nitro.config, with an example based on maintainer guidance.

## Test plan

- [x] Docs-only change

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/1.docs/8.storage.md
+++ b/docs/1.docs/8.storage.md
@@ -105,41 +105,60 @@ You can find the driver list on [unstorage documentation](https://unstorage.unjs
 
 The `driver` field accepts a built-in driver name (for example `"redis"`) or a **path** to a custom driver module. You cannot pass a driver function or object directly in config — Nitro imports drivers as separate modules so your config stays separate from runtime code.
 
-Create a driver file that default-exports a driver created with `defineDriver`:
+Create a driver file that **default-exports** a driver factory from `defineDriver`. To build a driver from scratch, implement the [unstorage driver interface](https://unstorage.unjs.io/guide/custom-driver) (`hasItem`, `getItem`, `setItem`, and so on):
 
-```ts [drivers/upstash.ts]
+```ts [drivers/kv.ts]
 import { defineDriver } from "unstorage";
-import { upstashDriver } from "unstorage/drivers/upstash";
 
-export default defineDriver((opts = {}) => {
-  const driver = upstashDriver(opts);
+export default defineDriver((opts: { prefix?: string } = {}) => {
+  const data = new Map<string, string>();
+  const prefix = opts.prefix || "";
+
   return {
-    ...driver,
-    async setItem(key, value, opts) {
-      const ttl = opts?.ttl ? { ...opts, ttl: Math.ceil(opts.ttl / 1000) } : opts;
-      return driver.setItem!(key, value, ttl);
+    name: "kv",
+    options: opts,
+    hasItem(key) {
+      return data.has(prefix + key);
+    },
+    getItem(key) {
+      return data.get(prefix + key) ?? null;
+    },
+    setItem(key, value) {
+      data.set(prefix + key, value);
+    },
+    removeItem(key) {
+      data.delete(prefix + key);
+    },
+    getKeys() {
+      return [...data.keys()].map((k) =>
+        prefix ? k.slice(prefix.length) : k,
+      );
+    },
+    clear() {
+      data.clear();
     },
   };
 });
 ```
 
-Reference it by path in your Nitro config:
+Reference it by path in your Nitro config (options besides `driver` are passed into the factory):
 
 ```ts [nitro.config.ts]
 import { defineConfig } from "nitro";
 
 export default defineConfig({
   storage: {
-    upstash: {
-      driver: "./drivers/upstash.ts",
-      url: process.env.KV_REST_API_URL,
-      token: process.env.KV_REST_API_TOKEN,
+    kv: {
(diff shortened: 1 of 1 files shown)


docs/1.docs/8.storage.md
```diff
@@ -101,6 +101,48 @@ Then, you can use the redis storage using the `useStorage("redis")` function.
...
 
+### Custom drivers
+
+The `driver` field accepts a built-in driver name (for example `"redis"`) or a **path** to a custom driver module. You cannot pass a driver function or object directly in config — Nitro imports drivers as separate modules so your config stays separate from runtime code.
+
+Create a driver file that default-exports a driver created with `defineDriver`:
+
+```ts [drivers/upstash.ts]
+import { defineDriver } from "unstorage";
+import { upstashDriver } from "unstorage/drivers/upstash";
```
pi0: This is bad for the example of custom drivers / relavant to wrap to patch but making from scratch, ony needs implemeneting the interface

pi0x:
Thanks for taking this on @vittorioexp, and sorry for the slow reply. #3847 is a real gap — there's no "custom drivers" section in the storage docs today. Unfortunately I can't merge this version, because I tested both of its central claims against `main` and neither holds anymore.

**1. The recommended relative path doesn't build.**

The guide says to use `driver: "./drivers/kv.ts"` and that the path is "relative to your project root". Driver strings are passed through as import specifiers (`src/utils/storage.ts`) and emitted inside the virtual module `#nitro/virtual/storage` (`src/build/virtual/storage.ts`). There is no real file backing that importer, so the bundler has nothing to resolve `./` against. Building the test fixture with exactly the config from this PR:

```
[UNRESOLVED_IMPORT] Could not resolve './drivers/kv.ts' in #nitro/virtual/storage
```

`~/drivers/kv.ts` fails the same way. Only an **absolute** path works — that one builds and the driver ends up in `.output/server/_chunks/storage.mjs`. So the parenthetical "(or an absolute path)" is the only accurate half, and the form the docs lead with is the broken one.

**2. The example uses an API that no longer exists.**

The driver example imports `defineDriver` from `unstorage`. Nitro v3 is on unstorage v2, which removed it:

```
[MISSING_EXPORT] "defineDriver" is not exported by "unstorage/dist/index.mjs"
```

In unstorage v2 a driver is just a plain factory function returning the driver object. This PR was written against v1 semantics.

---

**What we'd take instead**

The cheapest correct fix for #3847 is probably not a new hand-rolled-driver walkthrough at all. The [Runtime configuration](https://nitro.build/docs/storage#runtime-configuration) section already shows mounting an arbitrary driver from a plugin, which sidesteps path resolution entirely and works today. A short `### Custom drivers` heading that points at it would close the issue.

If you'd rather keep the config-path approach, a rewrite would need:

- absolute paths only (e.g. `fileURLToPath(new URL("drivers/kv.ts", import.meta.url))`), with no relative-path advice
- a plain factory function, no `defineDriver`
- a note that custom drivers are outside Nitro's [driver dependency detection](https://nitro.build/docs/storage#driver-dependencies), so any third-party library has to be passed via the driver's `lib` option — the same caveat the `::important` block in the runtime-configuration section already makes

Happy to review a follow-up if you want to pick it up. Closing this one for now.

<sub>🤖 This review was done with the help of Claude Code; the build failures above were reproduced locally against `main` before posting.</sub>

Issue #4472, opened 2026-07-23
Vite dev: SSR serves stale modules after edits — no reload for shared modules, missing file associations, and dev-worker reload() reuses cached evaluations


brenelz:
### Environment

- nitro: 3.0.260610-beta (also reproduced against current `main` sources)
- vite: 8.0.10
- vite-plugin-solid: 3.0.0-next.15 (solid-js 2.0.0-beta.23, SSR via a `ssr` service environment)
- Node.js 22 / macOS
- dev runner: node-worker (default)

### Reproduction

Any Vite + Nitro app where SSR renders modules that are also part of the client graph (i.e. every route component). With the dev server running:

1. Load a page (SSR renders it).
2. Edit the route component (or any module shared between client and server envs).
3. Reload the browser: the server still renders the **old** module — hydration keys/markup no longer match the freshly HMR-updated client, producing hydration mismatches (unclaimed server nodes, and in our case a hard `Cannot read properties of null (reading 'nextSibling')` crash during hydration).
4. Only restarting the dev server picks up the change.

This is the same end-user symptom as #4020 (closed, but the fix never landed — `reload()` on `main` still doesn't clear the cache) and #4043. Tracing the full chain with `NITRO_DEBUG=1` and instrumented `hotUpdate` hooks turned up **three** independent gaps, each of which is sufficient to keep SSR stale:

### Describe the bug

**1. `hotUpdate` never sends `full-reload` for shared modules** (`src/build/vite/plugin.ts`)

Modules that also exist in a client environment go into `sharedModules` and are neither invalidated in the server graph nor do they trigger `env.hot.send({ type: "full-reload" })`. Editing a route component therefore never reloads the server runner at all. (This is #4043; PR #4044 addressed the invalidation half but was closed unmerged.)

**2. `modules` is empty for server environments anyway — file→module associations are missing**

Even if (1) were fixed, the `modules` array `hotUpdate` receives for the server environments is empty: `environment.moduleGraph.getModulesByFile(file)` returns nothing, although the module is present in `idToModuleMap` under its resolved id (verified: id `/abs/path/src/routes/Index.tsx` exists while `getModulesByFile('/abs/path/src/routes/Index.tsx')` is empty). Since Vite's own file-change invalidation also goes through `getModulesByFile`, the server env's *transform cache* is never invalidated for edited files either. Modules that enter the graph via the module-runner `fetchModule` path apparently never get registered in `fileToModulesMap`.

[...]

PR #4473, opened 2026-07-23
fix(vite): clear module runner cache before dev worker reload

brenelz:
### 🔗 Linked issue

Resolves item (3) of #4472. Same root cause as #4020 (closed, but the fix never landed). Related to #4043.

### ❓ Type of change

### 📚 Description

`ViteEnvRunner.reload()` in the dev worker re-imports the environment entry without clearing the `ModuleRunner`'s `evaluatedModules` cache.

The runner's `fetchModule` protocol answers `{cache: true}` for any module whose transform is already populated on the Vite side at fetch time (`packages/vite`'s `fetchModule`: `if (options.cached && cached) return { cache: true }`). During a dev-worker reload, plugin-side crawls can repopulate transforms for large parts of the graph before the runner re-fetches them (observed with vite-plugin-solid's `?assets` transform), so the re-import walks the entry but keeps stale evaluations for the very modules that changed — the server renders old code until the dev server is restarted, causing SSR/client hydration mismatches.

Vite's own `full-reload` handler in the module runner clears `evaluatedModules` before re-importing ([`hmrHandler.ts`](https://github.com/vitejs/vite/blob/main/packages/vite/src/module-runner/hmrHandler.ts)); this PR does the same in the worker's `reload()`.

Verified against a Vite 8 + vite-plugin-solid SSR app: with this change (plus a `full-reload` actually being sent for shared modules — see #4472 items 1–2), edits to SSR-rendered modules are reflected on the next request without restarting the dev server, and hydration is clean. `NITRO_DEBUG=1` traces confirm all app modules re-execute after reload instead of being served from the stale cache.

Includes a regression test in the `vite:hmr` suite: the fixture gains a `dep-crawler` plugin that re-transforms `dep.ts` as a side effect of transforming `api/crawled.ts` (the way asset-collecting plugins crawl imports), plus a plugin that hard-invalidates the server module graph on `dep.ts` edits (as framework staleness workarounds do in the wild). Without `evaluatedModules.clear()` the crawled transform makes `fetchModule` answer `{cache: true}` for `dep.ts` and the test times out on the stale response; with the fix it passes.

### 📝 Checklist


pi0:
Would be nice to add a regression test to vite-hmr fixture!

pi0 (approved):

PR #4477, opened 2026-07-23
feat(types): opt-in flat NitroFetchRequest to bound $fetch type instantiation

### LINKED ISSUE #4476: Typed $fetch route union causes TS2589/TS2321 (excessive type instantiation) on apps with many routes
### Environment

- nitropack `2.13.4` (v2) — also reproduced on `main` (v3), same `src/types/fetch/fetch.ts` shape
- TypeScript 5.9.x, and `@typescript/native-preview` (tsgo)
- Real app: a Nuxt 4 project whose generated `InternalApi` has **~846 routes**

### Reproduction

Any project with a large generated `InternalApi` (hundreds of routes) that calls typed `$fetch` / `useFetch` / `useAsyncData` / `event.$fetch` with a string literal. On the real app it produces **TS2589** (`Type instantiation is excessively deep and possibly infinite`) and, under tsgo, a flood of **TS2321** (`Excessive stack depth comparing types`) on those call sites. A minimal repro is a Nitro app with a few hundred route handlers + one `$fetch('/some/route')`.

### Describe the bug

`NitroFetchRequest` is the **union of all route keys** and is used as the *generic constraint* of `$Fetch` (`R extends NitroFetchRequest`). At every typed call site the URL literal is re-scored against that union via `MatchedRoutes` / `AvailableRouterMethod` / `TypedInternalResponse`. With hundreds of routes this instantiates enormously — TypeScript's `tsc` sits just under its 5M instantiation valve, and Go-native `tsgo` (stricter, different stable type ordering) tips over it → TS2589/TS2321. It is not an app bug: the generated code is correct; the cost is intrinsic to scoring the union per call.

This is the same family as the TypeScript compiler issues below, where the maintainers consider the behavior **intended** (stable type ordering) and **close the issues as `NOT_PLANNED`, pointing to the libraries to make their types less pathological** (e.g. TypeBox fixed its own types in #929):

- microsoft/typescript-go#929 (closed — fixed upstream in the library, not the compiler)
- microsoft/typescript-go#4465 (closed `NOT_PLANNED` — "it's the stable type ordering")
- microsoft/typescript-go#787 (stable type ordering is a deliberate design decision)

So Nitro is the right place to offer a mitigation.

### Proposed fix (PR attached)

An **opt-in**, non-breaking escape hatch: a `NitroFetchConfig` interface that, when a project sets `{ flatRequest: true }`, resolves `NitroFetchRequest` to `string` and stops the per-call-site route-union scoring.

```ts
declare module "nitropack" {
  interface NitroFetchConfig { flatRequest: true }
}
```

Default behavior is **unchanged** (the union) — existing users keep full autocomplete + typing. When opted in, **response typing is preserved** (`MatchedRoutes<R>` still resolves the return from the URL literal); only request-key autocomplete is traded away. Measured on the real app: it removes the entire TS2589/TS2321 class on both `tsc` and `tsgo` with zero new errors.

PR: opt-in flat `NitroFetchRequest` (targets `main`).

Killian-Aidalinfo:
### 🔗 Linked issue

Closes #4476

### ❓ Type of change

- [x] 🚀 Enhancement (opt-in, non-breaking)

### 📚 Description

Apps with a large generated `InternalApi` (hundreds of routes) hit **TS2589** (`Type instantiation is excessively deep`) and, under `tsgo`, a flood of **TS2321** on typed `$fetch` / `useFetch` / `useAsyncData` / `event.$fetch` calls. `NitroFetchRequest` is the union of all route keys used as the generic constraint (`R extends NitroFetchRequest`), so every call site re-scores the URL literal against the full union via `MatchedRoutes`/`AvailableRouterMethod`/`TypedInternalResponse`. `tsc` sits just under its 5M instantiation limit; `tsgo` (stricter, stable type ordering) tips over → the whole diagnostic class.

This is the same family the TypeScript team closes as intended / `NOT_PLANNED`, pointing to libraries to fix their types (microsoft/typescript-go#929, #4465, #787) — so the mitigation belongs here.

**This PR adds an opt-in, non-breaking escape hatch** — a `NitroFetchConfig` interface:

```ts
declare module "nitropack" {
  interface NitroFetchConfig { flatRequest: true }
}
```

When opted in, `NitroFetchRequest` resolves to `string`, which stops the per-call-site union scoring. **Default is unchanged** (the union) so existing users keep full request autocomplete + typing. When enabled, **response typing is preserved** (`MatchedRoutes<R>` still resolves the return type from the URL literal); only request-key autocomplete is traded away.

### ✅ Why this approach

- **Non-breaking**: gated behind an interface users must explicitly augment; default resolves to the current union.
- **Keeps response typing**: `TypedInternalResponse`/`MatchedRoutes` are untouched.
- **Minimal**: one type-only change in `src/types/fetch/fetch.ts`.

Alternatives rejected: making `MatchedRoutes` non-recursive or emptying `InternalApi` destroys response typing (trades TS2589 for a flood of TS2339/TS18046); changing only the `$Fetch` default `R` has no effect (the union must stay in the constraint for autocomplete, and that union is exactly what over-instantiates).

### 🧪 Measured

On a real Nuxt 4 app (~846 routes): removes the entire TS2589/TS2321 class on both `tsc` and `tsgo` with **zero new errors**, response types preserved.

> Note: the same change applies cleanly to the `v2` branch (nitropack 2.13.x) if a backport is wanted.

### DIFF
--- a/src/types/fetch/fetch.ts
+++ b/src/types/fetch/fetch.ts
@@ -5,13 +5,23 @@ import type { MatchedRoutes } from "./_match.ts";
 // An interface to extend in a local project
 export interface InternalApi {}
 
+// Opt-in configuration interface to augment in a local project, e.g.
+//   declare module "nitropack" {
+//     interface NitroFetchConfig { flatRequest: true }
+//   }
+export interface NitroFetchConfig {}
+
 // TODO: upgrade to uppercase for h3 v2 types and web consistency
 type RouterMethod = Lowercase<HTTPMethod>;
 
-export type NitroFetchRequest =
+type TypedNitroFetchRequest =
   | Exclude<keyof InternalApi, `/_${string}` | `/api/_${string}`>
   | Exclude<FetchRequest, string>
   | (string & {});
+
+export type NitroFetchRequest = NitroFetchConfig extends { flatRequest: true }
+  ? string
+  : TypedNitroFetchRequest;
 
 export type MiddlewareOf<
   Route extends string,


Killian-Aidalinfo:
Hi @pi0 , I opened this PR to fix the type-instantiation issues we're seeing with the Go version of vue-tsc (tsgo).

There have already been several issues opened on the Microsoft side, but they were closed as intended / not planned, with the recommendation that libraries patch their own types instead.

The approach here is intentionally minimal and opt-in. It adds an escape hatch for users hitting these limits while keeping the current behavior by default. Response typing is still preserved, the only trade-off is request-path autocomplete when the option is enabled.

I tested it on a fairly large Nuxt 4 project (~846 routes), and it gets rid of the TS2589/TS2321 errors on both tsc and tsgo without introducing any new ones.

Let me know what you think 🙂

pi0:
Thanks for PR. Typed fetch is being removed from v3 #4572

PR #4478, opened 2026-07-23
fix: try to resolve server entry also from `server/index`

### LINKED ISSUE #4455: Possibility of recognizing `server/index.ts` as a server entry?
### Describe the feature

Would be nice to have this imo as an extension of Nitro's low-config philosophy. I saw https://github.com/nitrojs/nitro/pull/4313 but it seems like that only enables `/server/server.ts` alongside `/server.ts` usage?

rel: https://github.com/nuxt/nuxt/issues/35484

### Additional information

lazerg:
### 🔗 Linked issue

Resolves #4455

### ❓ Type of change

### 📚 Description

`server.ts` in the project root is auto-detected as the server entry, and #4313 extended that to a `server` dir, but `server/index.ts` still isn't picked up. This adds `index` to the resolved suffixes so the folder form (`server/index.ts`) is detected the same way.

### 📝 Checklist

- [x] I have updated the tests accordingly.


lazerg:
Thanks for the cleanup, no objection here. The actual fix in `paths.ts` (the `suffixes` addition) is still in place, so the behavior change stands on its own without the unit test. Happy to add a lighter test back if you'd rather have one, just say the word.

lazerg:
@pi0 when you have a moment, could you take a look at this, the paths.ts fix is still in place and ready for review.

pi0 (approved):

pi0:
Thanks for working on this. Idea looks nice but thinking more it might conflict with some user code. better to limit magic...

PR #4480, opened 2026-07-25
fix(dev): proxy websocket upgrades for `devProxy` routes with `ws`

### LINKED ISSUE #4269: nitro.devProxy[...].ws: true is silently ignored — WebSocket upgrades bypass devProxy and hit the worker
### Environment

- `nitropack@2.13.4` (current `latest` on npm; behavior identical on `main`)
- Node.js 22, dev server only

### Reproduction

```ts
// nuxt.config.ts (or any nitro consumer)
  nitro: {
    devProxy: {
      '/maildev': {
        target: 'http://127.0.0.1:1080/maildev',
        changeOrigin: true,
        ws: true, // <-- documented option, has no effect
      },
    },
  },
})
```

Run anything behind the proxy that opens a WebSocket — e.g. MailDev's socket.io endpoint:

```
GET /maildev/socket.io/?EIO=4&transport=websocket   Upgrade: websocket
```

### Expected

The upgrade is forwarded to `http://127.0.0.1:1080/maildev/socket.io/...`, MailDev responds `101 Switching Protocols`, and the WS pipe is established. This is what the `ProxyServerOptions.ws` field — re-exported from `httpxy` and documented as the WS-proxying knob — would lead a user to expect.

### Actual

The upgrade lands in the Nitro worker, which has no handler for that path, so it logs:

```
[request error] [fatal] [GET] http://localhost:3000/maildev/socket.io/?EIO=4&transport=websocket
ℹ Error: Page not found
```

The httpxy `ProxyServer` instance receives no `upgrade` event and never gets a chance to call its `.ws()` method, even though it was constructed with `ws: true`.

### Root cause

Two independent places in `src/dev/`:

**1. `devProxy` only wires the proxy into the HTTP-request path**

[`src/dev/app.ts#L78-L86`](https://github.com/nitrojs/nitro/blob/v2.13.4/src/dev/app.ts#L78-L86):

```ts
const routes = Object.keys(this.nitro.options.devProxy).sort().reverse();
for (const route of routes) {
  let opts = this.nitro.options.devProxy[route];
  if (typeof opts === "string") {
    opts = { target: opts };
  }
  const proxy = createHTTPProxy(opts);
  app.all(route, proxy.handleEvent); // ← HTTP only
}
```

…and `createHTTPProxy` at [`src/dev/app.ts#L145-L165`](https://github.com/nitrojs/nitro/blob/v2.13.4/src/dev/app.ts#L145-L165) only exposes `handleEvent`, which calls `proxy.web()`:

```ts
function createHTTPProxy(defaults: ProxyServerOptions = {}) {
  const proxy = createProxyServer({ xfwd: true, ...defaults });
  return {
    proxy,
    async handleEvent(event: H3Event, opts?: ProxyServerOptions) {
      try {
        return await fromNodeHandler((req, res) => {
          return proxy.web(req as IncomingMessage, res as ServerResponse, opts);
        })(event);
      } catch (error: any) { /* ... */ }
    },
  };
}
```

`proxy.ws()` is never called and the `proxy` instance is not exported back to the caller in any way that gets reached during an upgrade.

**2. The dev server forwards every upgrade straight to the worker**

[`src/dev/server.ts#L125-L146`](https://github.com/nitrojs/nitro/blob/v2.13.4/src/dev/server.ts#L125-L146):

```ts
async upgrade(req: IncomingMessage, socket: Socket, head: any) {
  if (!this.#manager.upgrade) {
    throw new HTTPError({ status: 501, statusText: "Worker does not support upgrades." });
  }
  return this.#manager.upgrade({ node: { req, socket, head } });
}

listen(opts?: Partial<Omit<ServerOptions, "fetch">>): Server {
  const server = serve({ ...opts, fetch: this.fetch, gracefulShutdown: false });
  this.#listeners.push(server);
  if (server.node?.server) {
    server.node.server.on("upgrade", (req, sock, head) => this.upgrade(req, sock, head));
  }
  return server;
}
```

There is no check against the `devProxy` route table — every upgrade goes to `this.#manager.upgrade(...)` (the worker), regardless of whether the request path matches a `devProxy` prefix.

### Type-level confusion

The config type at [`src/types/config.ts#L560`](https://github.com/nitrojs/nitro/blob/v2.13.4/src/types/config.ts#L560) is `Record<string, string | ProxyServerOptions>`, where `ProxyServerOptions` is re-exported from `httpxy` and includes `ws: boolean`. So `ws: true` autocompletes, type-checks, and silently does nothing.

### Suggested fix

In `NitroDevServer.upgrade` (or in `listen`'s upgrade listener), before calling `this.#manager.upgrade`, match `req.url` against the sorted-descending `devProxy` routes. If one matches, call the corresponding `proxy.ws(req, socket, head, opts)` from the httpxy instance created in `app.ts`. This means `createHTTPProxy` needs to also expose `proxy` (or a `handleUpgrade` wrapper) and `app.ts` needs to register those alongside the route handlers — e.g. keyed by route prefix, looked up in `server.ts`.

Happy to send a PR if maintainers agree on the shape.

### Workaround

For consumers hitting this today: patch `nuxt.server.upgrade` in a Nuxt module (or the equivalent underlying `NitroDevServer.upgrade` in non-Nuxt setups) and tunnel matching upgrades to the target using `node:http`.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nitrojs/nitro/issues/4269

### ❓ Type of change

### 📚 Description

this is also reproducible on v2 (see https://github.com/nuxt/cli/issues/107).

the issue is that `devProxy` with `ws: true` proxies HTTP fine but the upgrade falls through to the dev worker and gets answered as a normal request

### 📝 Checklist

### DIFF
--- a/src/dev/app.ts
+++ b/src/dev/app.ts
@@ -2,6 +2,7 @@ import type { Nitro } from "nitro/types";
 import type { H3Event, HTTPHandler } from "h3";
 import { createProxyServer, type ProxyServerOptions } from "httpxy";
 import type { IncomingMessage, ServerResponse } from "node:http";
+import type { Socket } from "node:net";
 import { H3, toEventHandler, serveStatic, fromNodeHandler, HTTPError } from "h3";
 import { joinURL } from "ufo";
 import mime from "mime";
@@ -20,6 +21,8 @@ import devErrorHandler, {
 export class NitroDevApp {
   nitro: Nitro;
   fetch: (req: Request) => Response | Promise<Response>;
+
+  #wsProxies: { route: string; proxy: ReturnType<typeof createHTTPProxy> }[] = [];
 
   constructor(nitro: Nitro, catchAllHandler?: HTTPHandler) {
     this.nitro = nitro;
@@ -97,6 +100,9 @@ export class NitroDevApp {
       }
       const proxy = createHTTPProxy(opts);
       app.all(route, proxy.handleEvent);
+      if (opts.ws) {
+        this.#wsProxies.push({ route, proxy });
+      }
     }
 
     // Main handler
@@ -105,6 +111,29 @@ export class NitroDevApp {
     }
 
     return app;
+  }
+
+  /**
+   * Proxy a WebSocket upgrade request if it matches a `devProxy` rule with `ws` enabled.
+   *
+   * @returns `true` if the socket was handed to a proxy, `false` if the caller should handle it.
+   */
+  proxyUpgrade(req: IncomingMessage, socket: Socket, head: any): boolean {
+    if (this.#wsProxies.length === 0) {
+      return false;
+    }
+    const path = (req.url || "/").split("?")[0]!;
+    const match = this.#wsProxies.find(
+      ({ route }) => path === route || path.startsWith(route.endsWith("/") ? route : `${route}/`)
+    );
+    if (!match) {
+      return false;
+    }
+    match.proxy.proxy.ws(req, socket, {}, head).catch((error) => {
+      this.nitro.logger.error(`Failed to proxy WebSocket upgrade for \`${path}\`:`, error);
+      socket.destroy();
+    });
+    return true;
   }
 }
 
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -123,6 +123,9 @@ export class NitroDevServer extends NitroDevApp implements RunnerRPCHooks {
   // #region Public Methods
 
   async upgrade(req: IncomingMessage, socket: Socket, head: any) {
+    if (this.proxyUpgrade(req, socket, head)) {
+      return;
+    }
     if (!this.#manager.upgrade) {
       throw new HTTPError({
         status: 501,
--- a/test/unit/dev-proxy-ws.test.ts
+++ b/test/unit/dev-proxy-ws.test.ts
@@ -1,0 +1,122 @@
+import { createServer, request, type Server } from "node:http";
+import type { AddressInfo, Socket } from "node:net";
+import type { Nitro } from "nitro/types";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { NitroDevApp } from "../../src/dev/app.ts";
+
+const sockets = new Set<Socket>();
+
+function listen(server: Server): Promise<number> {
+  server.on("connection", (socket) => {
+    sockets.add(socket);
+    socket.on("close", () => sockets.delete(socket));
+  });
+  return new Promise((resolve, reject) => {
+    server.on("error", reject);
+    server.listen(0, "127.0.0.1", () => resolve((server.address() as AddressInfo).port));
+  });
+}
+
+/** Minimal upgrade-aware target: replies `101` then echoes raw frames. */
+function createUpgradeTarget(upgrades: string[]): Server {
+  const server = createServer((_req, res) => res.end("http"));
+  server.on("upgrade", (req, socket) => {
+    upgrades.push(req.url!);
+    socket.write(
+      "HTTP/1.1 101 Switching Protocols\r\nUpgrade: websocket\r\nConnection: Upgrade\r\n\r\n"
+    );
+    socket.on("data", (chunk) => socket.write(chunk));
+  });
+  return server;
+}
+
+function upgradeRequest(port: number, path: string) {
+  return new Promise<{ status: number; echo: string }>((resolve, reject) => {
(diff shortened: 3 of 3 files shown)


pi0x:
## Triage notes

I rebased this branch onto latest `main` locally. It rebases cleanly, no conflicts, and the diff is unchanged. `pnpm typecheck` and `pnpm fmt` are clean, and `test/unit/dev-proxy-ws.test.ts` passes (3/3, ~0.3s).

I could not push the rebase from my environment, so nothing was pushed. Everything below is review feedback only.

### Overlap with other open PRs

Three open PRs touch dev-server WebSocket upgrades. They are **not** the same fix:

- **This PR (#4480)** — `nitro dev` (the standalone `NitroDevServer`). Routes an upgrade to a `devProxy` rule with `ws: true` before falling through to the worker.
- **#4376** — also `nitro dev`, but about *which transport* is used to reach the worker (Bun/Deno have no Node upgrade socket). Different problem, same file.
- **#4291** (draft) — the **`vite dev`** path (`src/build/vite/dev.ts`), narrowing the `vite-*` subprotocol skip. A separate dev server entirely.

They do not overlap logically, but two of them do collide in practice:

1. **#4376 would silently undo this PR.** It changes `listen()` so the Node `"upgrade"` listener is attached only when `features.websocket` is enabled, and on Bun/Deno it replaces the listener with a crossws proxy that never consults `devProxy`. `devProxy` + `ws: true` does not require Nitro's own websocket feature, so after #4376 this fix would stop working for users who only want proxying. Whichever lands second should keep the `devProxy` check on a path that runs regardless of `features.websocket`, and on all runtimes.
2. Textually, #4376 rewrites the same `upgrade()` / `listen()` block, so one of the two will need a manual rebase.

#4291 does not conflict with this PR.

### Review findings

**1. `vite dev` is not covered.** `nitro dev` uses `NitroDevServer`, but Vite users go through `configureViteDevServer` in `src/build/vite/dev.ts`. That path already uses the same `NitroDevApp` for HTTP (`ctx.devApp.fetch`), so `devProxy` HTTP works there, but its own `"upgrade"` listener never calls `proxyUpgrade`. It is also only installed when `features.websocket` is on. So on `vite dev`, `devProxy` + `ws: true` still hangs. I did not patch it here, because that would collide with #4291, which rewrites exactly that listener — but it is probably worth doing in one of the two PRs.

**2. WS matching is a prefix match, HTTP matching is exact.** `devProxy` routes are registered with `app.all(route, ...)`, and h3 matches that path exactly. I confirmed `/proxy/example/sub` does *not* match `/proxy/example` and falls through to the worker. This PR matches `path === route || path.startsWith(route + "/")` for upgrades. For the linked issue (#4269, MailDev socket.io) that is the behaviour the reporter wants, but it makes WebSocket proxying more permissive than HTTP proxying: `/maildev/socket.io/?transport=websocket` gets proxied, while the socket.io HTTP polling request on the same prefix still 404s. Worth deciding deliberately — either keep it and note the asymmetry, or fix the HTTP side too (`app.all(joinURL(route, "/**"), ...)`).

Good news: the PR does *not* strip the matched prefix from `req.url`, which is correct here — h3 does not strip it on the HTTP side either, so both sides stay consistent. (The workaround posted in #4269 stripped it, but that was based on v2/h3 v1 mount semantics.)

**3. Error logging may get noisy.** `httpxy`'s returned promise also rejects when the *client* socket errors after a successful upgrade (`res.on("error", reject)` inside `_createProxyFn`). A browser hard-closing a proxied socket gives `ECONNRESET`, which would print `Failed to proxy WebSocket upgrade ...` as an error even though nothing is wrong. Consider `logger.warn`/`debug`, or ignoring `ECONNRESET`/`EPIPE`.

**4. The test covers `NitroDevApp.proxyUpgrade` directly**, not the `NitroDevServer.upgrade()` wiring added in `src/dev/server.ts`. That is fine and fast, just noting the 3-line `server.ts` change is untested.

### Improvement I prepared (not pushed)

One commit documenting the option, since `ws` is what the issue reporter expected to work and it is not mentioned anywhere in `docs/`:

```md
// docs/3.config/0.index.md, `devProxy` section
"/proxy/socket": { target: "http://localhost:3002", ws: true },

Set `ws: true` on a route to also proxy WebSocket upgrade requests to the target.
```

Nothing else looked wrong. The core approach (keep the httpxy instance around per `ws` route and consult it before handing the socket to the worker) matches what the issue suggested and is the right shape.

PR #4481, opened 2026-07-25
fix(dev): proxy subpaths for `devProxy` rules

danielroe:
### ❓ Type of change

### 📚 Description

noticed this when trying to reproduce a bug on nuxt/cli:

on v2 a single `devProxy` rule proxies the whole prefix, but on v3 only the exact route (and its trailing-slash form) is proxied; every subpath falls through to the app.

```ts
 export default defineConfig({
   serverDir: './server',
   devProxy: {
     '/proxy': { target: 'http://localhost:8080' },
   },
 })
```

hitting `/proxy` works with v2 (target sees '/') and v3 (target sees '/proxy') but in v3 hitting `/proxy/foo/bar?x=1` doesn't work - it 404s...

this is an attempt to resolve but you may have a different approach - or maybe you want to move into routerules (?)

### 📝 Checklist

### DIFF
--- a/test/unit/dev-proxy.test.ts
+++ b/test/unit/dev-proxy.test.ts
@@ -1,0 +1,90 @@
+import { createServer, type Server } from "node:http";
+import type { AddressInfo } from "node:net";
+import { serve, type ServerHandler } from "srvx";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import type { Nitro } from "nitro/types";
+import { NitroDevApp } from "../../src/dev/app.ts";
+
+describe("dev proxy", () => {
+  let target: Server;
+  let devServer: ReturnType<typeof serve>;
+  let devURL: string;
+
+  beforeAll(async () => {
+    target = createServer((req, res) => {
+      res.setHeader("content-type", "application/json");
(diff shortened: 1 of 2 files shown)


test/unit/dev-proxy.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  it("should proxy the exact route", async () => {
+    expect(await req("/proxy")).toMatchObject({ status: 200, body: '{"url":"/"}' });
+  });
+
+  it("should proxy the trailing slash form", async () => {
+    expect(await req("/proxy/")).toMatchObject({ status: 200, body: '{"url":"/"}' });
+  });
+
+  it("should proxy nested subpaths with a query string", async () => {
+    expect(await req("/proxy/foo/bar?x=1")).toMatchObject({
```
pi0, line 65: I think this is a correct (breaking change) that `/proxy` matches `/proxy` while `/proxy/**` anything nested nitro v2 behavior was consequence of following connect middleware

danielroe: so we should tell people to set /proxy/** when they intend to proxy all subpaths?

PR #4488, opened 2026-07-28
fix(prerender): write files atomically to avoid torn output

### LINKED ISSUE #4487: Prerender: two routes resolving to the same output file are neither deduped nor written atomically
# Environment

- Nitro: 2.13.4
- Node: v24.11.1
- OS: Linux 6.8.0-136-generic, 16 cores
- Package manager: pnpm 10.30.0

Relevant config (full file in the reproduction):

```ts
preset: 'static',
prerender: {
  crawlLinks: false,
  routes: ['/other/index.html', '/other'],
  concurrency: cpus().length * 4,   // Nitro's default is 1; see below
},
```

---

# Reproduction

https://github.com/Eschricht/nitro-prerender-file-collision

```bash
pnpm install
node repro.mjs
```

`nitropack` is the only dependency — no framework is involved. (Noting this because
the template steers framework-related reports to `nuxt/nuxt`. There is a Nuxt-side
aspect, covered at the end, but the two defects below reproduce with Nitro alone.)

---

# Describe the bug

Two prerender routes can resolve to the same output file. Nitro does not detect
this — it renders both and writes both to that one path:

| route | resolved `fileName` | render size |
| --- | --- | --- |
| `/other/index.html` | `other/index.html` | 300 bytes |
| `/other` | `other/index.html` | 256 bytes |

`/other` resolves that way via `autoSubfolderIndex`, which defaults to `true`:

```ts
const htmlPath = route.endsWith("/") || nitro.options.prerender.autoSubfolderIndex
  ? joinURL(route, "index.html")
  : route + ".html";
```

`/other/index.html` ends with `.html`, so it is used verbatim. Both land on
`other/index.html`.

Deduplication is keyed on the route string, never on the resolved `fileName`
(`src/prerender/prerender.ts`):

```ts
const generatedRoutes = new Set();
if (generatedRoutes.has(route) || skippedRoutes.has(route)) return false;
generatedRoutes.add(route);
...
await writeFile(filePath, dataBuff!);
```

This produces two distinct problems.

### 1. One render is silently discarded, and which one is arbitrary

Both routes are rendered and both are logged as prerendered successfully. Only one
survives in the output, and which one depends on completion order, so the file's
content varies between otherwise identical builds. This is present at every
concurrency, including Nitro's default of `1`.

### 2. Above concurrency 1, the file can tear

`fs.writeFile` truncates at `open`, not at `write`. Both writers open the path
(each truncating), then each writes at offset 0. When the shorter write lands last,
the file keeps the longer writer's length and the longer writer's trailing bytes
survive:

```html
<!DOCTYPE html><html><head><link rel="canonical" href="https://example.com/other"><meta property="og:url" content="https://example.com/other"></head><body><h1>/other</h1><script type="application/json" id="__DATA__">{"path":"/other"}</script></body></html>:"/other/index.html"}</script></body></html>
```

That is the complete 256-byte `/other` render, followed by bytes 256..300 of the
300-byte `/other/index.html` render. The leak begins mid-JSON-string because it is
a byte offset, not a token boundary. A browser renders those 44 bytes as visible
text on the page.

Measured **3 of 14 builds torn** in the reproduction. The build exits 0 and nothing
warns, in either case.

---

# Additional context

### The write mechanism, independent of Nitro

```js
const a = await fsp.open('f', 'w')          // O_TRUNC at open
const b = await fsp.open('f', 'w')          // O_TRUNC at open
await a.write(long, 0, long.length, 0)      // 300 bytes
await b.write(short, 0, short.length, 0)    // 256 bytes at offset 0
// -> 300-byte file: short content + 44 stale bytes of long content
```

### Suggested fixes

- Dedup prerender routes by resolved output `fileName`. At minimum, warn when two
  routes target one file — that alone would make this self-diagnosing.
- Write via temp file + `rename`. Overlapping writers then produce last-one-wins
  rather than a torn file. Worth doing independently of the dedup change, since it
  removes the corruption for any future path that reaches the same state.

### Reproduction scaffolding, disclosed

Three knobs in the repro exist only to make an intermittent race observable in few
builds; none causes the bug, and the README documents each. Briefly: route order
(the longer render must be issued first for the shorter write to land last), a
rendezvous barrier (the two colliding renders must reach `writeFile` in the same
event-loop tick — real handlers do heavy variable work so their completions cluster
naturally, whereas this trivial handler's order otherwise collapses to route
insertion order), and filler routes for threadpool load.

For what it's worth, the same bug needed none of that scaffolding in a real app —
see below.

### Where this bites in practice

Frameworks generate the colliding pair without the user asking. Nuxt adds a
prerender route `/index.html` whenever `ssr: false`, on top of the page route `/`
— both resolve to `index.html` — and raises `prerender.concurrency` from Nitro's
`1` to `cpus().length * 4`.

In an ordinary Nuxt SPA, with no scaffolding of any kind, this tore roughly **1
build in 10**, silently, and shipped to a production static deploy. The leaked bytes
were a fragment of Nuxt's `__NUXT_DATA__` script cut mid-attribute, which the
browser rendered as visible text. SPA hosting that rewrites unmatched paths to
`/index.html` then shows it on every route lacking its own static file, which
presents as "random pages" rather than "one corrupt file" — it took a while to
recognise as a build problem at all.

There may be a Nuxt-side question too — whether `/index.html` should be enqueued at
all when `/` is already prerendered — but it seems secondary to me: if Nitro dedups
by resolved output path, the extra route becomes harmless duplicate work rather than
a source of corruption. The concurrency override itself is noted as an undocumented
divergence from Nitro's defaults in nuxt/nuxt#30067, though neither that issue nor
any other I could find covers this collision. Happy to file separately on
`nuxt/nuxt` if you'd prefer it handled there.

---

# Logs

```sh
# both routes are prerendered, both reported successful, both write other/index.html
[nitro] ℹ Prerendering 42 routes
[nitro]   ├─ /other/index.html (5ms)
[nitro]   ├─ /other (5ms)
[nitro] ℹ Prerendered 42 routes in 0.296 seconds
[nitro] ✔ Generated public .output/public

# repro.mjs inspecting .output/public/other/index.html across builds
run  1: 300 bytes, winner /other/index.html 1 __DATA__ tag(s), 0 trailing byte(s)
run  2: 256 bytes, winner /other            1 __DATA__ tag(s), 0 trailing byte(s)   # same input, different output
run  3: 300 bytes, winner /other            1 __DATA__ tag(s), 44 trailing byte(s) <-- TORN
```

sobol-sudo:
### 🔗 Linked issue

Partially resolves #4487 — this covers the "not written atomically" half. The dedup half is left out deliberately, see below.

### ❓ Type of change

### 📚 Description

Two prerender routes can resolve to the same output file (for example `/other` and `/other/index.html` via `autoSubfolderIndex`). With `prerender.concurrency > 1` their writes overlap, and because `writeFile` truncates and writes in place, the result can be a file torn between both payloads rather than either one of them.

The interleaving is: writer A truncates, writer B truncates, B writes 256 bytes, A writes 300 bytes — leaving a 300 byte file whose first 256 bytes came from B and whose last 44 came from A. That is exactly the shape reported in #4487, where it corrupted roughly 1 in 10 production static builds of a Nuxt SPA and shipped silently.

This writes to a sibling temp file and renames it into place. `rename` is atomic within a filesystem, and the temp file is created in the same directory as the destination, so overlapping writers now produce last-one-wins instead of a torn file. The temp file is removed if the write fails.

Reproducing the current behaviour locally, 200 pairs of concurrent writes of a 300 byte and a 256 byte payload to one path:

```
torn: 172/200
sizes seen: 300b x198, 256b x2
```

With this change, 0/200.

I left the dedup half of #4487 out on purpose. Which of two colliding routes should win is a design call, and deduping needs the resolved `fileName`, which is only known after the route has already been rendered — so it seemed better to leave that to you rather than guess. The atomic write stands on its own: it removes the corruption for any path that reaches this state, which is what the reporter suggested doing independently.

### 📝 Checklist

`test/unit/fs.test.ts` covers the regression along with the basic write behaviour. The concurrency test fails on `main` immediately (run 0) and passes with the change. Locally after `pnpm build`: `tsc --noEmit` clean, `test/unit` 178 passed / 2 skipped, `oxlint` and `oxfmt --check` clean.

### DIFF
--- a/src/utils/fs.ts
+++ b/src/utils/fs.ts
@@ -1,5 +1,5 @@
 import type { Nitro } from "nitro/types";
-import { stat, mkdir, writeFile as fspWriteFile } from "node:fs/promises";
+import { stat, mkdir, rename, rm, writeFile as fspWriteFile } from "node:fs/promises";
 import { dirname } from "pathe";
 import consola from "consola";
 import { colors } from "consola/utils";
@@ -47,9 +47,21 @@ function _compilePathTemplate(contents: string) {
     });
 }
 
+let _tmpFileCounter = 0;
+
 export async function writeFile(file: string, contents: Buffer | string, log = false) {
   await mkdir(dirname(file), { recursive: true });
-  await fspWriteFile(file, contents, typeof contents === "string" ? "utf8" : undefined);
+  // Write to a sibling temp file and rename it into place. `rename` is atomic within a
+  // filesystem, so two writers racing for the same path produce last-one-wins instead
+  // of a file torn between both payloads.
+  const tmpFile = `${file}.${process.pid}-${_tmpFileCounter++}.tmp`;
+  try {
+    await fspWriteFile(tmpFile, contents, typeof contents === "string" ? "utf8" : undefined);
+    await rename(tmpFile, file);
+  } catch (error) {
+    await rm(tmpFile, { force: true });
+    throw error;
+  }
   if (log) {
     consola.info("Generated", prettyPath(file));
   }
--- a/test/unit/fs.test.ts
+++ b/test/unit/fs.test.ts
@@ -1,0 +1,57 @@
+import { mkdtemp, readFile, readdir, rm, writeFile as fspWriteFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { join } from "pathe";
+import { writeFile } from "../../src/utils/fs.ts";
+
+describe("writeFile", () => {
+  let dir: string;
+
+  beforeEach(async () => {
+    dir = await mkdtemp(join(tmpdir(), "nitro-fs-"));
+  });
+
+  afterEach(async () => {
+    await rm(dir, { recursive: true, force: true });
+  });
+
+  it("writes contents and creates missing parent directories", async () => {
+    const file = join(dir, "nested/deeply/index.html");
+
+    await writeFile(file, "<h1>hello</h1>");
+
+    expect(await readFile(file, "utf8")).toBe("<h1>hello</h1>");
+  });
+
+  it("leaves no temporary files behind", async () => {
+    await writeFile(join(dir, "index.html"), "<h1>hello</h1>");
+
+    expect(await readdir(dir)).toEqual(["index.html"]);
+  });
+
+  it("overwrites an existing file", async () => {
+    const file = join(dir, "index.html");
+    await fspWriteFile(file, "old");
+
+    await writeFile(file, "new");
+
+    expect(await readFile(file, "utf8")).toBe("new");
+  });
+
+  it("never leaves a file torn between two concurrent writers", async () => {
+    // Two prerender routes can resolve to a single output file, and with
+    // `prerender.concurrency > 1` their writes overlap. A non-atomic write then
+    // leaves the longer payload's tail after the shorter payload's body.
+    const file = join(dir, "other/index.html");
+    const long = Buffer.from("L".repeat(300));
+    const short = Buffer.from("S".repeat(256));
+
+    for (let i = 0; i < 100; i++) {
+      await Promise.all([writeFile(file, long), writeFile(file, short)]);
+
+      const written = await readFile(file);
+      const isWhole = written.equals(long) || written.equals(short);
+      expect(isWhole, `torn after ${written.length} bytes on run ${i}`).toBe(true);
+    }
+  });
+});
--- a/test/unit/fs.test.ts
+++ b/test/unit/fs.test.ts
@@ -1,4 +1,4 @@
-import { mkdtemp, readFile, readdir, rm, writeFile as fspWriteFile } from "node:fs/promises";
+import { mkdir, mkdtemp, readFile, readdir, rm, writeFile as fspWriteFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { join } from "pathe";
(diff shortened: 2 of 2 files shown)


pi0x:
Thanks for this, and for the very clear write-up.

**Rebase**

I rebased the branch onto the latest `main` and force-pushed it. There were no conflicts. The diff is still exactly your two files.

**Review**

I checked the fix and it looks correct.

- The temp file is created next to the target file, so the `rename` stays inside one filesystem and is atomic.
- The temp name uses the pid plus a counter, so two writers in the same process, or in two processes, cannot pick the same name.
- The failed write is cleaned up and the error is re-thrown, so callers still see it.

I also confirmed the regression test really is a regression test. I put the old non-atomic `writeFile` back and ran `test/unit/fs.test.ts`: the concurrency test fails right away. With your change all tests pass.

**What I added**

One small commit on top: a test for the cleanup path. Your tests covered the happy path, but nothing covered the `rm` in the `catch` block. The new test makes the `rename` fail (the target path is an existing directory) and then checks that no `.tmp` file is left behind.

**Notes for the maintainers, not things you need to change**

1. There is a second, almost identical `writeFile` in `src/presets/_utils/fs.ts` that is still not atomic. It is not used by the prerenderer, and presets write their config files one at a time, so this fix does not need it. But the two copies now behave differently.
2. Replacing a file by `rename` gives the new file default permissions instead of keeping the old file's mode. For prerender output this does not matter.
3. If the build is killed in the middle of a write, a `*.tmp` file can be left in `publicDir`. That is a fair trade for not shipping a corrupted page, but it is worth knowing.

Leaving the dedup half out was the right call in my opinion, and your #4491 covers it.

PR #4490, opened 2026-07-28
fix(nitro): vercel function rules - preserve relative symlinks

AndrewBarba:
## Summary

- preserve relative symlink targets when Vercel `functionRules` copies the traced server function
- add a regression using Nitro’s existing multi-version nf3 fixture

## Context

nf3 emits relative links from package locations into `node_modules/.nf3`. Node’s `fs.cp` resolves those links by default, so copying `__server.func` into a route-specific function rewrites them to absolute paths in the temporary build directory. Once that directory is removed, the deployed function contains dangling dependency links.

Using `verbatimSymlinks: true` keeps the traced dependency graph self-contained in each copied function directory.

Related downstream report and workaround: https://github.com/vercel/eve/pull/1279

## Testing

- regression confirmed failing before the fix and passing afterward
- `pnpm fmt`
- `pnpm typecheck`
- `pnpm vitest run test/presets/vercel.test.ts` — 139 passed, 2 skipped, 2 todo
- full suite attempted with `TZ=UTC pnpm test`; the Rollup run reached 1,025 passing tests and stopped on an unrelated Vite HMR assertion receiving two identical `full-reload` events instead of one

### DIFF
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -642,6 +642,7 @@ async function createFunctionDirWithCustomConfig(
   await fsp.cp(serverDir, funcDir, {
     recursive: true,
     mode: constants.COPYFILE_FICLONE,
+    verbatimSymlinks: true,
     filter: (src) => basename(src) !== ".vc-config.json",
   });
   const mergedConfig = defu(overrides, baseFunctionConfig);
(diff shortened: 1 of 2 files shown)


RihanArfan (approved):

pi0 (approved):
Thanks and sorry for delay!

PR #4491, opened 2026-07-29
fix(prerender): warn and keep first output when two routes collide

sobol-sudo:
### 🔗 Linked issue

Resolves the dedup half of #4487. The other half, the non-atomic write, is #4488 — the two are independent and this branch is cut from `main`, not from that one.

### ❓ Type of change

### 📚 Description

Two prerender routes can resolve to a single output file — `/other` and `/other/index.html` both become `other/index.html` under `autoSubfolderIndex`. Nitro renders both and writes both to that one path, so the file that ships is whichever render happened to finish last, and nothing in the output says a route was discarded.

Nuxt produces exactly this pair without the user asking: it adds a prerender route `/index.html` alongside the page route `/` whenever `ssr: false`, and raises `prerender.concurrency` above 1.

The resolved `fileName` is only known after a route has been rendered, so this cannot be deduped when routes are queued. This tracks the file each route resolved to and, when a second route resolves to a file that is already taken, warns and skips the write instead of replacing the first route's output:

```
WARN  Routes `/other` and `/other/index.html` both prerender to `/other/index.html`. Keeping the output of `/other`.
```

The skipped route is marked `skip`, so it already renders as `(skipped)` in the prerender log through `formatPrerenderRoute`. The file is claimed synchronously before the write is awaited, so a concurrent render resolving to the same file sees it as taken.

Which of the two routes wins is still the one that finished first, and with `concurrency > 1` that is not deterministic. Making it deterministic would mean ordering the writes, which seemed like a bigger change than this warrants — the point here is that the build no longer silently discards a render, and the warning names both routes so the duplicate can be removed from the route list. Happy to go further if you would prefer the winner to be pinned to route order.

### 📝 Checklist

`test/prerender/collision.test.ts` builds a small fixture with both routes and asserts the warning fires once, that only one route claims the file, and that the file holds that route's render. It fails on `main` (no warning, both routes recorded as written) and passes with this change.

Locally: `test/unit` 174 passed / 2 skipped, `tsc --noEmit` clean, `oxlint` and `oxfmt --check` clean.

### DIFF
--- a/src/prerender/prerender.ts
+++ b/src/prerender/prerender.ts
@@ -111,9 +111,10 @@ export async function prerender(nitro: Nitro) {
   const failedRoutes = new Set<PrerenderRoute>();
   const skippedRoutes = new Set();
   const displayedLengthWarns = new Set();
-  // Resolved output file -> the route that produced it. Two routes can resolve to
-  // one file (`/other` and `/other/index.html` both become `other/index.html`),
-  // which is only known once each of them has been rendered.
+  // Output path -> the route that produced it. Two routes can resolve to one file
+  // (`/other` and `/other/index.html` both become `other/index.html`), which is only
+  // known once each of them has been rendered. Keyed by the resolved path rather
+  // than `fileName`, so aliases pointing at one file still collide.
   const routeByOutputFile = new Map<string, string>();
 
   const publicAssetBases: string[] = nitro.options.publicAssets
@@ -299,7 +300,7 @@ export async function prerender(nitro: Nitro) {
 
     // Write to the disk
     const filePath = join(nitro.options.output.publicDir, _route.fileName);
-    const writtenBy = routeByOutputFile.get(_route.fileName);
+    const writtenBy = routeByOutputFile.get(filePath);
     if (writtenBy !== undefined) {
       // Writing again would replace the first route's output with this one, so the
       // build result would depend on which of the two rendered first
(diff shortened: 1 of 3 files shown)


pi0x:
Thanks. This is the other half of #4487 and it pairs well with your #4488.

**Rebase**

I rebased the branch onto the latest `main` and force-pushed it. No conflicts.

**Review**

The approach looks right to me.

- The check runs after the `prerender:generate` hook, so a user who renames `fileName` in that hook still gets the correct result.
- The map is keyed by the joined output path, not by `fileName`, so two spellings of the same file still collide.
- The file is claimed before the `await`, so a second render finishing at the same time sees it as taken.
- `generatedRoutes` already stops the same route being rendered twice, so a route cannot collide with itself.
- The losing route is not pushed to `_prerenderedRoutes` and shows as skipped in the log. That is the same shape as the existing "cannot write" branch.

I ran your test against `main` and it fails there, and passes with the change. Good regression test.

**What I added**

One small commit on top. Your test only had colliding routes, so nothing proved the warning is not over-eager. I added a third route to the same build that resolves to its own file, and an assertion that it is still written. This keeps a future change from warning on routes that do not actually collide.

**Two things for the maintainers to decide**

1. This warning will fire on every Nuxt SPA prerender build, because Nuxt adds `/index.html` next to `/` when `ssr: false`. That is exactly the bug being surfaced, but it does mean a lot of users will suddenly see a warning for something they did not write themselves. It may be worth checking with the Nuxt side before this lands.
2. As you say, the winner is "whoever finished first", which is not stable when `concurrency > 1`. The warning makes it visible, which is the important part, but the output is still non-deterministic between builds.

One very small thing, not worth a change on its own: the collision branch is checked before `canWriteToDisk`, so a route that could never be written to disk anyway would still report a collision. In practice such a route would not produce the same file name, so I left it alone.

PR #4499, opened 2026-07-30
build: re-export `FastResponse` from `nitro` and srvx types from `nitro/types`

### LINKED ISSUE #4498: [Feature] Re-Export `srvx` similar to `h3`
### Describe the feature

Re-Export the root `srvx` path in nitro to make general utilities availables in nitro

### Additional information

schplitt:
closes #4498

I deemed the rest of the exports of srvx not necessary for usage in nitro so I only included the root package

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "./context": "./dist/runtime/context.mjs",
     "./database": "./dist/runtime/database.mjs",
     "./h3": "./lib/h3.mjs",
-    "./srvx": "./lib/srvx.mjs",
     "./meta": "./dist/runtime/meta.mjs",
     "./package.json": "./package.json",
     "./runtime-config": "./dist/runtime/runtime-config.mjs",
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "./context": "./dist/runtime/context.mjs",
     "./database": "./dist/runtime/database.mjs",
     "./h3": "./lib/h3.mjs",
+    "./srvx": "./lib/srvx.mjs",
     "./meta": "./dist/runtime/meta.mjs",
     "./package.json": "./package.json",
     "./runtime-config": "./dist/runtime/runtime-config.mjs",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "./context": "./dist/runtime/context.mjs",
     "./database": "./dist/runtime/database.mjs",
     "./h3": "./lib/h3.mjs",
+    "./srvx": "./lib/srvx.mjs",
```
pi0: Exporting all can be risky against breaking change surface (srvx 0.12 to 0.13 could drop some for example), instead we can simply export `FastResponse` from main `nitro` subpath

schplitt: Alright. I know I needed `FastURL` and `ServerRequest` aswell as  the types for those + `ServerRequestContext`, `ServerRuntimeContext` in the past

These can go under the `nitro` subpath aswell?

pi0: we have nitro/types subpath for types. Not sure about ServerRequestContext i _guess_ it needs explicit declaration for srvx to extend

schplitt: Updated the exports :)

src/types/srvx.ts
```diff
@@ -1 +1,7 @@
-export type { ServerRequest } from "srvx";
+export type {
+  FastResponse,
+  FastURL,
```
pi0: i don't think we need to export these types. FastURL type _is_ URL same for FastResponse which is also already exported from main

```suggestion
```

schplitt: Yea right! Thanks for being patient

src/runtime/nitro.ts
```diff
@@ -23,6 +23,9 @@ export {
 } from "h3";
 export type { H3Event, EventHandlerRequest, EventHandlerWithFetch } from "h3";
 
+// srvx
+export type { FastResponse, FastURL } from "srvx";
```
pi0:
```suggestion
export type { FastResponse } from "srvx";
```

pi0 (approved):

PR #4506, opened 2026-08-04
fix(dev): await worker shutdown before replacing it

### LINKED ISSUE #2735: Close Hook not awaited

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/32928

related: #4088
partly resolve https://github.com/nitrojs/nitro/issues/2735

### ❓ Type of change

### 📚 Description

we currently don't await when shutting down a worker, meaning tasks or plugins shutdown can overlap with new startup (e.g. database connection). we've implemented waiting (+ timeout) in nuxt/cli for the nuxt dev server (https://github.com/nuxt/cli/pull/1424) but I thought we should probably do the same for nitro workers.

let me know if you don't think this is desired behaviour, or if you want to move it to `env-runner` instead 🙏 

### 📝 Checklist

### DIFF
--- a/src/dev/server.ts
+++ b/src/dev/server.ts
@@ -37,6 +37,7 @@ export class NitroDevServer extends NitroDevApp implements RunnerRPCHooks {
   #buildError?: unknown;
   #reloadPromise?: Promise<void>;
   #shuttingDown: boolean = false;
+  #closing: boolean = false;
 
   constructor(nitro: Nitro) {
     super(nitro, async (event) => {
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks, lets do it here as hotfix, might make sense to move to env-runner indeed!

PR #4393, opened 2026-06-29
fix(build): escape dynamic route segments in Rollup chunk names

enlorik:
### 🔗 Linked issue

Fixes #4289

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

`getChunkName` returns `_routes/<path>.mjs` for server route handlers, deriving the path from `routeToFsPath`. Routes with dynamic segments (e.g. `/api/applications/:id/context`) produce paths containing square brackets (`[id]`), which Rollup rejects as invalid `chunkFileNames` placeholders with `"[id]" is not a valid placeholder`.

Applied `.replace(/\[([^\]]*)\]/g, "_$1_")` to the route-derived path inside `getChunkName` before building the chunk name. `routeToFsPath` is unchanged.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #4508, opened 2026-08-05
fix(rollup): escape dynamic route segments in chunk names

RezaRahemtola:
### ❓ Type of change

### 📚 Description

`getChunkName` returns `_routes/<path>.mjs` for route handlers, and `routeToFsPath` converts dynamic segments like `:id` to `[id]`. Rollup rejects `[id]` as an invalid `chunkFileNames` placeholder, so building with the Rollup builder fails on nested dynamic routes like `server/api/applications/[id]/context.get.ts`.

It looks like Rolldown (the default builder) tolerates the brackets, which is why this slipped through.

This PR encodes brackets (`[` and `]`) as `%5B`/`%5D` in the chunk name so it's a valid `chunkFileNames` pattern.
This also avoids any collision with literal route segments named `_id_` (the concern that closed the similar PR I found #4393).
Also updated the unit test and added a nested dynamic route case.

### 📝 Checklist

### DIFF
--- a/test/unit/chunks.test.ts
+++ b/test/unit/chunks.test.ts
@@ -149,8 +149,12 @@ describe("getChunkName", () => {
     );
   });
 
-  it("returns _routes/<path>.mjs for dynamic route", () => {
+  it.each([
+    ["rolldown", "_routes/api/users/[id].mjs"],
+    ["rollup", "_routes/api/users/%5Bid%5D.mjs"],
+  ])("returns _routes/<path>.mjs for dynamic route (%s)", (builder, expected) => {
     const n = createNitro({
+      options: { builder },
       routing: {
         routes: {
           routes: [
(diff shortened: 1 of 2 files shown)


test/unit/chunks.test.ts
```diff
@@ -162,10 +162,32 @@ describe("getChunkName", () => {
       },
     } as any);
     expect(getChunkName(createChunk("route", ["/src/routes/api/users/[id].ts"]), n)).toBe(
-      "_routes/api/users/[id].mjs"
+      "_routes/api/users/%5Bid%5D.mjs"
```
pi0: I guess we could limit this behavior to rolldown? this seems a rollup limit, [id] in filesystem is valid and much more readable. (not sure also if for rollup we could somehow make it working...)

RezaRahemtola: Indeed, limited the encoding to the rollup builder in the latest commit [fix(build): encode dynamic route chunks for rollup only](https://github.com/nitrojs/nitro/pull/4508/commits/2693f7b3e6bab160bcc93eaf6d243f044b71f20e)

pi0 (approved):

PR #4509, opened 2026-08-07
fix(vite): skip server bundle for static presets

### LINKED ISSUE #3843: Static preset doesn't work [Nitro 3.0.1-alpha.1]
### Environment

Nitro: 3.0.1-alpha.1
Vite: 7.2.6

### Reproduction

I am using this example from nitro repo with only one change - adding the `static` preset. 

Example: https://github.com/nitrojs/nitro/tree/main/examples/vite-ssr-vue-router
Change:
```
//
import vue from "@vitejs/plugin-vue";
import { defineConfig } from "vite";
import devtoolsJson from "vite-plugin-devtools-json";
import { nitro } from "nitro/vite";

export default defineConfig((_env) => ({
    plugins: [patchVueExclude(vue(), /\?assets/), devtoolsJson(), nitro()],
    environments: {
        client: {
            build: { rollupOptions: { input: "./app/entry-client.ts" } },
        },
        ssr: { build: { rollupOptions: { input: "./app/entry-server.ts" } } },
    },
    nitro: {
        preset: "static",
        prerender: {
            routes: ["/", "about"],
        },
    },
}));

// Workaround https://github.com/vitejs/vite-plugin-vue/issues/677
function patchVueExclude(plugin, exclude) {
    const original = plugin.transform.handler;
    plugin.transform.handler = function (...args) {
        if (exclude.test(args[1])) return;
        return original.call(this, ...args);
    };
    return plugin;
}
```

### Describe the bug

Building with non-static preset such as default, 'vercel', 'cloudflare_pages' succeed.
Building with static preset such as 'static', 'vercel-static, 'cloudflare_pages_static' fail with this error message:

```
error during build:
rollupOptions.input should not be an html file when building for SSR. Please specify a dedicated SSR entry.
    at resolveRollupOptions (file:///home/syahzuan/projects/personal/vue-ssg/nitro-v3-vue-ssr-example/node_modules/vite/dist/node/chunks/config.js:33451:73)
    at buildEnvironment (file:///home/syahzuan/projects/personal/vue-ssg/nitro-v3-vue-ssr-example/node_modules/vite/dist/node/chunks/config.js:33506:25)
    at Object.build (file:///home/syahzuan/projects/personal/vue-ssg/nitro-v3-vue-ssr-example/node_modules/vite/dist/node/chunks/config.js:33899:25)
    at buildEnvironments (file:///home/syahzuan/projects/personal/vue-ssg/nitro-v3-vue-ssr-example/node_modules/nitro/dist/_build/vite.plugin.mjs:186:16)
    at async Object.buildApp (file:///home/syahzuan/projects/personal/vue-ssg/nitro-v3-vue-ssr-example/node_modules/vite/dist/node/chunks/config.js:33893:5)
    at async CAC.<anonymous> (file:///home/syahzuan/projects/personal/vue-ssg/nitro-v3-vue-ssr-example/node_modules/vite/dist/node/cli.js:629:3)
error: script "build" exited with code 1
``` 

### Additional context

I am a beginner learning to integrate Nitro at a lower level without framework. Apology in advance if I am missing something.

### Logs

```sh

```

birkskyum:
### 🔗 Linked issue

- Resolves #3843

### ❓ Type of change

### 📚 Description

`buildEnvironments` builds the `nitro` environment unconditionally. Static presets define no entry, so the environment has no input, Vite falls back to resolving `index.html`, and rejects it for a server consumer:

```
rolldownOptions.input should not be an html file when building for SSR. Please specify a dedicated SSR entry.
```

This affects every static preset (`static`, `github-pages`, `gitlab-pages`, `vercel-static`, `cloudflare-pages-static`) on the Vite builder. It fires *after* prerendering, so `.output/public` is already complete and correct when the build dies.

Reproduced on current `main` with the reporter's repro from #3843 (the `vite-ssr-vue-router` example plus `nitro: { preset: "static", prerender: { routes: ["/", "/about"] } }`): fails before this change, builds clean afterwards with both routes prerendered and no server bundle.

#### Why this fix

The rolldown and rollup builders already guard the server build with `if (!nitro.options.static)` (`src/build/rolldown/prod.ts`, `src/build/rollup/prod.ts`); this brings the Vite builder in line. `writeBuildInfo` already accepts `output: ... | undefined`, so no signature change was needed.

The environment is also marked as built. Vite's default `builder.buildApp` is a no-op, and `buildApp()` treats "no environment was built" as "no plugin handled the build", falling back to building all of them. Without that line, a static build with no other environments hits the same error again, from Vite's own frame.

Two alternatives were tried and rejected:

- **Omit the `nitro` environment when static.** Redirects the fallback at `client`, which also has no input, giving `UNRESOLVED_ENTRY: Cannot resolve entry module index.html`. It would also need a `!options.dev` guard, since `dev.ts` reads `server.environments.nitro`.
- **Fix it in Vite instead.** Arguably the better long-term shape, since the fallback could skip environments with no resolvable input, or key off whether a `buildApp` hook was registered rather than off mutated state. But every static preset is broken today, and this guard is correct regardless of what Vite does.

#### Tests

Adds `test/vite/static-preset.test.ts` with a minimal fixture, asserting the build succeeds, no server bundle is emitted, and routes are prerendered to the public dir. It fails before this change with the error above.

Reported downstream by SolidStart users building SPAs (`solidStart({ ssr: false })` + `preset: "static"`), where this is currently the only blocker to a server-free static deploy.

### 📝 Checklist


pi0 (approved):
Thanks ❤️

Issue #4518, opened 2026-08-08
Import attributes don't survive the rollup builder: emitted with the removed `assert` key, and dropped in `.ts` sources on v2


agantelin:
### Environment

- Nitro `2.13.4` (v2, rollup) and `3.0.260610-beta` (main, `builder: "rollup"`)
- Node.js v24.19.0 — the failure applies to every Node.js ≥ 22
- Windows 11, npm

### Reproduction

https://github.com/agantelin/nitro-import-attributes-repro — four self-contained cases, each with
its own `npm install` and the exact commands in the README.

### Describe the bug

Two independent defects. Both are quiet: the build is green, the server starts, and only the route
that touches the module answers 500.

#### 1. `output.importAttributesKey` is never set

Rollup 4 re-prints the import attributes of external modules with its default key, `assert`:

```js
// source
() => import("@iconify-json/mdi/icons.json", { with: { type: "json" } })
// .output/server/chunks/routes/index.mjs
() => import('@iconify-json/mdi/icons.json', { assert: { type: 'json' } })
```

`assert` was removed from Node.js in **22.0.0** (nodejs/node#52104, "esm: drop support for import
assertions"). In the dynamic form the emitted code isn't a syntax error, it just no longer carries an
attributes bag, so Node refuses the module:

```
TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module "…/data.json" needs an import attribute of "type: json"
→ HTTP 500
```

Anything whose attribute is load-bearing (`type: "json"`, wasm) is unloadable this way; attributes a
runtime ignores are of course harmless.

`git grep importAttributesKey` is empty on both branches. On `main` this affects the rollup builder
and the rollup branch of `src/build/vite/bundler.ts` — the default builder is rolldown, which emits
`with` correctly, so on v3 you have to opt into rollup to hit it. On v2 rollup is the only builder.

Rollup's default flips in v5 (rollup/rollup#6248, merged to the `rollup-5` branch on 2026-01-30;
the user-facing report is rollup/rollup#5685, open since 2024-09), and `output.importAttributesKey`
is the documented workaround until then. No 5.x is published.

On the risk of setting it: the only runtimes that take `assert` but reject `with` are Node.js
16.14–18.19 and 20.0–20.9 (`with` landed in 21.0.0 / 20.10.0 and was backported to 18.20.0) — all
below `engines.node` (`^20.19.0 || >=22.12.0`), so the only person who could notice is someone
[...]

PR #4519, opened 2026-08-09
fix(rollup): keep import attributes in the build output

agantelin:
### 🔗 Linked issue

#4518

### ❓ Type of change

### 📚 Description

Two defaults in `getRollupConfig`, both still overridable by user config:

- `output.importAttributesKey: "with"`. Rollup 4 re-prints the import attributes of external
  modules with its default key, `assert`, which Node.js removed in 22.0.0 (nodejs/node#52104).
  Rollup flips this itself in v5 (rollup/rollup#6248, merged to `rollup-5`) and documents the option
  as the workaround until then. It goes in the `defu` defaults, so `rollupConfig.output` still wins.
  Nothing on the supported matrix loses: the runtimes that take `assert` but reject `with` are all
  below `engines.node`.
- `supported: { "import-attributes": true }` on the esbuild plugin. esbuild silently drops import
  attributes for every `esXXXX` target (checked up to `es2025`) and the target here is `es2019`, so
  attributes written in `.ts`/`.js` never reach Rollup at all. Merged with the user's own
  `esbuild.options.supported` rather than spread past it, so setting an unrelated feature flag
  doesn't quietly drop this one.

Without these, a module that marks a JSON import external produces code no supported runtime can
load, while the build stays green: pages render and only the affected route answers 500.
`@nuxt/icon`'s `serverBundle.externalizeIconsJson` is the case I hit — it generates a correct
`with { type: "json" }` import and has no way to know it will be rewritten.

### 📝 Checklist

- [ ] I have updated the documentation accordingly. — neither the `es2019` target nor
      `importAttributesKey` is documented today; this changes no documented behaviour.

### Verification

`test/unit/import-attributes.test.ts` covers three things, each checked by breaking the
corresponding line one at a time: the emitted attribute key, an explicit `rollupConfig.output`
setting surviving the merge, and a `.ts` source keeping its attributes through the esbuild
transform. `pnpm vitest run test/unit`, `tsc --noEmit`, eslint and `prettier -c` are green.

A `test/fixture/` regression route wouldn't be portable — the case needs an *external* JSON import,
which the worker and edge presets can't load at all — so the regression is pinned at the config
level instead. Happy to move it if you'd rather have it elsewhere.

End-to-end, including the runtime 500 before and 200 after:
https://github.com/agantelin/nitro-import-attributes-repro (cases 2 and 3)

### DIFF
--- a/src/rollup/config.ts
+++ b/src/rollup/config.ts
@@ -137,6 +137,8 @@ export const getRollupConfig = (nitro: Nitro): RollupConfig => {
       },
       inlineDynamicImports: nitro.options.inlineDynamicImports,
       format: "esm",
+      // `assert` was removed in Node.js v22; Rollup 5 defaults to `with` (rollup/rollup#6248).
+      importAttributesKey: "with",
       exports: "auto",
       intro: "",
       outro: "",
@@ -294,6 +296,11 @@ export const getRollupConfig = (nitro: Nitro): RollupConfig => {
       target: "es2019",
       sourceMap: nitro.options.sourceMap,
       ...nitro.options.esbuild?.options,
+      supported: {
+        // esbuild drops import attributes for every `esXXXX` target.
+        "import-attributes": true,
+        ...nitro.options.esbuild?.options?.supported,
+      },
     })
   );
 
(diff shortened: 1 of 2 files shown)


agantelin:
Pushed an update (force-push, same two source lines): added a test that an explicit
`rollupConfig.output.importAttributesKey` from user config survives the merge, and trimmed the
rationale comments to one line each. `test/unit` is 44 green, `tsc --noEmit` and `prettier -c` clean.

PR #4520, opened 2026-08-09
fix(rollup): emit import attributes with the `with` key

agantelin:
### 🔗 Linked issue

#4518

### ❓ Type of change

### 📚 Description

Rollup 4 re-prints the import attributes of external modules with its default key, `assert`, which
Node.js removed in 22.0.0 (nodejs/node#52104). Rollup itself flips the default in v5
(rollup/rollup#6248, merged to `rollup-5`), and documents `output.importAttributesKey` as the
workaround until then; no 5.x is published.

This only bites the rollup paths — the default rolldown builder emits `with` already — so it's
`builder: "rollup"` and the rollup branch of the vite builder:

- `src/build/rollup/config.ts` — the default in the builder's output config; it sits in the object
  `defu` merges *under* `nitro.options.rollupConfig`, so user config still wins.
- `src/build/vite/bundler.ts` — same default, applied with `??=` after the merge, because in that
  branch `commonConfig` is merged last and a literal would outrank user config.

Symptom: the build is green, the server starts, and the route that imports the module answers 500
with `TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]`. I hit it through `@nuxt/icon`'s
`serverBundle.externalizeIconsJson`, which generates a correct `with { type: "json" }` import.
Setting the key costs nothing on the supported matrix: the runtimes that take `assert` but reject
`with` are all below `engines.node`.

### 📝 Checklist

- [ ] I have updated the documentation accordingly. — `importAttributesKey` isn't documented; after
      this the assets docs from #4431 describe what actually ships on the rollup path too.

### Verification

- `test/unit/import-attributes.test.ts` covers three things, each checked by reverting the
  corresponding line one at a time: the rollup builder's default, the vite builder's rollup branch,
  and that an explicit `rollupConfig.output.importAttributesKey` from user config survives both.
- End-to-end on `nitro@3.0.260610-beta` with `builder: "rollup"`: `assert` → HTTP 500
  `ERR_IMPORT_ATTRIBUTE_MISSING`, `with` → 200. Case 4 in
  https://github.com/agantelin/nitro-import-attributes-repro
- A `test/fixture/` regression route can't express this: the case needs an *external* JSON import,
  which the worker and edge presets can't load, so the fixture suite would fail for other reasons.
- What I could not check end to end is a real vite *build* — `builder: "vite"` doesn't get that far
  on Windows for me (`ERR_UNSUPPORTED_ESM_URL_SCHEME … Received protocol 'c:'`, unrelated to this
  change, but you may want to know). The emitted config is pinned by the test above.


agantelin:
Pushed an update (force-push, same two source lines):

- The test now also covers the vite builder's rollup branch and the user-override contract, so both
  changed lines are pinned rather than just the rollup one. Each of the three assertions was checked
  by reverting its line individually.
- Trimmed the rationale comments to one line each.

Nothing else changed. Still unverified on my side: an actual vite build, which doesn't start on
Windows (`ERR_UNSUPPORTED_ESM_URL_SCHEME`, unrelated to this PR).

pi0 (approved):
Thanks!

PR #4522, opened 2026-08-09
fix(node): call `close` hooks on server shutdown

### LINKED ISSUE #4502: `node_server` preset does not call the `close` hook on shutdown (regression from v2)
### Environment

nitro 3.0.260610-beta, node 24/26, preset `node_server`, srvx 0.11.22

### Reproduction

Plugin with `app.hooks.hook('close', () => console.log('CLOSE HOOK RAN'))`, then
`nitro build` + `node .output/server/index.mjs` and send SIGTERM. Hook never runs.
Under `nitro dev` + Ctrl+C it does run.

### Describe the bug

The `close` hook never fires on SIGTERM/SIGINT with the `node_server` preset, so plugin
`close` handlers (flush, checkpoint, close resources) are silently skipped in production.
Works in `nitro dev`, which hides it. Worked in v2.

v2 `node-server.ts` called `setupGracefulShutdown(listener, nitroApp)`
(`runtime/internal/shutdown.ts`), whose `onShutdown` ran `callHook('close')`. v3
`node-server.ts` just calls srvx `serve()`. Shutdown is now srvx's `gracefulShutdownPlugin`,
which only does `server.close()` and never touches Nitro hooks. The only v3 callers of
`callHook('close')` are the dev preset and prerenderer, not node_server.

tarikermis:
### 🔗 Linked issue

Resolves #4502

### ❓ Type of change

### 📚 Description

The `node_server` preset no longer called Nitro's runtime `close` hook on SIGTERM or SIGINT, so cleanup handlers could be skipped in production. Nitro v2 used `setupGracefulShutdown(listener, nitroApp)` and ran the hook after connections drained; v3 delegates shutdown to srvx, which closes the server without calling Nitro hooks.

This wraps `server.close()` in the node server and cluster entries so the runtime hook runs once, is awaited, and still runs if the underlying close rejects. It runs after the server closes, matching the v2 order. Bun and Deno are unchanged because they did not call the hook in v2 either.

The issue reproduction now works as expected: build the `node_server` fixture, start it, send SIGTERM, and the close-hook marker appears before shutdown completes.

### Verification

- Added a regression test that spawns the built server, sends SIGTERM, and checks that the hook ran without an unhandled rejection.
- `pnpm vitest run test/presets/node.test.ts`: 66 passed with both rolldown and rollup.
- Manual built-server SIGTERM check passes.
- Lint, format, and typecheck pass.

The hook shares srvx's normal shutdown budget. The fixture scheduler can keep the process alive after the server closes, so the test kills it after the assertion; that behavior already existed. `NITRO_SHUTDOWN_*` remains out of scope.

### 📝 Checklist

- [ ] I have updated the documentation accordingly. (No docs change: this restores the documented close-hook behavior.)

Best Regards, Tarik


pi0:
Thanks for PR ❤️ Moving to #4574

PR #4532, opened 2026-08-16
fix(bun): call `close` hooks on server shutdown

### LINKED ISSUE #4479: Nitro close hook is not called on shutdown with the Vite integration and Bun preset
### Environment

- Nitro: `3.0.260610-beta`
- Runtime preset: `bun`
- Integration: `nitro/vite`
- Runtime: Bun

### Description

When using Nitro through the Vite integration with the Bun preset, callbacks registered on the Nitro runtime `close` hook are not called when the production server shuts down.

The srvx server handles `SIGINT` and prints `Server closed successfully.`, but the Nitro `close` hook is never invoked.

### Reproduction

Vite configuration:

```ts
import { nitro } from 'nitro/vite'

export default {
  plugins: [
    nitro({
      preset: 'bun',
      plugins: ['src/plugin.ts']
    })
  ]
}
```
Nitro runtime plugin:
```ts
import { definePlugin } from 'nitro'

export default definePlugin((app) => {
  app.hooks.hook('close', async () => {
    console.log('Nitro close hook called')
    // Dispose application resources
  })
})
```
Build and start the server:
```
vite build
PORT=3330 bun .output/server/index.mjs
```
Then press Ctrl+C.
Observed output:
```
Server closed successfully.
error: interrupted by SIGINT
```
`Nitro close hook called` is never printed.

### Describe the bug

The runtime plugin is correctly included in .output/server/index.mjs and successfully registers the callback.  

However, the generated Bun entry passes nitroApp.fetch to serve(), while srvx handles graceful shutdown independently.  

The srvx shutdown does not appear to be connected to the Nitro runtime close hook.

### Additional context

_No response_

### Logs

```sh

```

hamodywe:
Resolves #4479.

## The problem

The `bun` preset delegates shutdown to srvx (`gracefulShutdown`, default on), which closes the server on `SIGINT`/`SIGTERM` but has no callback into Nitro's runtime `close` hook. So a plugin that registers `nitroApp.hooks.hook("close", ...)` for resource cleanup is silently skipped when a production Bun server shuts down — the server prints "closed successfully" and the hook never runs.

This is the same root cause as #4502 (the `node_server` preset), fixed for node in #4522: Nitro v2 ran the hook after connections drained via `setupGracefulShutdown`; v3 hands shutdown to srvx, which closes the server without touching Nitro hooks.

## The change

Wrap `server.close()` so it runs the `close` hooks after the underlying close, using the exact pattern from #4522 (`useNitroHooks().callHook("close")`, guarded so it fires once, errors logged not thrown). srvx calls this `close()` on its signal handlers, so the hook now runs on `SIGINT`/`SIGTERM`.

## Scope

- This PR covers the **`bun`** preset (issue #4479); #4522 covers `node_server` / `node_cluster`.
- **`deno_server` has the identical gap** and is not covered here — happy to add it to this PR, or it can follow separately.
- `test/fixture/server/plugins/close.ts` is the same fixture #4522 adds; if these land in either order the identical file merges cleanly. If you'd prefer, I can rebase to depend on #4522 instead of including it.

## Tests

Added a `close`-hook test to `test/presets/bun.test.ts` (gated on `runIf(hasBun)`, skipped on Windows like the other preset shutdown tests): it builds the fixture, starts `.output/server/index.mjs` under Bun, sends `SIGTERM`, and asserts the fixture's `close`-hook marker is printed and no `unhandledRejection` occurs. It mirrors the node test in #4522.

### DIFF
--- a/src/presets/bun/runtime/bun.ts
+++ b/src/presets/bun/runtime/bun.ts
@@ -3,7 +3,7 @@ import type { ServerRequest } from "srvx";
 import { serve } from "srvx/bun";
 import wsAdapter from "crossws/adapters/bun";
 
-import { useNitroApp } from "nitro/app";
+import { useNitroApp, useNitroHooks } from "nitro/app";
 import { startScheduleRunner } from "#nitro/runtime/task";
 import { trapUnhandledErrors } from "#nitro/runtime/error/hooks";
 import { resolveWebsocketHooks } from "#nitro/runtime/app";
@@ -40,6 +40,22 @@ const server = serve({
   },
   plugins: [...tracingSrvxPlugins],
 });
+
+// Run `close` hooks on server shutdown (srvx closes the server on `SIGINT`/`SIGTERM`)
+const closeServer = server.close.bind(server);
+let closeHooksCalled = false;
+server.close = async (closeActiveConnections?: boolean) => {
+  try {
+    await closeServer(closeActiveConnections);
+  } finally {
+    if (!closeHooksCalled) {
+      closeHooksCalled = true;
+      await useNitroHooks()
+        .callHook("close")
+        ?.catch((error) => console.error("[close]", error));
+    }
+  }
+};
 
 trapUnhandledErrors();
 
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for PR ❤️ Moving to #4574

PR #4549, opened 2026-08-22
fix(dev): resolve public assets dynamically in the worker

### LINKED ISSUE #4500: Server side  fetch  during dev and production  runs different
### Environment

nodejs 

### Reproduction

https://stackblitz.com/edit/github-yjfey7sb?file=server%2Froutes%2Findex.ts

### Describe the bug

During development the  the server side fetch can not resolve files in the public directory. 

Once the app is build the server side fetch can resolve the public files and everything works as expected 

### Additional context

This was found in nuxt: https://github.com/nuxt/nuxt/issues/13857

### Logs

```sh

```

meta-syntax:
### 🔗 Linked issue
Resolves #4500

### ❓ Type of change

### 📚 Description

In dev, fetching a public asset from inside the server (e.g. `fetch("/some-asset.txt")` in an event handler, or `serverFetch`) returns 404, while the same URL works from the browser.

**Root cause:**
the `#nitro/virtual/public-assets-data` template globs `output.publicDir` at build time, but in dev nothing is ever copied there (`copyPublicAssets` only runs for `nitro build`).
So the asset manifest baked into the dev worker bundle is always empty, and the static handler responds 404.
External requests are unaffected because the dev server process serves static dirs itself, before proxying to the worker — internal fetch never goes through that path.

**Fix:**
in dev, the `#nitro/virtual/public-assets` template no longer imports the (empty) build-time manifest.
Instead it embeds the configured `publicAssets` source directories and resolves assets per request with a `statSync` lookup, so the worker sees the same files as the dev server process — including files added or removed after startup, without a rebuild.

Notes:

- No `etag` is generated in dev, matching the dev server's own static handling (`createServeStaticDirHandler`), which relies on `mtime`/`size` only.
- `mime` is used by the generated dev runtime code, so it is added to `runtimeDependencies` and `tracePkgs` (it was already a dev dependency, used by the dev server).
- Production behavior is unchanged.

Tests:
added a fixture route that fetches a public asset internally.
It is covered by the shared `serveStatic` suite for prod presets, and by a new dev context with `serveStatic: true` (the default dev test context disables `serveStatic`, so the bug was not observable there).
Dev test contexts now listen on a random port so two contexts can coexist.

### 📝 Checklist

### DIFF
--- a/build.config.ts
+++ b/build.config.ts
@@ -31,6 +31,7 @@ const tracePkgs = [
   "defu", // used by open-api runtime
   "destr", // used by node-server and deno-server
   "get-port-please", // used by dev server
+  "mime", // used by dev public assets runtime
   "rendu", // used by HTML renderer template
   "scule", // used by runtime config
   "source-map", // used by dev error runtime
--- a/src/build/virtual/public-assets.ts
+++ b/src/build/virtual/public-assets.ts
@@ -95,9 +95,6 @@ export default function publicAssets(nitro: Nitro) {
         );
 
         if (nitro.options.dev) {
-          // Assets are resolved dynamically in dev: the build-time snapshot in
-          // `public-assets-data` is always empty (nothing is copied to
-          // `output.publicDir` in dev) and would not follow file changes anyway.
           const publicAssetDirs = nitro.options.publicAssets.map((dir) => ({
             baseURL: withTrailingSlash(joinURL(nitro.options.baseURL, dir.baseURL || "/")),
             dir: dir.dir,
@@ -105,7 +102,7 @@ export default function publicAssets(nitro: Nitro) {
 
           return /* js */ `
 import { statSync, promises as fsp } from 'node:fs'
-import { resolve } from 'node:path'
+import { resolve, relative, isAbsolute, sep } from 'node:path'
 import mime from 'mime'
 
 const publicAssetDirs = ${JSON.stringify(publicAssetDirs)}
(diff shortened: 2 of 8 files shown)


pi0x:
- Rebased onto main. Clean, no conflicts. `pnpm fmt` and `pnpm typecheck` pass. I could not push from my environment, so nothing was pushed.
- Your test is a real regression test: it fails on main (404) and passes with the fix. Nice.
- `test/presets/nitro-dev.test.ts`, `test/presets/node.test.ts` and all of `test/vite` pass.
- One real problem: the dev template imports `node:fs` and calls `statSync`. The `cloudflare-dev` preset runs the dev worker in miniflare (workerd), where that module will not load. There is no CI test for that runner, so it would fail silently for Cloudflare users. Please gate the new branch on a node-capable dev runner and keep the old manifest path otherwise.
- Smaller note: the `.gz`/`.br`/`.zst` handling in the dev branch is dead code, since nothing pre-compresses assets in dev.

meta-syntax:
Thanks for the review.

Pushed two changes:
- The `node:fs` lookup is now used only when the dev runner is not `miniflare` (resolved as in `dev/server.ts`: `devServer.runner` → `NITRO_DEV_RUNNER` → `node-worker`).
  For `miniflare`, the build-time manifest path is kept as before. Added unit tests for both paths in `test/unit/public-assets.test.ts`.
- Removed the `.gz`/`.br`/`.zst` handling from the dev branch.

Also made the dev reader treat `ENOENT` as a missing asset, mirroring the node reader.

Rebased onto main. `nitro-dev`, `node` and `test/vite` pass locally.

PR #4551, opened 2026-08-23
fix(routing): encode non-ASCII characters in route patterns before registering

Vincentdevreede:
### 🔗 Linked issue

_No existing issue found_

### ❓ Type of change

### 📚 Description

`event.url.pathname` is always encoded by the URL parser, no matter how the request was written. Nitro's router compares that pathname against the literal route pattern it was given, so a route registered with literal Unicode text, for example a file named `について.get.ts` under `server/routes`, never matches on a live server, and gets silently skipped during static prerendering too.
h3 already solves this in its own route registration through `normalizeRoute`, which percent-encodes non-ASCII characters before handing the pattern to rou3. Nitro's router calls rou3 directly and bypasses that step, which is where this gap comes from.

This PR adds `encodeNonAsciiRoute` in `src/utils/route.ts` and applies it in `Router._update`, the single place nitro registers routes into rou3, mirroring what h3 already does at the same point in its own registration path.
The scope is narrower than [h3's normalizeRoute](https://github.com/h3js/h3/blob/1892ee9cae06533c7db72a5213bceda61cc1d58d/src/utils/internal/path.ts#L197), though. It only encodes non-ASCII characters, not the other things `normalizeRoute` also handles. Could be a follow-up later?

### 📝 Checklist

- [x] I have linked an issue or discussion. => _None_
- [x] I have updated the documentation accordingly. => _None_

### DIFF
--- a/docs/1.docs/5.routing.md
+++ b/docs/1.docs/5.routing.md
@@ -149,7 +149,11 @@ Hello nitro/is/hot!
 
 You can append the HTTP method to the filename to force the route to be matched only for a specific HTTP request method, for example `hello.get.ts` will only match for `GET` requests. You can use any HTTP method you want.
 
-Supported methods: `get`, `post`, `put`, `delete`, `patch`, `head`, `options`, `connect`, `trace`.
+Supported methods: `get`, `post`, `put`, `delete`, `patch`, `head`, `options`, `query`, `connect`, `trace`.
+
+::note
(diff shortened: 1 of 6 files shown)


pi0x:
Sorry — this PR was closed by mistake, not by you.

An automated rebase run of ours force-pushed `main` over this branch. That emptied the diff, so GitHub closed the PR, and it also turned off maintainer edit access, so we could not put your commits back or reopen it.

Your work is safe. It is rebased on `main` and reopened as #4600, with your commit and authorship intact.

If you would rather keep this PR, restore your branch and it will reopen:

```sh
git fetch https://github.com/nitrojs/nitro.git recover/4551-nonascii-route-encoding
git push --force origin FETCH_HEAD:fix/nonascii-route-encoding
```

We will close #4600 if you do. Again, sorry for the churn.

PR #4554, opened 2026-08-24
feat(presets): add `platformatic` preset

p-dubovitsky:
### 🔗 Linked issue

https://github.com/nitrojs/nitro/discussions/4553

### ❓ Type of change

### 📚 Description

Adds a built-in `platformatic` preset.

The preset extends `node-server` and loads `@platformatic/nitro/scheduler` when `scheduledTasks` are configured. Applications without scheduled tasks are unaffected, and `experimental.tasks` remains an explicit opt-in.

### 📝 Checklist


src/presets/platformatic/preset.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const platformatic = defineNitroPreset(
+  {
+    extends: "node-server",
+    hooks: {
+      "build:before": async (nitro) => {
+        if (Object.keys(nitro.options.scheduledTasks).length === 0) {
+          return;
+        }
+        await installModules(nitro, ["@platformatic/nitro/scheduler"]);
```
pi0, line 12: We need make this a built-in feature. Nitro presets do not require external dependencies.

pi0x:
I rebased this onto latest `main` locally. The only conflict was `docs/2.deploy/0.index.md`, where `main` had recapitalised the provider names; I kept that and added Platformatic to the list. Typecheck, lint and `test/unit/platformatic-preset.test.ts` all pass. I could not push, so the branch here is unchanged.

@pi0's review point is still the blocker: the preset installs `@platformatic/nitro/scheduler`, an external package, so it is not self-contained. That needs resolving before this can land.

One thing I fixed: the docs used `defineNitroConfig` from `nitro/config`. In v3 it is `import { defineConfig } from "nitro"`.

Note: #4025, #3970 and #3554 also add presets and touch the same generated files, so whoever lands first forces the rest to rebase.

PR #4565, opened 2026-08-29
fix(types): emit bare specifier for auto-imports from subpaths

danielroe:
### ❓ Type of change

### 📚 Description

this was introduced by me in #1528 🙈

basically auto-imports coming from a package `exports` subpath (`nitro/h3`, for example) end up as a relativised file path like `../../../node_modules/nitro/h3`, which isn't a real file, so TypeScript resolves the whole declaration to `any`

I think emitting the bare specifier in that case is the smallest fix, letting package `exports` and its `types` condition do the resolving, and keeping the relativised path only where there's no subpath

### 📝 Checklist


pi0:
removing auto imports from core #4573

PR #4585, opened 2026-09-06
docs(examples): use `rel` instead of `type` for modulepreload links

cprecioso:
The Vite SSR examples create `<link type=modulepreload>` tags but it's actually `<link rel=modulepreload>` [(MDN link)](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/modulepreload). Straightforward fix.

---

Disclosure: the patch was written by an AI agent (Claude Code), reviewed and driven by me.


pi0 (approved):
Thanks!

PR #4588, opened 2026-09-07
fix(static): respond with 404 when a public asset is missing on disk

### LINKED ISSUE #29078: 500 error occurs when trying to access any static file directly in production mode while using a CDN to serve public files.
### Environment

- Operating System: `MacOS Sonoma Version 14.6.1`
- Node Version:     `v20.11.0`
- Nuxt Version:     `3.13.2`
- CLI Version:      `3.13.2`
- Nitro Version:    `2.9.7`
- Package Manager:  `npm@10.2.4`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

1. open https://stackblitz.com/edit/github-q544zj?file=package.json
2. open dev tools
3. stop dev server
4. run `npm run build`
5. run `npm run remove-icon` to remove favicon from the build
6. run `npm run start`
7. navigate to [project-url]/favicon.ico in the stackblitz browser

### Describe the bug

There is information about the usage of `cdnUrl` in the official documentation: https://nuxt.com/docs/api/nuxt-config#cdnurl.
It states that `cdnUrl` is `an absolute URL to serve the public folder from (production-only).`
So, I wanted to move all my static files from the public folder to a CDN. I set `cdnUrl` and moved the `.output/public` folder to the CDN. Everything works great, except for one thing - if I try to access any file directly in the browser that was in the public directory during the build process, I get a 500 error and see this message in the Nuxt terminal:
`[nuxt] [request error] [unhandled] [500] ENOENT: no such file or directory, open '/home/projects/gqairmgblw.github/.output/public/[file_name]'.`
It seems like this is not the expected behavior. Shouldn't I be getting a 404 error instead?

You can reproduce this issue with any file, such as the `favicon.ico` file, as I described in the `Reproduction` section. The problem occurs with any file, not just `favicon.ico`.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #33033: H3Error: ENOENT: no such file or directory, open 'xxx.js.map'
### Environment

- Operating System: Darwin
- Node Version:     v24.3.0
- Nuxt Version:     4.0.3
- CLI Version:      3.28.0
- Nitro Version:    2.12.4
- Package Manager:  pnpm@10.12.4
- Builder:          -
- User Config:      compatibilityDate, devtools, typescript, sourcemap
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/Arman19941113/nuxt-app

#### 1. Build and serve

```bash
take git@github.com:Arman19941113/nuxt-app.git
ni
nr build
rm .output/public/_nuxt/**.js.map
node .output/server/index.mjs
```

#### 2. Open Chrome DevTools

Ensure "Enable JavaScript source maps" is turned on in settings.

### Describe the bug

When the browser requests a source map file for an existing JS file, the server throws a runtime error.

* **Case 1:** Requesting `*.js.map` for an existing JS file → server crashes with a runtime error (**unexpected behavior**)
* **Case 2:** Requesting `*.js.map` for a non-existent JS file → server correctly responds with **404**

I understand that setting the sourcemap config to `"hidden"` can prevent this, but even if the file is requested (e.g., via `curl`), the server should respond with **404** instead of crashing.

Image

### Additional context

_No response_

### Logs

```shell-script

```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/33033
resolves https://github.com/nuxt/nuxt/issues/29078

### ❓ Type of change

### 📚 Description

when a file can't be read because it can't be read from disk, we can return a 404 instead of a 500 error, which seems wrong.

alternatively, we could _always_ return a 404 when the asset can't be read, if you didn't want to tie this too closely to fs semantics.

... or you may have another better idea!

### 📝 Checklist

### DIFF
--- a/src/build/virtual/public-assets.ts
+++ b/src/build/virtual/public-assets.ts
@@ -142,7 +142,7 @@ import { resolve, dirname } from 'node:path'
 import assets from '#nitro/virtual/public-assets-data'
 export function readAsset (id) {
   const serverDir = dirname(fileURLToPath(globalThis.__nitro_main__))
-  // Public dirs are often pruned after the build (uploaded to a CDN, sourcemaps
+  // Public dirs might be pruned after the build (uploaded to a CDN, sourcemaps
   // stripped), so a manifest entry with no file behind it reads as absent, not broken.
   return fsp.readFile(resolve(serverDir, assets[id].path)).catch((error) => {
     if (error?.code !== 'ENOENT') { throw error }
--- a/src/build/virtual/public-assets.ts
+++ b/src/build/virtual/public-assets.ts
@@ -142,9 +142,10 @@ import { resolve, dirname } from 'node:path'
 import assets from '#nitro/virtual/public-assets-data'
 export function readAsset (id) {
   const serverDir = dirname(fileURLToPath(globalThis.__nitro_main__))
+  // Public dirs are often pruned after the build (uploaded to a CDN, sourcemaps
+  // stripped), so a manifest entry with no file behind it reads as absent, not broken.
   return fsp.readFile(resolve(serverDir, assets[id].path)).catch((error) => {
     if (error?.code !== 'ENOENT') { throw error }
-    console.warn(\`[nitro] Public asset "\${id}" is missing on disk (\${assets[id].path})\`)
   })
 }`;
       },
--- a/src/build/virtual/public-assets.ts
+++ b/src/build/virtual/public-assets.ts
@@ -142,7 +142,10 @@ import { resolve, dirname } from 'node:path'
 import assets from '#nitro/virtual/public-assets-data'
 export function readAsset (id) {
   const serverDir = dirname(fileURLToPath(globalThis.__nitro_main__))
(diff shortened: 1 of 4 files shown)


src/build/virtual/public-assets.ts
```diff
@@ -142,7 +142,11 @@ import { resolve, dirname } from 'node:path'
 import assets from '#nitro/virtual/public-assets-data'
 export function readAsset (id) {
   const serverDir = dirname(fileURLToPath(globalThis.__nitro_main__))
-  return fsp.readFile(resolve(serverDir, assets[id].path))
+  // Public dirs are often pruned after the build (uploaded to a CDN, sourcemaps
```
pi0:
```suggestion
  // Public dirs might be pruned after the build (uploaded to a CDN, sourcemaps
```

pi0 (approved):

danielroe:
thank you @pi0 - this is a better approach ❤️

Issue #4242, opened 2026-05-02
Prerendered `index.html` not served statically by Vercel preset — root URL goes through SSR function


MaxMrt:
### Question

  I'm running into the following behavior with the Vercel preset and wanted to ask whether this is a known issue, or whether
  others have observed the same — before filing it as a bug.

  ### Setup

  A Nuxt 4 project (Nitro under the hood) deployed to Vercel, with:

  ```ts
  routeRules: {
    '/': { prerender: true },
    '/foo/**': { prerender: true },
  }
  ```

  The build produces both `.output/public/index.html` and `.output/public/foo/some-slug/index.html` as expected.

  ### What I observe

  Detail routes are served as pure static files from the Vercel Edge CDN. Response headers include `accept-ranges: bytes`,
  `etag`, `last-modified`, `content-disposition` — clear signs of CDN file delivery, no SSR function involved.

  The root URL (`/`) on the same deploy behaves differently: it appears to be rendered by an SSR function on cache miss.
  Headers include `x-powered-by: Nuxt` and `cache-control: public, max-age=3600`, and crucially do **not** include
  `accept-ranges` / `etag` / `last-modified`.

  ```bash
  curl -D - -o /dev/null https://<deployment>/
  # x-powered-by: Nuxt
  # cache-control: public, max-age=3600
  # (no etag/accept-ranges/last-modified)

  curl -D - -o /dev/null https://<deployment>/foo/some-slug
  # accept-ranges: bytes
  # etag: "..."
  # last-modified: ...
  # (no x-powered-by)
  ```

  The first cache miss after deploy or after the 1h CDN cache expires therefore goes through a cold function start instead of
   being delivered from the static file. Detail pages are not affected.

  ### What seems to cause it

  Looking at the generated `.vercel/output/config.json`, the overrides differ:

  ```json
  "overrides": {
    "index.html":               { "path": "" },
    "foo/some-slug/index.html": { "path": "foo/some-slug" }
  }
  ```

  The root override has `"path": ""` (empty string), while detail routes get a proper non-empty path. Combined with the later
   route `{ "src": "/", "dest": "/index" }`, requests to `/` fall through to the SSR function instead of being matched by the
   filesystem handler.

  This looks similar to nuxt/nuxt#14888 (closed 2022) which was fixed by nitrojs/nitro#500 — possibly a regression?

  ### Question

  - Has anyone else observed this on recent Nitro / Nuxt versions with the Vercel preset?
[...]

PR #4592, opened 2026-09-07
fix(vercel): match ISR route rules against the route pattern

### LINKED ISSUE #4447: Vercel preset: observability-route generation checks ISR route rules against the compiled regex string, not a request path. Competing functions for dynamic routes, dangling `config.json` routes for static/catch-all
### Environment

- nitropack `2.13.4` (lockfile-resolved, latest release; not yet verified against v3 `main`)
- Nuxt `4.4.8` / `@nuxt/nitro-server@4.4.8`
- Node `v25.9.0`, Bun `1.3.14` for install/build
- Preset: `vercel` (Build Output API v3), `compatibilityDate: '2025-07-15'` (the threshold that enables observability-route generation; `'2025-07-14'` disables it)

Relevant config — one dynamic and one static/catch-all `isr` rule are enough to show both defects:

```ts
routeRules: {
  '/schedule': { isr: 300 },       // static → dangling config.json route
  '/users/:id': { isr: 60 },       // dynamic → competing plain + -isr functions
  '/catchall/**': { isr: 60 },     // catch-all → dangling config.json route
}
```

### Reproduction

https://github.com/MichaelDickBasecom/nuxt-vercel-isr-404-repro

```sh
bun install
NITRO_PRESET=vercel bun run build
node scripts/inspect-build.mjs
```

`scripts/inspect-build.mjs` reads `.vercel/output/` and lists (1) every `config.json` route whose `dest` resolves to no emitted function or static file, and (2) every route pattern for which both a plain function and an `-isr` function were emitted. No deployment needed — both defects are visible in the build output itself.

Note: the repo's main config includes a `'/**': { isr: 60 }` wildcard, which masks the competing-functions defect (the wildcard makes the skip apply to everything). To see both defects, comment out the `'/**'` rule in `nuxt.config.ts` and rebuild — the logs below are from exactly that configuration.

### Describe the bug

Since #3474, the Vercel preset emits per-route "observability" functions and matching `config.json` route entries. For ISR-ruled paths the per-route function must be skipped — those paths are served through the ISR rewrite machinery instead. The first round of clashes between the two features was fixed in 2.12.6 (#3560, #3562, #3563; symptom report nuxt#33140).

A residual gap survived that fix: `generateFunctionFiles` decides whether to skip an o11y function by calling `_getRouteRules(route.src)`, where `route.src` is the **compiled regex string** (e.g. `/users/(?<id>[^/]+)`), not a request path. Matching route rules against a regex string instead of a path breaks in both directions:

1. **Dynamic routes slip through the skip** — `/users/(?<id>[^/]+)` doesn't match the `'/users/:id'` rule as a path, so the skip never applies and the build emits **both** `users/[id].func` **and** `users/[id]-isr.func`, plus a `config.json` route entry for each.
2. **Static and catch-all routes are skipped, but their `config.json` route entries are still emitted** — the entry's `dest` then points at a function that was never created (e.g. `{"src": "/schedule", "dest": "/schedule"}` with no `schedule.func` in the output).

### Current impact

Both defects are **latent** in current output: the o11y route entries happen to sort after the ISR rewrites, so they're shadowed for any path both regexes match (verified by inspecting the emitted route order). But the correctness of every ISR-ruled route now hangs on emission order rather than on the entries being right, and per Build Output API semantics a matched `dest` that resolves to nothing hard-404s without falling through. The shadowing also isn't complete: the o11y catch-all src `/catchall/?(?<slug>.+)` makes the slash optional, so a request like `/catchallx` matches it while *not* matching the ISR src `/catchall/(?:.*)` — landing on the dangling dest (a platform-level 404 instead of the app's rendered 404; inferred from the Build Output API docs, not verified on a deployment).

The competing functions are near-free as symlinks, but with `functionRules`-style full copies each redundant function is a full server bundle (see #4233 for the size impact of copied functions).

### Suggested fix

Perform the ISR-skip check against the route's request path/pattern (known at generation time, before regex compilation) instead of the compiled `route.src` string — and when a function's creation is skipped, skip its `config.json` route entry too, so no `dest` can point at a function that doesn't exist.

### Additional context

- Found while dissecting the "ISR route serves a cached 404 naming `/<route>-isr`" symptom family (same repro repo). Important negative result so the two don't get conflated: **this generation bug is not the cause of those 404s** — disabling o11y routes entirely via `compatibilityDate: '2025-07-14'` changes nothing about the ISR 404 behavior, which is a runtime URL-restoration gap reported separately (#4446).
- I searched existing issues and found no report of the post-2.12.6 residual: #3560/#3562/#3563 fixed the 2.12.5 clashes, #1540 (route-rule symlink `EEXIST`) and #4233 (functionRules copy size) are adjacent but different defects.
- Everything above was checked on 2.13.4 (latest release at time of writing). I have not audited v3 `main`'s generation path for the same pattern — happy to do so if useful.

### Logs

```sh
`node scripts/inspect-build.mjs` after `NITRO_PRESET=vercel bun run build` on nitropack 2.13.4 (2026-07-18), with the `'/**'` wildcard rule removed so both defects are visible (trimmed to the relevant lines):


== functions emitted (name → symlink?):
  __fallback
  about  -> symlink
  index  -> symlink
  schedule-isr  -> symlink
  users/[id]  -> symlink
  users/[id]-isr  -> symlink
  users/[id]/_payload.json-isr  -> symlink
  catchall/[...]-isr  -> symlink
  ...

== DANGLING routes (dest resolves to no function/static output):
  {"src":"/secret","dest":"/secret"}
  {"src":"/schedule-pq","dest":"/schedule-pq"}
  {"src":"/schedule","dest":"/schedule"}
  {"src":"/query","dest":"/query"}
  {"src":"/nested/deep","dest":"/nested/deep"}
  {"src":"/flaky-ttl","dest":"/flaky-ttl"}
  {"src":"/flaky","dest":"/flaky"}
  {"src":"/catchall/?(?<slug>.+)","dest":"/catchall/[...slug]"}

== COMPETING outputs (plain o11y function alongside -isr function):
  users/[id].func  AND  users/[id]-isr.func


Emitted route order for the same build, showing the shadowing (ISR rewrite first, o11y entry later) and the src mismatch on the catch-all:


routes[10]: {"src": "(?<__isr_route>/users/(?<id>[^/]+))", "dest": "/users/[id]-isr?__isr_route=$__isr_route"}
routes[22]: {"src": "(?<__isr_route>/schedule)",           "dest": "/schedule-isr?__isr_route=$__isr_route"}
routes[23]: {"src": "(?<__isr_route>/catchall/(?:.*))",    "dest": "/catchall/[...]-isr?__isr_route=$__isr_route"}
routes[31]: {"src": "/schedule",                           "dest": "/schedule"}            ← dangling
routes[41]: {"src": "/users/(?<id>[^/]+)",                 "dest": "/users/[id]"}          ← competing
routes[42]: {"src": "/catchall/?(?<slug>.+)",              "dest": "/catchall/[...slug]"}  ← dangling, slash optional
```

ShreeBohara:
### 🔗 Linked issue

Resolves #4447

### ❓ Type of change

### 📚 Description

The observability-function skip for ISR routes was decided with `_getRouteRules(route.src)`, and `route.src` is the compiled PCRE rather than a path. A `:param` segment compiles to `(?<id>[^/]+)`, and the literal `/` inside that character class makes rou3 split the value into extra segments, so a dynamic route never matched its own rule:

```js
routeRules.matchAll("", "/users/:id")           // { isr: 60 }
routeRules.matchAll("", "/users/(?<id>[^/]+)")  // undefined
"/users/(?<id>[^/]+)".split("/")                // ["", "users", "(?<id>[^", "]+)"]
```

A catch-all compiles to `.+`, which has no slash in it, so it did match and its function was skipped correctly. That asymmetry is why the issue reports two different symptoms from one cause: dynamic routes emit a plain function competing with the ISR one, while static and catch-all routes emit a `config.json` entry whose `dest` was never created.

Reproduced against `main` (1a7c9b8e) with the reporter's three-rule config on the `vercel` preset:

```
before                                          after
  users/[id].func + users/[id]-isr.func           users/[id]-isr.func only
  {"src":"/schedule","dest":"/schedule"}          removed
  {"src":"/catchall/?(?<slug>.+)", ...}           removed
  /plain, / -> function + route entry             unchanged
```

The dangling entries were not fully shadowed by the ISR rewrites either: the o11y catch-all src `/catchall/?(?<slug>.+)` makes the separator optional, so `/catchallx` matched it without matching the ISR src `/catchall/(?:.*)` and landed on a `dest` with no function behind it. It now falls through to the fallback route.

**Why the check moved instead of the argument being corrected.** `generateBuildConfig(nitro, o11Routes)` runs at the top of `generateFunctionFiles`, before the skip loop, so passing the right value at the call site would still have left the `config.json` entries for skipped functions. `getObservabilityRoutes` already filters prerendered routes and still holds the uncompiled pattern, so doing it there makes both consumers read the same filtered list and they cannot drift apart. The in-loop check is removed rather than adjusted.

Tests are five cases in `test/unit/vercel-observability-routes.test.ts`. Three fail before and pass after (dynamic, static + catch-all, and a route covered by a wildcard ISR rule); two are controls that pass either way (a rule without ISR, and no matching rule) to guard against over-filtering.

Not addressed here: `routeFuncRouter?.match("", route.src)` a few lines below passes the same compiled `src`, so `functionRules` should mismatch on dynamic routes for the same reason. I have not verified the impact there and left it out of scope.

The analysis and patch were agent-assisted, as noted on the issue.

### DIFF
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -176,13 +176,7 @@ export async function generateFunctionFiles(nitro: Nitro) {
   if (o11Routes.length === 0) {
     return;
   }
-  const _getRouteRules = (path: string) =>
-    defu({}, ...nitro.routing.routeRules.matchAll("", path).reverse()) as NitroRouteRules;
   for (const route of o11Routes) {
-    const routeRules = _getRouteRules(route.src);
-    if (routeRules.isr) {
-      continue; // #3563
-    }
     const funcPrefix = resolve(nitro.options.output.serverDir, "..", route.dest);
     const funcDir = funcPrefix + ".func";
 
@@ -631,6 +625,17 @@ export function getObservabilityRoutes(nitro: Nitro): ObservabilityRoute[] {
   );
 
   // Sort routes by how much specific they are
+  // ISR-ruled paths are served through the ISR rewrite machinery, so they get
+  // neither an observability function nor a `config.json` route entry (#3563).
+  // Rules have to be matched against the route pattern rather than the compiled
+  // `src`: a `:param` segment compiles to `(?<id>[^/]+)`, and the `/` inside
+  // that character class splits into extra path segments, so a dynamic route
+  // never matched its own rule (#4447).
+  const hasISRRule = (route: string) =>
+    Boolean(
+      (defu({}, ...nitro.routing.routeRules.matchAll("", route).reverse()) as NitroRouteRules).isr
+    );
+
   const routePatterns = [
     ...new Set([
       ...(nitro.options.ssrRoutes || []),
@@ -638,7 +643,9 @@ export function getObservabilityRoutes(nitro: Nitro): ObservabilityRoute[] {
         .filter((h) => !h.middleware && h.route)
         .map((h) => h.route!),
     ]),
-  ].filter((route) => !prerenderedPaths.has(route.replace(SURROUNDING_SLASH_RE, "")));
+  ].filter(
+    (route) => !prerenderedPaths.has(route.replace(SURROUNDING_SLASH_RE, "")) && !hasISRRule(route)
+  );
 
   const staticRoutes: string[] = [];
   const dynamicRoutes: string[] = [];
--- a/src/presets/vercel/utils.ts
+++ b/src/presets/vercel/utils.ts
@@ -624,7 +624,6 @@ export function getObservabilityRoutes(nitro: Nitro): ObservabilityRoute[] {
       .map((route) => route.route.replace(SURROUNDING_SLASH_RE, ""))
   );
 
-  // Sort routes by how much specific they are
   // ISR-ruled paths are served through the ISR rewrite machinery, so they get
   // neither an observability function nor a `config.json` route entry (#3563).
   // Rules have to be matched against the route pattern rather than the compiled
@@ -636,6 +635,7 @@ export function getObservabilityRoutes(nitro: Nitro): ObservabilityRoute[] {
       (defu({}, ...nitro.routing.routeRules.matchAll("", route).reverse()) as NitroRouteRules).isr
     );
 
+  // Sort routes by how much specific they are
   const routePatterns = [
     ...new Set([
       ...(nitro.options.ssrRoutes || []),
--- a/test/unit/vercel-observability-routes.test.ts
+++ b/test/unit/vercel-observability-routes.test.ts
@@ -1,21 +1,29 @@
 import { describe, expect, it } from "vitest";
-import type { Nitro, NitroEventHandler, PrerenderRoute } from "nitro/types";
+import type { Nitro, NitroEventHandler, NitroRouteRules, PrerenderRoute } from "nitro/types";
 
 import { getObservabilityRoutes } from "../../src/presets/vercel/utils.ts";
+import { Router } from "../../src/routing.ts";
 
 function createNitroStub(opts: {
   compatibilityDate?: string;
   handlers?: NitroEventHandler[];
   ssrRoutes?: string[];
   prerenderedRoutes?: PrerenderRoute[];
+  routeRules?: Record<string, NitroRouteRules>;
 }): Nitro {
+  const routeRules = new Router<NitroRouteRules>();
+  routeRules._update(
+    Object.entries(opts.routeRules || {}).map(([route, data]) => ({ route, method: "", data }))
+  );
   return {
     scannedHandlers: opts.handlers || [],
     _prerenderedRoutes: opts.prerenderedRoutes,
+    routing: { routeRules },
     options: {
       compatibilityDate: { default: opts.compatibilityDate || "2025-07-15" },
       handlers: [],
       ssrRoutes: opts.ssrRoutes || [],
+      routeRules: opts.routeRules || {},
     },
   } as unknown as Nitro;
 }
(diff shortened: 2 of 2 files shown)


pi0x:
Thanks for the very clear write-up. I went through this PR for the maintainers.

**Rebase.** I rebased your branch onto the latest `main` and force-pushed it. There were no conflicts. The change is unchanged apart from one small tidy-up (below).

**Review.** I checked the fix and it looks right to me.

- `getObservabilityRoutes` is only used from `generateFunctionFiles`, and the result feeds both `generateBuildConfig` and the function-writing loop. So filtering there really does remove the dangling `config.json` entry *and* the competing function in one place. That is the correct spot.
- `defu({}, ...matchAll(...).reverse()).isr` keeps the same rule-merging order the old code used, so precedence does not change.
- I confirmed your tests fail on `main` and pass with the fix. Three fail before, all pass after. The two control tests are a good idea.

**One small change I made.** The line `// Sort routes by how much specific they are` ended up above your new ISR comment, away from the code it talks about. I moved your ISR comment above it so each comment sits next to its own code. No logic changed. I kept the explanation text, because it explains *why* and matches the style of the nearby `#4242` comment.

**About the follow-up you mentioned.** You are right about `routeFuncRouter?.match("", route.src)`. It has the same problem: `route.src` is the compiled pattern, so a `functionRules` entry will not match a dynamic route. The ISR loop just above it already does this correctly, using the raw rule key (`routeFuncRouter?.match("", key)`). Fixing it needs the raw route pattern to be carried on `ObservabilityRoute`, plus a preset-level test. I agree with you that it is out of scope here, so I left it alone. It would make a good separate PR.

**Smaller notes, no action needed:**

- The route-rules router is built with `nitro.options.baseURL` prefixed onto every rule, but handler route patterns do not carry the base. So with a non-default `baseURL` the ISR rule may still not match. This is not a regression from your PR (the old code was worse), just something to be aware of.
- `isr: 0` is falsy, so such a route is kept here, while the ISR rewrite is still emitted at the `value.isr !== undefined` check. That is a pre-existing edge case, unrelated to this PR.

`pnpm typecheck` and `pnpm fmt` are clean, and `test/unit/vercel-observability-routes.test.ts` and `test/unit/vercel-isr.test.ts` both pass.

PR #4600, opened 2026-09-07
fix(routing): normalize route patterns like h3

pi0x:
Replaces #4551 (closed by accident, see below).

Nitro registers routes into rou3 directly and overrides h3's route lookup, so patterns never went through h3's normalization in `on()`. A route like `/café` or `/について/**` was registered as-is, but it is matched against `event.url.pathname`, which is percent-encoded, so it never matched.

`Router` now passes every pattern (and the base URL) through h3's `normalizeRoute` (exported since h3 `2.0.1-rc.32`). This percent-encodes non-ASCII and other characters that request paths carry encoded, decodes needless escapes (`/%40handle` -> `/@handle`) and resolves dot segments, while leaving rou3 syntax alone.

Normalization can be turned off with `new Router(baseURL, { normalize: false })`. Two routers use this because they are matched against other patterns, not request paths: `nitro.routing.routeRules` (the runtime rules matcher is compiled separately by `h3/rules`) and Vercel's `functionRules` router.

Tests cover the router matching the same paths as a plain h3 app, and route rules keyed by a non-ASCII pattern in a Vite fixture.

---

**About #4551:** while rebasing the v3 PR backlog, an automated run force-pushed `main` over its head branch. That emptied the diff, closed the PR and revoked maintainer-edit access on the fork, so it could not be reopened. Sorry @Vincentdevreede, this was our mistake. Your original commit is kept here with authorship preserved. If you would rather keep ownership, restore your branch and we will close this one:

```sh
git fetch https://github.com/nitrojs/nitro.git recover/4551-nonascii-route-encoding
git push --force origin FETCH_HEAD:fix/nonascii-route-encoding
```

🤖 Generated with AI assistant


Vincentdevreede:
@pi0x no problem 👍

pi0 (approved):

PR #4612, opened 2026-09-11
fix(vite): warn when the SSR entry is outside the auto-detected directories

### LINKED ISSUE #4591: vite: prerendering silently emits no HTML when the SSR entry is outside the auto-detected directories
### Environment

- Nitro: `3.0.260903-beta`
- Vite: `8.2.2`
- React: `19.2.0`, Node.js: `v24.20.0`
- macOS: `26.6.2`, arm64
- Minimal React SSR app using `nitro/vite` with `prerender.routes: ["/"]`. No Nuxt or other framework.
- Also inspected Nitro main at `c5177e9218cdd113c6c6a6a74b2924b2354af120`; the detection logic below is unchanged there. The reproduction was executed against the published beta listed above.

### Reproduction

Ready-to-run repository: https://github.com/cprecioso/nitro-prerender-entry-repro

```sh
git clone https://github.com/cprecioso/nitro-prerender-entry-repro.git
cd nitro-prerender-entry-repro
npm ci
npm run repro
```

The check builds the same app three times, changing only the Vite config, and exits with code 1 while the behaviour reproduces.

The app is a trivial React SSR setup. The SSR entry is still named `entry-server.tsx`; the only variable is that it sits at `src/ssr/` rather than `src/`.

### Describe the bug

Nitro auto-detects the Vite SSR entry by probing for `./entry-server` in `<rootDir|scanDirs>/{app,src,}/`. When the entry is anywhere else, no `ssr` service is registered and no renderer is wired up, but **the build still exits 0**. The prerenderer resolves every route to `[404]`, emits no HTML at all, and prints no warning that the app has no renderer.

`.output/public/` ends up containing only the client JS chunk, with no HTML entry point, while the build reports success and suggests `npx vite preview`. Moving the file up one directory, with no other change, prerenders correctly.

Expected: one of

1. Nitro warns or errors when there is no renderer and no `ssr` service, rather than emitting a successful build with zero HTML.
2. Every prerendered route resolving to `[404]` is treated as a build failure.
3. There is a discoverable, documented way to point at an SSR entry in a non-standard location.

**Workaround.** Declaring the entry as the `ssr` environment input works:

```ts
environments: {
  ssr: {
    build: { rollupOptions: { input: "src/ssr/entry-server.tsx" } },
  },
  client: {
    build: { rollupOptions: { input: "src/entry-client.tsx" } },
  },
},
```

`setupNitroContext` skips auto-detection when `userConfig.environments.ssr` is defined and resolves the env's input as the `ssr` service instead, after which `configResolved` still auto-wires `renderer.handler` to the internal `ssr-renderer`.

This was hard to arrive at. `environments.ssr` does appear in the Solid and Vue Router examples, but always pointing at a file *named* `entry-server`, and it is introduced there as a framework requirement ("SolidJS requires explicit `ssr` and `client` environment configuration") rather than as the way to relocate the SSR entry. [`docs/1.docs/61.vite.md`](https://github.com/nitrojs/nitro/blob/c5177e9218cdd113c6c6a6a74b2924b2354af120/docs/1.docs/61.vite.md) documents the auto-detection rule and `environments.client.build.rollupOptions.input` for the client entry, with no `ssr` counterpart mentioned.

### Additional context

Secondary, and possibly the more useful fix: `renderer.handler` looks like the right option and fails confusingly.

The [renderer docs](https://nitro.build/docs/renderer) describe `renderer.handler` as "Path to a custom renderer handler module", so it is the natural thing to reach for when the handler is not auto-detected. Setting `renderer: { handler: "src/ssr/entry-server.tsx" }` routes the file into Nitro's own rolldown bundle, which has no Vite plugins, so the `?assets=` virtual modules (and JSX, and CSS imports) cannot be resolved. It fails identically regardless of where the file lives, including the auto-detected location. Logs below.

It would help if `renderer.handler` rejected a path belonging to a Vite environment with an explanatory error, or if the docs noted that under the Vite plugin the SSR entry is configured via `environments.ssr` and never via `renderer.handler`.

Relevant source:

- Entry auto-detection: [`src/build/vite/plugin.ts#L427-L440`](https://github.com/nitrojs/nitro/blob/c5177e9218cdd113c6c6a6a74b2924b2354af120/src/build/vite/plugin.ts#L427-L440)
- Renderer auto-wiring, which requires `ctx.services.ssr?.entry`: [`src/build/vite/plugin.ts#L198-L206`](https://github.com/nitrojs/nitro/blob/c5177e9218cdd113c6c6a6a74b2924b2354af120/src/build/vite/plugin.ts#L198-L206)

Happy to open a PR for whichever direction you prefer.

Prepared with AI assistance; the reproduction was executed locally.

### Logs

```sh
# npx vite build   (entry at src/ssr/entry-server.tsx, no environments.ssr)

[nitro] ◐ Building [Nitro] (preset: node-server, compatibility: 2026-09-07)
[nitro] ✔ Generated public .output/public
[nitro] ℹ Initializing prerenderer
[nitro] ℹ Prerendering 1 routes
[nitro]   ├─ / (109ms)
  │ └── [404]
[nitro] ℹ Prerendered 0 routes in 0.123 seconds

[nitro] ✔ You can preview this build using npx vite preview

$ echo $?
0
$ find .output/public -type f
.output/public/assets/entry-client-Ba_RLnZH.js


# npx vite build --config vite.config.renderer-handler.ts

error during build:
Error: Build failed with 2 errors:

[UNLOADABLE_DEPENDENCY] Could not load src/ssr/entry-server.tsx?assets=ssr
   ╭─[ src/ssr/entry-server.tsx:7:26 ]
   │
 7 │ import serverAssets from "./entry-server?assets=ssr";
   │                          ────────────┬────────────
   │                                      ╰──────────── No such file or directory (os error 2)
───╯

[UNLOADABLE_DEPENDENCY] Could not load src/entry-client.tsx?assets=client
   ╭─[ src/ssr/entry-server.tsx:6:26 ]
   │
 6 │ import clientAssets from "../entry-client?assets=client";
   │                          ───────────────┬───────────────
   │                                         ╰─────────────── No such file or directory (os error 2)
───╯

    at aggregateBindingErrorsIntoJsError (node_modules/rolldown/dist/shared/error-BP8lJHle.mjs:48:18)
    at async prerender (node_modules/nitro/dist/_chunks/nitro.mjs:743:2)
    at async buildEnvironments (node_modules/nitro/dist/vite.mjs:121:2)
```

pengzh1:
### 🔗 Linked issue

Resolves #4591

### ❓ Type of change

### 📚 Description

Under the Vite plugin, Nitro auto-detects the SSR entry by probing for `entry-server` in the root, `app/` and `src/` of the root dir and scan dirs (`setupNitroContext` in `src/build/vite/plugin.ts`). When the entry lives anywhere else, no `ssr` service is registered, so `configResolved` skips wiring the default renderer — yet the build still exits 0. The prerenderer then resolves every configured route to `[404]`, emits no HTML, and prints nothing that explains why (`.output/public` contains only the client chunk).

This PR implements direction (1) from the issue: the auto-detection failure branch now emits an explicit warning stating that no renderer is wired up, that prerendered routes will resolve to 404 without emitting HTML, and both remedies — moving `entry-server` into a detected directory, or pointing `environments.ssr.build.rollupOptions.input` at it explicitly (the workaround documented in the issue).

A warning (rather than an error) was chosen deliberately: apps that intentionally have no SSR entry (pure client builds) still build fine; the warning only makes the previously silent failure observable. No detection algorithm changes.

### 📝 Checklist

### Testing

- Added `test/vite/prerender-missing-ssr.test.ts` (+ fixture with the SSR entry deliberately placed at `src/ssr/entry-server.ts`).
- **Red (before the fix):** resolving the config for the fixture asserts both halves of #4591 — `environments.ssr` stays unregistered (passes), and no warning is emitted (fails with `expected '' to contain 'ssr entry'`).
- **Green (after the fix):** the same resolution emits the new warning; `pnpm vitest run test/vite` passes (69 passed / 1 skipped) and `test/unit` shows only the 6 known Windows path baseline failures unrelated to this change. `pnpm fmt` and `pnpm typecheck` are clean.

### DIFF
--- a/src/build/vite/plugin.ts
+++ b/src/build/vite/plugin.ts
@@ -450,6 +450,13 @@ async function setupNitroContext(
       if (ssrEntry) {
         ctx.services.ssr = { entry: ssrEntry };
         ctx.nitro!.logger.info(`Using \`${prettyPath(ssrEntry)}\` as vite ssr entry.`);
+      } else {
+        ctx.nitro!.logger.warn(
+          "No vite ssr entry found: no renderer is wired up, so SSR and prerendering are " +
+            "disabled and prerendered routes resolve to 404 without emitting any HTML. " +
+            "Move `entry-server` to the root, `app/` or `src/` directory, or point " +
+            "`environments.ssr.build.rollupOptions.input` at it explicitly."
+        );
       }
     } else {
       let ssrEntry = getEntry(userConfig.environments.ssr.build?.rollupOptions?.input);
--- a/test/vite/prerender-missing-ssr-fixture/src/ssr/entry-server.ts
+++ b/test/vite/prerender-missing-ssr-fixture/src/ssr/entry-server.ts
@@ -1,0 +1,9 @@
+// SSR entry deliberately placed *outside* the auto-detected directories
+// (`<root|app|src>/entry-server.*`) to reproduce #4591.
+export default {
+  fetch(_req: Request) {
+    return new Response("<!doctype html><h1>ssr</h1>", {
+      headers: { "content-type": "text/html" },
+    });
+  },
+};
--- a/test/vite/prerender-missing-ssr.test.ts
+++ b/test/vite/prerender-missing-ssr.test.ts
@@ -1,0 +1,50 @@
+import { fileURLToPath } from "node:url";
+import { describe, expect, test } from "vitest";
+import { consola } from "consola";
+import { nitro } from "../../src/vite.ts";
+
+const { resolveConfig } = (await import(
+  process.env.NITRO_VITE_PKG || "vite"
+)) as typeof import("vite");
+
+const rootDir = fileURLToPath(new URL("./prerender-missing-ssr-fixture", import.meta.url));
+
+// #4591: when the SSR entry lives outside the auto-detected directories
+// (`<root|app|src>/entry-server.*`), no `ssr` service is registered and no
+// renderer is wired up, but the build used to succeed silently — prerendered
+// routes all resolve to `[404]` and no HTML is emitted, with no hint why.
+describe("vite: missing SSR entry", () => {
+  test("warns when the SSR entry is outside the auto-detected directories", async () => {
+    const warnings: string[] = [];
+    const reporter = {
+      log: (logObj: { type: string; args: unknown[] }) => {
+        if (logObj.type === "warn") {
+          warnings.push(logObj.args.join(" "));
+        }
+      },
+    };
+    const originalReporters = [...consola.options.reporters];
+    consola.setReporters([reporter]);
+    let resolved;
+    try {
+      resolved = await resolveConfig(
+        {
+          root: rootDir,
+          configFile: false,
+          plugins: [nitro()],
+          nitro: { prerender: { routes: ["/"] } },
+        } as any,
+        "build"
+      );
+    } finally {
+      consola.setReporters(originalReporters);
+    }
+
+    // The entry at `src/ssr/entry-server.ts` is not auto-detected, so no `ssr`
+    // environment (service) is registered.
+    expect(resolved.environments?.ssr).toBeUndefined();
+
+    // The silent part of #4591: this situation must at least produce a warning.
+    expect(warnings.join("\n")).toContain("ssr entry");
+  });
+});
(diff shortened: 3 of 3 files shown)


pengzh1:
Friendly ping on this one — it's been ready for review for a few days now (checks are green and the earlier review feedback has been addressed). If a `warn` isn't the right shape here I'm happy to rework it, just let me know.

pengzh1:
Closing this for now — it's been sitting through a few review cycles without feedback, and I'd rather keep the queue clean than hold a stale branch open. The branch stays on my fork (head 29e55c6, rebase-ready) and I'm happy to reopen or rebase it if there's interest later. Thanks for the earlier review pass on the assertion hardening.

PR #4630, opened 2026-09-14
fix(rollup): load handler meta staged under file:// URLs

### LINKED ISSUE #4564: NOENT when loading .js server route from layer via nitro-handler-meta virtual module
### Describe the bug

When a Nuxt layer contains .js server route files, Nitro scans and registers them correctly but then fails to load them via the nitro-handler-meta virtual module with an ENOENT error. The file physically exists at the reported path. Renaming the file to .ts fixes the issue immediately.

### Error message

```
[nitro 4:39:32 PM]  ERROR  Error: Could not load nitro-handler-meta:file:///pathToProject/Client/packages/Payment/server/api/payment/bankRegister2.post.js (imported by virtual:#nitro-internal-virtual/server-handlers-meta): ENOENT: no such file or directory, open 'file:///pathToProject/Client/packages/Payment/server/api/payment/bankRegister2.post.js'


undefined
```

### To reproduce

Nuxt monorepo with multiple layers in extends
One layer contains a .js server route (e.g. server/api/hello.post.js)
Run nuxt dev
Nitro throws the ENOENT error above
Expected behavior

.js server route files in layers should load correctly, the same as .ts files.

### Actual behavior

Nitro registers the .js handler during scan but fails to open the file when building server-handlers-meta. .ts files in the exact same location work without any issue, indicating the file path is valid and the problem is specific to how the handler-meta virtual plugin resolves .js files from layer directories.

### Workaround

Rename all .js server route files in layers to .ts.

### Another scary workaround 

error disappear for .js files if layer folder name has the same start characters which basic layer. So if in my monorepo I have layers organized as:

/packages/App
/packages/Base
/packages/Auth

and I am running app from App which has `extend:['../Auth','../Base']` in nuxt.config
the error emerges for first .js server route found in Auth layer,

when I change layer folder name (!) to BaseAuth (or BaseAnything) : 

/packages/App
/packages/Base
/packages/BaseAuth

and I am running app from App which has now `extend:['../BaseAuth','../Base']`

error is gone... 

### Environment

 nuxt info

│ │  Operating system      macOS 25.5.0                                                    │
│ │  CPU                   Apple M4 Pro (12 cores)                                         │
│ │  Node.js version       v22.22.2                                                        │
│ │  nuxt/cli version      3.37.0                                                          │
│ │  Package manager       pnpm 11.24.0                                                    │
│ │  Nuxt version          4.5.2                                                           │
│ │  Nitro version         2.13.4                                                          │
│ │  Builder               vite 8.2.2                                                      │
│ │  Config                $meta, compatibilityDate, devtools, experimental, extends, ssr  │
│ │  Modules               -                                                               │

nikolas-sapa:
### 🔗 Linked issue

Fixes #4564

### ❓ Type of change

- [x] 🐞 Bug fix (a non-breaking change which fixes an issue)

### 📚 Description

When a Nuxt layer contains `.js` server routes, nitro scans and registers them but then fails to load them via the `nitro-handler-meta` virtual module:

```
ERROR Error: Could not load nitro-handler-meta:file:///path/to/project/Client/packages/Payment/server/api/payment/bankRegister2.post.js (imported by virtual:#nitro-internal-virtual/server-handlers-meta): ENOENT: no such file or directory, open 'file:///...'
```

The chain:

1. `#nitro-internal-virtual/server-handlers-meta` generates `import ...Meta from "<handler>?meta"` for every scanned handler.
2. The handlers-meta plugin's `resolveId` asks the host bundler to resolve the `<handler>` specifier. For files **outside the project root** (which is the normal case for server routes in Nuxt layers), the bundler reports the resolved id as a `file://` URL.
3. The plugin builds `\0nitro-handler-meta:file://...` and later loads it with `readFile("file://...")` — but `fs.readFile` does not understand URL strings, so it throws ENOENT even though the file exists.

Renaming the file to `.ts` made the error disappear only because the layer's resolver then returned a plain path; the file format itself was never the problem. It also explains the issue's second workaround: with a layer folder sharing a prefix with the base layer, the bundler's resolver caches/normalizes differently and stops returning the URL form.

This converts `file://` ids back to paths with `fileURLToPath` before reading, mirroring what the externals plugin already does (`src/rollup/plugins/externals.ts:41`).

### 📝 Checklist

- [x] I have read the [Contributing Guide](https://github.com/nitrojs/nitro/blob/main/CONTRIBUTING.md).
- [x] I have included a change request file using the `$ ` prefix (not applicable — bug fix with linked issue).
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.


pi0:
unreliable AI

PR #4632, opened 2026-09-15
feat: apply server entry middleware and plugins on all presets

pi0x:
Follow-up to #4620.

Server entry `middleware`, `plugins` and `error` options were only passed to the srvx server of the `node`, `bun` and `deno` presets. They worked locally but were silently ignored on serverless, edge and worker presets (Cloudflare, Vercel, Netlify, AWS Lambda, ...).

These options are now applied by `useNitroApp().fetch`, so they behave the same on every preset, in the Vite dev server, during prerendering and when calling `useNitroApp().fetch()` directly.

```ts
// server.ts
import { defineServerEntry } from "nitro";

export default defineServerEntry({
  middleware: [
    async (req, next) => {
      const res = await next();
      res.headers.set("x-powered-by", "nitro");
      return res;
    },
  ],
  fetch(req) {
    return new Response("Hello!");
  },
});
```

- Presets that start a srvx server (`node-server`, `node-cluster`, `bun`, `deno-server`, `nitro-dev`) pass all options to the native srvx adapter with the new raw `nitroApp["~fetch"]` handler. A small srvx plugin (registered by `resolveServeOptions`, and explicitly by `nitro-dev`) then points `nitroApp.fetch` to that server's middleware chain, so direct calls get the same middleware without running plugins again.
- Other presets need no changes. Plugins receive a minimal server object (`runtime: "generic"` and `options`) once, on first request, and middleware is composed around the app handler.
- Internal `serverFetch()` and `fetch("/...")` calls and WebSocket upgrades skip the entry middleware, as before.
- Listener and runtime specific options (`port`, `tls`, `maxRequestBodySize`, `trustProxy`, `gracefulShutdown`, `node`, `bun`, `deno`) still only apply to the Node.js, Bun and Deno servers.
- Presets without a srvx server add no runtime code when there is no server entry. The rolldown unminified minimal bundle budget is raised by 1 kB (~0.3 kB over the rounding edge). `srvx/generic` is not used: it adds ~3 kB and overrides the platform `request.waitUntil` (h3js/srvx#306).

Docs: `docs/1.docs/6.server-entry.md` ("Server options" section).

🤖 Generated with AI assistant


pi0:
not needed after https://github.com/nitrojs/nitro/pull/4654

Issue #4555, opened 2026-08-24
Default error handler throws ERR_INVALID_URL on an unparsable Host header


sergiooak:
### Environment

- nitropack: 2.13.4 (`latest`)
- h3: 1.15.11
- Node.js: 22.17.0
- preset: `node-server`
- Reproduced on Windows 11. The same error also appears on a Linux production deployment.

### Reproduction

https://github.com/sergiooak/nitro-host-header-error-handler

```sh
npm install
npm run build
node .output/server/index.mjs
# then, in another shell:
curl -s -H 'Host: a b' http://127.0.0.1:3000/x
```

### Describe the bug

The default error handler builds the request URL from the raw `Host` header. When the header is not a valid URL authority, `new URL()` throws inside the handler, and the throw is not caught there.

Flow:

1. A request arrives with a `Host` header that `new URL()` cannot parse, for example a value containing a space.
2. The route does not match, so Nitro runs the default error handler to render the 404.
3. `defaultHandler` calls `getRequestURL(event, { xForwardedHost: true, xForwardedProto: true })`.
4. `getRequestURL` runs `new URL(path, \`${protocol}://${host}\`)` with the raw `Host`. `new URL()` throws `TypeError [ERR_INVALID_URL]`.
5. The throw escapes the error handler as a rejected promise. `trapUnhandledNodeErrors` catches it and logs the full stack. The client gets a 404 with a reduced body. The process stays up.

Every request with such a `Host` logs one full stack trace.

Root cause:

- h3 `getRequestURL` builds the URL from the unvalidated `Host` header (`getRequestHost` returns `event.node.req.headers.host`).
- Nitro `defaultHandler` (`src/runtime/internal/error/prod.ts`) calls `getRequestURL` with no `try/catch`, so a throw in the error handler is not contained.

### Additional context

Which `Host` values trigger it: `new URL('http://' + host)` throws for authority values containing characters such as a space, `[`, `]`, `<`, `%`, or `::`. It does not throw for `"` or `/`. Vulnerability scanners commonly send values like `"></A></ADDRESS>"><script>alert(document.domain)</script><ADDRESS><A "/xss"'`, which throws because of the space.

Suggested fix: wrap the `getRequestURL` call in `defaultHandler` in `try/catch` and fall back to `event.path`, or validate the host in `getRequestURL`. The same class of bug was handled in `@hono/node-server` (CVE-2024-32652).

[...]

PR #4640, opened 2026-09-19
fix(error): don't throw on unparsable Host header in default error handlers

nikolas-sapa:
### 🔗 Linked issue

Resolves #4555

### ❓ Type of change

### 📚 Description

The default error handlers build the request URL from the raw `Host` header via `getRequestURL(event, { xForwardedHost: true, xForwardedProto: true })`. For a header that `new URL()` cannot parse — `Host: a b` from the issue's reproduction — that call throws *inside the error handler itself*, and the throw escapes as an unhandled rejection: every such request logs a full stack trace through `trapUnhandledNodeErrors` and the client receives a reduced body instead of the proper error payload. The `Host` header is unvalidated input, so this is reachable by anyone with `curl -H 'Host: a b'`.

Both handlers are affected identically (`src/runtime/internal/error/prod.ts` and `dev.ts`), so the fix adds one shared helper, `getRequestURLOrFallback`, in `internal/error/utils.ts`: it resolves the URL exactly as before, and on an unparsable authority falls back to a synthetic `http://localhost` base while preserving path and query. That keeps all three URL consumers working in the degraded case — the 404 baseURL redirect logic (pathname/search), the `[request error]` log line, and the response body's `url` field.

For every request with a valid `Host` or `x-forwarded-host`, behaviour is byte-identical to before; the handler only diverges where it previously threw.

**Testing.** `test/unit/error-handler.test.ts` covers: valid Host (unchanged URL), the malformed-Host degradation (no throw, fallback authority, status intact), path+query preservation, x-forwarded-host precedence (also unchanged), and the dev handler. Reverting the source change turns the three degradation tests red — the handler throws `TypeError [ERR_INVALID_URL]` before reaching a response — while the two valid-header tests stay green either way. `test/unit` passes (46 tests, including the pre-existing files), and `tsc --noEmit` / eslint / prettier are clean.

### 📝 Checklist


pi0:
unreliable AI

Issue #4446, opened 2026-07-18
Vercel preset: ISR URL restoration has no fallback — a failed restoration renders a cacheable 404 for the internal `-isr` path


MichaelDickBasecom:
### Environment

- nitropack `2.13.4` (lockfile-resolved, latest release) — the gap also exists in v3 `main` (`presets/vercel/runtime/isr.ts`, introduced via #3851)
- Nuxt `4.4.8` / `@nuxt/nitro-server@4.4.8`
- Node `v25.9.0` (repro script needs ≥ 20.11 for `import.meta.dirname`), Bun `1.3.14` for install/build
- Preset: `vercel` (Build Output API v3)

Relevant config — any single `isr` route rule is enough:

```ts
routeRules: {
  '/schedule': { isr: 300 },
}
```

### Reproduction

https://github.com/MichaelDickBasecom/nuxt-vercel-isr-404-repro

```sh
bun install
NITRO_PRESET=vercel bun run build
node scripts/local-repro.mjs   # exits non-zero if any case diverges
```

The StackBlitz/CodeSandbox starters can't exercise this bug: it lives in the *built Vercel function's* runtime URL restoration, and reproducing it means invoking `.vercel/output/functions/__fallback.func` with the exact request shapes Vercel's router/prerender cache produces. `scripts/local-repro.mjs` does precisely that — it serves the built function locally and sends it each invocation shape, so the failure is deterministic with **no deployment needed**. The repo's CI runs this script on every push against the frozen lockfile: https://github.com/MichaelDickBasecom/nuxt-vercel-isr-404-repro/actions/workflows/local-repro.yml

A live deployment with all route-rule shape variants (static / dynamic param / catch-all / nested / `passQuery`) plus self-monitoring is linked from the README: https://nuxt-vercel-isr-404-repro.vercel.app

### Describe the bug

On Vercel, Nitro implements `isr` route rules by rewriting the request to an internal prerender function (`/schedule` → `/schedule-isr?__isr_route=/schedule`) and reconstructing the original URL inside the function. Reconstruction is mandatory: Vercel invokes the function with `req.url` set to the rewritten dest path, and with a default `isr: <seconds>` rule the query string is stripped, so the URL-encoded groups in the undocumented `x-now-route-matches` header are the **only** carrier of the original URL.

The restoration in 2.13.4 (`presets/vercel/runtime/vercel.mjs`, decompiled for clarity) trusts that single carrier and gives up silently:

```js
[...]

PR #4641, opened 2026-09-19
fix(vercel): harden ISR URL restoration against carrier variants

nikolas-sapa:
### 🔗 Linked issue

Resolves #4446

### ❓ Type of change

### 📚 Description

This implements the fix design prioritized in #4446 — full credit to the reporter for the analysis, the carrier-shape taxonomy and the fail-safe reasoning; the code follows their proposal (items 1 and 2).

The ISR fallback function recovered the original URL from a single carrier: the `__isr_route` group of `x-now-route-matches`, and — only when the header was absent *entirely* — the destination query. Two gaps remain:

1. **Header present without `__isr_route`** (e.g. positional-only serialization like `1=schedule`): the header branch ran and found nothing, and the `if/else` shadowed the query fallback — so the rewrite's identical query value was never consulted. The internal `-isr` route rendered a 404, which Vercel stores as a *valid* ISR outcome and replays with `x-vercel-cache: HIT` — cache poisoning.
2. **Bare invocation** (no header, no query): neither branch rewrote, same cacheable-404 outcome.

Changes to `presets/vercel/runtime/vercel.ts`:

- Consult the destination query whenever the header carrier yields no `__isr_route` — strictly additive, since the rewrite writes the identical value into the dest query. Every carrier shape that works today resolves through the exact same code path as before.
- When no carrier resolves and the request targets an internal `-isr` path, respond **503** instead of rendering the internal route. As measured in the issue on a live deployment: 404/410 are stored by the ISR cache, 401/5xx are never stored and any existing good copy keeps serving (`x-vercel-cache: STALE`) — turning an unrecognized future carrier shape into a transient miss instead of permanent poisoning.

One trade-off, explicit in the issue: a *user* route literally named `…-isr` hit bare would now get a 503 from this function. That is the accepted cost of the fail-safe, and external requests to such paths land on separate cache keys anyway.

**Testing.** Four runtime tests against the built fixture function (the harness invokes `functions/__fallback.func` directly): header carrier (unchanged), query carrier (unchanged), header-without-`__isr_route` falling back to the query (fails on `main` with a 404 for the internal path), and bare invocation returning 503 for both the page and `_payload.json` variants. Reverting the runtime change turns exactly the two bug-pinning tests red while the two working-path tests stay green — pinning rather than describing. Full file: 55 passed. eslint / prettier / `tsc --noEmit` clean.

The v3 `main` equivalent (`presets/vercel/runtime/isr.ts`, via #3851) has the same gaps — happy to port this once the v2 approach is agreed.

### 📝 Checklist


pi0:
unreliable AI

PR #4651, opened 2026-09-22
fix(vercel): mark header-only route entries as `continue`

danielroe:
### ❓ Type of change

### 📚 Description

currently setting a headers route rule means we clobber any other rules, e.g. caching rules, and always fall to a function

discovered while debugging an issue with @larbish and @atinux

### 📝 Checklist


danielroe:
cc: @pi0 we likely need to pin actions on v2 branch 🙏 

> Error: The actions actions/checkout@v6 and actions/setup-node@v6 are not allowed in nitrojs/nitro because all actions must be pinned to a full-length commit SHA.

pi0 (approved):

PR #4652, opened 2026-09-22
fix(vercel): mark header-only route entries as `continue`

danielroe:
### ❓ Type of change

### 📚 Description

port of https://github.com/nitrojs/nitro/pull/4651 to main

currently setting a headers route rule means we clobber any other rules, e.g. caching rules, and always fall to a function

discovered while debugging an issue with @larbish and @atinux

### 📝 Checklist


pi0x:
On top of the first commit: 1e42875 orders header-only routes from least to most specific so the more specific rule wins under `continue`, and 2c1e248 drops `continue` for header rules with `redirect: false` so a broader redirect (e.g. `/foo/**`) no longer redirects them at the edge. Verified with a real Vercel deployment against main: public asset immutable caching now works, header precedence is correct, and ISR/CDN caching is unchanged.

_Written with AI assistance._

pi0 (approved):

